Issue #93, opened 2022-04-08
Split `engine.py`


betodealmeida:
`engine.py` has a lot of logic that is not related to executing queries. We should move the functions `get_query_for_node` and `get_query_for_sql` (as well as associated functions) to a new file `sql/build.py`, since they are used to build queries from a node or from SQL.

zzzzz1st:
I'll give that a try.

betodealmeida:
Thanks, @zzzzz1st! I'll take a look at your PR. Feel free to reach out if you have any questions, I'm still working on the developer docs.

PR #95, opened 2022-04-09
Splitted engine.py for issue #93

zzzzz1st:
I just splitted the engine.py. I didn't write the test cases.

### DIFF
--- a/src/datajunction/engine.py
+++ b/src/datajunction/engine.py
@@ -1,58 +1,35 @@
 """
 Query related functions.
 """
-
 import ast
 import operator
 import re
 from datetime import datetime, timezone
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Tuple
 
 import sqlparse
 from sqlalchemy import text
-from sqlalchemy.engine import create_engine as sqla_create_engine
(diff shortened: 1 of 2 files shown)


betodealmeida:
Thanks for the PR, @zzzzz1st! 

Some of the tests are failing because some imports need to be updated. For example:

```
src/datajunction/api/queries.py:23: in <module>
    from datajunction.engine import get_query_for_sql, process_query
E   ImportError: cannot import name 'get_query_for_sql' from 'datajunction.engine' (/home/runner/work/datajunction/datajunction/src/datajunction/engine.py)
```

You can check the failing tests locally by running `make test`, or if that doesn't work for you (it required `pyenv` to be already installed and configured) you can try tox:

```bash
pip install -e '.[testing]'
pip install tox
tox
```

Or using `pytest` directly:

```bash
pip install -e '.[testing]'
pytest --cov=src/datajunction -vv tests/ --doctest-modules src/datajunction
```

It would also be good to split the current tests we have in `tests/engine_test.py`, moving the corresponding tests into `tests/sql/build_test.py`.

zzzzz1st:
I reopened this PR as a new PR. I also divided the test cases on the new one. 

Thanks Beto.

Issue #43, opened 2022-04-04
Add more functions


betodealmeida:
Currently the only supported functions are `COUNT` and `MAX`: https://github.com/DataJunction/datajunction/blob/main/src/datajunction/sql/functions.py

aliirz:
Hi @betodealmeida  I will work on this as soon as I get my env setup. Having some difficulty at the moment.

betodealmeida:
@aliirz that's awesome! Let me know if you need any help. I recently added the `CONTRIBUTING.rst` file with some instructions, but I know they're pretty sparse.

betodealmeida:
I would suggest taking a look at the functions supported by SQLAlchemy (https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/functions.py#L1195 and onwards).

aliirz:
Thanks. I am having some trouble setting up the env I've been following the guide in contribution document but I still can't run the tests.

betodealmeida:
@aliirz can you post the error message you get?

aliirz:
> @aliirz can you post the error message you get?

Looks like some issue with the imports absolute vs relative paths:

```
ImportError while loading conftest '/Users/aliraza/projects/datajunction/tests/conftest.py'.
tests/conftest.py:17: in <module>
    from datajunction.api.main import app
E   ModuleNotFoundError: No module named 'datajunction'
make: *** [test] Error 4

```

I've followed the instructions in the contribution docs to the dot :(

aliirz:
@betodealmeida  not a problem, looking forward to help maintain

betodealmeida:
@aliirz are you using Windows, MacOS, or Linux?

aliirz:
@betodealmeida  I am on Mac OS 12.3.1

betodealmeida:
@aliirz thanks, I'll update the docs soon!

aliirz:
@betodealmeida  appreciate it. i can troubleshoot too if you can provide some guidance on what I might be doing wrong

betodealmeida:
@aliirz, another way to run the tests is via Docker, maybe you can give it a try?

1. Run the `docker-compose.yml file` with `docker compose up` in a terminal.
2. In another terminal, run `docker exec -u root -it dj bash`, this will put you inside the main container.
3. In that container, run `pip install -e '.[testing]'`.
4. Run `make test` and `make check`.

[...]

PR #96, opened 2022-04-09
feat: SQAlchemy dialect

betodealmeida:
```python
from sqlalchemy.engine import create_engine

query = 'SELECT "core.comments.user_id", "core.num_comments" FROM metrics GROUP BY "core.comments.user_id"'
engine = create_engine("dj://localhost:8000/0")
connection = engine.connect()

for row in connection.execute(query):
    print(row)
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -104,6 +104,8 @@ docs =
 #     awesome = pyscaffoldext.awesome.extension:AwesomeExtension
 console_scripts =
     dj = datajunction.console:run
+sqlalchemy.dialects =
+    dj = datajunction.sql.sqlalchemy.dialect:DJDialect
 
 [tool:pytest]
 # Specify command line options as you would do when invoking pytest directly.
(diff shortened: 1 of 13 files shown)


betodealmeida:
@agorajek a big milestone! :)

agorajek:
Wow, so much good stuff. So does it mean that, if we can translate to SQLAlchemy then we get all the SQL Alchemy built-in translations for free?

betodealmeida:
> Wow, so much good stuff. So does it mean that, if we can translate to SQLAlchemy then we get all the SQL Alchemy built-in translations for free?

Not really, this just allows us to query metrics using SQL and a "virtual" table called `metrics`. The stack is :

```
[SQLAlchemy dialect => DB API => sqloxide => transpile.py] => [SQLAlchemy query object => DB-specific SQL]
\-------------------------- DJ --------------------------/    \--------------- SQLAlchemy ---------------/
```

The translation happens on the last step, so we still need to add more methods to `transpile.py` in order to support `JOIN`s (#81, which I'm currently working on), `WITH`, `ORDER BY`, etc. And also add more functions (#43).

But this PR means that in theory we can already connect Superset to the DJ metric repo and query metrics.

PR #129, opened 2022-04-18
fix: `DATE_TRUNC`

betodealmeida:
Found a problem with `DATE_TRUNC`, we were passing the resolution as a `TextClause` and the identifier preparer would print it without quotes (since it was not a string). Now the query renders correctly, with the resolution quoted as a string should be.

Also fixed a problem with `ORDER BY` not working with columns, only metrics.

### DIFF
--- a/tests/sql/transpile_test.py
+++ b/tests/sql/transpile_test.py
@@ -484,7 +484,7 @@ def test_date_trunc() -> None:
 
     assert (
         query_to_string(get_function(function, source, dialect="postgresql"))
-        == "date_trunc(second, anon_1.col)"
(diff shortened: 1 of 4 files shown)


tests/sql/transpile_test.py
```diff
@@ -484,7 +484,7 @@ def test_date_trunc() -> None:
 
     assert (
         query_to_string(get_function(function, source, dialect="postgresql"))
-        == "date_trunc(second, anon_1.col)"
```
betodealmeida, line 487: D'oh, the unit tests was showing the bug, I should've noticed this.

PR #138, opened 2022-04-22
Add dates and datetimes to filter api.

agorajek:
### Summary
A small addition to the filter functionality to cover the Date and Datetime column types.

### Test Plan
Mostly just unit tests. Any other way I should test it?

### Deployment Plan
not sure


src/datajunction/sql/build.py
```diff
@@ -89,13 +91,20 @@ def get_filter(columns: Dict[str, SqlaColumn], filter_: str) -> BinaryExpression
...
+            value = parse(value)
+        except Exception as ex:
+            raise Exception(f"Invalid date or datetime value: {value}") from ex
+    else:
+        try:        
+            value = ast.literal_eval(value)
+        except Exception as ex:
+            raise Exception(f"Invalid value: {value}") from ex
+
+    print('ccv', comparison, column, value)
```
betodealmeida, line 107:
```suggestion
```

src/datajunction/sql/build.py
```diff
@@ -89,13 +91,20 @@ def get_filter(columns: Dict[str, SqlaColumn], filter_: str) -> BinaryExpression
...
     column = columns[name]
-
-    try:
-        value = ast.literal_eval(value)
-    except Exception as ex:
-        raise Exception(f"Invalid value: {value}") from ex
-
     comparison = COMPARISONS[operator_]
+
+    if type(column.type) in [Date, DateTime]:
```
betodealmeida, line 96: Eventually I think we'll need to use the Python type here:

```python
if ColumnType(column.type.python_type.upper()) in {ColumnType.DATE, ColumnType.DATETIME}:
```

Since I think some databases defined their own specific types for Date and DateTime.

betodealmeida (approved):

src/datajunction/sql/build.py
agorajek: Done.

betodealmeida (approved):

PR #139, opened 2022-04-24
chore: add integration test

betodealmeida:


.github/workflows/python-integration.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches: [ main ]
+
+jobs:
+  integration:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.8, 3.9, '3.10']
```
agorajek, line 16: Why only the last one is in quotes?

.github/workflows/python-integration.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
```
agorajek, line 19: What v2 are we talking here?

docker-compose-ci.yml
```diff
@@ -0,0 +1,211 @@
+#
```
agorajek, line 1: Could you please clarify the difference between docker-compose and docker-compose-ci? Also, is it possible to make one as an extension of the other instead of copy-pasta?

agorajek (commented):
The title is almost self explanatory, but I am wondering when will they run or anything else about them...

.github/workflows/python-integration.yml
betodealmeida: Good point, I think we could quote them all... traditionally they weren't quoted and were treated as a float. So when 3.10 came out it was considered as 3.1 by the action.

.github/workflows/python-integration.yml
betodealmeida: The action is defined here: https://github.com/actions/checkout

I'm using v2 because that's what I used in the past, but I just noticed there's now a v3.

docker-compose-ci.yml
betodealmeida: Oh, man... this has been a nightmare to figure out. `docker-compose.yml` defines local volumes, so that the data is persisted in your disk. This way when you spin it down and bring it back up again the data is still there. It also defines a "symlink" volume for the DJ repo:

```yaml
dj:
  volumes:
    - .:/code  # map current directory to /code
```

For some reason this doesn't work in CI, so I created a separate file without these volumes. Now it works, but the integration tests are still failing and I don't know why.

I'm going to refactor this whole PR and split it into 2-3, for now I just want to get the integration test passing.

betodealmeida:
> The title is almost self explanatory, but I am wondering when will they run or anything else about them...

You can run the locally with `make integration` (or using `pytest` directly, of course). And they will run in CI, right now as a non-required check.

betodealmeida:
I split this into:

- https://github.com/DataJunction/datajunction/pull/140
- https://github.com/DataJunction/datajunction/pull/141
- https://github.com/DataJunction/datajunction/pull/142
- https://github.com/DataJunction/datajunction/pull/143

betodealmeida:
We still need to figure out a way of running the integration test(s) as a GitHub action.

docker-compose-ci.yml
betodealmeida: @agorajek I split the PR into 4, and still didn't get CI integration tests to work. Going to give up for now, since we can run integration tests manually at least.

PR #145, opened 2022-04-28
Add min() to DJ SQL.

agorajek:
That's it. To not loose the touch with the project. 

Interestingly... the new `tests/integration/basic_test.py` have been failing on my side, but I assume it is something with my sqlalchemy. We'll see...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ ENV/
 
 # https://pypi.org/project/python-dotenv/
 .env
+
+# VS Code
+.vscode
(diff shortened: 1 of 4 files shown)


betodealmeida:
> Interestingly... the new `tests/integration/basic_test.py` have been failing on my side, but I assume it is something with my sqlalchemy. We'll see...

Oh weird, are you running `docker compose up` before running the tests?

betodealmeida:
It's also weird that so many tests failed... I'll take a look.

betodealmeida:
I checked out your branch and the tests work (as expected), so I just restarted them.

betodealmeida (approved):
I have no idea why the tests are failing.

PR #146, opened 2022-04-29
fix: ping SQLAlchemy due to regression introduced in 1.4.36

betodealmeida:


betodealmeida:
See https://github.com/tiangolo/sqlmodel/issues/315

PR #151, opened 2022-04-30
feat: DB migrations

betodealmeida:
Use alembic for DB migrtions.

Closes https://github.com/DataJunction/datajunction/issues/150


betodealmeida:
@agorajek I've added Alembic to handle database migrations. You need to remove the old Postgresl database:

```bash
% docker compose down
% docker volume rm -f datajunction_postgres_metadata
% docker compose up
```

For some reason the first time I did this the watchdog died, so I had to <kbd>ctrl+c</kbd> and do run `docker compose up` again.

PR #155, opened 2022-05-01
AVG and SUM functions

aliirz:
Added a couple of functions listed in #43


betodealmeida:
This is awesome! Thanks, @aliirz!

PR #168, opened 2022-06-13
Ping each database when evaluating candidates for query.

agorajek:
### Summary
As per title, we now check if the database is active before returning it to the user as a potential query source. This prevents getting queries that can not be run at the moment.

### Test Plan
In my setup Druid db is broken, but defined as the lowes cost db, so I was getting an error before the fix. After this change I am getting a query from the next best db - Postgres.

### Deployment Plan
auto

### DIFF
--- a/src/datajunction/sql/dag.py
+++ b/src/datajunction/sql/dag.py
@@ -118,13 +118,18 @@ def get_database_for_nodes(
     if not databases:
         raise Exception("No valid database was found")
 
+    active_databases = [db for db in databases if db.ping()]
+
+    if not active_databases:
+        raise Exception("No active database was found")
+
     if database_id is not None:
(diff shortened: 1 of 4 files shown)


agorajek:
Oh wow. I just realized I didn't add a reviewer when I made my PR...

src/datajunction/sql/dag.py
```diff
@@ -118,13 +118,18 @@ def get_database_for_nodes(
     if not databases:
         raise Exception("No valid database was found")
 
+    active_databases = [db for db in databases if db.ping()]
```
betodealmeida, line 121: Let's parallelize this in the future.

betodealmeida (approved):

PR #175, opened 2022-08-05
feat: Superset DB engine spec

betodealmeida:


src/datajunction/superset.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Since DJ doesn't have tables per se, a ``SELECT *`` query doesn't make sense.
+        """
+        message = SELECT_STAR_MESSAGE.replace("'", "''")
+        return f"SELECT '{message}' AS warning"
+
+    @classmethod
+    def get_metrics(  # pylint: disable=unused-argument
+        cls,
+        database: "Database",
```
agorajek, line 86: What does this type mean `"Database"` ?

agorajek (approved):
awesome_sauce

src/datajunction/superset.py
betodealmeida: This a Superset `Database` object... we only use it for type annotation, so instead of importing the actual `Database` class and using it here (which would require to have Superset installed) we can just reference it as a string and do the import under a `if TYPE_CHECKING` check:

https://github.com/DataJunction/datajunction/pull/175/files#diff-4ef254f7258961794702764475cd8683b37116cf4c0c24d803c76907556fe945R18-R19.

It's a little bit clunky. But usually when you have cyclic dependencies you use the type as a string. For example, this will fail:

```python
class Parent:
    child: Child  # NameError: name 'Child' is not defined

class Child:
    parent: Parent
```

So we do:

```python
class Parent:
    child: "Child"

class Child:
    parent: "Parent"
```

There are better ways of doing this, starting in Python 3.7:

```python
from __future__ import annotations

class Parent:
    child: Child  # Now it works!

class Child:
    parent: Parent
```

This will be the standard in 3.11. Since DJ supports 3.8+, maybe we should add the `from __future__ import annotations` to relevant files.

PR #177, opened 2022-08-06
feat: parallelize `do_ping` and add timeout

betodealmeida:
This PR improves the `get_database_for_nodes` function, making the calls to `do_ping` in each database run in parallel. It also introduces a timeout.

With the new behavior the pings will run in parallel, but the function still tries to return the fastest DB (and not the one with the fastest ping). For example, if we have these 2 databases:

- DB 1, cost = 1, ping = 3 seconds
- DB 2, cost = 10, ping = 1 second

In this case the `do_ping` for DB 2 will return first, but the function will wait until `do_ping` for DB 1 finishes, returning DB 1 as the fastest DB.

On the other hand, if we have:

- DB 1, cost = 1, ping = 10 seconds
- DB 2, cost = 10, ping = 1 second

Since the default timeout is 5 seconds the function will timeout before the `do_ping` for DB 1 finishes, returning DB 2 as the fastest DB.

Note that this doesn't improve the best case scenario (where the fastest DB is online), but improves cases where the fastest DB is offline.

### DIFF
--- a/src/datajunction/sql/dag.py
+++ b/src/datajunction/sql/dag.py
@@ -2,6 +2,7 @@
 DAG related functions.
 """
 
+import asyncio
 import operator
 from collections import defaultdict
 from io import StringIO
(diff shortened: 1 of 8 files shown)


agorajek (approved):
Wow. So much good code for such a little feature.

src/datajunction/sql/dag.py
```diff
@@ -118,18 +122,61 @@ def get_database_for_nodes(
...
+    for ``settings.do_ping_timeout`` it will return the fastest database that is online.
+    """
+    # sort by cheapest
+    sorted_databases = sorted(databases, key=operator.attrgetter("cost"))
+
+    # create tasks to ping all of the databases in parallel
+    pings = [asyncio.create_task(database.do_ping()) for database in sorted_databases]
+    pending = set(pings)
+
+    loop = asyncio.get_running_loop()
```
agorajek, line 150: Just one question... why not `get_event_loop()`. It looks like `get_running_loop()` will fail if there is no current loop running, and `get_event_loop()` will start one.

betodealmeida: Good question! In this case we're inside a coroutine (`async def`), so we know we're going to be inside an event loop, since it's the only way to call an async function as far as I know. The docs mention:

> Because this function has rather complex behavior (especially when custom event loop policies are in use), using the [get_running_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) function is preferred to [get_event_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) in coroutines and callbacks.

PR #180, opened 2022-09-02
refactor: dimension nodes have expressions

betodealmeida:
Change dimension nodes so they must have an `expression`, making them more similar to transform nodes than source nodes.


agorajek:
Looks good.

agorajek (approved):

PR #182, opened 2022-09-02
chore: rename `node.expression` to `node.query`

betodealmeida:
Renamed the attribute to "query", since it's more appropriate.

### DIFF
--- a/ARCHITECTURE.rst
+++ b/ARCHITECTURE.rst
@@ -2,7 +2,7 @@
 Architecture
 ============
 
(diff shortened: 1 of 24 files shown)


agorajek (approved):
Does the `alembic/versions/...file` get generated automatically or did you write it by hand?

betodealmeida:
> Does the `alembic/versions/...file` get generated automatically or did you write it by hand?

Sort of! The right way to create a migration is:

1. Update the models.
2. Enter the docker container (`docker exec -it dj bash`) and run `alembic revision --autogenerate -m "Description of the migration"`.
3. Then edit the file generated to make sure it's correct.

In this case the migration was trying to change too much stuff, so I trimmed it down to what you see in the PR.

Issue #98, opened 2022-04-10
Command to add databases


betodealmeida:
We could have a command to add databases:

```bash
% dj add-database dev.test_db sqlite:////tmp/file.db --description="A simple SQLite database" --cost=10
```

This would create the file `dev/test_db.yaml` with the configuration for the database:

```yaml
description: A simple SQLite database
URI: sqlite:////tmp/file.db
read_only: false
cost: 10
```

PR #183, opened 2022-09-06
add-database command

samredai:
This is an attempt to solve issue #98, adding an `add-database` command.
```
dj add-database DATABASE [REPOSITORY] --uri=URI [-f --loglevel=INFO --description=<str> --read-only=<bool> --cost=COST]
```

### DIFF
--- a/src/datajunction/cli/add_database.py
+++ b/src/datajunction/cli/add_database.py
@@ -1,0 +1,45 @@
+"""
+Add a database to an existing repository
+"""
+
+import logging
+from pathlib import Path
+from typing import Optional
+
+import yaml
+
+from datajunction.errors import DJError, DJException, ErrorCode
+from datajunction.models.database import Database
+
+_logger = logging.getLogger(__name__)
+
+def create_databases_dir(repository: Path):
+    """
+    Creates a databases directory in the repository if one doesn't already exist
+    """
+    database_dir_path = repository / Path("databases")
+    database_dir_path.mkdir(exist_ok=True)
+
+async def run(repository: Path, database: str, uri: str, description: Optional[str] = None, read_only: Optional[bool] = None, cost: Optional[float] = None) -> None:
+    """
+    Add a database.
+    """
+    create_databases_dir(repository)
+    db_config_path = repository / Path(f"databases/{database}.yaml")
+    if Path(db_config_path).exists():
+        raise DJException(message=f"Database configuration already exists", errors=[DJError(message=f"{db_config_path} already exists", code=ErrorCode.ALREADY_EXISTS)])
+
+    db_kwargs = {
+      "database": database,
+      "URI": uri,
+    }
+    if description:
+      db_kwargs["description"] = description
+    if read_only:
+      db_kwargs["read_only"] = read_only
+    if cost:
+      db_kwargs["cost"] = cost
+
+    db = Database(**db_kwargs)
+    with db_config_path.open("w", encoding="utf-8") as output:
+        yaml.safe_dump(db.to_yaml(), output, sort_keys=False)
--- a/src/datajunction/console.py
+++ b/src/datajunction/console.py
@@ -3,9 +3,17 @@ DataJunction (DJ) is a metric repository.
 
 Usage:
     dj compile [REPOSITORY] [-f] [--loglevel=INFO] [--reload]
+    dj add-database DATABASE [REPOSITORY] --uri=URI [-f --loglevel=INFO --description=<str> --read-only=<bool> --cost=COST]
 
 Actions:
     compile                 Compile repository
+    add-database            Add a database to an existing repository
+
+Arguments:
+    REPOSITORY              Path to a DJ repository
+    DATABASE                Name of a database
+    URI                     URI to a database
+    COST                    The relative cost as compared to other databases
 
 Options:
     -f, --force             Force indexing.    [default: false]
@@ -24,6 +32,7 @@ from docopt import docopt
 
 from datajunction import __version__
 from datajunction.cli import compile as compile_
+from datajunction.cli import add_database as add_database_
 from datajunction.errors import DJException
 from datajunction.utils import get_settings, setup_logging
 
@@ -45,7 +54,7 @@ async def main() -> None:
         repository = Path(arguments["REPOSITORY"])
 
     try:
-        if arguments["compile"]:
+        if arguments.get("compile"):
             try:
                 await compile_.run(
                     repository,
@@ -54,6 +63,19 @@ async def main() -> None:
                 )
             except DJException as exc:
                 _logger.error(exc)
+        elif arguments.get("add-database"):
+            try:
+                await add_database_.run(
+                    repository,
+                    database=arguments["DATABASE"],
+                    uri=arguments["--uri"],
+                    description=arguments["--description"],
+                    read_only=arguments["--read-only"],
+                    cost=arguments["--cost"]
+                )
+            except DJException as exc:
+                _logger.error(exc)
+
     except asyncio.CancelledError:
         _logger.info("Canceled")
 
--- a/src/datajunction/models/node.py
+++ b/src/datajunction/models/node.py
@@ -167,7 +167,7 @@ class Node(SQLModel, table=True):  # type: ignore
         Extra validation for node data.
         """
         if self.type == NodeType.SOURCE:
-            if self.expression:
+            if self.query:
                 raise Exception(
                     f"Node {self.name} of type source should not have an expression",
                 )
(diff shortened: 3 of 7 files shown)


src/datajunction/models/node.py
```diff
@@ -167,7 +167,7 @@ def extra_validation(self) -> None:
         Extra validation for node data.
         """
         if self.type == NodeType.SOURCE:
-            if self.expression:
+            if self.query:
```
samredai, line 170: Fixed this after noticing it was causing some test failures, looks like from a very recent change. There's a similar line change in `node_test.py`.

src/datajunction/cli/add_database.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Optional
+
+import yaml
+
+from datajunction.errors import DJError, DJException, ErrorCode
+from datajunction.models.database import Database
+
+_logger = logging.getLogger(__name__)
+
+def create_databases_dir(repository: Path):
```
betodealmeida, line 16:
```suggestion
def create_databases_dir(repository: Path) -> None:
```

src/datajunction/cli/add_database.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    database_dir_path = repository / Path("databases")
+    database_dir_path.mkdir(exist_ok=True)
+
+async def run(repository: Path, database: str, uri: str, description: Optional[str] = None, read_only: Optional[bool] = None, cost: Optional[float] = None) -> None:
+    """
+    Add a database.
+    """
+    create_databases_dir(repository)
+    db_config_path = repository / Path(f"databases/{database}.yaml")
```
betodealmeida, line 28: If you change `create_databases_dir` to return the path you can do:

```suggestion
    databases_dir = create_databases_dir(repository)
    db_config_path = (databases_dir / database).with_suffix(".yaml")
```

src/datajunction/cli/add_database.py
```diff
@@ -0,0 +1,45 @@
...
+    if Path(db_config_path).exists():
```
betodealmeida, line 29:
```suggestion
    if db_config_path.exists():
```

src/datajunction/cli/add_database.py
```diff
@@ -0,0 +1,45 @@
...
+    db_kwargs = {
+      "database": database,
+      "URI": uri,
+    }
+    if description:
+      db_kwargs["description"] = description
+    if read_only:
+      db_kwargs["read_only"] = read_only
+    if cost:
+      db_kwargs["cost"] = cost
```
betodealmeida, line 41: Use consistent indentation here, should always be 4 spaces.

src/datajunction/console.py
```diff
@@ -24,6 +32,7 @@
 
 from datajunction import __version__
 from datajunction.cli import compile as compile_
+from datajunction.cli import add_database as add_database_
```
betodealmeida, line 35: No need for the underscore here, the only reason we used in line 34 is because `compile` is a Python builtin. Conventionally we use the underscore suffix on variables that would shadow builtins or reserved keywords.

```suggestion
from datajunction.cli import add_database

File: src/datajunction/console.py [hunk 7]
```diff
@@ -54,6 +63,19 @@ async def main() -> None:
                 )
             except DJException as exc:
                 _logger.error(exc)
+        elif arguments.get("add-database"):
+            try:
+                await add_database_.run(
```
betodealmeida, line 68:
```suggestion
                await add_database.run(
```

tests/cli/add_database_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+from datajunction.cli import add_database
+from datajunction.errors import DJException
+
+@pytest.mark.asyncio
+async def test_add_database_minimum_requirements() -> None:
+    """
+    Test adding a database with the minimum required arguments
+    """
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
betodealmeida, line 17: It's better to use the `fs` fixture, it creates a fake filesystem and then you can create any directories you want. See an example here:

https://github.com/DataJunction/datajunction/blob/e90897cf7b733aef682cee5d2a59b1e94c2fa714/tests/cli/compile_test.py#L42

Using `fs` should be faster (because it simulates the disk in memory) and is usually less troublesome to run in the CI/CD IMHO.

tests/cli/add_database_test.py
```diff
@@ -0,0 +1,79 @@
...
+    Test adding a database using required and optional arguments
+    """
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        await add_database.run(
+            Path(tmpdirname),
+            database="testdb",
+            uri="testdb://test",
+            description="This is a description",
+            read_only=True,
+            cost=11.0
```
betodealmeida, line 44: Python allow trailing commas and they're great, since they allow quickly reordering lines without having to move the comma around.

```suggestion
            cost=11.0,
```

We use an autoformatter that should take care of this.

betodealmeida (changes requested):
This is awesome!

I think the tests will fail because of lint and reduced coverage. You can fix lint by running the pre-commit tests locally:

```bash
pip install pre-commit
pre-commit install
pre-commit run --all
```

You can also check the test coverage by running:

```bash
pip install pytest
pytest --cov=src/datajunction -vv tests/ --doctest-modules src/datajunction --without-integration --without-slow-integration
```

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 exclude: '^docs/conf.py'
 
 repos:
-- repo: git://github.com/pre-commit/pre-commit-hooks
+- repo: https://github.com/pre-commit/pre-commit-hooks
```
samredai, line 4: I had to change this otherwise pre-commit would just hang.

tests/cli/add_database_test.py
samredai: Updated to use this and it worked great. Thanks!

src/datajunction/models/node.py
samredai: Fixed 172 👍

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
   - id: isort
 
 - repo: https://github.com/psf/black
-  rev: 21.12b0
+  rev: 22.8.0
```
samredai, line 40: I also had to update this otherwise I was hitting [this](https://github.com/psf/black/issues/2964) issue. After this change, I was able to run `make check` without any issues.

.pre-commit-config.yaml
samredai: I do, which is why this surprised me as well. It was almost as if github was being blocked but only through this mechanism (I used `git@` when cloning this repo for example). Let me see if I can snag the error I was seeing in the log files.

samredai: Hangs here:
```
pre-commit run --all-files
[INFO] Initializing environment for git://github.com/pre-commit/pre-commit-hooks.

```
and then after a while fails with this:
```
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: unable to connect to github.com:
    github.com[0: 140.82.113.4]: errno=Operation timed out


```

tests/console_test.py
```diff
@@ -158,3 +160,117 @@ def test_interrupt(mocker: MockerFixture) -> None:
...
+    mocker.patch(
+        "datajunction.console.get_settings",
+        return_value=Settings(
+            index="sqlite:///dj.db",
+            repository=Path(test_repo),
+        ),
+    )
+
+    await console.main()
+    await console.main()  # Run a second time to log an already exists exception
```
betodealmeida, line 276: One more thing, can you add a check like here:

https://github.com/DataJunction/datajunction/blob/e90897cf7b733aef682cee5d2a59b1e94c2fa714/tests/console_test.py#L102,L117

To test that `_logger.error` was called?

betodealmeida (approved):
This looks great Sam, thanks for addressing all the comments! I left just a small ask on one of the tests.

tests/console_test.py
samredai: Nice! I was a bit self-conscious about that one there and only implicitly knew it worked because of the coverage output. But what you linked is exactly what I was looking for! I updated the PR and also had to add an `__eq__` method to the `DJException` class [here](https://github.com/DataJunction/datajunction/pull/183/files#diff-63255a15e0e2e0c88eb0ddf47824d505ce575de93fac46086936047c3137d475R172) that checks that all attributes match (errors/warning lists, HTTP status code, etc).

betodealmeida:
sagan

PR #187, opened 2022-09-15
Fix comments example config (Closes #185)

### LINKED ISSUE #185: `NoResultFound` error when running `dj compile`
When running `dj compile`, it would return a no result found error that happens when adding dimension information to the columns table.
```
sqlalchemy.exc.NoResultFound: No row was found when one was required
```

samredai:
Small fix that closes #185 which turns out was just an issue with one of the example definitions.


betodealmeida (approved):
Ah, my bad! I changed the nodes and didn't update this.

PR #190, opened 2022-09-24
Extend core examples.

agorajek:
### Summary

This is just a starting PR in this area...

Main things:
1. Reorganized core examples and added some new nodes. I split the nodes into directories by type. I'd like to run it by you and see what you say. This way may be easier for people to comprehend the node types, but because DJ allows for full flexibility here we may want to present few different ways of laying things out.
2. The first added transform node `core.etl.country_agg` seems to be loading just fine, but the metric based on this new node fails with (not sure why, no time to debug today):
```
...File "/code/src/datajunction/sql/transpile.py", line 279, in get_projection
dj                 |     raise NotImplementedError(f"Unable to handle expression: {expression}")
dj                 | NotImplementedError: Unable to handle expression: Wildcard
```
3. Added DBT template source nodes... I wanted to get some stuff out asap so I didn't finish the queries yet.

```
DAG:
  * core.dim.countries
  | * core.num_comments
  | | * core.num_countries
  | | | * dbt.jaffle_shop.customers
  | | | * dbt.jaffle_shop.orders
  | | | * dbt.stripe.payments
  | | *   core.etl.country_agg
  | | |\
  | |/ /
  |/| /
  | |/
  | * core.src.comments
  |/
  * core.dim.users
  * core.src.users
```


samredai:
This is looking great so far. As I was looking at the new file structure I started wondering why "databases" is treated as a special type of entity. Is it possible that databases can be treated as a fifth type of node meaning `type: database`?

That way we could eliminate any directory naming requirements entirely (no need for a "nodes" directory and a "databases" directory).

agorajek:
@samredai my initial reaction was to just say "nah, we don't need no database nodes" but then I started thinking... and perhaps it is not a crazy idea. From the DAG perspective it makes total sense to have a "database" node type below the "source" nodes where we can store the general properties on any given database. But the other requirement that I think we need to stick to is that every node types should have a schema defined on it and in this case we could just start as simply as this:
```
$ cat my-little-postgres-db.yaml
description: A Postgres database
URI: ...
read_only: false
cost: 10
columns:
  catalog:
    type: STR
  schema:
    type: STR
  table:
    type: STR
```

And if you submit a `SELECT * FROM my-little-postgres-db` then you could get a list of tables back that exist in this database, using the individual catalog access. This could prove to be a useful feature. Wonder what @betodealmeida thinks about all this.

agorajek:
Oh, and sorry about all the unit tests failing. I haven't gotten used to the dev cycle pattern on this repo yet.

betodealmeida:
Yeah, I don't think a database can technically be a node in the current implementation, since they don't have a schema — something shared by all other nodes. But we could store their files in the same root directory as the nodes, if the idea is to clean up some clutter.

I like the idea of enriching the SQL interface to give as much information as possible about the system. I think we talked about having something like:

```sql
SELECT name, description FROM metrics;
```

At some point we should write a spec for the SQL interface with all these ideas!

betodealmeida:
>     2. The first added transform node `core.etl.country_agg` seems to be loading just fine, but the metric based on this new node fails with (not sure why, no time to debug today):
> 
> 
> ```
> ...File "/code/src/datajunction/sql/transpile.py", line 279, in get_projection
> dj                 |     raise NotImplementedError(f"Unable to handle expression: {expression}")
> dj                 | NotImplementedError: Unable to handle expression: Wildcard
> ```

You need to add some extra logic to `datajunction.sql.transpile.get_projection` to handle the wildcard, something like:

```python
      if "UnnamedExpr" in expression:
            expression = expression["UnnamedExpr"]
        elif "ExprWithAlias" in expression:
            alias = expression["ExprWithAlias"]["alias"]["value"]
            expression = expression["ExprWithAlias"]["expr"]
        elif expression == "Wildcard":
            pass  # we can pass expression unmodified in this case
        else:
            raise NotImplementedError(f"Unable to handle expression: {expression}")
```

agorajek:
WIP: moving nodes for unitt-tests into ./tests dir.

agorajek:
TODO: I will also rename the core example dirs to match the node types:
- src --> source
- etl --> transform
- dim --> dimension

docker-compose.yml
```diff
@@ -7,10 +7,10 @@ volumes:
   broker_var: {}
   coordinator_var: {}
   router_var: {}
-  druid_shared: {}
   postgres_data: {}
   postgres_metadata: {}
   redis_data: {}
+  druid_shared: {}
```
samredai, line 13: Do we need this here for when people run both this and the druid docker-compose file?

Sort of related, when I run `docker compose up` I'm getting:
```
service "dj" depends on undefined service druid_coordinator: invalid compose project
```
We need to remove that from the `dj` depends-on block.

samredai: After removing the `druid_coordinator` it worked. At first db migration was complaining no matter how many times I removed all volumes and tore everything down. Eventually I just did `docker system prune` and then everything worked. I guess my docker environment was just generally funky.
docker

agorajek: I am not sure how to do this best. For a no-druid setup we don't need the `druid_coordinator` but for the with-druid setup we should have it. I need to try and see if I can redefine the `dj` service component in the `docker-compose-druid.yml` when running `make docker-run-with-druid`...

agorajek: It worked... oh man this PR. I am burned out :)

agorajek:
oh @betodealmeida you added a commit w/o coverage to my pr. not nice ;)

examples/configs/nodes/dbt/stripe/payments.yaml
```diff
@@ -0,0 +1,19 @@
...
+    type: DATE
+  status:
+    type: STR
+  _etl_loaded_at:
+    type: DATETIME
+tables:
+  postgres:
+  - catalog: null
+    schema: jaffle_shop
+    table: orders
```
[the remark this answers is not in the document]
agorajek: Thanks @samredai 
nice_catch

agorajek:
This PR now resolves:
- issue 161
- issue 192

But it creates #200

agorajek:
Let's merge it in...

samredai (approved):

CONTRIBUTING.rst
```diff
@@ -259,32 +242,33 @@ To get data for a given metric:
...
 
 .. code-block:: sql
 
-    SELECT "core.num_comments"
+    SELECT "basic.num_comments"
     FROM metrics
-    WHERE "core.comments.user_id" < 4
-    GROUP BY "core.comments.user_id"
+    WHERE "basic.source.comments.user_id" < 4
+    GROUP BY "basic.source.comments.user_id"
```
betodealmeida, line 271: Thank you for keeping the docs up-to-date!

docker-compose.yml
```diff
@@ -7,11 +7,10 @@ volumes:
   broker_var: {}
   coordinator_var: {}
   router_var: {}
-  druid_shared: {}
   postgres_data: {}
   postgres_metadata: {}
   redis_data: {}
-
+  druid_shared: {}
```
betodealmeida, line 13: Do we need this here or could we add it to `docker-compose-druid.yml` instead?

examples/configs/nodes/basic/transform/country_agg.yaml
```diff
@@ -0,0 +1,14 @@
+description: Country level agg table
+type: transform
+query: |
+  SELECT country, 
+    -- COUNT(1) as num_comments,
```
betodealmeida, line 5: Should we get rid of these comments so it's not confusing?

examples/configs/nodes/dbt/source/stripe/payments.yaml
```diff
@@ -0,0 +1,23 @@
...
+  created:
+    type: DATE
+  _batched_at:
+    type: DATETIME
+tables:
+  postgres:
+  - catalog: null
+    schema: stripe
+    table: payments
+    cost: 10.0
```
betodealmeida, line 23: Maybe we could remove `cost` when there's just a single table? Since it's inherited from the database cost?

examples/configs/nodes/dbt/transform/customer_agg.yaml
```diff
@@ -0,0 +1,74 @@
+description: Country level agg table
+type: transform
+query: "SELECT \n    \"dbt.source.jaffle_shop.customers\".id\n    , \"dbt.source.jaffle_shop.customers\"\
```
betodealmeida, line 3: Yeah... the auto-formatting of SQL queries will be nice to have! 🙂

tests/sql/transpile_test.py
```diff
@@ -700,3 +700,14 @@ def test_unary_op() -> None:
...
     assert get_expression(expression, source) == -2.0
+
+
+def test_wildcard_projection_unimplemented() -> None:
+    """
+    Test for projection with a wildcard.
+    """
+    database = Database(id=1, name="sqlite", URI="sqlite://")
+    tree = parse_sql("SELECT * FROM a", dialect="ansi")
+    with pytest.raises(Exception) as excinfo:
```
betodealmeida, line 711:
```suggestion
    with pytest.raises(NotImplementedError) as excinfo:
```

tox.ini
```diff
@@ -10,5 +10,5 @@ deps =
     coverage
 commands =
     pip install -e .[testing]
-    coverage run --source src --parallel-mode -m pytest {posargs}
-    coverage html --fail-under 100 -d test-reports/{envname}/coverage-html
+    coverage run --source src --parallel-mode -m pytest {posargs} --without-integration --without-slow-integration
+    coverage html --fail-under 100 -d test-reports/{envname}/coverage-html
```
betodealmeida, line 14:
```suggestion
    coverage html --fail-under 100 -d test-reports/{envname}/coverage-html
```

betodealmeida (approved):

PR #203, opened 2022-10-06
fix: do not create tables on app start

betodealmeida:
Remove `create_db_and_tables` from app initialization.

Before we introduced Alembic migrations in https://github.com/DataJunction/datajunction/pull/151 tables for models were created by calling `create_db_and_tables` on app initialization. This is an idempotent call, so it was fine to call it every time.

With Alembic, tables are now created by a short-lived docker container:

https://github.com/DataJunction/datajunction/blob/973fddd54d3a40f09da4e567c450a5041619a612/docker-compose.yml#L62

This resulted in two processes creating tables, and depending on the order they first run the DB might be in a bad state:

1. If Alembic runs first it will create the tables and stamp the DB with the SHA for `head`. Then `create_db_and_tables` runs and does nothing (it's idempotent and the tables already exist).
2. If `create_db_and_tables` runs first, the tables are created but there's no Alembic stamp. When Alembic runs it things that the DB is not initialized (since there's no stamp), and tries to run the initial migration, which fails.

Turns out we're not using `create_db_and_tables` anywhere, so I just removed it. I also updated the docs, instructing people that want to run the DJ app directly (ie, not with Docker) that they have to run `alembic upgrade head` in order to create the tables.

The pre-commit hook also fixed a lot of lint.


betodealmeida:
Closes https://github.com/DataJunction/datajunction/issues/196.

agorajek (approved):
Looks good... but how do I setup these pre-commit hooks again. I thought I already did, but it doesn't look like it.

betodealmeida:
> Looks good... but how do I setup these pre-commit hooks again. I thought I already did, but it doesn't look like it.

I need to setup them up a part of CI/CD, created https://github.com/DataJunction/datajunction/issues/204.

PR #205, opened 2022-10-07
fix: run `pip-compile-multi` for strawberry

betodealmeida:
Fix new deps introduced in https://github.com/DataJunction/dj/pull/199 by running `pip-compile-multi`.


samredai (approved):
LGTM! I think we should add a section to the [Contributing](https://github.com/DataJunction/dj/blob/main/CONTRIBUTING.rst) doc that describes how `pip-compile-multi` is supposed to be used. I figured out how it works by looking at the comments at the top of the requirements text files but even after reading that I'm still a bit fuzzy on when exactly I should run it and how things like runtime dependencies are determined vs. test dependencies.

PR #206, opened 2022-10-08
s/datajunction/dj

agorajek:
This is a "simple" diff to rename all occurrences of "datajunction" to "dj". Two things to note:
1. `src/dj/fixes.py` lost coverage on line 13 so I move the `PYDRUID_INSTALLED = True` before the import statement
2. I tried and failed to also rename the `./src/dj` as `./dj`. Node compile tests started failing and I couldn't see the fix.

We can attempt #2 again during the sync or you can try on your own if you want to scratch that itch.

PR #207, opened 2022-10-09
chore: remove `src/` directory

betodealmeida:
As a follow up to https://github.com/DataJunction/dj/pull/206, this PR removes the `src/` directory.

@agorajek @samredai @CircArgs: you'll probably need to run `pip install -e .` again after this PR.


samredai (approved):
🙌

PR #208, opened 2022-10-09
chore: fix lint and enable it in CI

betodealmeida:
This PR fixes all the lint complaints and enables link checks on pull requests.

I also changed the shadowed GraphQL classes to use the Python convention of underscore suffix, instead of prefix.


betodealmeida:
Closes #204.

PR #210, opened 2022-10-10
Fix installation issues after the renames.

agorajek:
1. One down. `pip install .` and `dj` cli work ok now.
2. Still another issue left:
```
dj                 |   File "/usr/local/lib/python3.10/site-packages/uvicorn/supervisors/watchfilesreload.py", line 75, in __init__
dj                 |     self.watcher = watch(
dj                 | TypeError: watch() got an unexpected keyword argument 'yield_on_timeout'
```

PR #211, opened 2022-10-10
Pin uvicorn to 0.17.6

samredai:
This fixes the second issue mentioned in PR #210 by reverting the minor version upgrade of uvicorn, and pinning it to `0.17.6`. Upgrading uvicorn to `0.18.x` seems to cause the following error in the docker environment.
```py
TypeError: watch() got an unexpected keyword argument 'yield_on_timeout'
```


betodealmeida (approved):
Thanks!

PR #213, opened 2022-10-12
Auto format SQL query on every node update.

agorajek:
### Summary
Finally... pretty print all the SQL queries in the nodes, so we can enjoy looking at the configs :)

### Test Plan
Unit tests plus compiled all the current examples.


agorajek:
OK, I see all the pre-commit issues. Will fix.

betodealmeida (approved):
THANK YOU! 🕺🏼

Issue #225, opened 2022-10-14
Use SQLite for the standard docker compose setup


samredai:
The *docker-compose.yml* that's used by default when running `docker compose up` should provide the simplest DJ setup possible to allow for easier demos and core development. So instead of including a postgres container and storing the metadata there, it should rely on the default behavior of DJ to use SQLite if no `INDEX` is set in the .env file.

A postgres backed DJ server should still be optionally available using a docker compose override.

*Start a DJ server that's backed by SQLite*
```
docker compose up
```

*Start a DJ server that's backed by Postgres*
```
docker compose -f docker-compose.yml -f docker-compose.postgres.yml up
```

betodealmeida:
Implemented in #226.

PR #226, opened 2022-10-14
Use SQLite by default

samredai:
### Summary

Instead of using Postgres by default, this changes the default `docker compose up` environment to not use any `INDEX` value and not launch any `postgres_metadata` container. DJ will then fall back and use SQLite by default.

Using a postgres container to store the metadata (the old behavior) can still be achieved using an override that adds the `postgres_metadata` container and uses a DJ env file that includes an `INDEX` value that points to the container.

*To use postgres*:
```
docker compose -f docker-compose.yml -f docker-compose.postgres.yml up
```

### Test Plan

Ran `docker compose up` and ran queries through curl as well as the graphql API.

Also ran `docker compose -f docker-compose.yml -f docker-compose.postgres.yml up` and ran the same queries.

- [x] PR has an associated issue: #225 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/docker-compose.cockroachdb.yml
+++ b/docker-compose.cockroachdb.yml
@@ -1,4 +1,4 @@
-# WARNING: this is not a standaonle docker config file. It must be used in addition to the docker-compose.yml.
+# WARNING: this is not a standalone docker config file. It must be used in addition to the docker-compose.yml.
 #          If in doubt please use the Makefile commands to build DJ docker containers.
 version: "2.2"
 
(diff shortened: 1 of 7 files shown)


samredai:
cc: @CircArgs

docker-compose.cockroachdb.yml
```diff
@@ -1,3 +1,5 @@
+# WARNING: this is not a standaonle docker config file. It must be used in addition to the docker-compose.yml.
```
betodealmeida, line 1:
```suggestion
# WARNING: this is not a standalone docker config file. It must be used in addition to the docker-compose.yml.
```

docker-compose.postgres.yml
```diff
@@ -0,0 +1,40 @@
+# WARNING: this is not a standaonle docker config file. It must be used in addition to the docker-compose.yml.
```
betodealmeida, line 1:
```suggestion
# WARNING: this is not a standalone docker config file. It must be used in addition to the docker-compose.yml.
```

betodealmeida (approved):
So cool!

Issue #124, opened 2022-04-17
DQL: Add support for `GROUP BY 1`


betodealmeida:
Currently we can only group by names:

```sql
SELECT "core.users.preferred_language", "core.num_comments"
FROM metrics
GROUP BY "core.users.preferred_language"
```

We should be able to group by column indexes, ie:

```sql
SELECT "core.users.preferred_language", "core.num_comments"
FROM metrics
GROUP BY 1
```

betodealmeida:
A good way to test this is adding a variation of this test:

https://github.com/DataJunction/datajunction/blob/43df656853b410d6b02aa43c0fc6f8025cb7d5af/tests/sql/build_test.py#L1367-L1371

But replacing the SQL with:

```sql
SELECT "core.num_comments", "core.comments.user_id" FROM metrics
WHERE "core.comments.user_id" > 1
GROUP BY 2
```

The produced SQL should be identical to the original test:

```sql
SELECT count('*') AS "core.num_comments", "core.comments".user_id{space}
FROM (SELECT comments.user_id AS user_id, comments.comment AS comment{space}
FROM comments) AS "core.comments"{space}
WHERE "core.comments".user_id > 1 GROUP BY "core.comments".user_id
```

PR #233, opened 2022-10-21
Add support for GROUP BY using positions

samredai:
### Summary

This adds support for using positions in a `GROUP BY` clause. For example:

```sql
  SELECT country,
         COUNT(DISTINCT id) AS num_users
  FROM basic.source.users
  GROUP BY 1
```

### Test Plan

I added a test `test_get_query_for_sql_where_groupby_num` and updated the `country_agg.yaml` example node to use
`GROUP BY 1` instead of `GROUP BY country`, then confirmed that the same behavior remains.

- [x] PR has an associated issue: #124 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/transpile.py
+++ b/dj/sql/transpile.py
@@ -180,6 +180,30 @@ def get_orderby(
     return expressions
 
 
+def get_resolved_num_groupby(
+    tree: ParseTree,
+    groupby: List[Expression],
+) -> List[Expression]:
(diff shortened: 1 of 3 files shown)


samredai:
cc: @CircArgs

betodealmeida (approved):
This is awesome, @samredai !

Issue #234, opened 2022-10-23
Commit the OpenAPI spec to the repo


samredai:
Currently, the OpenAPI spec isn't committed to the repo. In order to see the spec, you have to start the server locally and go to the swagger `/openapi.json` URL. 

It would be convenient to have the spec in the repo, for example, to copy it into the online [Swagger Editor](https://editor.swagger.io/) as a discussion aid or even just to explore the file to understand the request and response models.

PR #235, opened 2022-10-23
Manually dispatched workflow to commit openapi spec

samredai:
### Summary

This PR adds a workflow that imports the DJ app and outputs the OpenAPI spec (which is retrieved by calling `app.openapi()`) to a file called  `openapi.json` that's then committed to the root of the repository. The workflow does not run automatically and for now, I think it's ok that we just manually run it whenever we update any of the API routes.

### Test Plan

I added the workflow to a personal repo and ran it a few times.

- [x] PR has an associated issue: #234 
- [x] `make check` passes

### Deployment Plan

The workflow will only run when triggered manually.

### DIFF
--- a/scripts/generate-openapi.py
+++ b/scripts/generate-openapi.py
@@ -4,7 +4,7 @@
 import argparse
 import json
 
-from dj.api.query.engine import app
+from dj.api.main import app
 
 
 def save_openapi_spec(f: str):
--- a/scripts/generate-openapi.py
+++ b/scripts/generate-openapi.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python3
+# pylint: skip-file
+
+import argparse
+import json
+
+from dj.api.query.engine import app
+
+
+def save_openapi_spec(f: str):
+    spec = app.openapi()
+    with open(f, "w") as outfile:
+        outfile.write(json.dumps(spec, indent=4))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate a file containing the OpenAPI spec for a DJ server",
+    )
+    parser.add_argument(
+        "-o",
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Nice! This should be in `.github/workflows` if we want to run it manually from the GitHub page, no?

commit-openapi-spec.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Generate OpenAPI Spec
+        run: |
+          python -c "
+          import json
+          from dj.api.query.engine import app
+
+          spec = app.openapi()
+          with open('openapi.json', 'w') as outfile:
+              outfile.write(json.dumps(spec, indent=4))
```
betodealmeida, line 27: Can we make this a script in dj/scripts/`, so we can run it manually if needed?

samredai: Great idea! I added `scripts/generate-openapi.py` and call that in the workflow.

samredai:
> Nice! This should be in `.github/workflows` if we want to run it manually from the GitHub page, no?

Whoops! Thanks for catching that 😅 (fixed)

scripts/generate-openapi.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+# pylint: skip-file
+
+import argparse
+import json
+
+from dj.api.query.engine import app
```
CircArgs, line 7: Should this be `dj.api.main`?

CircArgs (approved):
Just the one question 😊

scripts/generate-openapi.py
samredai: Oh right, thanks for catching this. I think my IDE did something funny and changed this. 😨 (fixed!)

PR #242, opened 2022-11-04
DJ SQL AST with a swappable sqloxide backend

CircArgs:
### Summary

Creating an AST structure for internal DJ use.
It includes code to use `sqloxide` as a backend parser and a 'frontend' for generating a sql string

### Test Plan
Unit testing until reaching 100% + manual testing

- [x] PR has an associated issue: #245 
- [x] `make check` passes

## Examples:

```python
from dj.sql.parsing.ast import *
from dj.sql.parsing.backends.sqloxide import parse
from dj.sql.parsing.frontends.sql_string import sql
```

```python
q = parse(
    """
select a.x as "x" from 'a' as z left join b on a.id = b.id
"""
)
```

```python
print(sql(q))
```

    SELECT a.x AS "x"
    FROM 'a' AS z
    LEFT JOIN b
            ON a.id = b.id

```python
q
```

    Query(ctes=[], select=Select(distinct=False, from_=From(table=Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")), joins=[Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))]), group_by=[], having=None, projection=[Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None))], where=None, limit=None))

```python
list(q.find_all(Named))
```

    [Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")),
     Table(name='a', quote_style="'"),
     Table(name='b', quote_style=None),
     Column(name='id', quote_style=None),
     Column(name='id', quote_style=None),
     Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None)),
     Column(name='x', quote_style=None)]

```python
print(sql(q))
```

    SELECT a.x AS "x"
    FROM 'a' AS z
    LEFT JOIN b
            ON a.id = b.id

```python
x_nodes = list(q.filter(lambda node: isinstance(node, Named) and node.name == "x"))
```

```python
x_nodes
```

    [Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None)),
     Column(name='x', quote_style=None)]

```python
def rename_nodes(node):
    if isinstance(node, Column):
        node.name = "new_name"
```

```python
q.apply(rename_nodes)
```

```python
print(sql(q))
```

    SELECT a.new_name AS "x"
    FROM 'a' AS z
    LEFT JOIN b
            ON a.new_name = b.new_name

```python
print(
    sql(
        parse(
            """with a as (select * from a),
b10 as (select * from b)

select a.x from a"""
        )
    )
)
```

    WITH
    a AS (SELECT *
    FROM a),
    b10 AS (SELECT *
    FROM b)
    
    SELECT a.x
    FROM a

```python
q = parse(
    """
select t.x as "x" from 'a' as z left join b on a.id=b.id
"""
)
```

```python
parse(sql(q))
```

    Query(ctes=[], select=Select(distinct=False, from_=From(table=Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")), joins=[Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))]), group_by=[], having=None, projection=[Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None))], where=None, limit=None))

```python
q.apply(print)
```

    Query(ctes=[], select=Select(distinct=False, from_=From(table=Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")), joins=[Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))]), group_by=[], having=None, projection=[Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None))], where=None, limit=None))
    Select(distinct=False, from_=From(table=Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")), joins=[Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))]), group_by=[], having=None, projection=[Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None))], where=None, limit=None)
    From(table=Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'")), joins=[Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))])
    Alias(name='z', quote_style=None, child=Table(name='a', quote_style="'"))
    Table(name='a', quote_style="'")
    Join(kind=<JoinKind.LeftOuter: 'LEFT JOIN'>, table=Table(name='b', quote_style=None), on=BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None)))
    Table(name='b', quote_style=None)
    BinaryOp(left=Column(name='id', quote_style=None), op=<BinaryOpKind.Eq: '='>, right=Column(name='id', quote_style=None))
    Column(name='id', quote_style=None)
    Column(name='id', quote_style=None)
    Alias(name='x', quote_style='"', child=Column(name='x', quote_style=None))
    Column(name='x', quote_style=None)

```python

```

## Timings

```
sqloxide : 35.8 µs
dj ast : +113 µs
sqlglot : 663 µs
```

## Similarity demo

```python
def query_similarity(q1: str, q2: str) -> float:
    """demo of an extremely primitive query similarity"""
    ast1 = parse(q1)[0]
    ast2 = parse(q2)[0]
    g, h = list(ast1.flatten()), list(ast2.flatten())
    i = [i for i in g if i in h]
    return round(len(i) / (len(g) + len(h) - len(i)), 3)
```

```python
query_similarity("select x from b", "select x from a")
```
    0.429


dj/sql/parsing/frontends/sql_string.py
```diff
@@ -0,0 +1,132 @@
...
+@sql.register
+def _(node: Select) -> str:
+    projection = ",\n\t".join(sql(exp) for exp in node.projection)
+    return f"""SELECT {"DISTINCT " if node.distinct else ""}{projection}
+{sql(node.from_)}
+{"WHERE "+sql(node.where) if node.where is not None else ""}
+{"GROUP BY "+", ".join(sql(exp) for exp in node.group_by) if node.group_by else ""}
+{"HAVING "+sql(node.having) if node.having is not None else ""}
+{"LIMIT "+sql(node.limit) if node.limit is not None else ""}
+""".strip()
```
betodealmeida, line 118: This is not super easy to read, it might be clearer if you build it in parts, eg:

```python
parts = ["SELECT"]
if node.distinct:
    parts.append("DISTINCT")
parts.extend((projection, sql(node.from_))
...
return " ".join(parts)
```

CircArgs: Thanks I like that! probably can be used a couple of other places too

samredai:
@CircArgs the flake8 fix was merged so if you rebase this with main, the tests shouldn't get that gitlab error anymore.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+    op: UnaryOpKind  # pylint: disable=C0103
+    expr: Expression
+
+    def __hash__(self) -> int:
+        return hash((UnaryOp, self.op))
+
+
+class BinaryOpKind(Enum):
+    """the accepted binary operations"""
+
```
samredai, line 235: It doesn't matter much but I think you can do `# pylint: disable=C0103` on this line and then `# pylint: enable=C0103` on line 252 to apply it over this range of lines.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+@dataclass(eq=False)  # type: ignore
+class Value(Expression):
+    """base class for all values number, string, boolean"""
+
+    value: Union[str, bool, float, int]
+
+
+@dataclass(eq=False)
+class Number(Value):  # Number
```
samredai, line 311: super nit: The comment doesn't seem needed here or for the next couple of classes.

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -0,0 +1,391 @@
...
+    UnaryOpKind,
+    Value,
+    Wildcard,
+)
+from dj.sql.parsing.backends.exceptions import DJParseException
+
+
+def match_keys(parse_tree: dict, *keys: Set[str]) -> bool:
+    """
+    match a dictionary having exact keys
```
samredai, line 37: nit: I would either make the docstring more specific (`match a parse tree having exact keys`) or make the argument name more general (`d: dict`). Same for the next function.

tests/conftest.py
```diff
@@ -18,6 +18,8 @@
 from dj.config import Settings
 from dj.utils import get_project_repository, get_session, get_settings
 
+from .sql.parsing.queries import *  # pylint: disable=W0401,W0614
```
samredai, line 21: Imo we should list out all of the imports explicitly here and accept the overhead of having to update it as we add more queries. It makes it easier to figure out what's happening when trying to understand the query fixtures for the first time.

samredai (approved):
I left a number of comments but I don't think any of them are blocking-worthy so I approved it. This is great and awesome job on the coverage!

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -0,0 +1,391 @@
+"""
```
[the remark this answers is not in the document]
samredai: Discussed in the weekly sync: `backends` contain logic for converting to a DJ AST representation and `frontends` contain logic for converting from a DJ AST.

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -0,0 +1,391 @@
...
+parsing backend turing sqloxide output into DJ AST
```
agorajek, line 2: s/turing/turning

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -0,0 +1,391 @@
...
+    """
+    return set(parse_tree.keys()) in keys
+
+
+def match_keys_subset(parse_tree: dict, *keys: Set[str]) -> bool:
+    """
+    match a dict having a subset of keys
+    """
+    tree_keys = set(parse_tree.keys())
+    return any(key <= tree_keys for key in keys)  # pragma: no cover
```
agorajek, line 47: Why no unit test coverage here? I can understand the Exception cases below, but not sure about this one.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+
+    def __hash__(self) -> int:
+        return hash((Join, self.kind))
+
+
+@dataclass(eq=False)
+class From(Node):
+    """a from that belongs to a select"""
+
+    table: Union[Table, Alias]
```
agorajek, line 459: Do we currently support an aliased Query here? e.g.:
```
SELECT t1.a, t2.b FROM (SELECT * FROM t) t1 JOIN t2 ON t1.c = t2.c
```

dj/sql/parsing/frontends/sql_string.py
```diff
@@ -0,0 +1,139 @@
...
+
+@sql.register
+def _(node: Query) -> str:
+
+    ctes = ",\n".join(f"{cte.name} AS ({sql(cte.child)})" for cte in node.ctes)
+    return (
+        f"""{'WITH' if ctes else ""}
+{ctes}
+
+{("(" if node.subquery else "")+sql(node.select)+(")" if node.subquery else "")}
```
agorajek, line 136:
```suggestion
{"("+sql(node.select)+")" if node.subquery else sql(node.select)}
```

agorajek (commented):
This is amazing. I had a few comments in-line, but most of them minor. I'll defer the go-ahead to @betodealmeida . Thanks for coding this up @CircArgs!

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Type,
+    TypeVar,
+    Union,
+)
+
+from dj.utils import flatten
```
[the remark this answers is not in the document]
CircArgs: Hmm I originally had it in the file actually. I think I made it slightly more generic and then moved it out, but you're right it's only used here so I'll move it back.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+    @property
+    def parents(self) -> Set["Node"]:
+        """
+        get the parents of the node
+        """
+        try:
+            return self._parents  # type: ignore
+        except AttributeError:
+            self._parents = set()
+            return self._parents
```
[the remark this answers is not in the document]
CircArgs: I think I did that because Node isn't a dataclass so I can't use `set` in `field` like I would for other empty defaults. I think it's bad practice to initialize a set (or any reference type) as a class attribute? I also cannot use an init since dataclasses won't respect it. Here's an example:

```python
class Node(ABC):
    _parents: set = set()
    def __init__(self):
        self._parents = set()
        
@dataclass
class A(Node):
    name: str
        
@dataclass
class B(Node):
    value: int

a=A("yo")
a._parents.add(5)

b=B(6)

b._parents #{5}
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+
+    def add_parents(self, *parents: "Node") -> "Node":
+        """
+        add parents to the node
+        """
+        for parent in parents:
+            self.parents.add(parent)
+        return self
+
+    def add_self_as_parent(self) -> "Node":
```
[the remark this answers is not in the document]
CircArgs: Yeah right now this is causing some ugliness too with a bunch of `cast`s I'm doing for mypy right now so I'd like other solutions. What happens in the backend (ofc only sqloxide right now) is that when a node wraps another it calls the `add_self_as_parent` method which iterates through all the children and makes the wrapping node the parent of all the nodes it is wrapping (direct descendants only also through iterators - if in list, etc). `add_parents` could be used when manipulating the ast I figured so it is separate.

dj/sql/parsing/ast.py
CircArgs: how can you be sure the class represents a Number without that comment?

dj/sql/parsing/ast.py
CircArgs: yeah much cleaner. Thank you!

dj/sql/parsing/backends/sqloxide.py
CircArgs: great question. I think the coverage is confused here. That sounds a little crazy but I'll try to explain what I think is happening and hopefully you can correct me. (and @samredai  explained some of what coverage might be thinking here as well so perhaps he can add).

The coverage missing statement was "47->exit" suggesting the function may exit early at that line. It's not clear there is any branching. The goal with this line was to allow for early stopping since the `any` can lazily go through the generator comprehension and just stop at the first `True`. As @samredai suggested to me, the coverage may be taking into account the "early stopping branching" logic. I ignored coverage here as if that's the case it seemed irrelevant.

dj/sql/parsing/ast.py
CircArgs: we didn't but this felt big enough to make sure it's here from the start so I've added it

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,491 @@
...
+        return id(self)
+
+
+@dataclass(eq=False)
+class Query(Expression):
+    """overarching query type"""
+
+    ctes: List[Alias["Select"]]
+    select: "Select"
+    subquery: bool
```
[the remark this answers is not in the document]
CircArgs: I've removed this and now just check for parents where relevant (string frontend)

dj/sql/parsing/backends/sqloxide.py
agorajek: That makes sense. I haven't seen this kind of case in a while :)

dj/sql/parsing/ast.py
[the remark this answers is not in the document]
agorajek: oh I see. I don't have a good suggestion for you then. @betodealmeida may...

agorajek (approved):
It looks like you have one `fix end of files` issue. Other than that I think this is a go!

dj/sql/parsing/backends/sqloxide.py
samredai: Yeah I'm pretty sure that's exactly what's happening here since if you cast the generator to a tuple, causing the generator to iterate to completion, the branching coverage miss goes away. I learned something new here, coverage is much smarter than I thought!

betodealmeida:
I wanted to review today but didn't have time, sorry! I'll take a look tomorrow morning!

samredai:
The PR that's auto-generated to add the openapi spec is failing the tests because it doesn't have a new line. That's why it's triggering a `fix end of line` failure here. I opened PR #248 which adds `openapi.json` to the ignore list. Once that's merged, this PR can be rebased. Sorry for the noise!

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+def flatten(maybe_iterables: Any) -> Iterator:
+    """
+    flattens `maybe_iterables` by descending into items that are Iterable
+    """
+
+    def single_item_generator(item: Any) -> Generator[Any, None, None]:
+        yield item
+
+    if not isinstance(maybe_iterables, (list, tuple, set, Iterator)):
+        return single_item_generator(maybe_iterables)
```
betodealmeida, line 32: `single_item_generator` could be implemented using the `iter()` builtin:

```suggestion
    if not isinstance(maybe_iterables, (list, tuple, set, Iterator)):
        return iter([maybe_iterables])
```

dj/sql/parsing/ast.py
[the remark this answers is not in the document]
betodealmeida: The example above works if you use [`__post_init__`](https://docs.python.org/3/library/dataclasses.html#post-init-processing) instead:

```python
-class Node(ABC):
      _parents: set = set()

      def __post_init__(self):
          self._parents = set()
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+
+        if nodes_only:
+            child_generator = iter(
+                filter(
+                    lambda child: isinstance(child, Node),
+                    child_generator,
+                ),
+            )
+
+        if nones:
```
betodealmeida, line 123:
```suggestion
        if not nones:
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+
+        Returns:
+            Iterator: returns all children of a node given filters
+                and optional flattening (by default Iterator[Node])
+        """
+        child_generator = iter(
+            self.__dict__[field.name]
+            for field in fields(self)
+            if (not field.name.startswith("_") if not obfuscated else True)
+        )  # exclude obfuscated fields if `obfuscated` is True
```
betodealmeida, line 111: I'd split this to increase readability a little bit, and you don't need the `iter()`:

```suggestion
        child_generator = (self.__dict__[field.name] for field in fields(self))
        if not obfuscated:
            child_generator = (field for field in child_generator if not field.name.startswith("_"))
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        if flat:
+            child_generator = iter(flatten(child_generator))
```
betodealmeida, line 113: `flatten` already returns an iterable, no need to `iter()` here:

```suggestion
            child_generator = flatten(child_generator)
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        except AttributeError:
+            self._parents = set()
+            return self._parents
+
+    def add_parents(self, *parents: "Node") -> "Node":
+        """
+        add parents to the node
+        """
+        for parent in parents:
+            self.parents.add(parent)
```
betodealmeida, line 70:
```suggestion
        self.parents.update(parents)
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        child_generator = iter(
+            self.__dict__[field.name]
+            for field in fields(self)
+            if (not field.name.startswith("_") if not obfuscated else True)
+        )  # exclude obfuscated fields if `obfuscated` is True
+        if flat:
+            child_generator = iter(flatten(child_generator))
+
+        if nodes_only:
+            child_generator = iter(
```
betodealmeida, line 116: No need for `iter()` here.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+                filter(
+                    lambda child: isinstance(child, Node),
+                    child_generator,
+                ),
+            )
+
+        if nones:
+            child_generator = iter(
```
betodealmeida, line 124: No need for `iter()`.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        """
+        return self.filter(lambda _: True)
+
+    def fields(
+        self,
+        flat: bool = True,
+        nodes_only: bool = True,
+        obfuscated: bool = False,
+        nones: bool = False,
+    ) -> Iterator:
```
betodealmeida, line 93: Can you be more specific with the `Iterator` type here?

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        return self.fields(flat=True, nodes_only=True, obfuscated=False, nones=False)
+
+    def filter(self, func: Callable[["Node"], bool]) -> Iterator["Node"]:
+        """
+        find all nodes that `func` returns `True` for
+        """
+        if func(self):
+            yield self
+        for node in chain(*[child.filter(func) for child in self.children]):
+            yield node
```
betodealmeida, line 145:
```suggestion
        yield from chain(*[child.filter(func) for child in self.children])
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        """
+        func(self)
+        for child in self.children:
+            child.apply(func)
+
+    def compare(self, other: "Node") -> bool:
+        """
+        compare two ASTs
+        """
+
```
betodealmeida, line 165:
```suggestion
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        primitives = {int, float, str, bool, type(None)}
+        return type(self) == type(other) and all(  # pylint: disable=C0123
+            s == o
+            if type(s) in primitives  # pylint: disable=C0123
+            else type(s) == type(o)  # pylint: disable=C0123
+            for s, o in zip(
+                (self.fields(False, False, False, True)),
+                (other.fields(False, False, False, True)),
+            )
+        )
```
betodealmeida, line 187: I'd also break this down to make it easier to understand:

```python
        if type(self) != type(other):
            return False
            
        for self_field, other_field in zip(...):
            if type(self_field) in primitives:
                # compare by value
                if self_field != other_field:
                    return False
            else:
                # compare by type
                if type(self_field) != type(other_field):
                    return False

        return True
```

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+
+    name: str
+    quote_style: Optional[str]
+
+    @property
+    def quoted_name(self) -> str:
+        """
+        get the name of the Named Node including the quotes if any
+        """
+        return f'{self.quote_style if self.quote_style else ""}{self.name}{self.quote_style if self.quote_style else ""}'  # pylint: disable=C0301
```
betodealmeida, line 212: You can default `quote_style` to an empty string, then you can always use it:

```python
@dataclass(eq=False)  # type: ignore
class Named(Expression):
    """An Expression that has a name"""

    name: str
    quote_style: str = ""
```

I'd also create a `quoted_name` property to simplify the logic:

```python
@property
def quoted_name(self) -> str:
    return f"{self.quote_style}{self.name}{self.quote_style}"
```

@agorajek do we need to worry about asymetrical quoting, eg:

```
`foo'
```

Ie, quoting where the opening quote is different than the closing one.

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+
+    def __hash__(self) -> int:
+        return hash(Function)
+
+
+@dataclass(eq=False)  # type: ignore
+class Value(Expression):
+    """base class for all values number, string, boolean"""
+
+    value: Union[str, bool, float, int]
```
betodealmeida, line 330: Can this be `None`?

dj/sql/parsing/ast.py
```diff
@@ -0,0 +1,517 @@
...
+        return f'{self.quote_style if self.quote_style else ""}{self.name}{self.quote_style if self.quote_style else ""}'  # pylint: disable=C0301
+
+    def alias_or_name(self) -> str:
+        """
+        get the name or alias of the node
+        """
+        if len(self.parents) == 1:
+            parent = tuple(self.parents)[0]
+            if isinstance(parent, Alias):
+                return parent.name
```
betodealmeida, line 221: Why do we return the parent name when there's a single parent?

dj/sql/parsing/frontends/sql_string.py
```diff
@@ -0,0 +1,138 @@
...
+def sql(node: Any) -> str:
+    """
+    return the ansi sql representing the sub-ast
+    """
+    raise Exception("Can only convert specific Node types to a sql string")
+
+
+@sql.register
+def _(node: UnaryOp) -> str:
+    return f"{node.op.value} {sql(node.expr)}"
```
betodealmeida, line 38: I think in your first approach you had `sql` implemented as a method of each node, instead of dispatcher... any reason for the change? We could implement it as `__str__` method of each node, which would simplify a lot of the logic here.

dj/sql/parsing/frontends/sql_string.py
```diff
@@ -0,0 +1,138 @@
...
+def _(node: Query) -> str:
+    subquery = bool(node.parents)
+    ctes = ",\n".join(f"{cte.name} AS ({sql(cte.child)})" for cte in node.ctes)
+    return (
+        f"""{'WITH' if ctes else ""}
+{ctes}
+{"("+sql(node.select)+")"  if subquery else sql(node.select)}
+    """.strip()
+        + "\n"
+    )
```
betodealmeida, line 138: Might be a small nit, but I find it really hard to read logic inside f-strings, I think it's clearer to have it outside the formatting:

```python
with_ = "WITH" if ctes else ""
select = f"({sql(node.select)})" if subquery else sql(node.select)
return f"""
    {with_} 
    {ctes}
    {select}
"""
```

betodealmeida (changes requested):
This looks great! I left a few comments on Python idioms and how to simplify some of the logic, and I think there's a place where @samredai pointed out that the logic is inverted.

dj/sql/parsing/ast.py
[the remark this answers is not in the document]
CircArgs: Clever I hadn't thought to use it on the Node class!

dj/sql/parsing/ast.py
CircArgs: Yeah, but I think mypy complains about the type changing as that variable gets reassigned so instead of `cast` or something I wrapper it in `iter`

dj/sql/parsing/ast.py
CircArgs: Same thing I was tryig to address in the other spot where it was mypy complaining

dj/sql/parsing/ast.py
CircArgs: This is borrowed from sqlglot that it is useful to have a method return the alias name if an alias exists otherwise return the node's name so this should only return the parent name if there is a single alias node as the parent

dj/sql/parsing/frontends/sql_string.py
CircArgs: First, it was just `sql` as you pointed out which mimics sqlglot behavior. Then, with the advent of that frontend/backend separation the reasoning was `something -> ast` = backend and `ast -> something` = frontend and `ast -> string` felt like it fit the frontend pattern. Then, using dispatch just made sense to me as a clean way to implement it (I probably would have done the same thing if it were possible to match on typedict for the sqloxide backend which would have been super clean imo). Does all of that make sense? I'm perfectly happy to have it as `__str__` instead like you said that would clean up a lot

samredai: @betodealmeida would the `__str__` representation of a node be more than just the SQL string fragment? I'd expect it would also contain some of the other attributes of the node (maybe).

Imo, even though it's nice to be able to just look at a concrete node implementation to see how it's casted to a string, it's also nice to look in one place to see all of the casting logic. That also makes it easy to create another "frontend"--you can just copy sql_string.py and edit from there.

dj/sql/parsing/frontends/sql_string.py
CircArgs: completely agree I think making ugly f-strings in in my dna

dj/sql/parsing/ast.py
samredai: I'm not sure but I think we'd always need to catch this during parsing to raise an error since it might destroy the rest of the parse tree. A subsequent validation step might also work but that feels too late as I'm wondering if the parser would raise anyway just on some side-effect downstream.

Postgres raises something like: ```
ERROR: unterminated quoted string at or near `
```

File: dj/sql/parsing/frontends/sql_string.py (hunk 29 above)
Comment (agorajek) [reply dj/sql/parsing/frontends/sql_string.py:38]: +1 on @samredai point. Perhaps the only thing I am not clear on is that `ast -> string` does not really clarify what translation we are using. If it was `ast -> ansi sql` and then self.__str__() that would be clear. Maybe we should just call the `sql_string` frontend as `ansi_sql` to clarify that this is really ANSI SQL converted to string.

File: dj/sql/parsing/ast.py (hunk 27 above)
Comment (agorajek) [reply dj/sql/parsing/ast.py:344]: Oh yes, we need to represent a NULL here. Using None may be an ok shortcut, otherwise we'd need a new type class.

File: dj/sql/parsing/frontends/sql_string.py (hunk 29 above)
Comment (agorajek) [reply dj/sql/parsing/frontends/sql_string.py:38]: This does not have to happen in the same PR. I am ok with merging in and renaming things later, if all agree.

File: dj/sql/parsing/ast.py (hunk 26 above)
Comment (CircArgs) [reply dj/sql/parsing/ast.py]: Made the default. The property exists already. As for the asymetrical quoting I did this little test with `sqloxide`:

```python
----> 1 parse_sql("select * from `a'", "ansi")

ValueError: Query parsing failed.
	sql parser error: Unterminated string literal at Line: 1, Column 17
```

dj/sql/parsing/ast.py
CircArgs: OK added the a node `IsNull` as an operator

agorajek:
hey @betodealmeida, is there anything left here to do?

betodealmeida:
Let me take another look... This week was crazy!

betodealmeida (approved):

Issue #251, opened 2022-11-24
Explicitly set network in docker compose instead of relying on the default


samredai:
When you run `docker compose up` it creates a default docker network that all of the services in that docker setup can connect to (described [here](https://docs.docker.com/compose/networking/)). Right now this calculates to `dj_default`, where `dj_` comes from the project name inferred from the directory name, and then `default` because that's the default.

Instead of letting this get inferred I think we should set both the project name and the network name explicitly. This lets us:

1. Use the docker compose file in a different directory and reliably get the same behavior.
2. If we have different docker compose setups running, like a query service, and want to connect to this docker compose environment, we can reliably know the network name to connect the containers to.

PR #252, opened 2022-11-25
Explicitly set dj name and core network

samredai:
### Summary

This explicitly sets a name (`dj`) and a network (`core`) for the docker compose setups. It allows any other docker container to reliably interact with this docker compose setup by joining the `dj_core` network (docker compose prefixes all networks it creates with the project's name).

### Test Plan

Ran `make check` and `make test`. Also ran `docker compose up` and checked that creating a simple standalone docker container connected to the `dj_core` network could ping the `dj` container.

- [x] PR has an associated issue: #251 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,18 @@
-version: "2.2"
-
+version: "3"
+name: "dj"
 volumes:
   postgres_data: {}
   redis_data: {}
 
+networks:
+  core:
+    driver: bridge
+
 services:
   dj:
     container_name: dj
+    networks:
+      - core
     environment:
       - DOTENV_FILE=.docker-env/.env
     build: .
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -1,12 +1,18 @@
-version: "2.2"
-
+version: "3"
+name: "dj"
```
agorajek, line 2: Probably don't need to quote the `dj` value here.

agorajek (approved):
LGTM, one comment in-line.

docker-compose.yml
samredai, line 2:
```suggestion
name: dj
```

samredai: Fixed 👍

Issue #253, opened 2022-11-25
When using SQLite, errors are raised about using objects across threads (even though everything works)


samredai:
I think we've all seen the errors in the logs when using SQLite that look something like:
```
sqlalchemy.exc.ProgrammingError: (sqlite3.ProgrammingError) SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 274910651584 and this is thread id 275013908224. (Background on this error at: https://sqlalche.me/e/14/f405)
```
This doesn't effect the functionality and everything still works. It's just a check that's enabled by default but can be disabled by adding `?check_same_thread=False` to the URI.

PR #254, opened 2022-11-25
Disable check_same_thread when using SQLite

samredai:
### Summary

This disables the check by SQLite that objects aren't used across threads mainly because the errors are noisy in the logs. I don't think this effects anything since all of the write operations to the metadata db are serialized. (read more [here](https://docs.python.org/3/library/sqlite3.html#sqlite3.connect))

### Test Plan

- [x] PR has an associated issue: #253 
- [x] `make check` passes


betodealmeida (approved):
👍🏼

Issue #255, opened 2022-11-30
Create CRUD endpoints for nodes


samredai:
In addition to definitions in YAML files that are read during compile, we should have CRUD endpoints we can use to create, read, update, and delete nodes. Compile could also use these endpoints where it essentially reads a YAML file and converts it to a POST request.

PR #256, opened 2022-11-30
Endpoint for creating a node

samredai:
### Summary

This adds a POST `/nodes/` endpoint that allows for creating a node through the API. The functionality here aims to be the same as what runs through the compile step that processes the yaml files (with the plan being to eventually replace the compile command with calls to the REST API.

I've removed the flags above and switched to the idea of having an environment attribute on the node to signal if it's in a staging environment or a production environment. The node data model now has an "environment" attribute with an enum value of "staging" or "production". I've summarized behaviors for the different validations below:

### Inferring Column Types
- `"environment": "published"` - The request will fail if it can't properly infer the columns.
- `"environment": "draft"` - The server will try to infer the columns and if it can't, it will still allow adding the node.

### Unknown Dimension Tag
I've left this out entirely so right now you can't add dimension tags when creating a node. I started to feel unsure about how this would play with the automatic inferring of column types. If it's ok with everyone, I think it'd be better to tackle this in a follow-up PR to focus in on this. Right now I'm leaning towards having a separate API call that adds a dimension tag to a column in an existing node and we can include this validation logic in that route.

### Auto Discover Source Node Columns
I've removed the ability to add tables when creating a source node (thanks @betodealmeida for highlighting to me that those are optional!). If everyone agrees I think it'd be great to keep it as a separate API call to attach a source node to a table since that feels more involved than the generic "node creation" that the `POST '/nodes/'` route should handle.

### Test Plan

- Ran `docker-compose up`
- Went to the swagger UI at [http:localhost:8000/docs](http:localhost:8000/docs)
- Submitted the following payloads to the `POST /nodes/` route

_request_
```json
{
   "name":"revenue",
   "description":"This is a source table containing revenue data about the product",
   "type":"source",
   "tables":{
      "postgres":[
         {
            "catalog":null,
            "schema_":"product",
            "table":"revenue"
         }
      ]
   },
   "columns":{
      "country":{
         "type":"STR"
      },
      "revenue": {
         "type": "FLOAT"
      }
   }
}
```
_response_
```json
{
  "name": "revenue",
  "description": "This is a source table containing revenue data about the product",
  "type": "source",
  "query": null,
  "id": 13
}
```
_request_
```json
{
   "name":"purchases_over_a_grand",
   "description":"This filters the revenue source node to only include purchases over $1,000",
   "type":"transform",
   "query": "SELECT country, revenue FROM revenue WHERE revenue > 1000.00"
}
```
_response_
```json
{
  "name": "purchases_over_a_grand",
  "description": "This filters the revenue source node to only include purchases over $1,000",
  "type": "transform",
  "query": "SELECT country, revenue FROM revenue WHERE revenue > 1000.00",
  "id": 14,
  "columns": [
    {
      "id": 68,
      "dimension_id": null,
      "dimension_column": null,
      "name": "country",
      "type": "STR"
    },
    {
      "id": 69,
      "dimension_id": null,
      "dimension_column": null,
      "name": "revenue",
      "type": "FLOAT"
    }
  ],
  "tables": []
}
```
_request_
```json
{
   "name":"total_revenue_from_purchases_over_a_grand",
   "description":"A total of product revenue where only purchases over $1,000 are included",
   "type":"metric",
   "query": "SELECT SUM(revenue) as total_revenue FROM purchases_over_a_grand"
}
```
_response_
```json
{
  "name": "total_revenue_from_purchases_over_a_grand",
  "description": "A total of product revenue where only purchases over $1,000 are included",
  "type": "metric",
  "query": "SELECT SUM(revenue) as total_revenue FROM purchases_over_a_grand",
  "id": 15,
  "columns": [
    {
      "id": 70,
      "dimension_id": null,
      "dimension_column": null,
      "name": "total_revenue",
      "type": "FLOAT"
    }
  ],
  "tables": []
}
```
- [x] `make check` passes

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -4,17 +4,35 @@ Node related APIs.
 
 import logging
 from datetime import datetime
-from typing import List, Optional
+from http import HTTPStatus
+from typing import List, Optional, Union
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from sqlmodel import Session, SQLModel, select
 
-from dj.models.column import ColumnType
-from dj.models.node import Node, NodeType
+from dj.models.column import Column, ColumnType
+from dj.models.node import Node, NodeBase, NodeEnvironment, NodeType
+from dj.models.source import ExistingSourceNode, SourceNodeCreator
+from dj.sql.inference import infer_columns
+from dj.sql.parse import get_dependencies
 from dj.utils import get_session
 
 _logger = logging.getLogger(__name__)
 router = APIRouter()
+
+
+class NodeCreator(NodeBase):
+    """
+    A create object for adding a new node
+    """
+
+
+class ExistingNode(NodeBase):
+    """
+    A node read from the DJ system
+    """
+
+    id: int
 
 
 class SimpleColumn(SQLModel):
@@ -50,3 +68,83 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
     List the available nodes.
     """
     return session.exec(select(Node)).all()
+
+
+@router.post("/nodes/", response_model=Union[ExistingSourceNode, ExistingNode])
+def create_node(
+    data: Union[SourceNodeCreator, NodeCreator],
+    session: Session = Depends(get_session),
+) -> Union[ExistingSourceNode, ExistingNode]:
+    """
+    Create a node
+    """
+    # Check if the node already exists
+    query = select(Node).where(Node.name == data.name)
+    node = session.exec(query).one_or_none()
+    if node:
+        raise HTTPException(
+            status_code=HTTPStatus.CONFLICT,
+            detail=f"Node already exists: {data.name}",
+        )
+
+    if data.type == NodeType.SOURCE:
+        return create_source_node(
+            session=session,
+            data=data,
+        )  # Source nodes can't be created with Node.from_orm
+    new_node = Node.from_orm(data)
+    new_nodes_parents = get_dependencies(new_node.query) if new_node.query else set()
+
+    # Lookup the parent nodes for the new node
+    new_node.parents = session.exec(
+        select(Node).where(Node.name.in_(new_nodes_parents)),  # type: ignore  # pylint: disable=no-member
+    ).all()
+    _logger.info("Found parent nodes for %s: %s", new_node.name, new_node.parents)
+
+    # Infer the node's column names and column types by inspecting the parent nodes
+    try:
+        new_node.columns = infer_columns(new_node.query, parents=new_node.parents)
+    except Exception as e:  # pylint: disable=broad-except
+        error_message = f"Cannot infer columns for {new_node.name}: {str(e)}"
+        _logger.error(error_message)
+        if new_node.environment == NodeEnvironment.PRODUCTION:
+            raise HTTPException(
+                status_code=HTTPStatus.PRECONDITION_FAILED,
+                detail=error_message,
+            ) from e
+
+    new_node.extra_validation()
+
+    session.add(new_node)
+    session.commit()
+    session.refresh(new_node)
+
+    return new_node  # type: ignore
+
+
+def create_source_node(session: Session, data: SourceNodeCreator) -> Node:
+    """
+    Create a source node
+    """
+
+    config = {
+        "name": data.name,
+        "description": data.description,
+        "type": data.type,
+        "environment": data.environment,
+        "columns": [
+            Column(
+                name=column_name,
+                type=ColumnType[column_data["type"]],
+            )
+            for column_name, column_data in data.columns.items()
+        ],
+        "parents": [],
+    }
+
+    new_source_node = Node(**config)
+    session.add(new_source_node)
+    session.commit()
+    session.refresh(new_source_node)
+
+    return new_source_node
--- a/dj/models/source.py
+++ b/dj/models/source.py
@@ -1,0 +1,33 @@
+"""
+Models for source node CRUD operations.
+"""
+from typing import Dict, List, Optional
+
+from typing_extensions import TypedDict
+
+from dj.models import Column, Table
+from dj.models.node import NodeBase
+from dj.typing import ColumnType
+
+
+class SourceNodeColumnType(TypedDict, total=False):
+    """
+    Schema of a column for a table defined in a source node
+    """
+
+    type: ColumnType
+    dimension: Optional[str]
+
+
+class SourceNodeCreator(NodeBase):
+    """A create object for adding a new source node"""
+
+    columns: Dict[str, SourceNodeColumnType]
+
+
+class ExistingSourceNode(NodeBase):
+    """A source read from the DJ system"""
+
+    id: int
+    columns: List[Column]
+    tables: List[Table]
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ ENV/
 
 # VS Code
 .vscode
+
+# Intellij
+.idea
--- a/.pylintrc
+++ b/.pylintrc
@@ -4,3 +4,9 @@
 # https://github.com/samuelcolvin/pydantic/issues/1961#issuecomment-759522422
 extension-pkg-whitelist=pydantic
 ignore=templates,docs
+
+[BASIC]
+good-names=
+    e,
+    ex,
+    i,
--- a/alembic/versions/2022_12_16_2349-07cbd423efcd_add_an_environment_attribute_to_nodes.py
+++ b/alembic/versions/2022_12_16_2349-07cbd423efcd_add_an_environment_attribute_to_nodes.py
@@ -1,0 +1,27 @@
+"""Add an environment attribute to nodes
+
+Revision ID: 07cbd423efcd
+Revises: f620c4521c80
+Create Date: 2022-12-16 23:49:44.715585+00:00
+
+"""
+# pylint: disable=no-member, invalid-name, missing-function-docstring, unused-import, no-name-in-module
+
+import sqlalchemy as sa
+import sqlmodel
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "07cbd423efcd"
+down_revision = "f620c4521c80"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("node", sa.Column("environment", sa.String))
+
+
+def downgrade():
+    op.drop_column("node", "environment")
--- a/dj/models/node.py
+++ b/dj/models/node.py
@@ -70,6 +70,20 @@ class NodeType(str, enum.Enum):
     DIMENSION = "dimension"
 
 
+class NodeEnvironment(str, enum.Enum):
+    """
+    Node environment.
+
+    A node can be in one of the following environments:
+
+    1. PRODUCTION is an environment where the DAG must remain valid
+    2. STAGING is an environment with fewer restrictions for adding nodes
+    """
+
+    PRODUCTION = "production"
+    STAGING = "staging"
+
+
 class NodeYAML(TypedDict, total=False):
     """
     Schema of a node in the YAML file.
@@ -82,15 +96,24 @@ class NodeYAML(TypedDict, total=False):
     tables: Dict[str, List[TableYAML]]
 
 
-class Node(SQLModel, table=True):  # type: ignore
+class NodeBase(SQLModel):
+    """
+    A base node.
+    """
+
+    name: str = Field(sa_column=SqlaColumn("name", String, unique=True))
+    description: str = ""
+    type: NodeType = Field(sa_column=SqlaColumn(Enum(NodeType)))
+    query: Optional[str] = None
+    environment: NodeEnvironment = NodeEnvironment.PRODUCTION
+
+
+class Node(NodeBase, table=True):  # type: ignore
     """
     A node.
     """
 
     id: Optional[int] = Field(default=None, primary_key=True)
-    name: str = Field(sa_column=SqlaColumn("name", String, unique=True))
-    description: str = ""
-
     created_at: datetime = Field(
         sa_column=SqlaColumn(DateTime(timezone=True)),
         default_factory=partial(datetime.now, timezone.utc),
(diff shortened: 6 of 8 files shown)


samredai:
cc: @betodealmeida @agorajek @CircArgs

agorajek (commented):
Makes sense to me. One naming comment.

dj/api/metrics.py
```diff
@@ -143,3 +144,21 @@ async def read_metrics_sql(
         database_id=create_query.database_id,
         sql=create_query.submitted_query,
     )
+
+@router.post("/metrics/", response_model=MetricRead)
```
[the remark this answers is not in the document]
samredai: I started with something similar, I think it was `/create/metric/` but I thought it looked very clean and intuitive to see the `GET /metrics/` and `POST /metrics/` right next to each other in the swagger UI. It's a recognized pattern to just rely on the HTTP verbs but I'm mostly impartial here. I'll let others chime in as well.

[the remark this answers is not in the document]
samredai: After the discussion I'm going to try to use a union from the openapi spec and try a single `POST /nodes/` for all node types.

samredai:
@CircArgs @betodealmeida @agorajek I updated this to use a single `POST /nodes/` route after our discussion and it seems to work nicely. I still have the different create models (`MetricCreate`, `DimensionCreate`) but fastapi recognizes the `Union` of them.

Currently it only supports creating a metric or dimension but once we settle on a design I'll include source and transform nodes. Here's how the schema shows up in the swagger UI.

Screen Shot 2022-12-03 at 6 31 05 PM

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
+    return new_node  # type: ignore
+
+
+def create_source_node(session: Session, data: SourceCreate) -> Node:
+    """
+    Create a source node
+    """
+    dj_settings = get_settings()
+
+    config = {
```
agorajek, line 149: Should we add a marker for the client provided source nodes, say `mock: true`. This setting would prevent the DJ Watchdog from reporting on these source nodes as errors, we could still check them but not wake anyone up if a mock source node is not matching the underlying databases/tables?

dj/models/source.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Dict, List, Optional
+
+from typing_extensions import TypedDict
+
+from dj.models import Column, Table
+from dj.models.node import NodeBase
+from dj.typing import ColumnType
+
+
+class SourceCreateTable(TypedDict, total=False):
```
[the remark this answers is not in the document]
samredai: Yeah you're right, I'm scratching my head trying to figure out what the naming represents, haha. I got it from the SQLModel docs that explain how to have [multiple models](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/#multiple-models-with-inheritance) to represent the full path through a fastapi CRUD request. I'll update them to your suggestions!

dj/api/nodes.py
samredai: That's a good point. I'll need to think about this some more but I added it as an item to the weekly sync. Allowing CRUD through the API changes a few things that we'd have to consider, for example we will no longer have the signal of file changes to know when DJ should perform a validation analysis on the DAG.

We also have to think about how we handle situations where an upstream node is deleted or modified, making a once valid node now invalid (so if it has a `mock:false` attribute we'd need to know to switch it to `mock:true`). We could run the validation across the entire DAG each time a CRUD API request completes but that seems like the extreme solution and won't scale nicely. This is a tricky item with some nuances but a big one so it's worth all of us brainstorming a ton on it.

dj/models/source.py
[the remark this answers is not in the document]
samredai: Ok how does this look? I tweaked some of your suggestions a tiny bit.
- SourceCreateTable -> SourceTableSchema
- SourceCreateColumn -> SourceColumnSchema
- SourceCreate -> SourceNodeCreator
- SourceRead -> ExistingSourceNode
- NodeCreate -> NodeCreator
- NodeRead -> ExistingNode

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
     List the available nodes.
     """
     return session.exec(select(Node)).all()
+
+
+@router.post("/nodes/", response_model=Union[SourceRead, NodeRead])
+def create_a_node(
```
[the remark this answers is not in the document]
samredai: Updated! 👍

betodealmeida:
> * **FAIL_ON_UNKNOWN_DIMENSION_TAG** - If true, raises an error and prevents adding a node if one of the nodes columns are tagged with a dimension for which no dimension node currently exists.
> 
> * **FAIL_ON_INFER_COLUMNS_ERROR** - If true, raises an error and prevents adding a node if any of the columns cannot be inferred from upstream nodes, i.e. the upstream node doesn't exist.

Would we even want this to be false? That could lead to an inconsistent DAG.

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
+    if data.type == NodeType.SOURCE:
+        return create_source_node(session=session, data=data)
+
+    new_node = Node.from_orm(data)
+    new_nodes_parents = get_dependencies(new_node.query) if new_node.query else set()
+
+    # Check the columns for any dimension references and do a metadata lookup of the dimension node
+    for column in new_node.columns:
+        if dimension_target := column.get("dimension"):
+            dimension_target, dimension_column = dimension_target.rsplit(".", 1)
```
betodealmeida, line 103: I wonder if we should use a different separator when indicating dimension, eg:

```yaml
dimension = foo.bar::column
```

dj/api/nodes.py
betodealmeida: I think ideally we want to run the validation across the DAG on every modification. Why do you think this wouldn't scale, @samredai? The validation can just track column lineage, so it should be fast and not require any database calls.

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
+                # Connect to the source to retrieve the table's column info
+                columns = get_table_columns(
+                    databases[database_name].URI,
+                    databases[database_name].extra_params,
+                    table_data["schema_"],
+                    table_data["table"],
+                )
+            else:
+                # Use explicitly defined columns if auto-discover is disabled
+                columns = explicitly_defined_columns
```
betodealmeida, line 192: Another thing we might want to consider here is allowing "patching" the columns, ie, "read the schema information from the database but for this particular column use this type". Maybe have an `overrides` parameter or something like that? Something to discuss.

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
+            config["columns"] = columns
+            config["tables"].append(  # type: ignore
```
betodealmeida, line 194: I'm confused here, you're adding this table to all nodes, not just those that don't have a database. But also... shouldn't we enforce that all tables have a known database? What's the use case here?

samredai:
@betodealmeida I definitely see the case for fixing the DAG validations and reflection to always be on, but I feel there's value in providing a path for "legging in", especially in the spirit of making the project more approachable.

- crawling...
  - FAIL_ON_UNKNOWN_DIMENSION_TAG = **False**
  - FAIL_ON_INFER_COLUMNS_ERROR = **False**
  - AUTO_DISCOVER_SOURCE_NODE_COLUMNS = **False**
- walking...
  - FAIL_ON_UNKNOWN_DIMENSION_TAG = **True**
  - FAIL_ON_INFER_COLUMNS_ERROR = **False**
  - AUTO_DISCOVER_SOURCE_NODE_COLUMNS = **False**
- jogging...
  - FAIL_ON_UNKNOWN_DIMENSION_TAG = **True**
  - FAIL_ON_INFER_COLUMNS_ERROR = **True**
  - AUTO_DISCOVER_SOURCE_NODE_COLUMNS = **False**
- _running!_
  - FAIL_ON_UNKNOWN_DIMENSION_TAG = **True**
  - FAIL_ON_INFER_COLUMNS_ERROR = **True**
  - AUTO_DISCOVER_SOURCE_NODE_COLUMNS = **True**

What do you think?

dj/api/nodes.py
samredai: That might just be an incorrect assumption on my part but I imagined it's more than just upstream column lineage, right? If we allow for draft/incomplete/mock nodes, then we'd have to check all of them every time a node is added to see if those nodes are now valid. We would even have to do this when a source node is added which has no upstream nodes. So I'm imagining what that would look like for a large DAG with many of these incomplete nodes.

dj/api/nodes.py
samredai: Ah thanks for reminding me to add this as a discussion topic. Right now this is just a placeholder.

There are two use cases I think are worth supporting here, both related to this idea of allowing people to step into a full DJ deployment. The first is allowing someone to create nodes and generate an ANSI SQL query when calling the `/metrics/` route, using nothing but the DJ server (no external databases). This would be very useful for tutorials, demos, and a way to get familiar with the core concepts around DJ nodes and how they interact with each other. The second is if someone wants to only leverage the ANSI SQL building capabilities of DJ, they can rely on just sending requests to the `/metrics/{node_id}/sql/` route without having to connect all of their source nodes to a database. They could always do that instead by importing the library into a python runtime and using it directly, but it would be nice to facilitate that as an optional mode of the REST API.

dj/api/nodes.py
samredai: +1, I like it!

dj/api/nodes.py
samredai: Ooh I really like that! I wonder if that would change anything about how the the parsing or converting the parse tree to a DJ AST works.

dj/models/source.py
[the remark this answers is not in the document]
agorajek: *Creator and Existing* makes sense. But I would not be afraid of longer class names for clarity so for the first two items i propose:
- SourceCreateTable -> SourceNodeTableSchema
- SourceCreateColumn -> SourceNodeColumnType

dj/api/nodes.py
agorajek: This syntax (::) reminds me of the Postgres notation for casting a value to a type. What exactly do we want to represent here?

dj/api/nodes.py
agorajek: I think we need to clarify the cases of creating source nodes that have nothing to do with reality. I am always for making the system more approachable but I can't think of a good use case yet. Is it basically letting people draw in it like in a scratchpad?

dj/api/nodes.py
samredai: Ah that's right, I knew that looked familiar. 😄

dj/models/source.py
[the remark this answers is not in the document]
samredai: Updated! Much clearer now, thanks @agorajek

betodealmeida (changes requested):

dj/api/nodes.py
betodealmeida: Doesn't have to be `::`, could be `:`?

betodealmeida: `foo.bar->column`
`foo.bar:column`

dj/api/nodes.py
```diff
@@ -50,3 +71,144 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
+                        description=(
+                            "A placeholder database for source "
+                            "nodes with no known database"
+                        ),
+                        URI="unknown://localhost:9999/unknown",
+                    ),
+                    columns=columns,
+                    **table_data,
+                ),
+            )
```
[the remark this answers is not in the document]
samredai: Woah, thanks @betodealmeida! That totally worked. I don't know what made me think tables are required for source nodes. This along with the idea of having a staging environment feels very clean. I'm going to update this PR to do that and remove the global configs to tweak validation.

dj/api/nodes.py
samredai: I've updated the PR with the idea from our discussion earlier to have a staging environment which has softer validation requirements for adding nodes. That punts the discussion in this thread to one of the next PRs when we add the logic to move a node from the staging environment to the production environment and have to perform the full validation. I like how much better this turned out, thanks @betodealmeida @agorajek @CircArgs!

samredai:
@CircArgs @betodealmeida @agorajek please have another look when you have a chance. I've updated the PR as well as the description to summarize the expected behavior.

agorajek:
> I've removed the ability to add tables when creating a source node

@samredai  about that... i think you are talking here about the internals rather than user perspective, but just wanted to confirm, because I feel like "registering a table in DJ" would really be equivalent to creating a source node for it and I am not sure if I want to have these two steps separated from the user side. I understand that there may be a need for an empty/staging source node but most of the time I think "adding a table to DJ" should be part of "creating a source node". No?

agorajek (approved):
LGTM. Just one question on the source node creation logic.

samredai:
> @samredai about that... i think you are talking here about the internals rather than user perspective, but just wanted to confirm, because I feel like "registering a table in DJ" would really be equivalent to creating a source node for it and I am not sure if I want to have these two steps separated from the user side. I understand that there may be a need for an empty/staging source node but most of the time I think "adding a table to DJ" should be part of "creating a source node". No?

@agorajek that's a fair point but I don't think making them separate API routes necessarily means it has to always be a two step action for the user. Whatever command the user calls (CLI command, UI click, etc) can make a sequence of calls to create the source node and immediately attach one or more tables to it. I'm not strongly tied to this idea but something doesn't feel right about everything in the `POST /nodes/` being very lightweight DJ metadata operations, but then when the user includes a table it triggers an if statement that makes calls to external databases that can fail for a variety of reasons.

agorajek:
@samredai yes, I think I am now convinced that we can start making source nodes w/o any table pointer being necessary right away.

samredai:
Rebased and squashed. Keeping this PR open since it needs some more changes given some recent discussions. We may after all want to track the "valid"/"invalid" state of a node with respect to the DAG in addition to keeping "production"/"staging" as functionally meaningful attributes. Looking forward to discussing this more in the new year. Happy holidays!

Issue #178, opened 2022-08-08
`SELECT * FROM metrics` should return metric names, descriptions, etc.


betodealmeida:

CircArgs:
@betodealmeida would you be able to add a bit of detail on this one? Even when selecting explicitly all metrics I get an error like `Metrics ... have non-shared parents` which I think makes sense to me but then this issue would not make sense

betodealmeida:
Yeah, this one definitely needs more context. I added it here so I wouldn't forget.

The SQL API allows you to fetch data from metrics by running queries like:

```sql
SELECT
    "basic.num_comments"            -- a metric node
  , "basic.dimension.users.gender"  -- a column from a dimension node
FROM metrics                        -- a virtual table that encapsulates all metrics
WHERE
    "basic.dimension.users.country" != 'BR'
GROUP BY
     basic.dimension.users.gender
```

The query will be translated into SQL that actually be executed. It happens in [this function](https://github.com/DataJunction/dj/blob/8477abcee0ba5742dcb2a4b7b49e6171cf23ee4c/src/datajunction/sql/build.py#L221), and will produce SQL that looks like:

```sql
SELECT
    COUNT(*) AS "basic.num_comments"
  , gender AS "basic.dimension.users.gender"
FROM
    comments
JOIN
    users
ON
    comments.user_id = users.id
...
```

It would be nice to expand the abstraction of the `metrics` table, and allow queries like:

```sql
SELECT name, description FROM metrics
```

But I think it requires some discussion first of how we want those queries to look like. Additionally, it would be nice to improve the SQL API so that it can also run the SQL used to define nodes. For example, the node `basic.num_comments` has this SQL:

```sql
SELECT SUM(user_id) FROM basic.source.comments
```

Currently, we can't send that SQL to the SQL API, since today it can only be used to query the `metrics` virtual table. It would be nice to have that functionality, since eventually we could have a web UI where people can run the SQL for their nodes as they're developing them.

CircArgs:
@betodealmeida thinking about `SELECT name, description FROM metrics` it seems like it could be confusing. Do you think there should be a differentiation like `metrics.metadata`? (credit to @samredai for this convention which makes a lot of sense to me). Either that or I suppose allow queries directly against the actual table so something like `select name, description from node where type='METRIC'`

betodealmeida:
Yeah, I think there are 3 separate cases for the SQL API:

[...]

PR #260, opened 2022-12-10
Ast refinements

CircArgs:
### Summary

This makes some small changes suggested but missed in the original AST PR.
We also add `Identifier` and `Name` to better handle compound identifiers.

### Test Plan

Updated tests to account for changes. A few more tests for 100% coverage to come before converting the draft

- [x] `make check` passes


dj/sql/parsing/ast.py
```diff
@@ -220,25 +230,61 @@ class Expression(Node):
...
+        return hash(self.name + self.quote_style)
+
+
+@dataclass(eq=False)
+class Identifier(Node):
+    """
+    Represents a sequence of names for something Named in a query
+    """
+
+    idents: List[Name]
```
samredai, line 263: Doesn't have to be in this PR but would it make sense for us to add helper methods right on this Identifier class to help find different elements within the identifier?

For example `i.database()`, `i.namespace()`, `i.table()`, `i.node()` (last two might be the same logic). For something like `foo.bar.baz`, `i.catalog()` would return `foo` and for `bar.baz`, `i.catalog()` would return `None`. That kind of stuff.

dj/sql/parsing/ast.py
```diff
@@ -220,25 +230,61 @@ class Expression(Node):
...
         """
-        return (
-            f"{self.quote_style}{self.name}{self.quote_style}"  # pylint: disable=C0301
-        )
+        return str(self.ident)
 
     def alias_or_name(self) -> str:
         """
-        get the name or alias of the node
+        get the alias of a node if it is the descendant of an alias otherwise get its own ident
```
samredai, line 287: `s/ident/identifier` Only because my mind read that as identity at first...unless it actually is supposed to be shorthand for identity :D

samredai: Oh I see, it did mean ident as in an instance of `Name`, right? Any reason why we don't just call that `names` everywhere, including for the attribute for the `Identifier` class? Does that clash with something else?

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -283,13 +291,16 @@ def parse_function(parse_tree: dict) -> Function:
...
         names = parse_tree["name"]
-        if len(names) != 1:
-            raise DJParseException("Expected a single name for Function")
+
         return cast(
             Function,
             Function(
-                names[0]["value"],
-                names[0]["quote_style"] if names[0]["quote_style"] is not None else "",
+                cast(
```
samredai, line 298: Just looked up `cast` and learned something new. Does this boil down to an alternative to including `# type: ignore` in other places? What made you use `cast` instead?

dj/sql/parsing/ast.py
CircArgs: I think the docstring should read something more clear even like "returns the string name of a parent alias if one exists, otherwise, return the node's own string name".

CircArgs: And then about the name thing, something doesn't feel right but I can't explain it and I can't ignore the logic of your proposal so I think I'll make the change!

dj/sql/parsing/backends/sqloxide.py
CircArgs: First I saw it in @betodealmeida's code. To me it seems like bad practice to handle it with the ignore if it can be otherwise handled with the cast. Otherwise, I wonder if the effects could cascade to where ignoring would still imply that the value is just `Node` and not be specific (in this case `Function`) - I sort of doubt this since all the attributes are still typed but at least this also clarifies the intend ad recognition of the actual problem which is that the base class `Node` defines the method and only specifies the `Node` type as the return type. I tried using the `Self` type but mypy complained that that could not be used as a type. Reading up on `Self` https://peps.python.org/pep-0673/ suggested it should be able to be used like that but apparently not yet or at least not with the versions we allow (I imported from typing-extensions)

dj/sql/parsing/ast.py
CircArgs: Part of the reason I would argue against this is because it is sort of escaping some other niceties of the AST like there being nodes you can rely on for each entity you would find in a query. You might see what I'm getting at with `Column` having a `_table` attribute and `table` property but that's as far as I went and didn't assume anything else about a database besides columns belong to tables. The idea there is that some downstream validation happens which resolves whether the name is a table or something, adds a table node after examining the column identifier and changes the column identifier to be just the column name. 

I was hesitant to suggest that everything would always be something like database.catalog/schema.table or table.column because I was imagining you could have something like `basic.metrics.users.num_users` as a dj node if I'm not mistaken but there's not really a database, catalog, etc there. Finally, I'm not sure what the ansi spec says about this or how databases would handle this, but sqloxide seems to have no problem dragging things out such as `"select a.b.c.d.e.f.g from x"` so it seems from that there could be arbitrarily long "compound identifiers" that don't necessarily match any schema like the database.catalog....

dj/sql/parsing/ast.py
CircArgs: I thought a little bit more to put my feelings to words and I'm thinking that when I see `names` in this scenario it feels like there's some uniqueness implied as though a list of names suggests that they're all things that refer to the node _independently_ as opposed to things that _altogether_ make up what you refer to the node as. Does that make sense?

dj/sql/parsing/ast.py
samredai: That makes sense, I had the same hesitation. It's a huge pain that it's not really standardized anywhere and here we have the additional pattern of `[namespace_part_a, namespace_part_b, ..., dj_node_name]` that we'd have to account for. You've convinced me it's better to leave that out of the AST.

dj/sql/parsing/ast.py
samredai: That's fair and I agree. I looked around to see if there's a naming pattern in the database space and found:

[Oracle](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm) calls them identifier parts: "...If an identifier includes multiple parts separated by periods, then each attribute can be up to 30 bytes long..."

[IBM](https://www.ibm.com/docs/en/informix-servers/14.10?topic=expressions-using-dot-notation) seems to call each one an identifier but also calls them components: "Dot notation (sometimes called the membership operator) qualifies a SQL identifier with another SQL identifier of which it is a component. You separate the identifiers with the period (.) symbol."

[MySQL](https://dev.mysql.com/doc/refman/8.0/en/identifier-qualifiers.html) calls each item an identifier but describes the last identifier as the name and all of the items before it as qualifiers: "An unqualified name consists of a single identifier. A qualified name consists of multiple identifiers."

Let's keep it how you have it now and maybe see how we all refer to them in conversations. 😄

samredai (approved):

CircArgs:
@agorajek @betodealmeida would appreciate any more suggestions on this before merging

agorajek:
Thanks for the PR @CircArgs . I have read through your discussion with @samredai and I am wondering if we should introduce some restrictions into our AST in order to stay sane. For example: node and column names should be alphanumeric + underscore only (so no dots allowed).

This way we would be able to make sense of the following:
- `SELECT a.b.c.d.e.f FROM a.b.c.d.e`
- `SELECT x.f FROM a.b.c.d.e AS x`
But these should fail:
- `SELECT a.b.c FROM a.b.c`
- `SELECT x.f.g FROM a.b.c x`

I still like the idea of node names being hierarchical, like the e.g. of `basic.metrics.users.num_users` but I understand it will complicate the parsing and the AST. If we can keep it going (with some conscious limitations, like the one above) I think we'll be able to do things like this: say we have a node directory/namespace like `basic.metrics.users` with say 3 metrics in it `m1`, `m2`, `m3`. Then running this DJ SQL:
`SELECT * FROM basic.metrics.users`
could result in a query that is effectivle translated to:
`SELECT * FROM dj.nodes WHERE name LIKE 'basic.metrics.users.%'`

Sorry, if I am slightly off topic here... I will have another look at your PR tomorrow.

CircArgs:
>For example: node and column names should be alphanumeric + underscore only (so no dots allowed).

I think you're on point. I'm trying to reconcile your point with something like `SELECT basic.metrics.users.num_users FROM metrics` which I think as things stand would be a valid dj query right ? I need to revisit our conclusions about namespacing, but that aside...

agorajek:
Ah yes. @CircArgs you reminded me of the 3 types of queries we want to support. All nicely described at https://github.com/DataJunction/dj/issues/178.

The one special case of SQL used by BI tools (SuperSet, Tableau) will be always using `FROM dj.metrics` clause and in that case the SELECT clause will indeed be able to carry the full metric node names so:
`SELECT basic.metrics.users.num_users FROM dj.metrics`
will indeed be valid. 

All right we'll need to codify these 3 SQLs... that will be fun :)

agorajek (approved):

CircArgs:
@agorajek so @samredai and I were discussing the concept of a `Namespace` node that acts as a catchall for anything higher than `b.c` would be caught in a `Namespace` so in `a.b.c` `a` would be part of what could be an arbitrarily long list of identifiers, `b` would be a table and `c` would be a column. Any thoughts on this?

samredai:
+1 to @agorajek's idea of introducing some reasonable restrictions to simplify the variations.

> @agorajek so @samredai and I were discussing the concept of a Namespace node that acts as a catchall for anything higher than b.c would be caught in a Namespace so, for example, in a.b.c, if in a SELECT, a would be part of what could be an arbitrarily long list of identifiers called Namespace, b would be a table and c would be a column. Any thoughts on this?

We could maybe take that even further and only allow a max namespace length of 1 for column names. So we would allow `SELECT a.b FROM a` but fail on `SELECT a.b.c FROM a.b`. That seems reasonable to me because it's easy enough to alias the table name -> `SELECT tbl1.c FROM a.b AS tbl1`. Would that then mean the [Column](https://github.com/DataJunction/dj/blob/main/dj/sql/parsing/ast.py#L422) class needs to be updated to have an optional alias attribute?

As for tables with deep namespaces, I imagine they'd be represented as a single `Table` node in the SQL AST where the name attribute is the last part and everything before it is lumped into a namespace attribute which is a list of names. So the table in `SELECT col1, col2 FROM a.b.c.d.e.f` would be:
```py
Table(
  name=Name("f"),
  namespace=[
    Name("a"),
    Name("b"),
    Name("c"),
    Name("d"),
    Name("e")
  ]
)
```
That works for any DJ query or any database query, right? For DJ metadata queries, the 3rd type, we could simply check if the first item in the namespace is `Name("dj")` and treat those as a special case. (Thanks @agorajek for linking to the issue, that was a good memory refresher).

agorajek:
Good comments @CircArgs and @samredai . I like the idea of the Namespace but I am not sure how it would function as a node.  Possible a good topic for Friday...

@samredai I agree `SELECT a.b.c FROM a.b` could be marked as error, since aliasing tables is easy.

dj/sql/parsing/ast.py
```diff
@@ -244,47 +247,84 @@ def __str__(self) -> str:
...
+        transform the namespace into a column
+            whose name is the last name in the namespace
+
+        if the namespace contains a single name,
+            the created column will have no namespace
+        otherwise, the remaining names for the column's namespace
+        """
+        if not self.names:
+            raise DJParseException("Namespace is empty")
+        col = Column(self.names.pop())
```
agorajek, line 291: You seem to use `add_self_as_parent()` in related places but not here. Why?

dj/sql/parsing/ast.py
```diff
@@ -244,47 +244,68 @@ def __str__(self) -> str:
...
-        return cast(Identifier, Identifier([self]).add_self_as_parent())
+        return cast(Namespace, Namespace([self]).add_self_as_parent())
 
     def __hash__(self) -> int:
         return hash(self.name + self.quote_style)
 
 
 @dataclass(eq=False)
-class Identifier(Node):
+class Namespace(Node):
```
samredai, line 273: This class implies that we'd go `Namespace -> Column` or `Namespace -> Table`, are there implications if we design it to go the other way and logically guarantee that a `Namespace` instance never includes the last item (column name or table name)? Something along the lines of:
```py
>>> t = Table("n1.n2.n3.n4.tbl1")
>>> t.name
"tbl1"
>>> t.namespace
Namespace(
    names=[
        Name(name="n1", quote_style=""),
        Name(name="n2", quote_style=""),
        Name(name="n3", quote_style=""),
        Name(name="n4", quote_style=""),
    ]
)

File: dj/sql/parsing/ast.py [hunk 6]
```diff
@@ -220,25 +233,98 @@ class Expression(Node):
...
+    """
+    Represents a sequence of names prececeding some Table or Column
+    """
+
+    names: List[Name]
+
+    def to_column(self) -> "Column":
+        """
+        transform the namespace into a column
+            whose name is the last name in the namespace
```
betodealmeida, line 283: Nit: the convention for docstrings (https://peps.python.org/pep-0257/#multi-line-docstrings) is to have the summary description in a single line. There are tools that will index the summary, and if it's spread in two lines they'll show only the first half.

dj/sql/parsing/ast.py
```diff
@@ -220,25 +233,98 @@ class Expression(Node):
...
+            raise DJParseException("Namespace is empty")
+        col = Column(self.names.pop())
+        if self.names:
+            col.add_namespace(self)
+        return col
+
+    def to_table(self) -> "Table":
+        """
+        transform the namespace into a Table
+            whose name is the last name in the namespace
```
betodealmeida, line 299:
```suggestion
        Transform namespace into Table with the last name in the namespace
```

dj/sql/parsing/backends/sqloxide.py
betodealmeida: Yeah, `Self` is recent: https://peps.python.org/pep-0673/

dj/sql/parsing/backends/sqloxide.py
```diff
@@ -434,14 +421,14 @@ def parse_oxide_tree(parse_tree: dict) -> Query:
     raise DJParseException("Failed to parse query")  # pragma: no cover
 
 
-def parse(sql: str) -> Query:
+def parse(sql: str, dialect: str = "hive") -> Query:
```
betodealmeida, line 424: Let's keep "ansi" as default?
```suggestion
def parse(sql: str, dialect: str = "ansi") -> Query:
```

betodealmeida (approved):
Looks great! I left some nits, and my only ask would be to leave "ansi" as the default dialect, you changed it to "hive".

dj/sql/parsing/backends/sqloxide.py
agorajek: +1

dj/sql/parsing/ast.py
```diff
@@ -419,22 +505,36 @@ def __hash__(self) -> int:
...
 
-    _table: Optional["Table"] = field(repr=False, default=None)
+    _ref: Union["Table", Namespace, None] = field(repr=False, default=None)
 
     @property
-    def table(self) -> Optional["Table"]:
+    def ref(self) -> Union["Table", Namespace, None]:
         """
-        return the table the column was referenced from
+        return the preceding Namespace or Table if one
```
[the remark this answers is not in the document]
samredai: I'm not sure but I think this meant one as in just the table name, so "mytable" instead of "mycatalog.mydb.mytable". Either way this docstring could be clearer.

dj/sql/parsing/ast.py
```diff
@@ -32,9 +35,21 @@ def flatten(maybe_iterables: Any) -> Iterator:
...
-    Base class for all DJ AST nodes.
+    A DJ AST enum
+    """
+
+    def __repr__(self) -> str:
+        return str(self)
+
+# typevar used for node methods that return self
+# so the typesystem can correlate the self type with the return type
+TNode = TypeVar("TNode", bound="Node")
```
samredai, line 48: nit: s/TNode/TypeNode (I think it'd be easier to remember what it is when seen in other places)

samredai:
> ...my only ask would be to leave "ansi" as the default dialect, you changed it to "hive".

Agreed the default should be ansi. Should we also allow switching to another parser dialect through a dj setting? It looks like sqloxide supports a nice set of different dialects.

Issue #245, opened 2022-11-17
DJ SQL AST


samredai:
Create a DJ SQL AST. 👀

CircArgs:
This is done with https://github.com/DataJunction/dj/pull/242

PR #263, opened 2022-12-22
Add ansi and sparksql query tests that run with --tpcds arg

samredai:
### Summary

This PR adds a set of tests (currently failing) that test parsing of all 99 tpcds queries in both ANSI and Spark-SQL dialects. They don't run by default so this won't break the build but as we iterate on the DJ SQL AST we can test it against tpcds using `make test PYTEST_ARGS="--tpcds"`. I confirmed with @CircArgs that this would be helpful.

We don't necessarily need to support 100% of all SQL syntax in DJ for a really solid v0 but this is helpful to know what exactly isn't yet supported. As we add support for more things (unions, implicit inner joins, etc.) we can incrementally add the passing queries to the default test suite.

### Test Plan

`make test PYTEST_ARGS="--tpcds"` to verify that the tpcds query tests run
`make test` to make sure that the tpcds query tests do not run

- [x] PR has an associated issue: #245 
- [x] `make check` passes

### Deployment Plan

N/A


agorajek:
This is great! Thanks @samredai

agorajek (approved):

agorajek:
For those of you who like tox this is how you can run these tests:
`tox tests/sql/parsing/queries/tpcds/test_tpcds.py::test_parsing_sparksql_tpcds_queries -- --tpcds`

CONTRIBUTING.rst
```diff
@@ -337,6 +337,35 @@ with a string expression. Pytest will only run tests which contain names that ma
...
+
+.. code-block:: sh
+
+    make test PYTEST_ARGS="--tpcds --pdb -k test_parsing_ansi_tpcds_queries[./ansi/query1.sql]"
+
+If you prefer to use tox, these flags all work the same way.
+
+.. code-block:: sh
+
+    tox tests/sql/parsing/queries/tpcds/test_tpcds.py::test_parsing_sparksql_tpcds_queries -- --tpcds
```
samredai, line 367: Thanks @agorajek! You gave me the idea to describe how to run these tests in the contributing doc and I added the tox command as well.

Issue #259, opened 2022-12-07
Port build to the DJ AST


CircArgs:
This issue is about the need to port [sql transpilation](https://github.com/DataJunction/dj/blob/main/dj/sql/build.py) from the existing `sqloxide->sqlalchemy` process to just passes over the DJ AST. This is because the AST is intended to act as a mutable object that can be passed around, modified and finally put through some "frontend" e.g. get a string to pass to some [query runner](https://github.com/DataJunction/djqs)

The port will ostensibly involve passes on an ast looking for dj nodes to be replaced by their queries as sub-asts.

PR #264, opened 2023-01-02
Extract deps

CircArgs:
Much thanks to @samredai for all the help with tests <3

### Summary

extraction of dependencies from node queries

Utilizes the dj sql ast to examine queries and extract references to nodes including nodes that do not exist

### Test Plan

- [x] PR has an associated issue: (a step on the way to https://github.com/DataJunction/dj/issues/259)
- [x] `make check` passes


dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
+"""
+Functions for building queries, from nodes or SQL.
```
agorajek, line 2: Copy/pasta... "Functions for extracting dependencies and validating node queries." ?

dj/sql/parsing/ast.py
```diff
@@ -53,56 +54,77 @@ class Node(ABC):
...
+        object.__setattr__(self, key, value)
+        for child in flatten(value):
+            if isinstance(child, Node) and not key.startswith("_"):
+                child.set_parent(self, key)
+
+    def get_nearest_parent_of_type(
+        self: "Node",
+        node_type: Type[TNode],
+    ) -> Optional[TNode]:
+        """traverse up the tree until you find a node of `node_type` or hit the root"""
```
agorajek, line 109: What's the purpose of this function? Why would you want to get a single node of some type? What if there are more nodes of the same type at the same distance?

agorajek (commented):
@CircArgs I have left few comments in the `extract.py` mostly to make the structure a bit cleaner. This would work but in its current shape is hard to read. Still, this is so much good work. Thank you!

so_good

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+        """
+        self.errors = []
+
+    def set_raise(self, raise_: bool):
+        """
+        Set whether to raise caught exceptions or accumulate them
+        """
+        self._raise = raise_
+
+    def catch(self, error: DJError, message: Optional[str]):
```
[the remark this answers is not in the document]
samredai: +1, I like add or maybe append. This is actually my bad. @CircArgs originally had this as a contextmanager where catch made sense as a name but I had to switch it to a normal function when I made it more compatible with the DJException class and didn't think to rename it.

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+    query = select(Node).filter(Node.name == node_name)
+    match = None
+    try:
+        match = session.exec(query).one()
+    except NoResultFound:
+        kind_msg = " or ".join(str(k) for k in kinds) if kinds else ""
+        CompoundBuildException().catch(
+            error=DJError(
+                code=ErrorCode.UNKNOWN_NODE,
+                message=f"No {kind_msg} node `{node_name}` exists.",
```
samredai, line 107: Something I just noticed, this would be more readable if the "No" in the beginning was a "does not exist" at the end.
Currently: `No NodeType.DIMENSION or NodeType.SOURCE or NodeType.TRANSFORM node revenue exists.`
Proposed: `NodeType.DIMENSION or NodeType.SOURCE or NodeType.TRANSFORM node revenue does not exist.`

I keep missing the "No" and reading "node revenue exists".

dj/sql/parsing/ast.py
CircArgs: The intention of this function was that some times you might find yourself with a table or column that belongs to an ast, but you don't necessarily know where. Maybe it is in the uppermost query or maybe it is buried into some subquery so you might ask for the nearest select. 

I expected that this might be useful while building if we needed to add joins for dimensions you want to make sure you are adding them to the correct query, but it is definitely vestigial right now (no building in this PR) and actually in the build code I've got I just call build on the subqueries so I probably don't need this method 🙃

> What if there are more nodes of the same type at the same distance?

I don't think this is possible is it? Nodes currently only have a single parent, so as you trace up the AST towards the root you'll always encounter a unique node of a specific type.

Ought I just get rid of this?

dj/construction/extract.py
CircArgs: Thanks! This file was monolithic and I've moved all the build stuff elsewhere and this PR turned to just extraction with this big function

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+                    return namespace
+        return namespace
+
+    # check projection
+    for col in chain(*(exp.find_all(Column) for exp in select.projection)):
+        check_col(col, table_deps.columns.projection)  # type: ignore
+
+    # check groupby, filters, and join ons
+    gbfo: List[
+        Tuple[List[Tuple[Column, Union[TableExpression, Node]]], Column, bool]
```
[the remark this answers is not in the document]
CircArgs: Yeah, it's a poorly named attempt to organize the variables I wanted for the loop here https://github.com/DataJunction/dj/blob/619987c49293a38cdf1cdc6169c3070ebb429201/dj/construction/extract.py#L395

Breaking down that `Tuple`

- `List[Tuple[Column, Union[TableExpression, Node]]]`: this will be what we append our pairs of resolved `Column`, `Node` to (Column here are AST nodes and Node is DJ Node 😵‍💫)... TableExpression should only be subquery so `Alias[Select]` or `Select` here so I might tighten that type up
- `Column`: This is the current column we are looking at and if it checks out we will append it as mentioned above
- `bool`: ugly imo, but I'm using this to differentiate between where a dimension would be referenced or not... This `extract_dependencies_from_select` design allows you to take an AST modified to include dimensions/filters referenced from the API and process them as if there was SQL submitted that directly referenced dimension nodes in a GROUP BY, WHERE, HAVING, or any JOIN's ON. Ofc the dimension would not be valid for the JOIN though so this `bool` is basically to indicate the column came from an ON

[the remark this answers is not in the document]
agorajek: I see. Yeah this is not easy to remember but I don't have a better suggestion here. We can always clean it up later...

dj/sql/parsing/ast.py
agorajek: > I don't think this is possible is it? Nodes currently only have a single parent, so as you trace up the AST towards the root you'll always encounter a unique node of a specific type.

Oh... inside a query that is. Makes sense.

> Ought I just get rid of this?

If you are not using it, no need to keep it.

dj/construction/extract.py
agorajek: I had the same exact thought and then I forgot to comment.

dj/sql/parsing/ast.py
betodealmeida: Wait... nodes with a `JOIN` will have multiple parents.

samredai: @betodealmeida "nodes" is being overloaded here. Nodes in the SQL AST only have single parents but DJ Nodes can definitely have multiple parents.

betodealmeida: Ah, sorry! I just got back home and I'm catching up. :)

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+        ]
+
+    if select.from_.joins:
+        for join in select.from_.joins:
+            gbfo += [
+                (table_deps.columns.ons, col, False) for col in join.on.find_all(Column)
+            ]
+
+    for add, col, dim_allowed in gbfo:
+        bad_namespace = False
```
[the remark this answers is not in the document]
CircArgs: attempted

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+    subqueries: List[Select] = []
+
+    # used to check need and capacity for merging in dimensions
+    dimension_columns: Set[Node] = set()
+    sources_transforms: Set[Node] = set()
+    dimensions_tables: Set[Node] = set()
+
+    # get all usable namespaces and columns from the tables
+    for table in tables:
+        namespace = ""
```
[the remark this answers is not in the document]
CircArgs: attempted

dj/construction/extract.py
```diff
@@ -0,0 +1,541 @@
...
+    multiple_refs = set()
+    for namespaces_cols in namespaces.values():
+        for col in namespaces_cols:  # type: ignore
+            if col in no_namespace_safe_cols:
+                multiple_refs.add(col)
+            no_namespace_safe_cols.add(col)
+
+    namespaces[""] = no_namespace_safe_cols - multiple_refs  # type: ignore
+
+    def check_col(col: Column, add: list) -> Optional[str]:
```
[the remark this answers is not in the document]
CircArgs: attempted

dj/construction/extract.py
CircArgs: Changed to be 

```python
No node `revenue` exists of kind NodeType.DIMENSION or NodeType.SOURCE or NodeType.TRANSFORM.
```
good? Now you don't even have to read the whole message for the idea "no node revenue"

samredai: Yes that's much clearer. I'm nitpicking now but having `does not exist` somewhere would make it a tiny bit clearer at first glance.

samredai (approved):

dj/construction/extract.py
agorajek: Much better.

agorajek (approved):
I think this is awesome for a first shot. GTG

dj/sql/parsing/ast.py
samredai: It's worth trying to come up with different names for these classes. When they both need to be imported to the same file it's confused me a couple of times and is solved by doing something like `from dj.sql.parsing.ast import Column as ASTColumn` which I don't think is the best solution. Any suggestions are welcome.

CircArgs: They are naturally namespaced already, we could use it like

```python
from dj.sql.parsing import ast
...
ast.Column(ast.Name('x'))
```

agorajek: I agree this is a small issue. And prefixing it either during import or in the name looks much cleaner. The same problem exists for the Node classes. We should clearly make them distinct from the DJ dag objects. Otherwise confusion is guaranteed.

PR #266, opened 2023-01-03
Add hugo powered versioned docs site

samredai:
### Summary

This replaces the `docs` directory with a hugo site and a scaffolding of the sections and pages that we can incrementally fill-in.

Since this is a big PR, let me outline what's happening here.
- `docs/themes` shouldn't require much review since it's just a vendoring/copy of the [hugo-book](https://github.com/alex-shpak/hugo-book) theme
- `docs/content` contains all of the actual content. Right now it's just the sections and mostly empty titled pages that we can fill in asynchronously
- `docs/content/0.1.0/docs/developers/docs-deployment.rst` is probably the main page to review. It's meant to cover everything about how the docs work and serve as a guide for a new contributor who wants to add/update/release docs
- `docs/build-docs.sh` is a small script that takes a few arguments to build the full docs site, including multiple versions. This will be used by our deployment pipelines once we set that up
- `docs/config` all of the hugo configuration
- `docs/static` additional assets such as the dj logo image and a small css file to style code snippets

### Test Plan

Changed into the `docs` directory and ran `hugo serve --contentDir content/0.1.0/`

- [x] PR has an associated issue: #265 
- [x] `make check` passes

### Deployment Plan

This PR doesn't include any docs deployment logic/configuration but the site can be launched locally.


CircArgs (approved):
Site looks great imo! Successfully ran for me using the instructions in the docs in this PR

agorajek:
The main page looks good, nice marketing spin! But the development docs page looks bad (see below). Is that expected or is my setup borked?

Screen Shot 2023-01-11 at 9 11 24 PM

@betodealmeida there goes the decision which docs library to use ;)

samredai:
@agorajek looks like you don't have `rst2html` which is required. It's used by hugo as an "external helper" because hugo doesn't have built-in support for rst. Instead rst2html is plugged in as a helper during rendering. Try `pip install rst2html`. It should look like this:

Screen Shot 2023-01-12 at 12 30 01 AM

agorajek:
That was it. Thanks @samredai . It looks great imo, but I would like to have @betodealmeida speak his mind on this.

Quick question on the content. Why in Getting Started you have `Requesting SQL/Data for a Cube` but nothing about the same requests for other nodes? Shouldn't we do `Requesting SQL/Data for a Node` instead?

samredai:
@agorajek wow good call out. I'm trying to remember why I didn't do that and can't think of any good reasons...I'm wondering if it should just be `Requesting SQL` and `Requesting Data` and then we cover it all within those pages. How does that sound?

betodealmeida:
I'll take a look tomorrow!

agorajek:
@samredai  Requesting SQL and Requesting Data sounds good.

agorajek:
@betodealmeida any chance for you to look at this?

PR #267, opened 2023-01-12
Add status and mode attributes to Node

samredai:
### Summary

This is a small PR that adds a `status` and a `mode` attribute to the `Node` data model. The doc strings describe the values allowed by the enums and a description of them.

```py
    """
    Node mode.

    A node can be in one of the following modes:

    1. PUBLISHED nodes must be valid and not cause any child nodes to be invalid
    2. DRAFT nodes can be invalid, have invalid parents, and include dangling references
    """
```

```py
    """
    Node status.

    A node can have one of the following statuses:

    1. VALID nodes - All references to other nodes and node columns are valid
    2. INVALID nodes - One or more parent nodes that are incompatible or do not exist
    """
```

This PR also breaks up the `Node` class into a `NodeBase` class and a `Node` class that inherits from that. The `NodeBase` represents what we expect to be provided by the user (so we can use it as a model for what users must provide to an endpoint). `Node` then adds in all of the attributes that will be set by the system or by the database. For example, `mode` is added as an attribute on `NodeBase` since the user will specify that, and `status` is added as an attribute on `Node` since that will be determined by the system.

### Test Plan

Ran `make test` and `make check` to ensure the existing tests aren't broken by the new attributes.

- [x] `make check` passes

### Deployment Plan

N/A


alembic/versions/2023_01_12_0141-20321580ea8a_add_status_and_mode_attributes_to_node.py
```diff
@@ -0,0 +1,35 @@
+"""Add status and mode attributes to Node
```
CircArgs, line 1: what's this file

samredai: Since I'm adding attributes, the metadata tables needed to be updated so I did an alembic revision. It's described here in the contributing doc: https://github.com/DataJunction/dj/blob/main/CONTRIBUTING.rst#database-migrations

dj/models/node.py
```diff
@@ -70,6 +70,34 @@ class NodeType(str, enum.Enum):
...
+    DRAFT = "draft"
+
+
+class NodeStatus(str, enum.Enum):
+    """
+    Node status.
+
+    A node can have one of the following statuses:
+
+    1. VALID nodes - All references to other nodes and node columns are valid
```
agorajek, line 93: nit: s/VALID nodes/VALID, same for INVALID. It reads better imo.

dj/models/node.py
```diff
@@ -82,15 +110,25 @@ class NodeYAML(TypedDict, total=False):
     tables: Dict[str, List[TableYAML]]
 
 
-class Node(SQLModel, table=True):  # type: ignore
+class NodeBase(SQLModel):
```
agorajek, line 113: I assume you considered `BaseNode` and decided to call it `NodeBase` ?

dj/models/node.py
```diff
@@ -82,15 +110,25 @@ class NodeYAML(TypedDict, total=False):
...
-    A node.
+    A base node.
     """
 
-    id: Optional[int] = Field(default=None, primary_key=True)
     name: str = Field(sa_column=SqlaColumn("name", String, unique=True))
     description: str = ""
+    type: NodeType = Field(sa_column=SqlaColumn(Enum(NodeType)))
+    query: Optional[str] = None
+    mode: NodeMode = NodeMode.PUBLISHED
```
agorajek, line 122: Shouldn't default mode be `DRAFT`?

dj/models/node.py
```diff
@@ -82,15 +110,25 @@ class NodeYAML(TypedDict, total=False):
...
+    status: NodeStatus = NodeStatus.INVALID
```
agorajek, line 123: You said in your PR summary that you added `status` to `Node` and here it's on `NodeBase`. Which side it true?

agorajek (changes requested):
Thank you @samredai for pushing forward. Few things to clarify in-line.

dj/models/node.py
samredai: Doh! Thanks for catching this, I had a typo in my force push and pushed to the `status-and-node` branch instead of the `status-and-mode` branch 😂 . Updated to have `status` on `Node`.

dj/models/node.py
samredai: This depends on the behavior we're expecting the API to have.

When someone makes a request to create a node, do we always assume it's to be published right away? Published requires that `status="valid"` and in the case where the node is invalid we would return an error with additional text along the lines of `set mode to "draft" if you'd like to create an invalid node`.

On the other hand if we default to draft mode, then the common flow is that users face little friction when adding a node, assuming that the simple checks pass (name is not already in use, SQL is syntactically valid, etc.), and then users are familiar that they have to publish the node at some point which is when they'd face the requirement of the node being valid.

dj/models/node.py
samredai: I hadn't considered `BaseNode` but I'm not entirely opposed to it. Although at first glance it might imply that there's a node type "base".

dj/models/node.py
samredai: Updated. Agreed it reads much better!

dj/models/node.py
agorajek: I see... makes sense. NodeBase it is.

dj/models/node.py
agorajek: Yeah... I think this will be clearer as we go. Probably OK to leave the stringer default for starters.

agorajek (approved):
LGTM

CircArgs (approved):

Issue #268, opened 2023-01-14
Add a `missing_parents` array to the Node class


samredai:
In order to support creating nodes that are invalid, a node would need to be able to track all parents that are broken links (parent nodes that do not exist). Therefore we need to add a `MissingParent` class and have a `missing_parents: List[MissingParent]` attribute on the metadata for nodes.

A couple notes:
- One of the requirements for a node's status to become `valid` is that it does not have any missing parents
- Each time a new node is created, we will need to search for any existing nodes that have that node's name as a missing parent. The new node should be added to the `parents` attribute of the existing node, removed from the `missing_parents` attribute of the existing node, and the existing node should be re-validated to detect if its `invalid` status can be flipped to `valid`. If the new node being added is invalid, it's safe to infer that the existing node is still `invalid` and skip that re-validation.

PR #269, opened 2023-01-14
Add `MissingParent` class and `missing_parents` node attribute

samredai:
### Summary

This adds a `MissingParent` metadata object and a `missing_parents: List[MissingParent]` attribute to the `Node` class.

Missing parents are stored in a separate table, with each missing parent having a unique ID and a join key to an existing node ID. It also has a name field that is a simple non-unique string field. If multiple invalid existing nodes have references to the same missing parent node name, each reference shows up in the table as an individual record with a unique ID.

Also, I've used `"cascade": "all, delete"` for the relationship so that when a node is deleted, all of that node's entries to the missing parents table is also cleaned up. (looking for a good review there from anyone more familiar with sqlalchemy's behavior).

### Test Plan

Ran `make check` and `make test`. Launched the API locally and did some light testing to confirm there are no obvious side-effects.

- [x] PR has an associated issue: #268 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/models/node.py
+++ b/dj/models/node.py
@@ -122,6 +122,36 @@ class NodeBase(SQLModel):
     mode: NodeMode = NodeMode.PUBLISHED
 
 
+class MissingParent(SQLModel, table=True):  # type: ignore
+    """
+    A missing parent node
+    """
+
+    id: Optional[int] = Field(default=None, primary_key=True)
+    name: str = Field(sa_column=SqlaColumn("name", String))
+    created_at: datetime = Field(
+        sa_column=SqlaColumn(DateTime(timezone=True)),
+        default_factory=partial(datetime.now, timezone.utc),
+    )
+
+
+class NodeMissingParents(SQLModel, table=True):  # type: ignore
+    """
+    Join table for missing parents
+    """
+
+    missing_parent_id: Optional[int] = Field(
+        default=None,
+        foreign_key="missingparent.id",
+        primary_key=True,
+    )
+    referencing_node_id: Optional[int] = Field(
+        default=None,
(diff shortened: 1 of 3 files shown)


dj/models/node.py
```diff
@@ -152,6 +182,15 @@ class Node(NodeBase, table=True):  # type: ignore
         },
     )
 
+    missing_parents: List[MissingParent] = Relationship(
+        link_model=NodeMissingParents,
+        sa_relationship_kwargs={
+            "primaryjoin": "Node.id==NodeMissingParents.referencing_node_id",
+            "secondaryjoin": "MissingParent.id==NodeMissingParents.missing_parent_id",
+            "cascade": "all, delete",
```
agorajek, line 190: I am not fully sure how SQL Alchemy resolves this but I think you DO want the cascading delete on the primary join, but I hope the cascading delete will not happen on the secondaryjoin. @betodealmeida any experience with this?

agorajek (commented):
Just one commend on the cascading delete.

dj/models/node.py
samredai: I spent a little time reading up on how this works and I think this would be ok as currently defined because it's actually going to create a dedicated join table. So what's being cascaded are just the join records. If a missing parent reference is directly deleted (which I don't think will ever happen), then all of those links will be removed from the join table but no nodes are deleted. More likely is that a node is deleted in which case all of the links from that node to all missing parents will be removed from the join table.

CircArgs (approved):

agorajek (approved):

PR #270, opened 2023-01-19
Build logic for nodes using the DJ SQL AST

CircArgs:
### Summary

compile refinements with build code on ast
ast refinements

#### Details
The overall design in this PR is

- an ast can be "compiled" which amounts to validating the sql to the point that DJ metadata can be added in
    - one can infer the types of expressions (this does not strictly require compile but 99% of the time it would - when nodes are referenced)
    - once compiled, one can extract "dependencies" straight from the ast (for example using the `.find_all` or `.filter` methods)
        - one can  create a `BuildPlan` which is used to select the "optimal"* database and followed when replacing deps in the query with the correct sub-ast
            - then, one can build which creates a query fully runnable in the "optimal" database

`*` where "optimal" means by cost or a manually selected one by id

We've added to the ast
- ability to replace nodes in the ast at will and recursively while automatically handling parent/children
    - this means you can have a `Table` that can appear in a `From` but also be pointed to by a `Column` (the column was determined during compile to come from that table) and replace the `Table` and have all instances replaced at once
- query conversion to selects
    - allows code to be written only with select in mind getting cte support for free
- minor fixes (some special attributes, sqloxide backend fixes)
 
### Test Plan

- [x] PR has an associated issue: https://github.com/DataJunction/dj/issues/259
- [x] `make check` passes


dj/construction/build_planning.py
```diff
@@ -0,0 +1,174 @@
+"""
+tools for planning the steps a node will need to take to be built to run
```
samredai, line 2: Suggested rewording:
```
"""
Functions for generating build plans for nodes
"""
```

samredai (approved):
I left a few small comments, mostly about the formatting of the doc strings (The common pattern used so far is newline and starting with a capital letter. This will eventually be parsed and show up as descriptions on the docs site's module section.

Overall the PR looks great and I've been playing around with it locally and it works awesome! Some of the functions are pretty large which isn't much of an issue since it's mostly internal logic that users won't be using directly. That being said, once this area of the library stabilizes it could benefit from refactoring into smaller functions. Does this PR include the rest of the logic needed for the "/nodes/validate" and the create/update nodes routes?

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+        for agg in aggs:
+            query.select.group_by += parse(
+                f"select * group by {agg}",
+                dialect,
+            ).select.group_by  # type:ignore
+
+    return query
+
+
+async def build_node(  # pylint: disable=too-many-arguments
```
agorajek, line 232: How about `build_query_ast_for_database` ?

agorajek (commented):
I wasn't able to finish my review today, will finish tomorrow. I left few minor comments already, but I am in awe of the speed of your work @CircArgs .

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+    session: Session,
+    query: ast.Query,
+    build_plan: BuildPlan,
+    build_plan_depth: int,
+    database: Database,
+    dialect: Optional[str] = None,
+) -> ast.Query:
+    """transforms a query ast by replacing dj node references with their asts"""
+    select = query.to_select()
+    build_select(session, select, build_plan, build_plan_depth, database, dialect)
```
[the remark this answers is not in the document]
CircArgs: It (like all the other steps in the pipeline) modifies the ast in-place. Do you think the return is deceptive?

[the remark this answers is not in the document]
agorajek: Yes, it fooled me. It's been ages since I modified input arguments in place through python functions purposefully. I'd like to hear what others think, but if you want to keep the structure of modifying in place then perhaps using a class and instance methods would be a compromise.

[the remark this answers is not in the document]
samredai: That class suggestion from @agorajek gave me an idea. Could this be a method on the `Select` class? Maybe it can even just be a one-liner that calls this stand-alone function (although if it's only used in the context of an instance of Select, it should probably live as a method there). Then you can just call `query.to_select().build(session, build_plan, build_plan_depth, database, dialect)`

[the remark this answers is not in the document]
agorajek: In the end it would be nice to have something like:
```
query.to_select().build_for_database(database).serialize_for_engine(dialect)
```
or maybe even
```
query.build_for_database(database).serialize_for_engine(dialect)
```
?

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+from dj.errors import DJException
+from dj.models.column import Column
+from dj.models.database import Database
+from dj.models.node import Node, NodeType
+from dj.sql.parsing import ast
+from dj.sql.parsing.backends.sqloxide import parse
+
+
+# flake8: noqa: C901
+def build_select(  # pylint: disable=too-many-arguments,too-many-locals,too-many-nested-blocks,too-many-branches
```
shangyian, line 24: To piggy-back off of what the linter is suggesting, might be nice to eventually refactor into a set of smaller functions that it calls, for better readability.

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+
+async def build_node(  # pylint: disable=too-many-arguments
+    session: Session,
+    node: Node,
+    dialect: Optional[str] = None,
+    database_id: Optional[int] = None,
+    filters: Optional[List[str]] = None,
+    aggs: Optional[List[str]] = None,
+) -> Tuple[ast.Query, Database]:
+    """transforms a query ast by replacing dj node references with their asts
```
shangyian, line 240: What about "Determines the optimal database to run the query in and builds the query AST appropriately"?

dj/sql/parsing/ast.py
```diff
@@ -679,6 +817,17 @@ class Query(Expression):
     select: "Select"
     ctes: List[Alias["Select"]] = field(default_factory=list)
 
+    def to_select(self) -> Select:
+        """compile ctes into the select and return the select
+
+        Note: This destroys the structure of the query which cannot be undone
```
shangyian, line 823: This message about the mutation seems a bit buried. Would it be reasonable to have another layer that does a deepcopy instead of changing the structure of the query, and then rename this method to be `_to_select`, to indicate that it's not meant for external use.

CircArgs: I can look into this again. I had some trouble getting a deepcopy to work as I expected and since this was meant purely as a utility to ditch ctes, I just tabled the copy. The different stages modify in-place anyways. Not sure what uses might be for people to do something like compile then put that ast aside, then build and then have both asts sitting around.

dj/construction/build.py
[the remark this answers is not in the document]
CircArgs: I think it's viable to add it as a method. I would still segregate this logic to a different file though don't you think and just call it in the method? I hope to do the same with the type inference e.g. allow calling `.type` on expression types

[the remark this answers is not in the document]
samredai: Yeah I think a different file for the function makes sense and then just using that function in the method. That way if we start refactoring into smaller functions, as long as we keep that top level function signature we don't have to touch any classes.

dj/sql/parsing/ast.py
CircArgs: Thinking now, `_to_select` makes the most sense. Thanks!

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+) -> ast.Query:
+    """add filters and aggs to a query ast"""
+    if filters:
+        filter_asts = (  # pylint: disable=consider-using-ternary
+            query.select.where and [query.select.where] or []
+        )
+
+        for filter_ in filters:
+            filter_asts.append(
+                parse(f"select * where {filter_}", dialect).select.where,  # type:ignore
```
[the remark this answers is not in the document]
CircArgs: Comment added

dj/sql/parsing/ast.py
CircArgs: done!

dj/construction/build.py
CircArgs: done

dj/construction/build.py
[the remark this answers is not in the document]
CircArgs: done!

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+    for i, exp in enumerate(select.projection):
+        if not isinstance(exp, ast.Named):
+            name = f"_col{i}"
+            aliased = ast.Alias(ast.Name(name), child=exp)
+            # only replace those that are identical in memory
+            select.replace(exp, aliased, lambda a, b: id(a) == id(b))
+    return query
+
+
+def add_filters_aggs(
```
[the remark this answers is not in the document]
CircArgs: done!

dj/construction/build.py
CircArgs: I elected for build_node_for_database. How's that?

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+                ),
+            )
+
+        for tbl in tbls:
+            select.replace(tbl, node_ast)
+
+    return select
+
+
+def build_query(  # pylint: disable=too-many-arguments
```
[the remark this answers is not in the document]
CircArgs: done + similar changes to the compile and extract functions

dj/construction/build.py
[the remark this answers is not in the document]
CircArgs: obfuscated the original functions in `build.py` with a leading `_`, removed return typing since it was confusing. added `.build` to `ast.Query`

dj/construction/build_planning.py
CircArgs: Done!

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+    session: Session,
+    select: ast.Select,
+    build_plan: BuildPlan,
+    build_plan_depth: int,
+    database: Database,
+    dialect: Optional[str] = None,
+) -> ast.Select:
+    """transforms a select ast by replacing dj node references with their asts"""
+
+    _, build_plan_lkp = build_plan
```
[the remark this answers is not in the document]
CircArgs: changed!

dj/construction/build.py
```diff
@@ -0,0 +1,275 @@
...
+# flake8: noqa: C901
+def build_select(  # pylint: disable=too-many-arguments,too-many-locals,too-many-nested-blocks,too-many-branches
+    session: Session,
+    select: ast.Select,
+    build_plan: BuildPlan,
+    build_plan_depth: int,
+    database: Database,
+    dialect: Optional[str] = None,
+) -> ast.Select:
+    """transforms a select ast by replacing dj node references with their asts"""
```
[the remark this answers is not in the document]
CircArgs: done!

dj/construction/build_planning.py
```diff
@@ -0,0 +1,174 @@
...
+)
+from dj.models.database import Database
+from dj.models.node import Node, NodeType
+from dj.sql.dag import get_cheapest_online_database
+from dj.sql.parsing import ast
+
+BuildPlan = Tuple[ast.Query, Dict[Node, Tuple[Set[Database], "BuildPlan"]]]  # type: ignore
+
+
+def get_node_materialized_databases(
```
agorajek, line 21: nit: `get_materialized_databases_for_node()` flows a bit better

dj/construction/inference.py
```diff
@@ -0,0 +1,165 @@
...
+    left_type = get_type_of_expression(expression.left)
+    right_type = get_type_of_expression(expression.right)
+
+    def raise_binop_exception():
+        raise DJParseException(
+            "Incompatible types in binary operation "
+            f"{expression}. Got left {left_type}, right {right_type}.",
+        )
+
+    return {
```
agorajek, line 115: Oh man, what a return statement :). Even if this dictionary is not used anywhere else I think I would still put it on a variable. Up to you.

dj/construction/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    """utility taking a namespace and name to make a possible name of a DJ Node"""
+    ret = ""
+    if namespace:
+        ret += ".".join(name.name for name in namespace.names)
+    if name:
+        ret += ("." if ret else "") + name
+    return ret
+
+
+def get_dj_node(
```
agorajek, line 29: Wait, we don't have a version of `get_dj_node` yet anywhere? @betodealmeida anything comes to mind?

agorajek (approved):
This is a big decent diff, which we should get in and iterate on top of. I don't want to block you anymore, but I left few small comments in-line.

dj/construction/build_planning.py
CircArgs: right, done!

dj/construction/build.py
CircArgs: Already working on a follow up to this PR and I will refactor this.

dj/construction/utils.py
samredai: @agorajek when I read your comment I was convinced as well that there must be a very similar function somewhere! I couldn't find one but then I realized, since we're using SQLModel/sqlalchemy, it's so easy to just grab a node.
```py
query = select(Node).filter(Node.name =="foo")
session.exec(query).one()
```
That's easy enough to do which I figure is why a dedicated function never existed. This `get_dj_node` function here though adds two things that we'd want DRY from what I can tell.
- If a node can't be found (specifically, during a build), instead of raising an error, a `DJError` is added to a `CompoundBuildException`. This is important during a build since at the end we want to convert these errors into "missing parent" references that eventually get attached to a `draft` node.
- It optionally lets you include a `NodeType` filter and if a node by that name is found of a different type, it raises a very clear error.

Maybe @CircArgs can describe how he's using the second functionality right now but I do think that would have some use in future. For example I can see a common mistake by a user when looking up metrics by name is to see a transform node name and try and plug that into a metrics search. It's nice that we can return `Node foo is of type TRANSFORM. Expected kind to be of METRIC.`

agorajek: Thanks for the explanation @samredai . All makes sense. I don't think we need a better name for this method.

Issue #274, opened 2023-01-26
Create a route that allows posting diff-based materialization availability states for nodes


samredai:
A few recent discussions have let to the idea of having an endpoint where external services can post the availability of materialized tables for a given node. This will allow materialization metadata to get updated asynchronously and then discovered during the building of SQL queries. The availability state should also include information about which partitions have been materialized and should be diff-based, meaning that a newly posted availability state is always merged with the existing availability state of the node.

```py
class AvailabilityState:
  id: int
  node_name: str
  catalog: str
  schema: str
  table: str
  valid_through_ts: int
  max_partition: str
  min_partition: str
```

agorajek:
Min and max partitions will likely need to be `{str: str}` dicts.

agorajek:
Never mind my previous comment. We concluded that list of strings will be ok for now.

PR #275, opened 2023-01-26
Add endpoint for posting availability state of materialized nodes

samredai:
### Summary

This adds a new endpoint `POST "/data/availability/{node_name}/"` that allows an external materialization service to post information about the availability of a node. The tests include many examples but here's an example of a payload that's setting an availability state for a node `revenue`.

_POST /data/availability/revenue_
```py
{
    "catalog": "test",
    "schema_": "accounting",
    "table": "revenue",
    "valid_through_ts": 20230125,
    "max_partition": "20230125",
    "min_partition": "20220101",
}
```

Availability states are merged meaning the `max_partition` value that's currently on a node is only moved forward if an availability state with a higher `max_partition` value is received. For `min_partition` it's only changed when a state with a lower `min_partition` is received. `valid_through_ts` can be moved forward or backward and could potentially be made optional (and default to `now()`). One thing that's challenging here is knowing how to handle different partition types. Here I make an insufficient assumption that the partition is a string and is sortable in a way that makes sense. If a dataset is partitioned by country however, I'm not sure what a sort of `max_partition` or `min_partition` would mean and if those values are even at all useful.

Another requirement enforced in the API endpoint is that source nodes may have availability states set, but they must be for tables pre-configured as source node tables. Since source tables are fundamentally external to DJ (the starting point for all other transformations), I'm not sure what it would mean to allow posting that a source table `prod.db.foo` is available under the name `prod.db.bar`. Therefore I added this restriction so if the source table is `prod.db.foo`, you can post availability states (for example to inform which partitions are available), however it must be for the table named `prod.db.foo`.

### Test Plan

Wrote a number of tests as well as started up the docker compose environment and tried out the following request:

_POST /data/availability/dbt.dimension.customers_
```json
{
  "catalog": "test",
  "schema_": "accounting",
  "table": "customers",
  "valid_through_ts": 20230125,
  "max_partition": "20230125",
  "min_partition": "20230101"
}
```
_response_
```json
{
  "message": "Availability state successfully posted"
}
```
I then checked the `GET "/nodes/"` route to ensure the availability set can be seen for that node.

- [x] PR has an associated issue: #274 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
+++ b/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
@@ -41,7 +41,9 @@ def upgrade():
             name="fk_nodeavailability_availability_id",
         ),
         sa.ForeignKeyConstraint(
-            ["node_id"], ["node.id"], name="fk_nodeavailability_node_id",
+            ["node_id"],
+            ["node.id"],
+            name="fk_nodeavailability_node_id",
         ),
         sa.PrimaryKeyConstraint("availability_id", "node_id"),
     )
--- a/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
+++ b/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
@@ -1,0 +1,52 @@
+"""Add availability state table and join table
+
+Revision ID: 84e6a136312d
+Revises: 7caaf4276ea2
+Create Date: 2023-01-25 16:01:38.166209+00:00
+
+"""
+# pylint: disable=no-member, invalid-name, missing-function-docstring, unused-import, no-name-in-module
+
+import sqlalchemy as sa
+import sqlmodel
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "84e6a136312d"
+down_revision = "7caaf4276ea2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "availabilitystate",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("catalog", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column("schema_", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column("table", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("valid_through_ts", sa.Integer(), nullable=False),
+        sa.Column("max_partition", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("min_partition", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_table(
+        "nodeavailabilitystate",
+        sa.Column("availability_id", sa.Integer(), nullable=False),
+        sa.Column("node_id", sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["availability_id"],
+            ["availabilitystate.id"],
+        ),
+        sa.ForeignKeyConstraint(
+            ["node_id"],
+            ["node.id"],
+        ),
+        sa.PrimaryKeyConstraint("availability_id", "node_id"),
+    )
+
+
+def downgrade():
+    op.drop_table("nodeavailabilitystate")
+    op.drop_table("availabilitystate")
--- a/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
+++ b/alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
@@ -38,10 +38,10 @@ def upgrade():
         sa.ForeignKeyConstraint(
             ["availability_id"],
             ["availabilitystate.id"],
+            name="fk_nodeavailability_availability_id",
         ),
         sa.ForeignKeyConstraint(
-            ["node_id"],
-            ["node.id"],
+            ["node_id"], ["node.id"], name="fk_nodeavailability_node_id",
         ),
         sa.PrimaryKeyConstraint("availability_id", "node_id"),
     )
--- a/dj/models/node.py
+++ b/dj/models/node.py
@@ -152,6 +152,44 @@ class NodeMissingParents(SQLModel, table=True):  # type: ignore
     )
 
 
+class AvailabilityStateBase(SQLModel):
+    """
+    An availability state base
+    """
+
+    catalog: Optional[str] = None
+    schema_: Optional[str] = Field(default=None)
+    table: str
+    valid_through_ts: int
+    max_partition: str
+    min_partition: str
+
+
+class AvailabilityState(AvailabilityStateBase, table=True):  # type: ignore
+    """
+    The availability of materialized data for a node
+    """
+
+    id: Optional[int] = Field(default=None, primary_key=True)
+
+
+class NodeAvailabilityState(SQLModel, table=True):  # type: ignore
+    """
+    Join table for availability state
+    """
+
+    availability_id: Optional[int] = Field(
+        default=None,
+        foreign_key="availabilitystate.id",
+        primary_key=True,
+    )
+    node_id: Optional[int] = Field(
+        default=None,
+        foreign_key="node.id",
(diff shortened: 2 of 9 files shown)


samredai:
cc: @agorajek @betodealmeida @shangyian @CircArgs

dj/models/node.py
```diff
@@ -152,6 +152,44 @@ class NodeMissingParents(SQLModel, table=True):  # type: ignore
...
+class AvailabilityStateBase(SQLModel):
+    """
+    An availability state base
+    """
+
+    catalog: Optional[str] = None
+    schema_: Optional[str] = Field(default=None)
+    table: str
+    valid_through_ts: int
+    max_partition: str
```
agorajek, line 164: @samredai if you want to take a shortcut for now and say that all the partition keys will be serialized in the correct order to a single string, that's ok, but we should at least make that comment in here. Ideally we have a list of values in here, ordered according to the partition key list, e.g. dateint, hourint, etc.

agorajek (approved):
@samredai this is good, but we need to address the multiple partition case. Because we are only comparing the values in this PR feel free to take the shortcut I suggested, but we need to make some code comments on this.

dj/models/node.py
samredai: Thanks @agorajek, I'm definitely leaning on you here in order to understand the exact information/functionality that needs to exist here. Since a table can be partitioned by multiple fields with a hierarchy, it seems like we may need a separate attribute specifically to determine that hierarchy, right? For example, let's take a table that has separate `year`, `month`, `day`, `country` partitions. We would need to know something like:
```yaml
partitions: ["year", "month", "day", "country"]
max_partition: "2023,01,26,AF"
min_partition: "2020,01,01,ZW"
```
We could also go all the way and include type information with a `datepart` type so we know how to properly compare the incoming `AvailabilityState` events to the existing `AvailabilityState` and do the merge properly. So maybe...
```yaml
partitions: [
    {"name": "year", "type": "datepart", "format": "yyyy"},
    {"name": "month", "type": "datepart", "format": "mm"},
    {"name": "day", "type": "datepart", "format": "dd"},
    {"name": "country", "type": "str"},
]
max_partition: 2023-01-26T00:00:00.000000  # datetime object
min_partition: 2020-01-01T00:00:00.000000  # datetime object
```
Then we'd know the last element isn't really important in determining the date range of the available data. Am I at least getting warmer? :)

agorajek: @samredai the partition spec in the form of an ordered list of field names should simply be part of the node attributes. I don't think we need to carry it inside the AvailabilityState. What we need to carry here is the values for all or some of the first partition keys (according to the order from the spec). This way we can keep these values as a dict (or a list) and serialize to a string for comparison whenever we need to compare two states.

One thing that I just realized is that when we update nodes (specifically their schemas or partition keys) we will need to invalidate their AvailabilityState (most likely) and count on the Reflection service to pick up the slack and let us know the new AvailabilityState on the changed nodes.

samredai: Got it, makes sense!

samredai: Updated the PR to use `List[str]` for the partition values. Surprisingly, python already sorts them the way you'd expect so I didn't have to change any of the merge logic. I added a comment there though about the need for considering the partition types and doing a cast before comparing.
```py
>>> max(["2022", "01", "US"], ["2019", "01", "ZW"])
['2022', '01', 'US']
>>> max(["2022", "01", "01"], ["2022", "02", "01"])
['2022', '02', '01']
```

dj/models/node.py
```diff
@@ -152,6 +152,44 @@ class NodeMissingParents(SQLModel, table=True):  # type: ignore
...
+
+    catalog: Optional[str] = None
+    schema_: Optional[str] = Field(default=None)
+    table: str
+    valid_through_ts: int
+    max_partition: List[str] = Field(sa_column=SqlaColumn(JSON))
+    min_partition: List[str] = Field(sa_column=SqlaColumn(JSON))
+
+
+class AvailabilityState(AvailabilityStateBase, table=True):  # type: ignore
```
CircArgs, line 168: Is this `AvailabilityState` something that ought to have logic in some method that can just say `True` or `False` with maybe some criterion to say "you can use me" or not?

dj/models/node.py
```diff
@@ -152,6 +152,44 @@ class NodeMissingParents(SQLModel, table=True):  # type: ignore
     )
 
 
+class AvailabilityStateBase(SQLModel):
+    """
+    An availability state base
```
shangyian, line 157: Does this also need a `last_updated_at` timestamp? To indicate when this availability state was last posted.

dj/models/node.py
samredai: Eventually yes but that's part of the downstream features that can be built on top of this information. For now this PR just facilitates external materializations announcing information on incremental materializations, i.e of only a single or a couple partitions. The external system then doesn't have to store the materialization state of each node. What it can instead do is just hit the DJ server, see that the max_partition is 20220120, and then materialize only the partitions for 20220120 to "today". Then push the new availability information to DJ so that it can merge it into the current availability state.

dj/models/node.py
samredai: I can add that, is there a specific feature that you think could use that or is it more for having operational information to present to the user? If it's the later I'm wondering if the `valid_through_ts` is a better indicator for users. I'm worried that even a backfill of a single day far in the past may technically update a `last_updated_at` value and create the appearance the data is processed up to that timestamp.

CircArgs (approved):

dj/models/node.py
agorajek: Agree with @shangyian that `updated_at` would be useful to know, if nothing else then for "operational reasons" ;) But we don't need the `last_` prefix though.

dj/models/node.py
agorajek: Agree with @samredai and I think we will only need to add a route/method to scratch the current state by simply deleting it, when someone say, makes a change to a node... And the moment I wrote the previous sentence down I started to think... but wait, what if the user creates a new version of a node and wants to put it on a new table. Perhaps in that case we don't need to invalidate the old node's data availability state. Good topic for our next brainstorming session.

agorajek (approved):
One minor comment agreeing with @shangyian . And one philosophical question semi-related to this PR. Approving so you can finish at your convenience.

samredai:
> but wait, what if the user creates a new version of a node and wants to put it on a new table. Perhaps in that case we don't need to invalidate the old node's data availability state. Good topic for our next brainstorming session.

amazing

dj/models/node.py
samredai: Updated the PR to add an `updated_at` attribute. I also added a test as a sanity check to make sure that through multiple calls it's actually updated. 😄

alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
```diff
@@ -0,0 +1,52 @@
...
+        sa.Column("valid_through_ts", sa.Integer(), nullable=False),
+        sa.Column("max_partition", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("min_partition", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_table(
+        "nodeavailabilitystate",
+        sa.Column("availability_id", sa.Integer(), nullable=False),
+        sa.Column("node_id", sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
```
shangyian, line 38: Let's add a name for this foreign key constraint. This makes it easier to identify in case we ever need to drop or modify it, as otherwise each database engine will just generate their own key name.

alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
```diff
@@ -0,0 +1,52 @@
...
+            ["availability_id"],
+            ["availabilitystate.id"],
+        ),
+        sa.ForeignKeyConstraint(
```
shangyian, line 42: Same for this one

alembic/versions/2023_01_25_1601-84e6a136312d_add_availability_state_table_and_join_.py
samredai: Great idea, I added it for the ones here and also the ones in the previous alembic upgrades.

PR #276, opened 2023-01-27
Add a POST /nodes/validate/ route

samredai:
### Summary

This adds a `POST /nodes/validate/` route that allows validating a given node definition. The route checks the following:
- The node's query can be parsed.
- The node is either in `published` mode and all upstream dependencies can be determined or is in `draft` mode.
- All column types can be inferred.

This function for this validation route will be reusable as part of the create/update routes. When those routes are added, they should start by running this function to ensure that the API has a consistent response (at least for invalid nodes) when someone is creating, updating, or just validating a node definition.

### Test Plan

Added tests to confirm the expected behaviors as well as ran `docker compose up` and submitted a number of requests to the `POST /nodes/validate` route.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -4,13 +4,23 @@ Node related APIs.
 
 import logging
 from datetime import datetime
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from fastapi import APIRouter, Depends
 from sqlmodel import Session, SQLModel, select
 
-from dj.models.column import ColumnType
-from dj.models.node import AvailabilityState, Node, NodeType
+from dj.construction.extract import extract_dependencies_from_node
+from dj.construction.inference import get_type_of_expression
+from dj.errors import DJError, DJException, ErrorCode
+from dj.models.column import Column, ColumnType
+from dj.models.node import (
+    AvailabilityState,
+    Node,
+    NodeBase,
+    NodeMode,
+    NodeStatus,
+    NodeType,
+)
 from dj.utils import get_session
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


samredai:
cc: @shangyian @betodealmeida @CircArgs @agorajek

dj/construction/build.py
```diff
@@ -189,12 +189,7 @@ def _build_query_ast(  # pylint: disable=too-many-arguments
     """
     select = query._to_select()  # pylint: disable=W0212
     _build_select_ast(session, select, build_plan, build_plan_depth, database, dialect)
-    for i, exp in enumerate(select.projection):
```
samredai, line 192: I moved this logic to a method on the `ast.Select` class, mainly because it only applies to that class. I also needed to use this in the validate route and figured DRY. (cc: @CircArgs)

CircArgs: looks good

tests/api/nodes_test.py
```diff
@@ -61,3 +61,249 @@ def test_read_nodes(session: Session, client: TestClient) -> None:
             "type": "INT",
         },
     ]
+
+
+class TestValidateNodes:  # pylint: disable=too-many-public-methods
```
CircArgs, line 66: do you think we should work on a universal fixture for these sessions? right now there's the `construction_session` maybe we can merge them

dj/api/nodes.py
```diff
@@ -44,6 +47,74 @@ class NodeMetadata(SQLModel):
...
+                DJError(
+                    code=ErrorCode.MISSING_PARENT,
+                    message="Node definition contains references to nodes that do not exist",
+                    debug={"missing_parents": list(missing_parents_map.keys())},
+                ),
+            ],
+        )
+
+    # Add aliases for any unnamed columns and confirm that all column types can be inferred
+    query_ast.select.add_aliases_to_unnamed_columns()
```
CircArgs, line 102: tangential: do we intend to write out the query again which would include these added aliases? I was also experimenting with some serialization just for [reference](https://github.com/CircArgs/dj/blob/8ea622c230437b8142ae8287ed4a6161df2ce11b/dj/sql/parsing/ast.py#L197)

tests/api/nodes_test.py
samredai: Great suggestion. If it's ok I'll merge them into a global fixture in a follow-up PR. Only because it'll be pretty noisy to add to this PR's diff.

dj/api/nodes.py
samredai: Hmm, this is a great point I hadn't thought of. I can see the upside of automatically aliasing unnamed queries and then saving them as already aliased. I'm wondering if there are any drawbacks. Downstream transformations will have to reference the generated column name, i.e. `_col1` but that was always true, right?

samredai:
The build failed with `RuntimeError: The Poetry configuration is invalid:` and I didn't even touch the poetry config file...

hella-confused

samredai:
build failures will be fixed after merging PR #281 and rebasing this.

dj/construction/exceptions.py
```diff
@@ -37,7 +37,7 @@ def set_raise(self, raise_: bool):
         """
         self._raise = raise_
 
-    def append(self, error: DJError, message: Optional[str]):
+    def append(self, error: DJError, message: Optional[str] = None):
```
CircArgs, line 40: thx

dj/api/nodes.py
CircArgs: yeah that was always true. the purpose of actually doing this aliasing in my mind is to make sure that there is consistency when a query is run regardless of the database.

dj/construction/exceptions.py
samredai: No problem but I'm pretty sure that's a bug I introduced when I replaced the context manager with this append method 😅

Issue #277, opened 2023-01-27
Generate an HTML coverage report when the `make test` command runs


samredai:
Adding `--cov-report=html` will generate an `htmlcov` directory in the root of the repo with a very useful html report that you can use for deeper investigation into missing coverage.

PR #278, opened 2023-01-27
generate an HTML coverage report when running tests

samredai:
### Summary

One-line change to add the `--cov-report=html` to the `make test` command. It created an `htmlcov` directory in the root of the project every time the tests run which can be launched with any kind of local server, i.e. live server in VSCode. I find it super helpful when you need to investigate a bit more deeply to identify missing coverage.

### Test Plan

Ran `make test` and made sure the `htmlcov` directory was created.

- [x] PR has an associated issue: #277 
- [x] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
👍

Issue #271, opened 2023-01-23
Improved AST support


CircArgs:
Allow the ast to support the following:

- [ ] `IN`
- [ ] `DISTINCT` functions
- [ ] window functions
- [ ] `ORDER BY`

CircArgs:
adding cross join and table valued functions based on https://github.com/DataJunction/dj/issues/393

PR #280, opened 2023-01-29
Ast refinements

CircArgs:
### Summary

Adding support for the next most commonly used ANSI SQL statements to the ast

### Test Plan

manual testing during development + new unit tests

- [x] PR has an associated issue: #https://github.com/DataJunction/dj/issues/271
- [x] `make check` passes

### DIFF
--- a/dj/construction/compile.py
+++ b/dj/construction/compile.py
@@ -282,7 +282,7 @@ def _validate_groupby_filters_ons_columns(
 
 
 # flake8: noqa: C901
-def compile_select_ast(
+def _compile_select_ast(
     session: Session,
     select: ast.Select,  # pylint: disable= W0621
 ):
@@ -389,18 +389,7 @@ def compile_select_ast(
         ),
     )
     for subquery in subqueries:
-        compile_select_ast(session, subquery)
-
-
-def compile_query_ast(
-    session: Session,
-    query: ast.Query,
-):
-    """
-    Get all dj node dependencies from a sql query while validating
-    """
-    select = query._to_select()  # pylint: disable=W0212
-    compile_select_ast(session, select)
+        _compile_select_ast(session, subquery)
 
 
 def compile_node(
@@ -414,5 +403,5 @@ def compile_node(
     if node.query is None:
         raise DJException(f"Cannot compile node `{node.name}` with no query.")
     query = parse(node.query, dialect)
-    compile_query_ast(session, query)
+    query.compile(session)
     return query
--- a/dj/construction/compile.py
+++ b/dj/construction/compile.py
@@ -3,7 +3,7 @@ Functions for transforming an AST using DJ information
 """
 
 from itertools import chain
-from typing import Dict, List, Optional, Set, Tuple, Union, cast
+from typing import Dict, Iterator, List, Optional, Set, Tuple, Union, cast
 
 from sqlmodel import Session, select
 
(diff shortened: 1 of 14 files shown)


samredai:
build failures will be fixed after merging PR #281 and rebasing this.

dj/construction/compile.py
```diff
@@ -422,4 +414,5 @@ def compile_node(
     if node.query is None:
         raise DJException(f"Cannot compile node `{node.name}` with no query.")
     query = parse(node.query, dialect)
-    return compile_query_ast(session, query)
+    compile_query_ast(session, query)
```
samredai, line 417: Would this fit as a method on the query class? Then this could be changed to `query.compile(session)`

dj/sql/functions.py
```diff
@@ -98,6 +98,11 @@ class Function:  # pylint: disable=too-few-public-methods
     def infer_type(*args: Any) -> ColumnType:
         raise NotImplementedError("Subclass MUST implement infer_type")
 
+    @staticmethod
+    @abc.abstractmethod
+    def infer_type_from_types(*args: Any) -> ColumnType:
+        raise NotImplementedError("Subclass MUST implement infer_type_from_types")
```
samredai, line 104: Since this is an `abstractmethod`, doesn't it already raise an error if the function isn't implemented in a child class? I noticed the you have full coverage on the PR though so this line must be getting hit...I'm missing something aren't I? What would happen if you just had an empty function here with a docstring?:
```py
    @staticmethod
    @abc.abstractmethod
    def infer_type_from_types(*args: Any) -> ColumnType:
        """
        Infer the return type of a function given the types of its arguments
        """
```

samredai: Actually just saw that the implementations don't have the same args part of the function signature. Maybe this shouldn't be an `abstractmethod` then and you can just drop the decorator. That should let you remove all of those `# type: ignore` comments on the implementations.

dj/sql/parsing/ast.py
```diff
@@ -539,14 +549,74 @@ def __str__(self) -> str:
     END)"""
 
 
+@dataclass(eq=False)
+class In(Expression):
+    """an in expression"""
```
samredai, line 554: nit: Doc string formatting
```
    """
    An in expression
    """
```
Same in a couple other places.

dj/sql/functions.py
CircArgs: I just mimicked the existing `infer_type` creating this method to fit with the ast inference

dj/sql/parsing/ast.py
CircArgs: did all of them

dj/construction/compile.py
CircArgs: I'm even still skeptical of the `build` method on `Query`. The steps really all work in concert with one another I'm not sure what the use case is to really be calling them out of the blue.

samredai: Of course either works but for me it's about the scope. From what I understand this function only works on query instances and manipulates them in place, the function name tries to make that clear but I think even clearer is a method on an instance that makes some change to the instance's state. Then even with a simple name like `compile`, it's clear that it doesn't operate on Column nodes, as an example.

CircArgs: okay I added the compile method to `Query`, everywhere now just uses the `build` and `compile` methods and I got rid of the old functions and consolidated the little bit of logic specific to `Query` into the methods

CircArgs: +to consolidate I stole the method to add aliases

samredai (approved):

Issue #283, opened 2023-01-30
Alembic upgrade bug for cockroachdb


samredai:
I was doing some docker housekeeping and realized the the database migration fails for the docker setup that uses cockroach DB. I looked into it and it seems to be caused by revision `f620c4521c80` which adds the `uuid` field for databases. This adds the column and then immediately updates the value of the UUID for the DJ database ID and the in-memory sqlite database.

For some reason I'm not completely sure of, cockroach DB doesn't like that the uuid is being added and then values are being set on it within the same session. It seems the bug is fixed by moving the setting of the value to a separate revision.

PR #284, opened 2023-01-30
Fixing some docker and alembic bugs

samredai:
### Summary

I noticed a bug while making sure all of the different docker overrides work properly. It turns out that there's a quirk with alembic and cockroach DB that causes an error when you add a column and try to set values on that column in the same revision. This PR fixes that by moving the portion that sets the UUID values to a separate alembic revision.

One other thing I noticed is that the `core` network was not set on the docker compose overrides which is required to allow them to communicate with the `dj` container in the main docker compose.

### Test Plan

Ran every variation of `docker compose up` with every override file and ensured that the migration works as well as the API.

- [x] PR has an associated issue: #283 
- [x] `make check` passes

### Deployment Plan

N/A


agorajek:
Looks good to me. I was trying to find why this breaks on cockroach, but I think we may be walking in a unique space here ;)

agorajek (approved):

Issue #286, opened 2023-01-31
Non-ISO format for datetimes in API responses breaks auto datestring parsers in other languages


samredai:
The UTC timestamps included in the API responses are not ISO 8601 formatted which makes them incompatible with most automatic date string parsers in other languages.

_Java_
```java
Text '2023-01-31T18:30:30.593485' could not be parsed at index 26
```

_Go_
```go
parsing time "\"2023-01-31T18:30:41.304941\"" as "\"2006-01-02T15:04:05Z07:00\"": cannot parse "\"" as "Z07:00"
```

PR #287, opened 2023-01-31
Extend pydantic datetime parser with a UTCDatetime parser

samredai:
### Summary

This is a fix for #286 which adds a `UTCDatetime` class that extends `datetime.datetime`. It then can be used as a drop-in replacement for `datetime.datetime` type hints and is usable by pydantic.

### Test Plan

I launched the server with `docker compose up` and queried it with generated Java and Go clients. I also ran curl commands against the server to make sure the date strings are returned as proper UTC ISO timestamps. I also ran queries against the graphql API at `/graphql` to confirm the date strings are showing up properly there as well.

- [x] PR has an associated issue: #286 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/graphql/database.py
+++ b/dj/api/graphql/database.py
@@ -4,6 +4,7 @@ GQL Database Models and related APIs.
 
 # pylint: disable=too-few-public-methods, no-member
 
+import datetime
 from typing import TYPE_CHECKING, List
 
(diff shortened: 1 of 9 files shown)


agorajek (approved):
cool-chuck

Issue #262, opened 2022-12-20
Node versioning...


agorajek:
When we update an existing node (say change the SQL of a transform that's connected up and down the dag) we should be conscious about how we handle that:
1) do we update in place ?
2) or do we just make the nodes immutable and help creating a new node from the original one?
3) do we make +1 version (and the node name stays the same) ?

Re #1 There are some node properties that we surely can allow to be updated in place.
Re #2 It should also be easy to make copies of nodes and create new ones under new names.

So the issue comes down to #3. We can offer a version strategy in which we would support various aspects of node management with version build. Let's think of it as a managed view layer:
1. If we support ONLY one active version, it would be obvious which one should be used by consumers.
2. We could help the node author to fill in all the gaps (columns inference + data backfills) so the DAG is valid, before they flip the active version.
3. Optionally we could support the API overrides, such that the consumers would have an escape hatch if the current active version of a node is not working for them.

samredai:
Thanks for opening this @agorajek, a lot of great points for us to think through. A working system I've seen for similar user-defined config objects usually includes four components. This may not be the best system but I figured it's good to lay this out here since it shares a good amount with how DJ is already designed now.
- `ref` - an alpha-numeric name that's used by users and other objects in the system to refer to this object. This is defined by the user at creation time and can never be changed. (equivalent to a node `name` in DJ)
- `display_name` - A very easily update-able display name that has no impact to functionality. A `ref` of "cars.sales.revenue" might have a `display_name` of "Revenue from Sales (Cars)" (doesn't currently exist in DJ)
- `uuid` - an automatically generated uuid that represents the object in a specific state. Users have no reason to be aware of it and any update to the object generates a new `uuid` for the unchanging `ref`.
- `version` - This increments whenever the object is updated and is always paired with a new definition entry with a new `uuid` but the same `ref`

To your point #1 in the first list, I think we can break that down into two lists. One list includes properties that we want to update in place without producing a new version--basically anything that has no impact to functionality and allowing rollback is not super important. Display name and description are two examples. The second list includes properties that we want to allow updating in place but definitely want that to produce a new version. The most obvious one being the query. With some clever data modeling we can enable these varying levels of flexibility.

agorajek:
Yes, a cool little list @samredai . Agree with your observations. I am not strict about it but I think for more approachable system I propose we use the terms in parents:
- ref (or name)
- display_name (or description)
- uuid (or id)
- version (no change)

Two questions:
[...]

PR #289, opened 2023-02-01
Adding a reference node to facilitate node versioning

shangyian:
### Summary

**Note:** This isn't completely ready, but I figured it was better to get it out there for feedback first.

In order to facilitate node versioning, this PR adds a `ReferenceNode`, which acts as a reference to all historical node versions, along with a reference to the current node version. 

All columns in the reference node are immutable (i.e., `name`, `type` etc), whereas those in the node itself can be modified and version tracked (i.e., `description`, `query` etc). In this approach, we continue to store links between node versions in the adjacency list tables, which allows us to have multiple "active" versions of a ReferenceNode.

This also adds the following endpoints:

* `POST /nodes/`: Create node
* `PUT /nodes/{name}/` Update an existing node, which may create a new node revision if there are changes

### Test Plan

I ran `docker compose up` and verified:
* The yaml configs continue to load 
* Some manual testing of creating and updating different node types (source, transform etc)

Also added some unit tests for the create/update endpoints

- [X] PR has an associated issue: #262  #255 
- [X] `make check` passes

### DIFF
--- a/alembic/versions/2022_04_30_1939-5f31ff8814e7_initial_migration.py
+++ b/alembic/versions/2022_04_30_1939-5f31ff8814e7_initial_migration.py
@@ -65,7 +65,7 @@ def upgrade():
         sa.Column("description", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
         sa.Column("expression", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
         sa.PrimaryKeyConstraint("id"),
-        sa.UniqueConstraint("name"),
+        sa.UniqueConstraint("name", name="uniq_node_name"),
     )
     op.create_index(op.f("ix_node_description"), "node", ["description"], unique=False)
     op.create_index(op.f("ix_node_expression"), "node", ["expression"], unique=False)
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -3,23 +3,31 @@ Node related APIs.
 """
 
 import logging
-from typing import List, Optional, Union
+from http import HTTPStatus
+from typing import Dict, List, Optional, Tuple, Union
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from sqlmodel import Session, SQLModel, select
 
 from dj.construction.extract import extract_dependencies_from_node
 from dj.construction.inference import get_type_of_expression
 from dj.errors import DJError, DJException, ErrorCode
+from dj.models import Database, Table
 from dj.models.column import Column, ColumnType
 from dj.models.node import (
     AvailabilityState,
+    CreateNode,
+    CreateSourceNode,
     Node,
     NodeBase,
     NodeMode,
+    NodeRevision,
     NodeStatus,
     NodeType,
+    UpdateNode,
 )
+from dj.models.table import CreateTable
+from dj.sql.parsing import ast
 from dj.utils import UTCDatetime, get_session
 
 _logger = logging.getLogger(__name__)
@@ -35,22 +43,57 @@ class SimpleColumn(SQLModel):
     type: ColumnType
 
 
-class NodeMetadata(SQLModel):
+class TableMetadata(SQLModel):
+    """
+    Output for table information.
+    """
+
+    id: Optional[int]
+    catalog: Optional[str]
+    schema_: Optional[str]
+    table: Optional[str]
+    database: Optional[Database]
+
+
+class NodeRevisionMetadata(SQLModel):
     """
     A node with information about columns and if it is a metric.
     """
 
     id: int
     name: str
+    version: str
+    reference_node_id: Optional[int]
     description: str = ""
 
-    created_at: UTCDatetime
     updated_at: UTCDatetime
 
     type: NodeType
     query: Optional[str] = None
     availability: Optional[AvailabilityState] = None
     columns: List[SimpleColumn]
+    tables: List[TableMetadata]
+
+
+class NodeMetadata(SQLModel):
+    """
+    A reference node that shows the current revision.
+    """
+
+    id: int
+    name: str
+    type: NodeType
+    current_version: str
+    created_at: UTCDatetime
+    current: NodeRevisionMetadata
+
+
+class NodeWithRevisions(NodeMetadata):
+    """
+    Output for a reference node with revision history.
+    """
+
+    revisions: List[NodeRevisionMetadata]
 
 
 class NodeValidation(SQLModel):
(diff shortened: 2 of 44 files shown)


alembic/versions/2022_04_30_1939-5f31ff8814e7_initial_migration.py
```diff
@@ -65,7 +65,7 @@ def upgrade():
         sa.Column("description", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
         sa.Column("expression", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
         sa.PrimaryKeyConstraint("id"),
-        sa.UniqueConstraint("name"),
+        sa.UniqueConstraint("name", name="uniq_node_name"),
```
shangyian, line 68: I had to give this one a name in order to remove it later

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+    session.add(ref_node)
+    session.commit()
+    session.refresh(ref_node)
+    return ref_node  # type: ignore
+
+
+@router.put("/nodes/{name}/", response_model=RefNodeMetadata)
+def update_node(
+    name: str,
+    data: Union[CreateSourceNode, NodeBase],
```
samredai, line 303: Since `NodeBase` requires `name: str`, it requires you to include the name twice when making an API call (the other one being the `{name}` in the route). We could remove the route parameter but if we want to keep that maybe we need to create an `UpdateNode` class that's basically `NodeBase` without the name attribute.

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+    ref_node = session.exec(query).one_or_none()
+    if not ref_node:
+        raise DJException(
+            message=f"A node with name `{name}` does not exist.",
+            http_status_code=404,
+        )
+
+    # Verify that the changes are valid
+    if ref_node.name != data.name:
+        raise DJException(message="Cannot change node name!")
```
samredai, line 325: Continuing the comment above about the two `name` parameters, solving that would also mean this check can be removed.

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+    if ref_node.type != data.type:
+        raise DJException(message="Cannot change node type!")
+
+    # If nothing has changed, do not create the new node revision
```
samredai, line 329: 🤩

dj/models/node.py
```diff
@@ -195,17 +209,41 @@ class NodeAvailabilityState(SQLModel, table=True):  # type: ignore
...
+        sa_relationship_kwargs={
+            "primaryjoin": "and_(ReferenceNode.id==Node.reference_node_id, "
+            "ReferenceNode.current_version == Node.version)",
+            "viewonly": True,
+            "uselist": False,
+        },
+    )
+
+
+class Node(NodeBase, table=True):  # type: ignore
```
samredai, line 236: A naming suggestion: Since `ReferenceNode` will be the common entry point that's used most around the codebase, why not rename that to `Node` and then rename `Node` here to `Revision`? That way `Node.revisions` gives you `List[Revision]`.

samredai (commented):
This is awesome @shangyian! I left some comments, the only major one being about the way we think of "latest" here. I feel it's a good assumption for us to keep that when users are looking at the current state of the DAG, they will always assume that their node's current version is by default looking at the current latest view of everyone else's node. Other behaviors can be very useful but I think they should be non-default.

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+
+    node_revision.extra_validation()
+
+    session.add(ref_node)
+    session.commit()
+    session.refresh(ref_node)
+    return ref_node  # type: ignore
+
+
+@router.put("/nodes/{name}/", response_model=RefNodeMetadata)
```
samredai, line 300: It just occurred to me that we could do this diff-based and signal that by using `PATCH` instead of `PUT` here. That way the user can just provide the fields they're trying to update, `{"query": "select 1"}`. This may make it easier in the future to enrich this to handle those (probably rare) race condition updates.

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+    if (
+        ref_node.current.query == data.query
+        and ref_node.current.description == data.description
+        and ref_node.current.mode == data.mode
+    ):
+        return ref_node
+
+    # Build the new version and find the new node version's parents
+    if data.type == NodeType.SOURCE:
+        node_revision = create_source_node_revision(data, session)
```
samredai, line 339: Adding the `UpdateNode` class without the name means you'll have to inject the name here before calling this function.
```py
node_revision = create_source_node_revision(CreateNode(name=ref_node.name, **data.dict()), session)
```

dj/models/node.py
agorajek: +1. I think what @samredai suggests is more intuitive. One adjustment to what he said could be `s/Revision/NodeRevision`. This way the class name on its own is also pretty informative.

dj/api/nodes.py
```diff
@@ -113,17 +146,208 @@ def validate_node(
...
+    _logger.info(
+        "Parent nodes for %s (v%s): %s",
+        data.name,
+        node_revision.version,
+        node_revision.parents,
+    )
+
+    # Infer the new node revision's column names and column types by inspecting the parent nodes
+    try:
+        node_revision.columns = infer_columns(
```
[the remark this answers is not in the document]
shangyian: Makes sense, I extracted that to a separate `validate` function so that we can pull the right values in.

dj/api/nodes.py
shangyian: Yep, I added an `UpdateNode` class, and since it's being used with `patch` instead of `put`, it actually just modifies the mutable fields to be `Optional` so that only a subset of them need to be provided.

dj/api/nodes.py
shangyian: +1, adding `UpdateNode` handles this already

dj/models/node.py
shangyian: @agorajek and @samredai  -- agreed on naming changes. I went ahead and did `ReferenceNode` -> `Node` and `Node` -> `NodeRevision`.

dj/api/nodes.py
shangyian: Makes sense. I modified this to do `patch` instead of `put`, and to support that created a single `UpdateNode` model with all optional fields.

shangyian:
@samredai and @agorajek - I've updated this PR based on earlier conversations. Each `NodeRevision` now refers to a `Node` parent, meaning that it defaults to the node's current revision when pulling in the dependencies graph. I think we can implement the ability to pin to a specific node revision in a separate PR (cc @matthewwardrop), as we'll also have to figure out the best way to let users specify that type of pinning. (i.e., do they do it in the query or do they pass in separate parameters that indicate versions?)

samredai (approved):
I pulled this down and played with all of the endpoints and this really looks great! Even the availability state expiration on new versions works as expected since they're tied to node revisions and not the actual node ref. One thing it's missing though is the alembic migration. If everyone is ok with it I think we should squash all of the versions that we have in there now. If I understand how it's supposed to be used, we only really need an alembic upgrade at points of version releases where we batch all the changes required to go from one version to the next. We should be able to just delete everything in the `alembic/versions/` directory and follow [these](https://github.com/DataJunction/dj/blob/main/CONTRIBUTING.rst#database-migrations) instructions to do a fresh init. Makes sense to me if you'd rather do that in a follow-up PR so I'm approving.

shangyian:
Oops, how did I miss the alembic migration! Pushed

PR #290, opened 2023-02-02
Generic column types

CircArgs:
### Summary

Makes `ColumnType` capable of handling complex types.

```python
        >>> ColumnType('list[INT]')
        'LIST[INT]'

        >>> ColumnType['INT']
        'INT'

        >>> ColumnType.List[ColumnType.Int]
        'LIST[INT]'

        >>> ColumnType.List[ColumnType.Int].args[0]
        'INT'

        >>> ColumnType.dict[ColumnType.INT, ColumnType.list[ColumnType.dict[ColumnType.INT, ColumnType.list[ColumnType.STR]]]]
        'DICT[INT, LIST[DICT[INT, LIST[STR]]]]'

        >>> ColumnType.Dict[ColumnType.str, ColumnType.list[ColumnType.int]].args[1].args[0]
        'INT'
```

### Test Plan

Full coverage. Ran Docker examples from scratch, examined api  request results. 

- [x] `make check` passes


dj/typing.py
```diff
@@ -46,27 +50,133 @@ class SQLADialect(Protocol):  # pylint: disable=too-few-public-methods
...
+        test = TYPE_PATTERN.match(type_)
+        if test is not None:
+            outer = test.group("outer")
+            inner = test.group("inner")
+            if outer not in COMPLEX_TYPES:
+                raise ColumnTypeError(f"{outer} is not a KNOWN complex type.")
+            inners = inner.split(",")
+            return ColumnType(outer)[inners]
+        if type_ not in PRIMITIVE_TYPES:
+            raise ColumnTypeError(f"Expected a primitive type but got {type_}.")
```
samredai, line 164: Should this error say that it's specifically expecting a primitive type? Seems it could also be from a typo while defining a complex type.
```py
>>> ColumnType('LIST[STR')
ColumnTypeError: Expected a primitive type but got LIST[STR.
```
```py
>>> ColumnType('LIST[STR]')
'LIST[STR]'
```

dj/typing.py
```diff
@@ -46,27 +50,133 @@ class SQLADialect(Protocol):  # pylint: disable=too-few-public-methods
...
+    "BOOL",
+    "DATETIME",
+    "DATE",
+    "TIME",
+    "TIMEDELTA",
+    "NULL",
+    "WILDCARD",
+}
+
+COMPLEX_TYPES = {"LIST": 1, "DICT": 2}
```
samredai, line 68: I think `ARRAY` instead of `LIST` is the more common term with databases. I checked postgres, spark, and trino and they all use that.

samredai (approved):
LGTM! Just had some small comments.

dj/typing.py
agorajek: Agree. Since we need to work with Spark and Trino for starters I would recommend `ARRAY` and `MAP` for your first two complex types. ROW (or named tuple) should be next in line.

dj/models/query.py
```diff
@@ -67,7 +67,7 @@ class ColumnMetadata(SQLModel):
     """
 
     name: str
-    type: ColumnType
+    type: str
```
agorajek, line 70: Good catch, but wondering why no tests need to be adjusted.

agorajek (commented):
Naming suggestion and few small nits in-line.

dj/models/query.py
CircArgs: `ColumnType` inherits from `str` like the original `Enum` so it can be passed around. The idea was that it's just a `str` with this extra validation logic that an be used like the original `Enum` so the behavior would be intuitive or at least how I thought people would expect it to be

dj/typing.py
CircArgs: changed it to a more generic exception message

dj/typing.py
CircArgs: added row and changed the other names.

dj/typing.py
```diff
@@ -46,27 +50,133 @@ class SQLADialect(Protocol):  # pylint: disable=too-few-public-methods
...
+
+    @classmethod
+    def _validate_type(cls, type_: str):
+        test = TYPE_PATTERN.match(type_)
+        if test is not None:
+            outer = test.group("outer")
+            inner = test.group("inner")
+            if outer not in COMPLEX_TYPES:
+                raise ColumnTypeError(f"{outer} is not a KNOWN complex type.")
+            inners = inner.split(",")
```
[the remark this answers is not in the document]
CircArgs: added the comment. I didn't implement anything to check the inner types yet

agorajek (approved):
@CircArgs, do you think it would be possible to make a more generic ROW() type that allows you to optionally name your elements? Say:
```
ROW([STR] Primitive_Type, ...) --> ROW("foo" STR, "bar" INT, "baz" TIME)
```
Later on we could loosen the requirement for the inner type to be more that Primitive, but for "99%" cases we'll be ok with just this.

CircArgs:
> @CircArgs, do you think it would be possible to make a more generic ROW() type that allows you to optionally name your elements? Say:
> 
> ```
> ROW([STR] Primitive_Type, ...) --> ROW("foo" STR, "bar" INT, "baz" TIME)
> ```
> 
> Later on we could loosen the requirement for the inner type to be more that Primitive, but for "99%" cases we'll be ok with just this.
@agorajek 
Sorry I missed this! I see what you mean. I think this is possible with some minor changes. I'll put an issue and assign myself to do this soon.

Issue #273, opened 2023-01-24
Node Query Raw parts


CircArgs:
### Problem statement
This is a proposal to add an escape hatch syntax into the sql allowed in DJ node queries. There are many instances where queries could contain some database specific syntax that we either could parse under a very specific dialect or may never even be able to parse.

A trivial example would be a function we don't support but we can parse the query
```sql
SELECT myspecialfunc(x) FROM a
```
when we look to validate this we will not know the function to be able to infer the type. It would be possible to say something like ```Unknown function `...` specify the type of the expression `myspecialfunc(x)``` 
and then have the means to create nodes (API route) accept specified types. There are more advanced cases though where that would not be possible.

Take for example 

```sql
> SELECT map_zip_with(map(1, 'a', 2, 'b'), map(1, 'x', 2, 'y'), (k, v1, v2) -> concat(v1, v2));
 {1:"ax",2:"by"}
```

from https://spark.apache.org/docs/latest/api/sql/index.html#map_zip_with

we have a combination of some specific functions as well as a lamba. I do not think any sqloxide dialects support this and so we would have to either expand our backends to maybe sqlglot which would introduce multiple extra layers of complexity from selecting default backend to the question of fallbacks... Besides, engines have not shown themselves to be too kind to following standards imo. Also, in the current state if new syntax comes out we automatically do not support it until we can completely parse it.

### Proposal

I think what could work is a dedicated syntax that just facilitates marking and escaping from parts of a query that can be typed expressions - essentially anywhere you could use a column or more complicated expression. 

Consider an Expression in sql which can ofc be many things (a single Column, Function, Case, a subquery, a primitive - int, str, bool, ...,) and will have a type that when creating a new node from sql we will need to know for validation. The Expression may contain other Expressions but most importantly it may contain Column(s). Columns are relevant because we would like to validate as part of a node creation that columns actually exist so that when the node enters the dag it doesn't invalidate potential parts of the dag that have it as a dependency (child DJ nodes).

To allow raw sql (unparsed) that still allows us type inference and column validation I propose something such as wrapping such expressions in a clear marking such as `${Expression : TYPE}` this signifies that this is a part of the sql that we need not worry about only that when we build the query we will dump `Expression` out and we can assume the type is `TYPE`. Further, should any sub-expressions exist within `Expression` that are parseable, they may be escaped with `{SubExpression}`

This is some pseudo metasyntax I think illustrates this further:
`${Expression({Expression})} | ${Expression}`

[...]

PR #292, opened 2023-02-02
raw expressions

CircArgs:
### Summary

Implements a small and very limited "escape hatch for defining typed expressions that may use some unparseable elements or unknown functions (custom functions, lambdas, ...)

The syntax introduced is parseable as standard sql:

`RAW('aggregate(array(1, 2, {col3}), 0, (acc, x) -> acc + x, acc -> acc * 10)', 'INT')`

The second argument must always parse to `String` and we validate that it is a `ColumnType`

The first is flexible and we use a string representation of it.

`{}` are used to escape columns, we extract, parse these columns, and they are traversable in the ast for any manipulation necessary. 

### Test Plan

full coverage, manual experiments

- [x] PR has an associated issue: #https://github.com/DataJunction/dj/issues/273
- [x] `make check` passes

### DIFF
--- a/dj/construction/inference.py
+++ b/dj/construction/inference.py
@@ -104,6 +104,11 @@ def _(expression: ast.Function):  # pragma: no cover
 
 
 @get_type_of_expression.register
+def _(expression: ast.Raw):
+    return expression.type_
+
+
+@get_type_of_expression.register
 def _(expression: ast.IsNull):
     return ColumnType.BOOL
 
--- a/dj/sql/parsing/ast.py
+++ b/dj/sql/parsing/ast.py
@@ -771,7 +771,7 @@ class Function(Named, Operation):
                 + query[end + last_start :]
             )
             last_start += start + len(expression_replace_pattern)
-        expressions = parser(
+        expressions = col_expression_strs[:] and parser(
             f"SELECT {', '.join(col_expression_strs)}",
             dialect,
         ).select.projection
--- a/dj/sql/parsing/ast.py
+++ b/dj/sql/parsing/ast.py
@@ -748,8 +748,10 @@ class Function(Named, Operation):
                 f"ColumnType not {self.args[1]} in {self}.",
             ) from exc
 
-        query = (
-            str(self.args[0]).strip("".join(name.quote_style for name in self.args[0].find_all(Name)))
+        query = (#pragma: no cover
+            str(self.args[0]).strip(
+                "".join(name.quote_style for name in self.args[0].find_all(Name)),
+            )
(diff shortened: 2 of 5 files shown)


samredai:
I pulled this down to start playing with it. I noticed one issue and on maybe issue.

First one is that the double quotes are included when the RAW expression in the AST is converted back to a string to render the query.
```py
>>> ast = parse("""SELECT RAW("ARRAY_MAX(MAP_VALUES(MAP_FILTER({foo}, (k, v) -> k= 'US')))", 'INT') FROM bar""")
>>> print(ast)
SELECT  "ARRAY_MAX(MAP_VALUES(MAP_FILTER(foo, (k, v) -> k= 'US')))" 
 FROM bar
```

The maybe issue is that it looks like there's a requirement that a column is included in the raw parts. So it raises an error when you don't have anything surround by curly braces. Rewriting the above to remove the curly braces around column `foo`:
```py
>>> parse("""SELECT RAW("ARRAY_MAX(MAP_VALUES(MAP_FILTER(foo, (k, v) -> k= 'US')))", 'INT') FROM bar""")
ValueError: Query parsing failed.
        sql parser error: Expected an expression:, found: EOF
```

For the second one, I'm trying to reason about whether or not that's actually ok. Is there ever a scenario where there's an un-parseable expression that has absolutely **no** references to any columns? Curious to hear if others know of an example.

agorajek:
@samredai I almost wanted to say you are right on the 2nd maybe issue. But then I realized that there is plenty of examples where you basically use something like `NOW()` or `generate_series()` or some other built-in generator UDFs.

samredai:
@agorajek ah right! I knew there had to be examples but I think my mind was trying to come up with a complicated one, haha.
```py
parse("""SELECT RAW("generate_series()", 'ARRAY[INT]') FROM foo""")
```

CircArgs:
@samredai Thanks! you're 100% right. there was a bug that required columns to appear in the expression but that should be fixed. Also, I have gotten rid of quotes. Just to reiterate, the first argument to raw - the raw expression - can be anything that would pass the initial parse but I'm sure most will only ever use `"` or `'`. In the case of `"` it will be parsed as a column and I can see the quote styles easily and strip. 

```python
>>>str(parse("""SELECT RAW('generate_series()', 'ARRAY[INT]') FROM foo""", "ansi"))
'SELECT  generate_series() \n FROM foo'

>>>str(parse("""SELECT RAW("generate_series()", 'ARRAY[INT]') FROM foo""", "ansi"))
'SELECT  generate_series() \n FROM foo'

```

samredai (approved):

Issue #291, opened 2023-02-02
Complex Column Type enhancements


CircArgs:
- [ ] validate generic types e.g. Map keys are hashable
- [ ] Row optional named args

PR #295, opened 2023-02-03
Column typing validation + named inner types

CircArgs:
### Summary

@agorajek suggestions from https://github.com/DataJunction/dj/pull/290

validates map key types and allows named types for row. Row still has full support for any inner types including other nested complex types

```python
>>> ColumnType.Row['int "number"'].args[0].name
'number'
```

### Test Plan
unit tests

- [x] PR has an associated issue: #https://github.com/DataJunction/dj/issues/291
- [x] `make check` passes

### DIFF
--- a/dj/typing.py
+++ b/dj/typing.py
@@ -65,7 +65,35 @@ PRIMITIVE_TYPES = {
     "WILDCARD",
 }
 
-COMPLEX_TYPES = {"ARRAY": 1, "MAP": 2, "ROW": -1}
+def process_array_args(*args: str)->Tuple["ColumnType"]:
+    if len(args)!=1:
+        raise ColumnTypeError(
+                f"ARRAY expects 1 inner type but got {len(args)}.",
+            )
+    return (ColumnType(args[0]),)
+        
+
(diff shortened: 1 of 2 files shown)


samredai (approved):
LGTM! One thing I noticed though is that the logical branches that raise a `ColumnTypeError` are all tested but I don't see any of the more boring tests for when the column type is instantiated correctly. Might be worth throwing a few of those in for good measure.

agorajek:
awesome_sauce

agorajek (approved):

CircArgs:
> LGTM! One thing I noticed though is that the logical branches that raise a `ColumnTypeError` are all tested but I don't see any of the more boring tests for when the column type is instantiated correctly. Might be worth throwing a few of those in for good measure.

they are actually doctests https://github.com/DataJunction/dj/blob/b9da54c1315e528f5b6ae9d0830c8e53fd4a7d5c/dj/typing.py#L142

samredai:
> they are actually doctests

Very cool! Sorry, I missed that. We should probably be using doctests more throughout the codebase. :)

Issue #298, opened 2023-02-04
Create catalogs and engines as entities with relationships


samredai:

agorajek:
Not sure if we want to go crazy with the versions here, but something like that will be useful. The trick will be to make it not stand in the way of users.

PR #299, opened 2023-02-04
Add models and routes for catalogs and engines

samredai:
### Summary

This adds models for catalogs and engines as well as API endpoints to list and create them. Catalogs have just a name and an `engines: List[Engine]` attribute. The `Engine` class is very simple with an `id`, `name`, and `version` attribute.

I guess we could follow up by making a `Schema` model which could carry a `catalog_id` attribute. Then for a `Table` we would only need to know the `schema_id` which can then be used to find the catalog and then the available engines that can be used to query that catalog/schema/table.

Screen Shot 2023-02-05 at 7 40 33 PM

### Test Plan

Wrote tests, ran `make test`, launched the API using `docker compose up` and confirmed that requests/responses behaved as expected.

- [x] PR has an associated issue: #298 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/catalogs.py
+++ b/dj/api/catalogs.py
@@ -79,7 +79,7 @@ def add_catalog(
 
     catalog = Catalog.from_orm(data)
     catalog.engines.extend(
-        filter_to_new_engines(
+        list_new_engines(
             session=session,
             catalog=catalog,
             create_engines=data.engines,
@@ -104,7 +104,7 @@ def add_engines_to_catalog(
     """
     catalog = get_catalog(session, name)
     catalog.engines.extend(
-        filter_to_new_engines(session=session, catalog=catalog, create_engines=data),
+        list_new_engines(session=session, catalog=catalog, create_engines=data),
     )
     session.add(catalog)
     session.commit()
@@ -112,7 +112,7 @@ def add_engines_to_catalog(
     return catalog
 
 
-def filter_to_new_engines(
+def list_new_engines(
     session: Session,
     catalog: Catalog,
     create_engines: List[EngineInfo],
--- a/dj/api/catalogs.py
+++ b/dj/api/catalogs.py
@@ -1,0 +1,132 @@
+"""
+Catalog related APIs.
+"""
+
+import logging
+from http import HTTPStatus
+from typing import List
+
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy.exc import NoResultFound
+from sqlmodel import Session, SQLModel, select
+
+from dj.api.engines import EngineInfo, get_engine
+from dj.models.catalog import Catalog
+from dj.utils import get_session
+
+_logger = logging.getLogger(__name__)
+router = APIRouter()
+
+
+class CatalogInfo(SQLModel):
+    """
+    Class for catalog creation
+    """
+
+    name: str
+    engines: List[EngineInfo] = []
+
+
+def get_catalog(session: Session, name: str) -> Catalog:
+    """
+    Return a Catalog instance given a catalog name
+    """
+    statement = select(Catalog).where(Catalog.name == name)
+    try:
+        catalog = session.exec(statement).one()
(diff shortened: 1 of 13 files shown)


dj/models/catalog.py
```diff
@@ -0,0 +1,42 @@
+"""
+Models for columns.
```
samredai, line 2:
```suggestion
Models for catalogs
```

dj/models/engine.py
```diff
@@ -0,0 +1,17 @@
+"""
+Models for columns.
```
samredai, line 2:
```suggestion
Models for engines
```

tests/api/catalog_test.py
```diff
@@ -0,0 +1,359 @@
...
+            },
+        ],
+    }
+
+
+def test_catalog_adding_a_new_catalog_then_attaching_engines(
+    client: TestClient,
+) -> None:
+    """
+    Test adding a catalog then attaching a catalog
```
samredai, line 170:
```suggestion
    Test adding a catalog then attaching engines
```

tests/api/catalog_test.py
```diff
@@ -0,0 +1,359 @@
...
+            },
+        ],
+    }
+
+
+def test_catalog_adding_without_duplicating(
+    client: TestClient,
+) -> None:
+    """
+    Test adding a catalog and having existing catalogs not re-added
```
samredai, line 217:
```suggestion
    Test adding a catalog and having existing engines not re-added
```

dj/api/engines.py
```diff
@@ -0,0 +1,86 @@
...
+    name: str, version: str, *, session: Session = Depends(get_session)
+) -> EngineInfo:
+    """
+    Return an engine by name and version
+    """
+    return get_engine(session, name, version)
+
+
+@router.post("/engines/", response_model=EngineInfo)
+def add_engine(
```
shangyian, line 63: Hmm, do we want to enable people to add arbitrary engines as a part of the API? For example, someone could come in and add 'Spark' as an engine to a Druid catalog, but that wouldn't make much sense. I think these settings actually make more sense just as config. Alternatively, if we could lock these specific parts of the API down, that works as well.

samredai: Good point, I like the latter option of thinking through how to lock down the API to just the server administrators. It's a pain having to restart the server whenever you want to add/remove/edit catalog related configurations and a lot of query engines like Spark and Trino are moving away from that towards a `CREATE|DROP CATALOG` command (which requires a lot of painful refactoring for them).

shangyian: +1 on convenience. Yeah, and along with server administration, we'll also have to think about how to deal with user management in general, since there isn't currently the concept of users.

shangyian (approved):

dj/api/catalogs.py
```diff
@@ -0,0 +1,132 @@
...
+    catalog.engines.extend(
+        filter_to_new_engines(session=session, catalog=catalog, create_engines=data),
+    )
+    session.add(catalog)
+    session.commit()
+    session.refresh(catalog)
+    return catalog
+
+
+def filter_to_new_engines(
```
agorajek, line 115: A bit confusing name imo. Why not just `list_new_engines` to follow the `list_catalogs` from above? I usually don't call API getters with ` list_` prefix but no strong opinion here, other than let's try to be consistent.

dj/api/engines.py
agorajek: For user management, we can discuss the following two BDP approaches: authorized_managers VS read/write/admin. Added to the agenda.

agorajek (approved):
Just a small comment on the `list_` methods. Otherwise - thank you!

dj/api/catalogs.py
samredai: Fixed! I updated it to `list_new_engines`

PR #307, opened 2023-02-07
Renaming reference nodes + adding naming conventions

shangyian:
### Summary

I noticed that there are still instances of 'reference node' lingering in the code. This PR renames all instances of reference node to node, and node to node revision. 

It also adds naming conventions for constraints in alembic migrations with `BaseSQLModel`. This means that using alembic's autogenerate tool will generate constraints with names that follow the defined convention by default.

### Test Plan

- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/alembic/versions/2023_02_07_1640-03aaf20e6ab6_remove_reference_node_usage.py
+++ b/alembic/versions/2023_02_07_1640-03aaf20e6ab6_remove_reference_node_usage.py
@@ -1,0 +1,53 @@
+"""Remove reference node usage
+
+Revision ID: 03aaf20e6ab6
+Revises: 06b0bef0a7e3
+Create Date: 2023-02-07 16:40:22.905524+00:00
+
+"""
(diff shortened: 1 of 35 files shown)


samredai (approved):
LGTM. Sort of side note but a thought I had while reviewing this. In API responses, should we just return the `NodeRevision` instead of having it nested under the "current" attribute of the `Node` class? It makes sense why the classes are structured like that but it looks like all of the top level `Node` information is also contained in the `NodeRevision` and we can encourage the assumption that if you requested the latest version of a metric, then the `version` attribute of the `NodeRevision` you get back is the `current_version` of the node definition. What do you think?

**edit**: Almost forgot `created_at` which is another `Node` level attribute we'd want to preserve in response payloads alongside the `updated_at` attribute.

shangyian:
@samredai Yep, I was thinking about doing exactly that, as it'll also help when we add the endpoints for getting a specific version of the node. That should just return the node revision attributes in a top-level container, since `current` doesn't make sense in that context anyway. I'll do that in a separate PR!

PR #309, opened 2023-02-07
type aliases and cast statement

CircArgs:
### Summary

add the `CAST` sql statement to the ast and add type aliases 

```python
ALIASES = {
    "DATETIME": "TIMESTAMP",
    "DECIMAL": "FLOAT",
    "INTEGER": "INT",
    "STRING": "STR",
    "LIST": "ARRAY",
    "DICT": "MAP",
}
```

### Test Plan

unit tests

- [ ] `make check` passes


samredai (approved):
LGTM. I also cloned this down and tested it out and it all seems to be working as expected. I'm a bit skeptical on if we should allow these type aliases though. It'll be clear to us as devs who know to come back to this code to look at the alias map, but for other systems building on top of DJ, wouldn't it be confusing when they start to wonder if there's a difference between `DECIMAL` vs. `FLOAT` or `DATETIME` vs. `TIMESTAMP` when they're looking at API logs? Especially since decimal and float have nuanced differences (although granted I'm sure it's fine here since I think those differences are immaterial to DJ's features).

CircArgs:
@samredai just to explain my rationale, it's motivated by how we had list, dict and datetime and to make them more inline with the canonical types we have now array, map and timestamp, but I assumed there could be either used in a query. It would also be fair imo to disallow them and in that case it's a matter of deleting a few lines here

samredai:
@CircArgs thanks that makes sense to me. We can always add constraints later if it it ends up being needed.

Issue #300, opened 2023-02-06
Metrics should have a user-friendly name


shangyian:
The current `Node.name` serves as a unique identifier / system identifier, and `Node.description` works as a longer description of the metric. However, in addition to description and a system identifier, metrics should also have a short, user-friendly name. 

Not sure about naming, perhaps `friendly_name`?

samredai:
+1, maybe `descriptive_name`?

shangyian:
I like `display_name`!

agorajek:
+1 on `display_name`. But to distinguish it from `name` and `description` perhaps it should:
- be unique, just like `name`
- be a limited length, unlike `description`

Not a strong opinion. Just trying to make the use of it match its purpose.

samredai:
@agorajek I agree with the limited length requirement but I don't know if it should be unique. It's not used in any logic in the code so it's essentially a free form "human-readable" text for UI purposes only. I don't even think it should be versioned. We don't have to worry about it breaking any links or bookmarks either since those will utilize the `name` or `id` values which can't be changed.

agorajek:
I agree it will be hard to maintain sensible uniqueness if the field is completely free form. I think we are gonna be OK, as long as we send the (unique by definition) metric name always next to a display name and use the metric name as an additional piece of info, or the only one, if `display_name` is not present. Wondering if @betodealmeida has any opinion?

shangyian:
I think `display_name` should be version controlled. I've definitely seen instances where users are confused after something gets renamed, so it would be better to have a record that we can point users towards if needed. 

I think uniqueness can be nice in the sense that we can force people to really think about duplicates. However, it's also possible for some metrics to be named the same but under different categories...

agorajek:
+1 on versioning `display_name`.

shangyian:
I made it unique here: https://github.com/DataJunction/dj/pull/311 But on second thought perhaps enforcing it is a bad idea, especially if we allow for some form of metric categorization.

betodealmeida:
`display_name` sounds good!

@shangyian do you think it would be confusing for users if the display name is not unique? We'd have to also show the name (which might be too technical, since it uses dot notation for directories, ie, foo.bar) or the description (which might be too verbose).

shangyian:
[...]

PR #311, opened 2023-02-08
Adding display name to node and modify responses for read node endpoints

shangyian:
### Summary

* Add `display_name` to nodes, providing a human-readable version of a node's name. 
* Add an endpoint for listing all revisions for a given node (`GET /nodes/{name}/revisions/`)
* Modify the response model for nodes to return a flattened `NodeRevision` output model along with the node's `created_at` timestamp. It now looks like this without the nesting:
```
{
  "node_revision_id": 3,
  "node_id": 3,
  "type": "source",
  "name": "basic.source.users",
  "display_name": "Users",
  "version": "1",
  "description": "A user table",
  "query": null,
  "availability": null,
  "columns": [
    {
      "name": "id",
      "type": "INT"
    },
    ...
  ],
  "tables": [
    ...
  ],
  "updated_at": "2023-02-08T16:43:28.935144+00:00",
  "created_at": "2023-02-07T17:38:50.343389+00:00"
}
```

### Test Plan

- [X] PR has an associated issue: #300
- [X] `make check` passes

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -7,6 +7,8 @@ from http import HTTPStatus
 from typing import Dict, List, Optional, Tuple, Union
 
 from fastapi import APIRouter, Depends, HTTPException
+from pydantic import BaseModel, Field
+from sqlalchemy.orm import joinedload
 from sqlmodel import Session, SQLModel, select
 
 from dj.construction.extract import extract_dependencies_from_node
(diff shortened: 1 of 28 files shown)


shangyian:
I've updated this PR to remove the uniqueness constraint on `display_name`.

dj/api/nodes.py
```diff
@@ -186,18 +206,16 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
...
-    name: str, *, session: Session = Depends(get_session)
-) -> NodeWithRevisions:
+@router.get("/nodes/{name}/", response_model=NodeMetadata)
+def read_node(name: str, *, session: Session = Depends(get_session)) -> NodeMetadata:
     """
-    List the specified node and include all revisions.
+    Show the active version of the specified node.
     """
-    statement = select(Node).where(Node.name == name)
+    statement = select(Node).where(Node.name == name).options(joinedload(Node.current))
```
samredai, line 218: Does this `joinedload` mean the query filters to the node but then only returns the current node revision that's linked? This is pretty cool.

dj/api/nodes.py
```diff
@@ -207,6 +225,25 @@ def read_node(
     return node  # type: ignore
 
 
+@router.get("/nodes/{name}/revisions/", response_model=List[NodeRevisionMetadata])
```
samredai, line 228: I know we're calling these revisions in the source code, but would this be more intuitive if it was "versions" here? Just the endpoint I mean, as in `"/nodes/{name}/versions/"`

samredai (approved):
Just a couple small comments but otherwise LGTM

dj/api/nodes.py
agorajek: I think this is a minor issue, but imo `revision` seems to be more technical term and I think it helps to see it as an internal feature (for change tracking) rather than external feature for supporting multiple versions of something.

agorajek (approved):
This is awesome. Thanks @shangyian !

dj/api/nodes.py
samredai: Agreed, I definitely don't think this should block this PR. I'll go ahead and merge this in.

PR #313, opened 2023-02-09
add agg and filter columns from api to projections

CircArgs:
### Summary
In the existing build, when aggs and filter dimensions are requested for a metric (or any node) we add them to the ast in the appropriate sections (group by, where) but do not add them to the projection. This implements that along with some parts of the compile refactored to facilitate this.

### Test Plan

Unit tests 
- [x] `make check` passes


samredai (approved):
🚀

Issue #308, opened 2023-02-07
A better way for attaching columns to dimensions for new nodes (through the API*)


samredai:
I noticed that attaching a dimension to a node as you're creating it is sort of awkward. When creating nodes, the user doesn't think about columns at all since columns and column types are inferred from the parsed SQL query. For example, here's an example request for creating a transform node.

_POST "/nodes/"_
```json
{
   "description":"Revenue data filtered to large payments only",
   "query":"SELECT payment_id, payment_amount, customer_id, account_type FROM revenue_source LEFT JOIN payment_type on payment_type = payment_type.id WHERE payment_amount > 1000000",
   "mode":"published",
   "name":"large_revenue_payments_only",
   "type":"transform"
}
```
_response_
```json
{
   "id":21,
   "name":"large_revenue_payments_only",
   "type":"transform",
   "current_version":"1",
   "created_at":"2023-02-07T18:58:04.385384+00:00",
   "current":{
      "id":21,
      "name":"large_revenue_payments_only",
      "version":"1",
      "node_id":21,
      "description":"Revenue data filtered to large payments only",
      "updated_at":"2023-02-07T18:58:04.386891+00:00",
      "type":"transform",
      "query":"SELECT payment_id, payment_amount, customer_id, account_type FROM revenue_source LEFT JOIN payment_type on payment_type = payment_type.id WHERE payment_amount > 1000000",
      "availability":null,
      "columns":[
         {
            "name":"payment_id",
            "type":"INT"
         },
         {
            "name":"payment_amount",
            "type":"FLOAT"
         },
         {
            "name":"customer_id",
            "type":"INT"
         },
         {
            "name":"account_type",
            "type":"STR"
         }
      ],
      "tables":[
         
      ]
   }
}
```
As you can see in the response, the column names were inferred from the parsed SQL and the upstream types were determined by the node dependencies implicit in the SQL query. What I find "awkward" here is how for the user to connect the `account_type` column to an `account_types` dimension node that already exists in the system, they'd have to essentially add an override of the columns section that would have been otherwise automatically inferred. We could tell them to perform a call to `PATCH "/nodes/large_revenue_payments_only"` afterwards and that would work, but it's sort of the same problem and seems like we're providing a blunt instrument for something that will be done very very frequently and could benefit from dedicated functional logic with contextual error messages for bad requests.

CircArgs:
[...]

PR #314, opened 2023-02-09
Add separate endpoints for adding tables and dimension labels

samredai:
### Summary

This adds two new endpoints:
```
http://localhost:8000/nodes/{name}/{column}/dimension/{dimension}/: Add a dimension to a column on a node
http://localhost:8000/nodes/{name}/table/: Add a table to a node
```

The sequence through the API can be seen in the updated [runbook](https://github.com/DataJunction/dj/blob/bb7be38f73fc75a9d9fbb4f2f392190bce40f6d8/notebooks/DJ%20Runbook%20-%20Creating%20Nodes.ipynb). The existing endpoint for creating nodes now purely focuses on creating nodes and doesn't incorporate tables or dimension labels. After creating nodes, tables and dimension labels are added in follow-up requests.

Another thing to consider here is that these linkages are not versioned and adding links happens only on the current version of the node. (Looking to follow-up with delete endpoints for these). We can always update this later to version these which would just be a small change to create a new `NodeRevision` each time, but I think there's something to the idea of these links being light and very easy to aggressively set and unset. There's always the requirement of "knowing what happened" but maybe activity logging could be something we consider as an alternative to always bumping the version.

### Test Plan

Ran `make check` and `make test`. Also ran `docker compose up` and tested that the endpoints work as expected.

- [x] PR has an associated issue: #308 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -1,0 +1,48 @@
+"""
+Helpers for API endpoints
+"""
+
+from typing import Optional
+
+from sqlmodel import Session, select
+
+from dj.errors import DJException
+from dj.models import Database
+from dj.models.node import Node, NodeType
+
+
+def get_node_by_name(
+    session: Session,
+    name: str,
+    node_type: Optional[NodeType] = None,
+) -> Node:
+    """
+    Get a node by name
+    """
+    statement = select(Node).where(Node.name == name)
+    if node_type:
+        statement.where(Node.type == node_type)
+    node = session.exec(statement).one_or_none()
+    if not node:
+        raise DJException(
+            message=(
+                f"A {'' if not node_type else node_type} "
+                f"node with name `{name}` does not exist."
+            ),
+            http_status_code=404,
+        )
+    return node
+
+
+def get_database_by_name(session: Session, name: str) -> Database:
+    """
+    Get a database by name
+    """
+    statement = select(Database).where(Database.name == name)
+    database = session.exec(statement).one_or_none()
+    if not database:
+        raise DJException(
+            message=f"Database with name `{name}` does not exist.",
+            http_status_code=404,
+        )
+    return database
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -7,10 +7,12 @@ from http import HTTPStatus
 from typing import Dict, List, Optional, Tuple, Union
 
 from fastapi import APIRouter, Depends, HTTPException
+from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
 from sqlalchemy.orm import joinedload
 from sqlmodel import Session, SQLModel, select
 
+from dj.api.helpers import get_database_by_name, get_node_by_name
 from dj.construction.extract import extract_dependencies_from_node
 from dj.construction.inference import get_type_of_expression
 from dj.errors import DJError, DJException, ErrorCode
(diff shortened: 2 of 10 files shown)


dj/api/nodes.py
```diff
@@ -207,64 +208,6 @@ def read_node(
...
-        Database.name.in_(database_names),  # type: ignore
-    )
-    databases = {db.name: db for db in session.exec(query).all()}
-    missing_databases = database_names - databases.keys()
-    if missing_databases:
-        raise DJException(message=f"Database(s) {missing_databases} not supported.")
-    return databases
-
-
-def create_source_node_revision(
```
samredai, line 230: Since the database check and dimension column additions are moved to the separate API endpoints, this function became just instantiating the `NodeRevision` so I removed this and moved the instantiation to the create node endpoint function.

dj/api/nodes.py
```diff
@@ -329,6 +284,75 @@ def create_node(
...
 
 
+@router.post(
+    "/nodes/{name}/{column}/dimension/{dimension}/",
+    response_model=List[Column],
+)
+def add_dimension_to_node(
+    name: str,
+    column: str,
+    dimension: str,
```
agorajek, line 294: Nit: we could simplify this call a bit for columns that keep the same name as the dimension they reference, by letting users skip `dimension` argument. If None we could simply assume `dimension = column`.

dj/api/nodes.py
```diff
@@ -329,6 +284,75 @@ def create_node(
...
+    )
+    session.add(table)
+    session.commit()
+    session.refresh(table)
+
+    node.current.tables.append(table)
+    session.add(node)
+    session.commit()
+    session.refresh(node)
+    return node
```
agorajek, line 353: Did you mean to say ?
`return node.current.tables`

agorajek (approved):
One bug (I think) and one nit, but approving to unblock you. Thanks for pushing on @samredai !

dj/api/nodes.py
samredai: Good catch, I actually think returning the node at all here isn't the right decision. It makes sense if your creating or updating a node, but in this case the user just wants confirmation that the dimension node has been associated with the column. I'm going to change this to be similar to the availability endpoint and have a response of:
```py
JSONResponse(
        status_code=200,
        content={"message": (
                f"Dimension node {dimension} has been successfully "
                f"linked to column {column} on node {name}"
            )}
    )
```

dj/api/nodes.py
samredai: Hmm, I really like that idea but I'm wondering how to deal with this little tradeoff. Since it's a route parameter, it must be required. We could do something like:
```py
@app.post("/nodes/{name}/{column}/dimension/") 
@app.post("/nodes/{name}/{column}/dimension/{dimension}/")
async def root(name: str, column: str, dimension: Optional[str]=None):
...
```
Which is almost perfect, but for the first route, `dimension` will still show as an optional query parameter that can be used as `?dimension=foo` which can be confusing because now there's two ways to do this. I'm remembering now that you brought up in the sync that there will potentially be many things we'd want to do to a column, that feels like we should go full query parameter mode then and just have one endpoint for linking information to columns.
```py
@app.post("/nodes/{name}/{column}/")
async def root(name: str, column: str, dimension: Optional[str]=None, .....<more args here later?>....):
...
```
Then requests would look like `POST "/nodes/some_node/a_column_on_the_node?dimension=some_dimension_node"`. Thoughts?

dj/api/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Optional
+
+from sqlmodel import Session, select
+
+from dj.errors import DJException
+from dj.models import Database
+from dj.models.node import Node, NodeType
+
+
+def get_node_by_name(session: Session, name: str, node_type: Optional[NodeType] = None):
```
shangyian, line 14: minor: return typing 'Node'

dj/api/nodes.py
```diff
@@ -312,7 +256,19 @@ def create_node(
...
+            description=data.description,
+            type=data.type,
+            columns=[
+                Column(
+                    name=column_name,
+                    type=ColumnType[column_data["type"]],
+                )
+                for column_name, column_data in data.columns.items()
+            ],
+            parents=[],
```
shangyian, line 270: Does tables need to be included here, i.e.: 
```
        tables=[
            Table(
                catalog=table.catalog,
                schema_=table.schema_,
                table=table.table,
                cost=table.cost,
                database=databases[table.database_name],
                columns=[],
            )
            for table in data.tables
        ],
```

dj/api/nodes.py
shangyian: I prefer the latter, as that makes it clearer that it's not required and we'll always have that consistency with all our endpoints (i.e., if it's in the query parameter, it may or may not be required, but if it's in the route path, it's always required).

dj/api/nodes.py
```diff
@@ -329,6 +285,87 @@ def create_node(
...
+                f"Dimension node {dimension} has been successfully "
+                f"linked to column {column} on node {name}"
+            ),
+        },
+    )
+
+
+@router.post("/nodes/{name}/table/")
+def add_table_to_node(
+    name: str, data: CreateTable, *, session: Session = Depends(get_session)
```
shangyian, line 333: Will the table always be a new table? Is it possible that someone would add a table that already exists and thus we should check to see if we already have records of the table and then only add it if needed?

shangyian (approved):

dj/api/nodes.py
samredai: Great catch, I'll add a check for an existing table on the given node.

dj/api/nodes.py
agorajek: Agree, the 2nd approach looks better and leaves more room for similar functionality later.

samredai: Updated the PR to go with the second approach although added in "columns", otherwise the API gets confused with another route we have. That means `POST "/nodes/{name}/columns/{column}/"`

samredai:
@shangyian @agorajek thanks for the reviews! I updated the PR with your feedback, rebased, and squashed. 👍

PR #315, opened 2023-02-10
Allow specifying specific dimension_column

samredai:
### Summary

This is a small fast-followup PR to #314. I realized I didn't fully implement the feedback offered by @agorajek [here](https://github.com/DataJunction/dj/pull/314#discussion_r1101074005). When adding a dimension and not specifying a dimension node name, the API will assume the dimension node you're looking to link has the same name as the column.

That also made me realize I didn't provide a way to specify a specific dimension node's column you want to link to (when not specified, there's a system default of `id`). So I also added an optional argument `dimension_column` that can be provided.

Besides those things, I did a little clean up by moving some logic out to a `get_columns` function in `dj/api/helpers.py`.

### Test Plan

Tested out the API, ran the runbook, ran `make check`, ran `make tests`

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -56,6 +56,7 @@ def get_column(node: NodeRevision, column_name: str) -> Column:
     for node_column in node.columns:
         if node_column.name == column_name:
             requested_column = node_column
+            break
 
     if not requested_column:
         raise DJException(
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -7,8 +7,8 @@ from typing import Optional
 from sqlmodel import Session, select
 
 from dj.errors import DJException
-from dj.models import Database
-from dj.models.node import Node, NodeType
+from dj.models import Column, Database
+from dj.models.node import Node, NodeRevision, NodeType
 
 
 def get_node_by_name(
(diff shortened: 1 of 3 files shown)


dj/api/nodes.py
```diff
@@ -327,16 +327,14 @@ def add_dimension_to_node(
...
     """
     Add information to a node column
     """
-    if not dimension:
-        raise DJException(
-            message="A dimension node must be specified",
-            http_status_code=400,
-        )
+    if not dimension:  # If no dimension is set, assume it matches the column name
+        dimension = column
```
samredai, line 337: cc: @agorajek

dj/api/helpers.py
```diff
@@ -46,3 +46,20 @@ def get_database_by_name(session: Session, name: str) -> Database:
...
+
+
+def get_column(node: NodeRevision, column_name: str) -> Column:
+    """
+    Get a column from a node revision
+    """
+    requested_column = None
+    for node_column in node.columns:
+        if node_column.name == column_name:
+            requested_column = node_column
```
agorajek, line 58: nit: I don't remember us doing this kind of optimization across the code, but I would have added `break` here if I was writing this.

tests/api/nodes_test.py
```diff
@@ -969,12 +969,27 @@ def test_adding_dimensions_to_node_columns(self, client: TestClient):
         assert response.status_code == 404
         data = response.json()
         assert data["message"] == (
-            "A column with name `non_existent_column` on node "
-            "`company_revenue` does not exist."
+            "Column non_existent_column does not " "exist on node company_revenue"
```
agorajek, line 972: What's this `" "` in the middle?

agorajek (approved):
Thanks @samredai, this will make things easier for users . Two nits in-line.

dj/api/helpers.py
samredai: Great idea, I added in a `break`.

tests/api/nodes_test.py
samredai: Thanks for catching this! It was the auto-formatter that realized these two strings fit on one line, lol

PR #316, opened 2023-02-10
Change catalog attribute on table to Catalog instance

samredai:
### Summary

This is a continuation of the refactoring towards a catalog and engine design to replace the database entity that currently exists. Specifically, this PR changes the catalog attribute on tables from a `str` to a `Catalog` instance. I've left it optional for now so that it can co-exist with the database based logic that exists now as we migrate. I've also added a `uri` attribute to the `Engine` class--we can use that to facilitate continued support for the sqlalchemy mechanism that currently works.

The API endpoint for adding tables is more strict in that it requires a catalog. Catalogs are not required to have engines and you can see in the updated runbook that it's very easy to create one.
```py
response = requests.post(
    f"{DJ_URL}/catalogs/",
    json={"name": "test"},
)
response.json()
```

I think we should eventually have a `default` catalog that's automatically created when a DJ server is initialized, and then we can make it optional to provide a catalog to the API when creating a table, using `default` when `catalog_name == None`.

For the SQL AST build plan (cc: @CircArgs), instead of looking for `Table` instances on nodes, then finding a common shared `Database` defined, we will need a slight change in that logic to find a common shared `Catalog`. We could either flat out not support cross-catalog joins and return an error when that's the only option, or we can be softer there by generating the SQL anyway and including some metadata in the response that implies it's a multi-catalog query (system clients can check for this metadata and make a decision on what to do). I'm not yet sure what's the best option there.

### Test Plan

Updated tests, ran `docker compose up` and tested the API. Also updated and ran the runbook.

- [x] PR has an associated issue: #298
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/alembic/versions/2023_02_10_0345-e19b5daa77a7_catalog_and_table_relationships.py
+++ b/alembic/versions/2023_02_10_0345-e19b5daa77a7_catalog_and_table_relationships.py
@@ -1,0 +1,68 @@
+"""catalog and table relationships
+
+Revision ID: e19b5daa77a7
+Revises: 2bdcec2d24a7
+Create Date: 2023-02-10 03:45:17.696922+00:00
+
+"""
+# pylint: disable=no-member, invalid-name, missing-function-docstring, unused-import, no-name-in-module
+
+import sqlalchemy as sa
+import sqlalchemy_utils
+import sqlmodel
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e19b5daa77a7"
+down_revision = "2bdcec2d24a7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("catalog") as batch_op:
+        batch_op.add_column(
+            sa.Column("uuid", sqlalchemy_utils.types.uuid.UUIDType(), nullable=True),
+        )
+        batch_op.add_column(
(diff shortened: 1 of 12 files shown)


agorajek (approved):
@samredai I thought more about the catalog/engine relationship and I think that we should just give the users what they want, as long as all tables mentioned in their query can be accessed through one engine. To be super true to reality we could add an attribute to our Engine entity called `cross_catalog` and that would make it clear if a given engine can support these queries. But as you noted if the client asks for a query no matter what, perhaps we could give it to them despite the value of the `cross_catalog` attribute.

One question in-line but feel free to merge and we can build on this.

shangyian (approved):
Nice addition! 

For cross-catalog joins, we need some way of deciding whether catalogs are compatible with each other, perhaps dependent on whether or not they can share the same engine. I would say for the time being we should err on the side of generating the SQL but provide a warning about the cross-catalog join.

samredai:
@agorajek @shangyian I agree with both of you. Even if we design it now to just find the common catalog and only warn when there is no common catalog, it shouldn't be too hard to extend that functionality to then do an analysis over the engines attached to the catalogs to find if there's a common engine (despite no common catalog). Thanks for the reviews!

CircArgs (approved):

Issue #301, opened 2023-02-06
Squash Alembic revisions


shangyian:
At this point the data models have all changed enough that for maintenance/cleanliness, we should consider squashing the current alembic revisions into a single revision, as suggested here: https://github.com/DataJunction/dj/pull/289#pullrequestreview-1284549578. 

However, if anyone has an active version of DJ deployed somewhere, where newer revisions need to be relayed on top of an existing database, we can close this issue.

betodealmeida:
Ah, good point... do you think we should wait until we have a 0.1 release? Or do it again by then?

shangyian:
Good point! Let's wait until an 0.1 release. I got ahead of myself and created a PR, but then shortly afterwards started a branch that creates another revision... 😄

PR #318, opened 2023-02-10
Squash alembic revisions

shangyian:
### Summary

Squash all alembic revisions. I tried to run the existing set of revisions against a CockroachDB cluster and came across a number of issues. For example, types cannot be recreated, but some of the existing revisions recreate `nodetype` when the enum is declared in the revision. In lieu of making changes to each of the individual revisions, I think it makes sense just to squash. (cc @betodealmeida)

@samredai I will wait for you to get your changes in before merging this. 

### Test Plan

Ran `alembic migrate head` against a Cockroach cluster.

- [X] PR has an associated issue: #301
- [X] `make check` passes


samredai:
@shangyian I agree with doing this now. We've been aggressively making changes to the metadata model but with the node versioning and catalog/engine changes that are now in, I don't see any more major planned changes. That being said, a few may have to sneak in before the 0.1, but no reason why we can't squash again at that point.

samredai (approved):
I've merged my PRs so this should be good to go in now. 🚀

agorajek:
Database migrations... so cool in theory, but so tough in practice.

shangyian:
@agorajek Indeed! And we won't have this luxury when we have a real database going...

Issue #321, opened 2023-02-11
Endpoints where common metrics/dimensions can be discovered given a set of metrics/dimensions


samredai:
Two common scenarios exist:
1. Someone has a set of **metrics** and wants to know what **dimensions** they share (and can therefore be collectively **grouped by**)
2. Someone has a set of **dimensions** and wants to know what **metrics** they share (and can therefore collectively **group**)

This information is available with the current design and just needs to be intuitively surfaced via one or more endpoints.

agorajek:
The 1st scenario is already supported thanks to @shangyian 's https://github.com/DataJunction/dj/commit/98f1565ffc64f1d796a01e16f8baf1a88d850483.

The title of this issue can be changed to focus on the common metrics.

Issue #320, opened 2023-02-11
We need a new "cube" node type that can group one or metrics with one or more dimensions


samredai:
Currently, the API lets you request any metric nodes grouped by any joinable dimension node (discovered by the dimension node labels on columns). We need a "cube" node type meant to exist further downstream than metrics and dimension nodes and represent various meaningful combinations of compatible metrics and dimensions, i.e. those requiring more aggressive materialization/caching to serve low-latency dashboards.

Issue #272, opened 2023-01-24
Support defining additional aggregations on top of metric nodes


shangyian:
For some of our metrics, we want to perform statistical inference downstream. 

The available inference models for each metric are stored externally to DJ, with context communicated via API calls. In order to streamline the computation there, we need to provide another layer of aggregations on top of the raw metric data. This set of aggregations will typically compute sufficient statistics (i.e., sum, sum of squares etc), but the exact aggregations will vary depending on the inference model. 

We also need to support two different types of source data:
* **Pre-Aggregated**: The metric data is already pre-aggregated to a selection of sufficient statistics (i.e., there's a table with a list of dimensions and all of the metric fields in that table are already aggregated with sufficient stats to the granularity of those dims, stored as `<metric_name>.sum`, `<metric_name>.sum_of_squares` etc) 
* **Raw**: The source data is raw metric data. Arbitrary sufficient statistics can be computed in a separate aggregation stage, configured on top of the metric node.

**Endpoint Changes:**

* For `POST /metrics` (metric creation), we need to:
  - Allow for configuration of the additional aggregations needed for sufficient stats
  - Allow for the ability to specify that certain columns already correspond to sufficient stats (meaning that these would not get recomputed)
* For `GET /metrics/{node_id}/sql`, add an extra argument `include_extra_aggregations` that would specify if the SQL returned should include the additionally configured aggregations for the metric or if it should be raw metric data only. This endpoint also needs to check if we even have raw metric data (in the pre-aggregated data case) and return an error indicating that no such data exists if it is requested.

samredai:
Thanks for highlighting these use cases! It sounds like at a high level there are two things here so I'll reply to each one separately.

**First thing: Adding pre-aggregated source nodes and defining metrics on top of it**: For this, I wonder if we can infer that a metric is pre-aggregated by looking at the query in the metric node definition. Currently a metric node is defined by a highly restrictive query that must select a single column from a single node, i.e. `SELECT SUM(foo) FROM some_node`. Since we're parsing the SQL during the build step, we can detect if a metric node's query is defined without any aggregation and I think it makes sense to make it a hard assumption that no aggregation function means the metric has been pre-aggregated. So for example a query `SELECT foo FROM some_node` would make DJ assume that `foo` must have been pre-aggregated (otherwise it's not a metric, right?). I think this gets us a mechanism for users to communicate that a metric is pre-aggregated without adding any additional attributes like `"pre-aggregated": "true"`, for example.

[...]

PR #322, opened 2023-02-11
Cube nodes and cube endpoints

samredai:
### Summary

This adds the cube node type and endpoints for viewing cube information given a cube name. I've updated the runbook with examples of creating cubes.
```py
response = requests.post(
    f"{DJ_URL}/nodes/",
    json={
        "cube_elements": ["total_revenue", "payment_type", "customers"],
        "description": "A cube of Total Revenue grouped by customer and payment type",
        "mode": "published",
        "name": "total_revenue_by_customer_payment_type",
        "type": "cube",
    },
)
response.json()
```

Although the relations are all here, the validation of cube elements only checks that:

- All elements are either metric nodes or dimension nodes (no source, transform, or other cube nodes)
- At least one metric node and one dimension node is included in the list of `cube_elements`

This doesn't actually check if the metrics and dimensions being added to a cube are joinable. Once #321 is addressed, that logic can also be used here as part of the validation step as well as while managing the status of cube nodes (valid/invalid). We can also add additional attributes here where it makes sense to support #272.

### Test Plan

Wrote some tests, added some working examples to the runbook, ran `docker compose up` and tested out the new endpoints.

- [x] PR has an associated issue: #320 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -1,7 +1,7 @@
 """
 Node related APIs.
 """
-
+import http.client
 import logging
 from http import HTTPStatus
 from typing import Dict, List, Optional, Tuple, Union
@@ -21,10 +21,11 @@ from dj.api.helpers import (
 from dj.construction.extract import extract_dependencies_from_node
 from dj.construction.inference import get_type_of_expression
 from dj.errors import DJError, DJException, ErrorCode
-from dj.models import Database, Table
+from dj.models import Catalog, Database, Table
 from dj.models.column import Column, ColumnType
 from dj.models.node import (
     AvailabilityState,
+    CreateCubeNode,
     CreateNode,
     CreateSourceNode,
     Node,
@@ -59,7 +60,7 @@ class TableMetadata(SQLModel):
     """
 
     id: Optional[int]
-    catalog: Optional[str]
+    catalog: Optional[Catalog]
     schema_: Optional[str]
     table: Optional[str]
     database: Optional[Database]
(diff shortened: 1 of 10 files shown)


CircArgs:
I guess we might consider moving the metric specific build stuff to compile and that way it could be directly used to check cube nodes

dj/api/nodes.py
```diff
@@ -214,7 +232,7 @@ def read_nodes(*, session: Session = Depends(get_session)) -> List[NodeMetadata]
     """
     List the available nodes.
     """
-    nodes = session.exec(select(Node).options(joinedload(Node.current))).all()
+    nodes = session.exec(select(Node).options(joinedload(Node.current))).unique().all()
```
shangyian, line 235: Hmm, why is `unique()` needed here? Isn't that already going to be the case given that we have a constraint on unique node names?

dj/api/nodes.py
```diff
@@ -224,7 +242,7 @@ def read_node(name: str, *, session: Session = Depends(get_session)) -> NodeMeta
     Show the active version of the specified node.
     """
     statement = select(Node).where(Node.name == name).options(joinedload(Node.current))
-    node = session.exec(statement).one_or_none()
+    node = session.exec(statement).unique().one_or_none()
```
shangyian, line 245: ^ same as above here

shangyian (approved):
Changes look good! Left some minor comments. 

We may need more metadata for Druid ingestion, but we can build on top of this later.

dj/api/nodes.py
samredai: Yeah I was surprised I need to do this as well. Let me know if there's a better solution but I was fighting this error once I added "cube_elements" to `NodeRevision`
```
sqlalchemy.exc.InvalidRequestError: The unique() method must be invoked on this Result, as it contains results that include joined eager loads against collections
```

samredai:
> I guess we might consider moving the metric specific build stuff to compile and that way it could be directly used to check cube nodes

@CircArgs great idea. It looks like we're going to have a small number of entry points for the build (maybe 3-5?), one of them being cube nodes. I imagine 90% of the logic will be common amongst all of them with some slight tweaks based on the entry point.

dj/api/nodes.py
shangyian: It looks like this change was added that requires `unique()`: https://github.com/sqlalchemy/sqlalchemy/issues/4395

Issue #323, opened 2023-02-11
Add endpoint to list all downstream nodes from a given node


shangyian:
This endpoint will be useful when users want to know what downstream metrics/transforms/dimensions will be affected if a given source or transform node is modified. It can also be used for discoverability purposes, i.e., finding all metrics that use a source node.

shangyian:
Turns out there's actually a better way to do this with a recursive CTE.

```
WITH RECURSIVE node_graph AS (
    SELECT
        parent_id AS path_start,
        rev.node_id AS path_end,
        rev.name AS node_name
    FROM noderelationship nr
    JOIN noderevision rev ON nr.child_id = rev.id
    WHERE parent_id=3

    UNION ALL

    SELECT
        ng.path_start,
        rev.node_id AS path_end,
        rev.name AS node_name
    FROM node_graph ng
    JOIN noderelationship nr ON ng.path_end = nr.parent_id
    JOIN noderevision rev ON nr.child_id = rev.id
)
SELECT DISTINCT * from node_graph;
```

PR #324, opened 2023-02-11
Add endpoint to list all downstream nodes from a given node

shangyian:
### Summary

Adds an endpoint to list all nodes that are downstream from the given node, filterable by `node_type`. 

### Test Plan

Added a unit test. Ran a few manual checks.

- [X] PR has an associated issue: #323
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -1,7 +1,7 @@
 """
 Node related APIs.
 """
-
+import collections
 import logging
 from http import HTTPStatus
 from typing import Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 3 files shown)


dj/api/nodes.py
```diff
@@ -546,3 +546,22 @@ def node_similarity(
...
+
+@router.get("/nodes/{name}/downstream/")
+def downstream_nodes(
+    name: str, *, node_type: NodeType = None, session: Session = Depends(get_session)
+) -> List[NodeMetadata]:
+    """
+    List all nodes that are downstream from the given node, filterable by type.
+    """
+    node = get_node_by_name(session=session, name=name)
+    queue = collections.deque([node])
```
samredai, line 559: I'm curious why you used a queue here instead of just a filter. Wouldn't a filter be clearer and not require the for-loop over the children?

shangyian: Do you mean a filter for all of the node's children in the NodeRelationship table? But we'd still need to process all of the its children's children, hence the queue. I may be missing something about that table though.

samredai: Ah I see, I was under the impression that `current.children` returned all children, not just the immediate ones. It makes sense because looking at when we create a node, we only include immediate parents. This might be ok but I'm wondering if we should update the `extract_dependencies_from_node` function to return all upstream parents. That would then make this endpoint work with just a filter (cc: @CircArgs). Approving this either way so that you're not blocked. 👍

samredai (approved):

dj/api/nodes.py
CircArgs: @samredai `extract_dependencies_from_node` at one time would recurse till hitting sources if you specified a depth, but I ultimately just made it do the direct dependencies because it wasn't clear if we would want to pay the cost upfront of finding all the dependencies or just the immediate ones and if we did want all of them I figured the user could easily just call it over and over. Should be an easy add though if we think we can use it in a few places and want to just consolidate

shangyian: @CircArgs yeah, I actually think we shouldn't store links to all children of a node unless we determine that we really need it from a performance perspective (i.e., people are making the same call over and over again). We'll probably need some more metadata on the node relationships as well to indicate that it's not an immediate child.

PR #325, opened 2023-02-11
Set pydantic response model for downstream endpoint

samredai:
### Summary

Not sure how the tests in github actions were passing but `make test` on `main` started failing for me locally with this:
```
E               pydantic.error_wrappers.ValidationError: 2 validation errors for NodeMetadata
E               response -> 0 -> current
E                 field required (type=value_error.missing)
E               response -> 1 -> current
E                 field required (type=value_error.missing)
```

@shangyian does this fix look right?

### Test Plan

`make test` and `make check` locally.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -620,10 +620,10 @@ def node_similarity(
     return JSONResponse(status_code=200, content={"similarity": similarity})
 
 
-@router.get("/nodes/{name}/downstream/")
+@router.get("/nodes/{name}/downstream/", response_model=List[Node])
 def downstream_nodes(
     name: str, *, node_type: NodeType = None, session: Session = Depends(get_session)
-) -> List[NodeMetadata]:
+) -> List[Node]:
     """
     List all nodes that are downstream from the given node, filterable by type.
     """


shangyian:
@samredai that's weird - it works for me when I run `make test` on `main` locally. 

Also it's meant to be `NodeMetadata` since that's the output model rather than `Node`, which is just the entire underlying data model.

shangyian:
Let me dig into what's happening though.

samredai:
> Also it's meant to be `NodeMetadata` since that's the output model rather than `Node`, which is just the entire underlying data model.

Do you then need to call `node.current` on what you get back from `get_node_by_name`? That function returns a `Node` (maybe we should add a `get_node_revision` helper function that takes an optional `version` arg and returns a `NodeRevision`).

shangyian:
> Do you then need to call node.current on what you get back from get_node_by_name? 

It should just be a `Node` object (`NodeMetadata` corresponds to `Node` and `NodeRevisionMetadata` corresponds to `NodeRevision`). I actually want to go through and rename all the `XYZMetadata` objects to be `XYZOutput` to indicate that it's for API response output, since `Metadata` is not very intuitive.

samredai:
> `NodeMetadata` corresponds to `Node` and `NodeRevisionMetadata` corresponds to `NodeRevision`

Right, right, I was mixing that up in my head. 🤦

samredai:
Huh, so a fresh pip install fixed it. 😕 

Sorry for the noise!

PR #326, opened 2023-02-12
Add __str__ for Catalog model

samredai:
### Summary

My recent PR that changed catalog from `str` to a `Catalog` instance introduced some bugs in the build where it's being used as a string to construct the table reference. This adds a `Catalog.__str__` method to fix this.

### Test Plan

`make test` and `make check`. Also ran a few query builds to make sure the tables in the FROM clauses are being rendered correctly.

- [ ] PR has an associated issue: #298 
- [ ] `make check` passes

### Deployment Plan

N/A


CircArgs (approved):
Do you think we should split this PR up?

PR #329, opened 2023-02-12
Use a single recursive CTE query to find all downstream nodes instead

shangyian:
### Summary

This PR updates the approach to finding all downstream children for a node. The new approach uses a single recursive CTE query, which eliminates the multiple round-trips to the database for each layer of the graph. 

Here's the CTE query in raw SQL (which I started with) and then translated to SQLAlchemy:

```
WITH RECURSIVE node_graph AS (
    SELECT
        parent_id AS path_start,
        rev.node_id AS path_end,
        rev.name AS node_name
    FROM noderelationship nr
    JOIN noderevision rev ON nr.child_id = rev.id
    WHERE parent_id=3

    UNION ALL

    SELECT
        ng.path_start,
        rev.node_id AS path_end,
        rev.name AS node_name
    FROM node_graph ng
    JOIN noderelationship nr ON ng.path_end = nr.parent_id
    JOIN noderevision rev ON nr.child_id = rev.id
)
SELECT DISTINCT * from node_graph;
```

### Test Plan

- [X] PR has an associated issue: #323
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -2,14 +2,26 @@
 Helpers for API endpoints
 """
 
-from typing import List, Optional, Tuple
 
+from typing import Dict, List, Optional, Tuple, Union
+
+from sqlalchemy.orm import joinedload
 from sqlmodel import Session, select
 
 from dj.construction.build import build_node_for_database
-from dj.errors import DJException
+from dj.construction.extract import extract_dependencies_from_node
+from dj.construction.inference import get_type_of_expression
+from dj.errors import DJError, DJException, ErrorCode
 from dj.models import Catalog, Column, Database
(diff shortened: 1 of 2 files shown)


samredai (approved):
LGTM! What do you think about putting this in a `get_downstream_nodes(session: Session, node_name: str)` function in `dj/api/helpers.py` and then just calling that here? There are few other places where we can re-use this. For example when a node is updated and becomes valid, we can use this to get the downstream nodes and check if any of the invalid ones have also become valid.

shangyian:
Makes sense to me! I also moved the `validate` function out to helpers.

Issue #312, opened 2023-02-08
Support defining engine-specific info for nodes with queries


shangyian:
Nodes with queries may be executed with different engines like Spark, Trino, or Druid. Some of these engines require additional configuration for execution - for example, users may want the ability to tune Spark memory parameters. 

We can add an optional field to a node called `engine_configuration` that accepts an arbitrary dictionary of properties to pass to an engine during execution time. This field should be version-controlled.

shangyian:
We'll need a new `MaterializationConfig` entity that stores these configurations linked up to a node. Only `TRANSFORM`, `DIMENSION`, and `CUBE` nodes will have this config, as other nodes won't need materialization.

### Model
* `node_revision_id`: reference to the node revision that this configuration is set for
* `engine_id`: stores a reference to the engine this is for
* `config`: the body of the config, an arbitrary dictionary of properties

### Endpoints
* `POST /nodes/{name}/materialization`
* `GET /nodes/{name}/` will include a list of materialization configs, if any

### Engines

* **Spark:** For Spark jobs this would be a dictionary of various spark properties. We should seed this with sensible defaults, but everything can be overridden by the user.

* **Druid:** For cube nodes, this would be the Druid ingestion spec. Some of the spec should be auto-generated based on the cube configuration and we should not allow direct user modification without changing the cube itself. However, other aspects of the spec (like`tuningConfig`) may need to be manually modified by the user. We can again supply sensible defaults.

PR #330, opened 2023-02-13
Adding materialization config to nodes + endpoint to create them

shangyian:
### Summary

Nodes can be materialized with different engines, some of which require additional configuration for execution. For example, users may want the ability to tune Spark memory parameters or alter the Druid `tuningConfig`. 

This change adds a new `MaterializationConfig` entity that stores these configurations linked up to a node. Source nodes will not have this config, as they don't need to be materialized (i.e., their materialization is not managed as a part of DJ). At some later point we may want to consider adding validation to these configs, although it will be engine-specific so may be difficult to pull off.

I also did some light refactoring to move some CRUD models to their respective model files.

### Test Plan

Wrote some unit tests. Added the scenarios for creating/updating materialization configs and creating an engine to the runbook.

- [X] PR has an associated issue: #312
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -6,134 +6,45 @@ import logging
 from http import HTTPStatus
 from typing import List, Optional, Union
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends
 from fastapi.responses import JSONResponse
-from pydantic import BaseModel, Field
 from sqlalchemy.orm import joinedload
-from sqlmodel import Session, SQLModel, select
+from sqlmodel import Session, select
 
 from dj.api.helpers import (
     get_catalog,
     get_column,
     get_database_by_name,
     get_downstream_nodes,
+    get_engine,
     get_node_by_name,
     validate_node_data,
 )
 from dj.errors import DJException
-from dj.models import Catalog, Database, Table
+from dj.models import Table
 from dj.models.column import Column, ColumnType
 from dj.models.node import (
-    AvailabilityState,
     CreateCubeNode,
     CreateNode,
     CreateSourceNode,
+    MaterializationConfig,
     Node,
+    NodeOutput,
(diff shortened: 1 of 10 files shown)


dj/api/main.py
```diff
@@ -26,13 +26,6 @@
 )
 from dj.api.graphql.main import graphql_app
 from dj.errors import DJException
-from dj.models.catalog import Catalog
-from dj.models.column import Column
-from dj.models.database import Database
-from dj.models.engine import Engine
-from dj.models.node import NodeRevision
-from dj.models.query import Query
-from dj.models.table import Table
```
samredai, line 35: I can't remember the exact reason mentioned, but I vaguely remember @betodealmeida explaining why these seemingly unused imports needed to be here for `main.py`.

dj/api/nodes.py
```diff
@@ -176,46 +87,108 @@ def validate_node(
...
+        config
+        for config in node.current.materialization_configs
+        if config.engine.name == data.engine_name
+    ]
+    if (
+        existing_config_for_engine
+        and existing_config_for_engine[0].config == data.config
+    ):
+        return JSONResponse(
+            status_code=200,
```
samredai, line 137: nit (and not sure how useful): Should we start to use 204 "No content" status codes for successful idempotent operations that didn't make any changes server side? I've seen this leveraged pretty nicely by front ends that see 204 and know to show a "toast" notification instead of going to a different page.

samredai (approved):
Looks good! Just a couple comments

dj/api/main.py
shangyian: Good to know, I added them back.

dj/api/nodes.py
shangyian: Yeah makes sense. I went ahead and updated it to a 204.

PR #333, opened 2023-02-14
DJ Roads Example Database

samredai:
### Summary

As we start shifting focus onto the dimension path logic, I figured we would benefit from having some richer example datasets so I've created this DJ Roads Example Database which contains 12 tables with various join keys. Here's an ERD of it.

image

This PR adds a `postgres_roads` container and an init SQL script that creates the tables and loads some sample records and it all happens as part of the standard `docker compose up`. I'm planning to follow-up with a dedicated notebook example that creates metrics and dimensions over the example data. Let me know what ya'll think.

### Test Plan

`docker compose up`, exec into the `postgres_roads` container and explored the tables using `psql`.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3"
-name: dj
 volumes:
   postgres_data: {}
+  postgres_roads_example: {}
   redis_data: {}
 
 networks:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3"
-name: dj
```
agorajek, line 2: Did you find this line causing issues on your side as well?

agorajek (approved):
@samredai this example looks great. We can do a session on building some metrics together.

docker-compose.yml
samredai: I didn't but I figured there's no reason to have it if it causes issues in some environments. From what I can tell it's just a display value.

PR #334, opened 2023-02-15
include uri in EngineInfo, update runbook

samredai:
### Summary

This is a small PR that include the `Engine.uri` attribute in the engine endpoints as well as adds some examples to the runbook of using it when adding engines.

### Test Plan

Ran the notebook end to end.

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
The run book is amazing.

PR #336, opened 2023-02-15
Dj query api

CircArgs:
### Summary

This pr aims to add the ability to execute via the api a "DJ SQL" that allows otherwise illegal references to dj nodes. For example

Example:
```SQL
    SELECT Avg(n_comments),
        age
    FROM   (SELECT basic.num_comments            AS n_comments,
                basic.dimension.users.country country,
                basic.dimension.users.age     AS age
            FROM   metrics
            GROUP  BY basic.dimension.users.country, basic.dimension.users.age)
    GROUP  BY age
```

Get's translated from the route to an intermediate query that would be a legal dj node query and ulitmately built into this thing

```sql
SELECT  Avg(n_comments) AS col0,
        age
 FROM (SELECT  basic_DOT_num_comments.cnt AS n_comments,
        basic_DOT_dimension_DOT_users.country AS country,
        basic_DOT_dimension_DOT_users.age AS age
 FROM (SELECT  COUNT(1) AS cnt,
        basic.comments.id AS basic_DOT_source_DOT_comments_DOT_id,
        basic.comments.user_id AS basic_DOT_source_DOT_comments_DOT_user_id,
        basic.comments.timestamp AS basic_DOT_source_DOT_comments_DOT_timestamp,
        basic.comments.text AS basic_DOT_source_DOT_comments_DOT_text
 FROM basic.comments

) AS basic_DOT_num_comments
LEFT JOIN basic.comments
        ON basic_DOT_num_comments.basic_DOT_source_DOT_comments_DOT_id = basic.comments.id AND basic_DOT_num_comments.basic_DOT_source_DOT_comments_DOT_user_id = basic.comments.user_id AND basic_DOT_num_comments.basic_DOT_source_DOT_comments_DOT_timestamp = basic.comments.timestamp AND basic_DOT_num_comments.basic_DOT_source_DOT_comments_DOT_text = basic.comments.text
LEFT JOIN (SELECT  basic.comments.id,
        basic.comments.full_name,
        basic.comments.age,
        basic.comments.country,
        basic.comments.gender,
        basic.comments.preferred_language,
        basic.comments.secret_number
 FROM basic.comments

) AS basic_DOT_dimension_DOT_users
        ON basic.comments.user_id = basic_DOT_dimension_DOT_users.id AND basic.comments.user_id = basic_DOT_dimension_DOT_users.id
 GROUP BY  basic_DOT_dimension_DOT_users.country, basic_DOT_dimension_DOT_users.age)

 GROUP BY  age
```
which successfully runs and returns a small table. 

The hope is for this to also be an escape hatch for users to test different things on the fly.

### Test Plan

- [x] `make check` passes


agorajek (commented):
Whoa, this looks cool @CircArgs, but I am wondering how did the content of dj_query.py got past our linters ;)

dj/construction/dj_query.py
```diff
@@ -0,0 +1,110 @@
...
+from dj.construction.build import build_ast_for_database
+from dj.construction.utils import amenable_name, get_dj_node, make_name
+from dj.models.database import Database
+from dj.models.node import NodeType
+from dj.sql.parsing import ast
+from dj.sql.parsing.backends.exceptions import DJParseException
+from dj.sql.parsing.backends.sqloxide import parse
+
+
+async def build_dj_metric_query(  # pylint: disable=R0914,R0912
```
samredai, line 18: I love where this is going! Some early feedback is I think we should split the validation in this function out to a separate function and then call that in a `validate_dj_query` function in `dj/api/helpers.py` and even a `"/sql/validate/"` endpoint.

dj/construction/dj_query.py
```diff
@@ -0,0 +1,110 @@
...
+            col_name,
+            {NodeType.METRIC},
+            raise_=False,
+        ):
+            parent_select = cast(ast.Select, col.get_nearest_parent_of_type(ast.Select))
+            if not getattr(parent_select, "_validated", False):
+                if len(parent_select.from_.tables) != 1 or parent_select.from_.joins:
+                    raise DJParseException(
+                        "Any SELECT referencing a Metric must source "
+                        "from a single unaliased Table named 'metrics.'.",
```
agorajek, line 44: `s/'metrics.'/'metrics'`

dj/construction/dj_query.py
```diff
@@ -0,0 +1,110 @@
...
+                try:
+                    metrics_ref_name = make_name(
+                        metrics_ref.namespace,  # type: ignore
+                        metrics_ref.name.name,  # type: ignore
+                    )
+                except AttributeError:
+                    metrics_ref_name = ""
+                if metrics_ref_name != "metrics":
+                    raise DJParseException(
+                        "The name of the table for a Metric select must be 'metrics'.",
```
agorajek, line 56: `s/for a Metric select/in a Metric query`

tests/cli/urls_test.py
```diff
@@ -30,6 +30,8 @@ def test_run(capsys: CaptureFixture) -> None:
 http://localhost:8000/metrics/{name}/: Return a metric by name.
 http://localhost:8000/metrics/{name}/data/: Return data for a metric.
 http://localhost:8000/metrics/{name}/sql/: Return SQL for a metric.
+http://localhost:8000/sql/data/: Return data for a DJ Query.
```
agorajek, line 33: Pls disregard the location of this comment, its just easy to see here. These new paths are confusing next to the existing `/queries/` path. How about ?
```
/queries/data/: Return data for a DJ Query.
/queries/sql/: Return SQL for a DJ Query.
```
Or did I miss the fact that these are special queries and you made a separate path on purpose?

agorajek (commented):
I also like where this is going :) Left few comments in line.

tests/cli/urls_test.py
samredai: Agreed the current paths are confusing but I also feel that `"/queries/sql/"` will be confusing to new-comers who don't fully understand giving SQL to get different SQL. Since we'll need an endpoint for validating a query without running it, maybe we can offer that as a way to get the built sql? Then we wouldn't need to specify `data` in the endpoint.
```
POST /query/: Return data for a DJ Query
POST /query/validate/: Validate a DJ Query
```
The second one would also return the built SQL. Thoughts?

samredai: Also I noticed the `GET "/queries/{query_id}/"` endpoint which should still work as expected. For consistency with my suggestions above, we should rename that to `GET "/query/{query_id}/"` (if we go with my suggestions, that is 
😄 ).

agorajek: I like your suggestion @samredai . Especially since I am not a fan of the REST paths starting with a plural term. I am not familiar with the REST standards so I won't push but imo this:
```
.../thing/{thing_id}
.../thing/list
```
is more intuitive than:
```
.../things/{thing_id}
.../things/
```

CircArgs: @samredai so is this correct?

1. `/query`: the user sends a str query to get data
2. `/query/validate`: the user sends a str query to see if it is valid and gets back the query we would run
3. `/query/{query_id}`: I don't actually know what this does haha

and so I should make the `/sql` routes I have right now conform to 1, 2?

shangyian: Isn't `/query/{query_id}` more for tracking the results of a specific query? It would only make sense after someone has actually kicked off a query run. One potential flow:
* `/query` -> user sends query str, DJ returns a reference to a `query_id` that identifies the query run
* `/query/validate` -> users sends query str to see if it's valid
* `/query/{query_id}` -> returns the status of the query run
* `/query/{query_id}/data` -> returns the query data

CircArgs: @shangyian that looks very sensible to me!

samredai: @CircArgs yep that's right. And @shangyian good catch, it makes sense to me to split out the data retrieval part to a `/query/{query_id}/data/` endpoint. Right now it just shows up in the status response when the status is "COMPLETED" but that's not ideal. I like the explicit follow-up call to get the actual data.

dj/api/query.py
```diff
@@ -0,0 +1,48 @@
+"""
+Metric related APIs.
```
samredai, line 2: s/Metric/DJ SQL

dj/api/query.py
```diff
@@ -0,0 +1,48 @@
...
+class DJSQL(SQLModel):
+    """
+    Class for DJ SQL request.
+    """
+
+    database_name: Optional[str]
+    sql: str
+
+
+@router.get("/query/validate", response_model=TranslatedSQL)
```
samredai, line 26: This should be `router.post` since it requires a body

dj/construction/dj_query.py
```diff
@@ -0,0 +1,230 @@
...
+        col_name,
+        {NodeType.METRIC},
+    ):
+        # if we found a metric node we need to check where it came from
+        parent_select = cast(ast.Select, col.get_nearest_parent_of_type(ast.Select))
+        if not getattr(
+            parent_select,
+            "_validated",
+            False,
+        ):  # pragma: no cover# we haven't seen this
```
samredai, line 50: Not sure what this comment means, could use some clarification/elaboration

dj/construction/dj_query.py
```diff
@@ -0,0 +1,230 @@
...
+        ]
+        metric_table_expression = ast.Alias(
+            ast.Name(metric_name),
+            None,
+            metric_select,
+        )
+        # we go through all the dep nodes directly in the metric's FROM
+        # we need to surface the node itself to join potential dims
+        # and to surface the node we need to source all its columns
+        # - in the metric for an implicit join
```
samredai, line 89: nit: Since the loop just calls the `_hoist_metric_source_tables`, these comments should probably just go in that function's doc-string

samredai (approved):
LGTM! Just a few comments but approving to unblock you. I think this PR could use a squash as well before merging.

Issue #339, opened 2023-02-16
Enable type inference for arbitrarily complex maps


shangyian:
The current setup disallows arbitrarily complex maps and doesn't do type inference when a map's contents is queried via the subscription operator `[]` (it instead raises a `Failed to parse Expression` exception).

Suppose we have a source node `foo` with a column `measures_map` of type `MAP[STR, FLOAT]`. This following query should be parseable and the type of `measure_name` be inferred from the upstream type of `measures_map` to be `FLOAT`:
```
SELECT measures_map["measures_name"] AS measure_name FROM foo
```

CircArgs:
Our AST is something of a hybrid between sqlglot's and sqloxide's. pdb'ing into the failure for your example, I see the sqloxide parse tree looks like `{'MapAccess': {'column': {'Identifier': {'value': 'measures_map', 'quote_style': None}}, 'keys': [{'Value': {'SingleQuotedString': 'measures_name'}}]}}`.

Based on that, I can see this going a couple of ways

- a. some optional attributes on the existing `Column` ast node to account for this
- A new node (inheriting from existing `Column`?) that is special:
    - b. special in that is has the custom attribute for this
    - c. special in that it is exactly a map access and then we need one for all other kinds 

imo b, a, c would be my order of preference and a, b are very close. c seems gross.

To start to implement this, there's the sort of orchestrator function for all expressions that forks to specialized parsers https://github.com/DataJunction/dj/blob/8c99b979fb7fc3e1140ed4292966253f0f2b1d27/dj/sql/parsing/backends/sqloxide.py#L101

then you might consider extending the column one https://github.com/DataJunction/dj/blob/8c99b979fb7fc3e1140ed4292966253f0f2b1d27/dj/sql/parsing/backends/sqloxide.py#L187 or implementing a new specialized one.

shangyian:
@CircArgs I added a new `MapSubscript` object that inherits from `Expression` and handles the parsing of the `MapAccess` parse tree in https://github.com/DataJunction/dj/pull/340. Do you think it should inherit from `Column` instead though?

CircArgs:
@shangyian 

> @CircArgs I added a new `MapSubscript` object that inherits from `Expression` and handles the parsing of the `MapAccess` parse tree in #340. Do you think it should inherit from `Column` instead though?

I don't think so since you have the column attribute on it. Maybe it should inherit from `Operator`?

PR #340, opened 2023-02-16
Enable type inference for arbitrarily complex maps

shangyian:
### Summary

1. In the current metadata database's `column` table, the `type` field is restricted to an `enum`, which disallows arbitrarily complex column types. This change removes that restriction.
2. Adds type inference for map subscription in a query (syntax `map_name["key_name"]`). 

This following query should be parseable and the type of `measure_name` be inferred from the upstream type of `measures_map`:
```
SELECT measures_map["measures_name"] AS measure_name FROM foo
```

@CircArgs if you could take a look at the changes, it'd be great, thanks!

### Test Plan

- [X] PR has an associated issue: #339
- [X] `make check` passes

### Deployment Plan

N/A


dj/sql/parsing/ast.py
```diff
@@ -1030,6 +1030,20 @@ def __str__(self) -> str:
...
 
 
+@dataclass(eq=False)
+class MapSubscript(Expression):
+    """
+    Map accessors
+    """
+
+    map_column: "Column"
+    keys: List[str]
```
samredai, line 1040: Would it be possible for this to be `List[Name]`? That way we can use the `quote_style` attribute on the [Name](https://github.com/DataJunction/dj/blob/main/dj/sql/parsing/ast.py#L469) class to preserve the type of quotes that were originally used when we re-render the SQL.

dj/sql/parsing/ast.py
```diff
@@ -1030,6 +1030,20 @@ def __str__(self) -> str:
         return prefix + str(self.name)
 
 
+@dataclass(eq=False)
```
CircArgs, line 1033: I think I spoke too soon about my option "c" in your issue. This looks good. Are there any other cases like this where there's some special syntax on a column that would parse to something other than `MapSubscript`?

dj/sql/parsing/ast.py
shangyian: Makes sense. Just added this change.

dj/typing.py
```diff
@@ -237,7 +233,7 @@ def _validate_type(cls, type_: str):
             inner = test.group("inner")
             if outer not in COMPLEX_TYPES:
                 raise ColumnTypeError(f"{outer} is not a KNOWN complex type.")
-            inners = inner.split(",")
+            inners = inner.split(",") if outer == ColumnType.MAP else inner
```
CircArgs, line 236: What's this doing? Was there a problem with map in particular? all the complex types take comma separated lists right?

CircArgs (approved):
Just some questions

dj/sql/parsing/ast.py
shangyian: The only type of map subscription syntax I can think of is `some_map["some_key"]`, but I suppose every dialect has their own version of this type of map access. In fact, I'm not even sure that `some_map["some_key"]` is ANSI SQL, but it turns out sqloxide supports it!

dj/typing.py
CircArgs: `Map` and `Row` are pretty similar right? it doesn't affect `Row`?

shangyian: Good point, it should include `Row` here as well

shangyian: Although I think for `Row` it gets more complex because we'll also need to handle cases where there are maps in rows, like:
```
ROW[STR, MAP[STR, STR]]
```

If it's just doing `inner.split(",")`, it would parse to `["STR", "MAP[STR,", "STR]"]`, which isn't right.

samredai (approved):

agorajek (approved):
Thanks @shangyian for the quick fix for the Maps. If we have any troubles with Row() we can  address it later.

shangyian:
Thanks all for reviewing! I'm going to merge this for now, but will need to revisit for the `ROW` case.

PR #341, opened 2023-02-17
Support type parsing for rows with complex types

shangyian:
### Summary

Type parsing didn't work for some rows with complex types before, i.e.,
```
ROW[MAP[STR, STR], STR, ARRAY[STR]]
```

This change adds support for these types of arbitrarily nested values in rows by switching from solely using a regex matcher to using a custom function that takes a given type string and parses the nested structure based on the parentheses. For example:
```
>>> from dj.typing import parse_type
>>> parse_type('ROW[MAP[STR, STR], STR, ARRAY[STR]]')
[['ROW', [['MAP', ['STR', 'STR']], 'STR', ['ARRAY', ['STR']]]]]
```

This nested list is then used to construct the `ColumnType` object.

### Test Plan

Added some unit tests.

- [X] PR has an associated issue: Related to #339
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/typing.py
+++ b/dj/typing.py
@@ -6,6 +6,7 @@ Custom types for annotations.
 
 from __future__ import annotations
 
+import collections
 import re
 from enum import Enum
 from types import ModuleType
@@ -79,6 +80,15 @@ def process_array_args(*args: str) -> Tuple["ColumnType"]:
     """
     Validate the args of an ARRAY
     """
+    if len(args) != 1:
+        raise ColumnTypeError(
+            f"{', '.join(args)} is not an acceptable type for ARRAY"
+            if args
+            else "ARRAY needs subtype defined.",
+        )
+    array_type = args[0]
+    if isinstance(array_type, list) and array_type[0] in COMPLEX_TYPES:
+        return (ColumnType(array_type[0])[array_type[1]],)
     return (ColumnType(args[0]),)
 
 
@@ -90,21 +100,35 @@ def process_map_args(*args: str) -> Tuple["ColumnType", "ColumnType"]:
         raise ColumnTypeError(
             f"MAP expects 2 inner types but got {len(args)}.",
         )
-    arg1, arg2 = ColumnType(args[0], "key"), ColumnType(",".join(args[1:]), "value")
-    if arg1 not in PRIMITIVE_TYPES - {"WILDCARD", "NULL"}:
-        raise ColumnTypeError(f"MAP key is not an acceptable type {arg1}.")
-    return arg1, arg2
+    key_args, value_args = args[0], args[1]
+
+    map_key_type = (
+        ColumnType(key_args[0], "key")[key_args[1]]
+        if isinstance(key_args, list)
+        else ColumnType(key_args, "key")
+    )
+    if map_key_type not in PRIMITIVE_TYPES - {"WILDCARD", "NULL"}:
+        raise ColumnTypeError(f"`{map_key_type}` is not an acceptable MAP key type.")
+
+    map_value_type = (
+        ColumnType(value_args[0], "value")[value_args[1]]
+        if isinstance(value_args, list)
+        else ColumnType(value_args, "value")
+    )
+    return map_key_type, map_value_type
 
 
 def process_row_args(*args: str) -> Tuple["ColumnType", ...]:
     """
     Validate the args of a ROW
     """
-
     ret = []
     for arg in args:
-        type_, name, *_ = (*arg.split(), None)
-        ret.append(ColumnType(type_, name and name.strip("\"' ")))
+        if not isinstance(arg, list):
+            type_, name, *_ = (*arg.split(), None)
+            ret.append(ColumnType(type_, name and name.strip("\"' ")))
+        else:
+            ret.append(ColumnType(arg[0])[arg[1]])
     return tuple(ret)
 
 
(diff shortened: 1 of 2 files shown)


dj/typing.py
```diff
@@ -227,20 +249,49 @@ def __getitem__(self, keys) -> "ColumnType":
...
-            outer = test.group("outer")
-            inner = test.group("inner")
-            if outer not in COMPLEX_TYPES:
-                raise ColumnTypeError(f"{outer} is not a KNOWN complex type.")
-            inners = inner.split(",") if outer == ColumnType.MAP else inner
-            return ColumnType(outer)[inners]
+        nested = parse_type(type_)[0]
+        if isinstance(nested, list) and len(nested) > 1:
+            if nested[0] not in COMPLEX_TYPES:
+                raise ColumnTypeError(f"{nested[0]} is not a KNOWN complex type.")
```
samredai, line 255: nit: I know this error message was already defined but the all caps looks odd

dj/typing.py
```diff
@@ -227,20 +249,49 @@ def __getitem__(self, keys) -> "ColumnType":
...
+
+    for i, token in enumerate(tokens):
+        next_token = tokens[i + 1] if i + 1 < len(tokens) else None
+        if token == "[":
+            stack.append(items)
+            items.append([])
+            items = items[-1]
+        elif token == "]":
+            if not stack:
+                raise ColumnTypeError("Unbalanced parentheses")
```
samredai, line 281: Would it be easy to include more information here in this error message, maybe even just dump the expression in?
`ColumnTypeError(f"Unbalanced parentheses: {expr}")`

dj/typing.py
```diff
@@ -79,6 +80,13 @@ def process_array_args(*args: str) -> Tuple["ColumnType"]:
     """
     Validate the args of an ARRAY
     """
+    if len(args) > 1:
```
agorajek, line 83: Perhaps `if len(args) != 1:` to cover a missing arg too? Although the message would probably need to change.

dj/typing.py
```diff
@@ -90,21 +98,35 @@ def process_map_args(*args: str) -> Tuple["ColumnType", "ColumnType"]:
...
-    return arg1, arg2
+    key_args, value_args = args[0], args[1]
+
+    map_key_type = (
+        ColumnType(key_args[0], "key")[key_args[1]]
+        if isinstance(key_args, list)
+        else ColumnType(key_args, "key")
+    )
+    if map_key_type not in PRIMITIVE_TYPES - {"WILDCARD", "NULL"}:
+        raise ColumnTypeError(f"MAP key is not an acceptable type {map_key_type}.")
```
agorajek, line 109: I think `'{map_key_type}' is not an acceptable map key type.` reads easier.

samredai (approved):
LGTM. Do we still need to add some logic for column type inference based on ROW/STRUCT types? I'm thinking about the `STRUCT<field1_name: field1_type, field2_name: field2_type, …>` type in Spark that looks like a map when accessed but the type can vary based on the key.

dj/typing.py
```diff
@@ -227,20 +249,49 @@ def __getitem__(self, keys) -> "ColumnType":
...
+            if nested[0] not in COMPLEX_TYPES:
+                raise ColumnTypeError(f"{nested[0]} is not a KNOWN complex type.")
+            return ColumnType(nested[0])[nested[1]]
         type_ = ALIASES.get(type_, type_)
         if type_ not in PRIMITIVE_TYPES:
             raise ColumnTypeError(f"{type_} is not an acceptable type.")
         return ColumnType(type_)
 
 
+def parse_type(expr: str) -> List:
```
agorajek, line 263: `s/parse_type/parse_nested_type` ?

dj/typing.py
CircArgs: @samredai  WHY

agorajek (approved):
lovin-it

Just few nits in-line.

CircArgs:
> LGTM. Do we still need to add some logic for column type inference based on ROW/STRUCT types? I'm thinking about the `STRUCT<field1_name: field1_type, field2_name: field2_type, …>` type in Spark that looks like a map when accessed but the type can vary based on the key.

Just for reference, here is how the `Row` looks with names right now:

`ColumnType.Row[ColumnType.STR, "INT id", ColumnType.ARRAY[ColumnType.bytes]]`

CircArgs (approved):

samredai:
> Just for reference, here is how the `Row` looks with names right now:
> 
> `ColumnType.Row["STR name", "INT age"]`

Nice! So this means if this column is accessed in a downstream node as `upstream_column.name`, then we'd be able to infer that as a type `STR`? I just remembered there's also `upstream_column.*` that expands the row out into columns, but I don't know how commonly that's used.

dj/typing.py
samredai: NOT EXACTLY SURE 😂

shangyian: I REMOVED IT!

dj/typing.py
shangyian: Makes sense, I'll dump the expression in

dj/typing.py
shangyian: Good idea! I added an altered error message for the missing arg case: `ARRAY needs subtype defined`.

shangyian:
> I just remembered there's also upstream_column.* that expands the row out into columns, but I don't know how commonly that's used.

When I get a chance I'll have a look at the column type inference for `ROW` types. That can be the third round of this set of changes 😃

PR #346, opened 2023-02-19
Various type inference and metric validation-related fixes

shangyian:
This PR contains various type inference and metric validation-related fixes:

**Fixing type inference for `CAST` expressions.**

For `CAST` expressions, the sqloxide parse tree looks like `{'Int': None}`, but there's no set up to parse the dict's keys, which this PR fixes.

**Adding type inference for `CASE WHEN` expressions**

For `CASE WHEN` expressions, there is no type inference, so this adds an inference check based on the results and else_result's types. It also verifies that both the if-true case and the if-false case have the same type, or type inference breaks. Related to https://github.com/DataJunction/dj/issues/271.

**Adding option to decline active database checks**

This adds a boolean `check_database_online` setting to the `GET /metrics/{name}/sql/` endpoint. Sometimes we just want to request the metric's SQL without also having active databases. This is especially useful in the testing phase, but will likely help users who just want to see SQL without setting up their tables and wiring together the db connections with DJ.

**Adds two new types**
Adds `DOUBLE` and `LONG`.

### Test Plan

- [X] PR has an associated issue: #271
- [x] `make check` passes

### Deployment Plan

N/A


dj/construction/build_planning.py
```diff
@@ -137,20 +140,27 @@ async def optimize_level_by_cost(build_plan: BuildPlan) -> Tuple[int, Database]:
...
+                        ),
+                    ),
                 )
                 some_db = True
             except Exception as exc:  # pylint: disable=broad-except    # pragma: no cover
                 if "No active database found" not in str(exc):
                     raise exc
 
-    if not some_db:  # pragma: no cover
+    if check_database_online and not some_db:  # pragma: no cover
```
samredai, line 156: Seeing this flag getting passed around to eventually be used here, I'm wondering if this `some_db` could just be returned at the end and the check can happen at the end instead of deep within this flow of logic. In other words, if this is updated so that `some_db` is blindly passed along, then you would only need the `check_database_online` argument for the `dj.api.helpers.get_query` function and you can move this check out to that function:

*psuedo-code, the dj.api.helpers.get_query function*
```py
async def get_query(........, check_database_online=False)
    ...
    query_ast, optimal_database, some_db = await build_node_for_database(...)
    if check_database_online and not some_db:
        raise Exception("No database found that can execute this query.")
```

samredai (commented):
I'm so excited to be able to build queries as examples without having an online database. I left a few comments, the main one being on if we could move the online database check to a post-build step so that it's truly separate and the check flag arg can be in one place.

samredai (approved):

PR #348, opened 2023-02-20
Add notebook example, modeling the roads example database

samredai:
### Summary

This adds an example notebook that creates source, dimension, transform, and metric nodes around the data in the `postgres_roads` container that's loaded with the DJ roads example database.

### Test Plan

`docker compose up`, `juptyer notebook` and then `Run All` in the notebook.

- [x] `make check` passes

### Deployment Plan

N/A


shangyian (approved):
Nice!

PR #354, opened 2023-02-24
Add status logic for draft nodes and missing parent resolution

samredai:
### Summary

This is mostly ready, just need to write a few tests and do a couple more passes over the logic. Here's the gist of this PR:

### Current Behaviors

Currently, the result of building a node provides a set of "missing parents", which are essentially any references to upstream nodes that don't exist. As part of validation, a node is never considered valid if it has any of these missing parent references. This status of `invalid`/`valid` is tracked as the status attribute on a `NodeRevision`. Another aspect of the current behavior is the consideration of the node's `mode` which is defined by the requester as either `published` or `draft`. A new node definition can never be added in `published` mode if validation finds it to be `invalid`. However a requester can add a node (or modify an existing one) if the node's mode is set to `draft`, even if its status is determined as `invalid`. So in summary, the requester determines the `mode` and the system determines the `status`. The former dictates the requirement of the latter, and the latter is determined during node validation.

### This PR

This PR addresses the fact that `invalid` nodes with missing parents will no longer be invalid once all of the missing parent nodes have been added. We store the links between missing parent references and the nodes that contain them. This PR adds logic to consider this status change to other nodes whenever a new node is added. When a node is added, DJ will:
- Find all nodes that have a missing parent reference to the new node being added (a.k.a the new node's "downstream nodes")
- Remove the missing parent reference and add a parent/child link between the new node and all of its downstream nodes
- Revalidate all of the downstream nodes
  - If they're now valid, switch the nodes status to `valid`
  - Otherwise, leave the nodes status to `invalid` (An example is when it still has other missing parents)

**note:** This PR only updates the "create" endpoint for a node--I'll follow up with updates to the "update" endpoint.
### Test Plan

...

- [x] `make check` passes

### Deployment Plan

N/A


dj/api/helpers.py
```diff
@@ -234,30 +246,85 @@ def validate_node_data(
...
+    validated_node.columns = []
+    type_inference_failed_columns = []
+    for col in query_ast.select.projection:
+        try:
+            column_type = get_type_of_expression(col)
+            validated_node.columns.append(
+                Column(name=col.name.name, type=get_type_of_expression(col)),
+            )
+        except DJParseException:  # p
+            type_inference_failed_columns.append(col.name.name)
```
shangyian, line 282: This will be useful!

dj/api/helpers.py
```diff
@@ -234,30 +246,85 @@ def validate_node_data(
...
+                missing_parents_map,
+                type_inference_failed_columns,
+            ) = validate_node_data(data=downstream_node_revision, session=session)
+            if not missing_parents_map and not type_inference_failed_columns:
+                downstream_node_revision.status = NodeStatus.VALID
+            session.add(downstream_node_revision)
+            downstream_references_updated += 1
+
+        session.delete(missing_parent)  # Remove missing parent reference to node
+    return downstream_references_updated
```
shangyian, line 330: Is this value just for auditing? As in, we don't actually use this count anywhere right?

shangyian (approved):

dj/api/helpers.py
samredai: Yeah good catch, I really just returned that because otherwise the function felt too opaque to the calling logic. I was thinking we could include it with the metadata that's returned when you create a node, that way you get at least some information it's had an effect to the downstream portion of the DAG. In the future we can enrich that even more by specifying which nodes were specifically updated and how.

dj/api/helpers.py
```diff
@@ -16,7 +16,9 @@
 from dj.errors import DJError, DJException, ErrorCode
 from dj.models import Catalog, Column, Database, Engine
 from dj.models.node import (
+    MissingParent,
```
agorajek, line 19: Not related to this PR, but the name difference between `MissingParent` and `NodeMissingParents` stands out.

agorajek (approved):
I like this. Thanks for pushing ahead!

dj/api/helpers.py
samredai: Yeah `NodeMissingParents` is a relationship table between `Node` and its `MissingParent`s. In the discussion in PR [#366](https://github.com/DataJunction/dj/pull/366#discussion_r1118230186) I proposed changing `MissingParent` to `MissingUpstream`. Then the relationship table would be `NodeMissingUpstreams`. Let me know if that sounds better.

dj/api/helpers.py
samredai: Changed this to return a list of nodes that have been newly switched from invalid to valid. That's then used to know how to propagate the effect and revalidate all downstream nodes.

dj/api/helpers.py
```diff
@@ -237,30 +249,123 @@ def validate_node_data(
...
+                (
+                    validated_node,
+                    _,
+                    missing_parents_map,
+                    type_inference_failed_columns,
+                ) = validate_node_data(data=node.current, session=session)
+                if not missing_parents_map and not type_inference_failed_columns:
+                    node.current.columns = validated_node.columns or []
+                    node.current.status = NodeStatus.VALID
+                    session.add(node)
```
samredai, line 367:
```suggestion
                    session.add(node.current)
```

PR #355, opened 2023-02-25
Move all models to the model module

shangyian:
### Summary

This change moves all of the models in the `api` module to the `models` module so that we can have some logical separation between the two. I originally had this as a part of the tagging PR but pulled it out for clarity.

### Test Plan

- [X] `make check` passes

### Deployment Plan

N/A


samredai (approved):
🧹

PR #362, opened 2023-02-26
Update Python Client - d93615f

agorajek:


shangyian:
This appears to be an endless cycle of PRs 🤔

agorajek:
Oh wow, I see. Perhaps one way to break it would be to specify the trigger for the workflow in a way to skip the `client` dir itself.

PR #365, opened 2023-02-26
Update Python Client - c3d7051

agorajek:


samredai:
Haha, looks like it's stuck in a loop, merging this will cause another PR to get opened. 😅

Not a huge deal for now but I think we need to set a directory match for this. It should only need to generate this PR if a change is made to the `dj/` directory.

agorajek:
This should help in breaking the vicious cycle.

agorajek:
Should we just delete this PR?

PR #366, opened 2023-02-26
Content for Node Dependencies concept page

samredai:
### Summary

This fills in the content for the "Node Dependencies" concept page. As I was writing it I realized the concept is so closely tied to the "Node Validity" page so I removed that and covered that in the "Node Status" section.

### Test Plan

N/A

- [x] `make check` passes

### Deployment Plan

N/A


docs/content/0.1.0/docs/dj-concepts/node-dependencies.rst
```diff
@@ -1,3 +1,53 @@
...
\ No newline at end of file
+---
+
+-----------------
+Node Dependencies
+-----------------
+
+Relationships between nodes are tracked by a DJ server. A node's position in the DJ DAG is determined by the node's definition,
+particularly the query. Node queries reference other DJ nodes as tables and this is what defines upstream and downstream
+dependencies for any given node. In other words--given a node, the other nodes it queries are its **upstream** dependencies
```
CircArgs, line 11: you use `upstream/downstream` here but refer to parents in the bottom section

CircArgs (changes requested):

docs/content/0.1.0/docs/dj-concepts/node-dependencies.rst
samredai: Which do you feel is more clear? I notice we've all started using upstream/downstream in our DJ syncs but in the source code we call these missing upstream dependencies `MissingParent`s. Sounds like I should rename that class to `MissingUpstreamNode` and we use upstream/downstream everywhere in the source code and docs.

CircArgs: I personally have no preference

CircArgs (approved):

docs/content/0.1.0/docs/dj-concepts/node-dependencies.rst
samredai: Updated to change "parent" to "upstream"

docs/content/0.1.0/docs/dj-concepts/node-dependencies.rst
```diff
@@ -1,3 +1,53 @@
...
----
\ No newline at end of file
+---
+
+-----------------
+Node Dependencies
+-----------------
+
+Relationships between nodes are tracked by a DJ server. A node's position in the DJ DAG is determined by the node's definition,
+particularly the query. Node queries reference other DJ nodes as tables and this is what defines upstream and downstream
```
shangyian, line 10: Should we remove the "as tables" part of "reference other DJ nodes as tables"? I'm coming at this from the angle of it potentially being confusing to talk about nodes and tables as if they were similar, when they're actually different things in DJ.

samredai: Good call, updated!

PR #367, opened 2023-02-27
Update generate-clients.yml

agorajek:
### Summary

1) Specified ./dj as the only dir that should be considered for triggering the `generate-clients.yml` workflow. 
2) Removed the old comment line.

### Test Plan

In the wash...

- [ ] `make check` passes

### Deployment Plan

n/a


samredai (approved):

shangyian:
Nice, thanks! I was also thinking that we might want to consider creating a separate repo for the client, but maybe it's not worth it at this stage.

Issue #342, opened 2023-02-17
Add interface for a custom query service


shangyian:
We should add an interface between a custom query service and DJ. DJ currently has the `queries` API endpoints which stores queries to the database, runs them, and manages their states. A custom query engine would largely replace this functionality and we need the mechanism to plug one in.

**Checklist**
- [x] Add query service interface to DJ
- [x] Build out djqs to conform to that interface
- [x] Add examples to runbook that demonstrate both working in conjunction

DJ would provide the following:
* `query`: the generated query in ANSI SQL
* `engine`: the query engine to use (Spark, Presto, Druid etc)
* `config`: query-specific config (pulled from `MaterializationConfig`)

A custom query service will provide the following endpoints:

**POST /queries** 

Save query for running
```
{
    "query": "SELECT ...",
    "engine": {
        "name": "spark",
        "version": "2.4.4"
    },
    "config": {}
}
```
Response:
```
{
    "id": "<query_uuid>"
}
```

**POST /queries/<query_uuid>/**

Run query, returns a query execution

Response:
```
{
    "query_uuid": "<query_uuid>",
    "execution_uuid": "<query_execution_uuid>"
}
```

**GET /queries/<query_uuid>/<execution_uuid>/**

Get query execution status.

Response:
```
{
    "query_uuid": "<query_uuid>",
    "execution_uuid": "<query_execution_uuid>",
    "created_at": "<creation_timestamp>",
    "started_at": "<creation_timestamp>",
    "ended_at": "<creation_timestamp>",
    "status": "CREATED" | "RUNNING" | "SUCCEEDED" | "FAILED" | "UNKNOWN",
    "data": {
        "output_table": null | {
            "name": "<table name>",
            "catalog": "<catalog name>",
            "database": "<database name>"
        },
        "direct": null | ...,
        "columns": {
            "<col_name>": "<col_type>"
        }
    }
}
```

**Remaining Questions**
* Should the query engine handle the conversion of ANSI SQL to any selection of dialects? Or should DJ?
* Should the query engine determine whether to save the query to a table or should it allow for a flag to indicate whether to do so?

samredai:
The endpoints look good to me. For the execution response, the `"direct"` attribute isn't clear to me, is that were the data would be if it's directly returned instead of offloaded to an output table? What do you think of changing `"data"` to `"result"` and then `"direct"` to `"data"`? Also I think columns will have to be an ordered array so that it can be applied when the data is returned which will be an array of arrays.
```json
[...]

PR #368, opened 2023-02-27
Add query service interface and allow for column inference when one is set

shangyian:
### Summary

**Query Service Interface**

This change sets up the interface between a query service and DJ through the `QueryServiceClient` class. The client class will serve as a common interface for any custom query services, including the one that [comes default with DJ](https://github.com/DataJunction/djqs). 

(As an addendum to this PR, we'll need to modify [djqs](https://github.com/DataJunction/djqs) to have the same `/table/<name>/columns` API endpoint, and indeed all the endpoints that are supported through this client.)

**Column Interface**

This also modifies the "add table to node" endpoint to allow for column inference when the user passes in empty columns and. valid query service is configured. This makes adding tables very easy, as users don't need to translate from the source table schema to DJ column types and can rely on the configured query service to do so.

### Test Plan

- [X] PR has an associated issue: #342
- [X] `make check` passes

### DIFF
--- a/.docker-env/.env
+++ b/.docker-env/.env
@@ -3,3 +3,4 @@ DESCRIPTION="A simple DJ server running in Docker"
 REPOSITORY=examples/configs
 REDIS_CACHE=redis://query_broker:6379/0
 CELERY_BROKER=redis://query_broker:6379/1
+QUERY_SERVICE=http://host.docker.internal:8001
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -43,8 +43,9 @@ from dj.models.node import (
     UpsertMaterializationConfig,
 )
 from dj.models.table import CreateTable
+from dj.service_clients import QueryServiceClient
 from dj.sql.parsing.backends.sqloxide import parse
-from dj.utils import Version, VersionUpgrade, get_session
+from dj.utils import Version, VersionUpgrade, get_query_service_client, get_session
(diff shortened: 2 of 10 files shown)


.docker-env/.env
```diff
@@ -3,3 +3,4 @@ DESCRIPTION="A simple DJ server running in Docker"
 REPOSITORY=examples/configs
 REDIS_CACHE=redis://query_broker:6379/0
 CELERY_BROKER=redis://query_broker:6379/1
+QUERY_SERVICE=http://host.docker.internal:8001
```
samredai, line 6: This won't work on linux unless we add the below to the `dj` service in the docker compose file.
```yaml
services:
  dj:
    ...
    extra_hosts:
      - "host.docker.internal:host-gateway"
```
Something about how `host.docker.internal` is specific to the docker desktop apps for mac and windows.

samredai (approved):
This is pure awesome. I've been scratching my head for a while thinking of the best way to properly define these interfaces, I really like the idea to use dependency injection. I wonder how this will look in the openapi spec. Once the djqs library is ready, its openapi spec will be a nice reference for anyone who wants to implement a query service.

samredai (approved):

.docker-env/.env
shangyian: Ah yes, good catch! Let me add that

PR #387, opened 2023-03-06
Endpoint for getting common dimensions for a set of metrics

samredai:
### Summary

This adds a `GET /metrics/common/dimensions/` endpoint to find the common dimensions that are shared by a set of metrics. In this PR I also added nodes from the roads example to the test client fixture. We can consolidate the various node creation through the tests to instead create it as part of the fixture so we have them all in one place, available to all tests.

For an example of finding common dimensions for a set of metrics, try the following request.

```sh
curl -X 'GET' \
  'http://localhost:8000/metrics/common/dimensions/?metric=total_repair_order_discounts&metric=total_repair_cost' \
  -H 'accept: application/json'
```
_response_
```json
[
  "repair_order_details.quantity",
  "repair_order.repair_order_id",
  "repair_order.dispatched_date",
  "repair_order.dispatcher_id",
  "repair_order.required_date",
  "repair_order_details.repair_order_id",
  "repair_order.order_date",
  "repair_order_details.price",
  "repair_order_details.repair_type_id",
  "repair_order_details.discount",
  "repair_order.hard_hat_id",
  "repair_order.municipality_id"
]
```
### Test Plan

Wrote tests, ran curl requests

- [x] PR has an associated issue: #321 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/metrics.py
+++ b/dj/api/metrics.py
@@ -6,11 +6,14 @@ from http import HTTPStatus
 from typing import List, Optional
 
 from fastapi import APIRouter, Depends, HTTPException, Query
+from sqlalchemy.exc import NoResultFound
 from sqlmodel import Session, select
 
 from dj.api.helpers import get_node_by_name, get_query
+from dj.errors import DJError, DJException, ErrorCode
 from dj.models.metric import Metric, TranslatedSQL
 from dj.models.node import Node, NodeType
+from dj.sql.dag import get_dimensions
 from dj.utils import get_session
 
 router = APIRouter()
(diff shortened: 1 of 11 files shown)


shangyian (approved):
Looks good!

dj/api/metrics.py
```diff
@@ -81,3 +84,46 @@ async def read_metrics_sql(
...
+                DJError(
+                    message=f"Metric node not found: {node_name}",
+                    code=ErrorCode.UNKNOWN_NODE,
+                ),
+            )
+
+    if errors:
+        raise DJException(errors=errors)
+
+    common = set(get_dimensions(metric_nodes[0]))
```
shangyian, line 125: This function only gets the first layer of dimensions right? We can probably update this to pull in further layers once there's better dimensions metadata, so that we're not pulling in a lot of extraneous dims.

samredai: Currently this pulls the same metadata as the metrics/ endpoint (which uses the same function). You've made me realize it will actually think similarly named columns on different metric parent nodes are a "common dimension" which I don't think is a safe assumption. For now I'm going to keep it this way--I'll start looking at how we can enable the sql and data endpoints to accept multiple metrics at once and build the appropriate SQL. That will inevitably add some finer logic for this.

PR #390, opened 2023-03-06
Fix column type parsing to handle named fields in `ROW`

shangyian:
### Summary

Two issues:
1. `ColumnType.ROW` doesn't parse correctly when the `ROW` has a named field with a complex type. For example, with `ROW[STR name, MAP[STR,STR] display, ROW[STR id]]`, it doesn't treat `display` as a field name for the `MAP`.
2. Named fields aren't output in the actual `ColumnType` string. This is needed so that we can store into the columns table for future use (otherwise these names will be lost after parsing it for the first time).

### Test Plan

Added some tests for named fields. Also verified locally with some tables with named struct columns.

- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/typing.py
+++ b/dj/typing.py
@@ -84,7 +84,7 @@ def process_array_args(*args: str) -> Tuple["ColumnType"]:
     """
     if len(args) != 1:
         raise ColumnTypeError(
-            f"{', '.join(args)} is not an acceptable type for ARRAY"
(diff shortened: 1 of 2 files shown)


samredai (approved):

CircArgs:
Thanks for fixing all my mistakes

CircArgs (approved):

PR #394, opened 2023-03-08
Add column annotations

shangyian:
### Summary

Still WIP. May remove `uniqueness_scope` until we find a better way to set up rules around this. Also need to add tests for tagging a column with more than one annotation, as well as tagging multiple columns with annotations.

This adds column-level annotations (aka global tags, but I chose `annotations` so as to not get this concept confused with the node-level `tag`). 

This setup allows users to add their own custom annotation types by calling `POST /annotations/`, which is functionality that we may want to eventually lock down. We also provide a set of default annotations -- `primary_key`, `event_time` etc, and if this set of defaults gets updated, the annotation types will be automatically refreshed in the database.

Annotating a column is achieved with the `POST /nodes/{node_name}/annotate/` endpoint, which lets users specify a set of columns and their annotations.

### Test Plan

- [ ] `make check` passes


dj/api/annotations.py
```diff
@@ -0,0 +1,127 @@
...
+def default_annotation_types(session: Session = Depends(get_session)):
+    """
+    Loads all the column annotation types that are supported by the system
+    by default into the database.
+    """
+    defaults = [
+        AnnotationType(
+            namespace=RESERVED_ANNOTATION_NAMESPACE,
+            name="primary_key",
+            description="Points to a column which is part of the primary key of the node",
```
agorajek, line 68: This just made me thought of partition keys. And the amazing thing is that you could almost implement them using the same mechanism, except the fact that list of partition key columns should be ordered... so should we sneak in a `position` attribute to our ColumnAnnotation/ColumnAttribute class and have a solved problem. Or should we keep partition attributes separate? I am leaning towards the latter, but I wanted to ask you all.

dj/models/annotation.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class AnnotationTypeBase(AnnotationTypeName):
+    """
+    Available annotation types for column metadata (base model).
+    """
+
+    namespace: str
+    description: str
+    uniqueness_scope: List[str] = Field(sa_column=SqlaColumn(JSON))
```
agorajek, line 36: This could probably be some sort of enum made of Column, Node, System, whatever else we want to have these attributes/annotations on, right?

agorajek (approved):
@shangyian thank you for this work! Left a question and a comment.

dj/models/annotation.py
shangyian: Yeah, makes sense! I added an enum and also some checks based on that.

dj/api/annotations.py
shangyian: Yeah, this is a good idea! It makes sense to me that partitions are just attributes on columns. Although I think this may be better off in a separate PR so that we can also add in more logic around enforcement of partition positions.

PR #395, opened 2023-03-08
Add query running functions to the query service client

samredai:
### Summary

This adds functions to the query service client for submitting queries and retrieving them (when running async).

### Test Plan

N/A

- [x] PR has an associated issue: #342 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/errors.py
+++ b/dj/errors.py
@@ -235,3 +235,12 @@ class DJDoesNotExistException(DJException):
 
     dbapi_exception: DBAPIExceptions = "DataError"
     http_status_code: int = 404
+
+
+class DJQueryServiceClientException(DJException):
+    """
+    Exception raised when the query service returns an error
+    """
+
+    dbapi_exception: DBAPIExceptions = "InterfaceError"
+    http_status_code: int = 500
--- a/dj/service_clients.py
+++ b/dj/service_clients.py
@@ -6,7 +6,7 @@ import requests
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
-from dj.errors import DJException
+from dj.errors import DJQueryServiceClientException
 from dj.models.column import Column
 from dj.models.query import QueryWithResults
 from dj.typing import ColumnType
@@ -106,8 +106,8 @@ class QueryServiceClient:  # pylint: disable=too-few-public-methods
                 "async_": async_,
             },
         )
-        if response.status_code >= 400:
-            raise DJException(
+        if not response.ok:
+            raise DJQueryServiceClientException(
                 message=f"Error response from query service: {response.text}",
             )
         query_info = response.json()
@@ -122,7 +122,7 @@ class QueryServiceClient:  # pylint: disable=too-few-public-methods
         """
         response = self.requests_session.get(f"/queries/{query_id}/")
         if not response.ok:
-            raise DJException(
+            raise DJQueryServiceClientException(
                 message=f"Error response from query service: {response.text}",
             )
         query_info = response.json()
--- a/dj/service_clients.py
+++ b/dj/service_clients.py
@@ -121,7 +121,7 @@ class QueryServiceClient:  # pylint: disable=too-few-public-methods
         Get a previously submitted query
         """
         response = self.requests_session.get(f"/queries/{query_id}/")
-        if response.status_code >= 400:
+        if not response.ok:
             raise DJException(
                 message=f"Error response from query service: {response.text}",
             )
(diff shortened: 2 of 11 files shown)


dj/service_clients.py
```diff
@@ -82,3 +84,46 @@ def get_columns_for_table(
...
+                "catalog_name": catalog,
+                "async_": async_,
+            },
+        )
+        if response.status_code >= 400:
+            raise DJException(
+                message=f"Error response from query service: {response.text}",
+            )
+        query_info = response.json()
+        return QueryWithResults(**query_info)
```
agorajek, line 114: Let's make it `QueryResults` please? :) 

`QueryWithResults` feels a bit weird. It makes me want to ask for `QueryWithoutResults` as well.

agorajek (approved):
Looks good. Just one silly request in-line.

dj/service_clients.py
samredai: Haha, every-time I see that I think of how much you're not a fan of that class name. 😂 I have a big refactor coming up to fully complete the move to the API and wiring up DJQS. Once that's in I'll remember to come back and update the class name here and in DJQS. 💯

betodealmeida: @agorajek but this is a query object with results! 🙂 

It has the query information (status, progress, submitted/executed query) and the results. It's almost identical to the `Query` payload, with the addition of the result, hence `QueryWithResults`.

https://github.com/DataJunction/dj/blob/110168faedd7d052f86107ff71c0c60dc648d93c/dj/models/query.py#L114

dj/service_clients.py
```diff
@@ -82,3 +84,46 @@ def get_columns_for_table(
...
+
+    def get_query(
+        self,
+        query_id: str,
+    ) -> QueryWithResults:
+        """
+        Get a previously submitted query
+        """
+        response = self.requests_session.get(f"/queries/{query_id}/")
+        if response.status_code >= 400:
```
betodealmeida, line 124: Nit:

```suggestion
        if not response.ok:
```

betodealmeida (changes requested):
@samredai I left a few comments, especially about the naked `DJException`. I'm going through the code base converting all of them to more specific exceptions, it would be nice if we could do the same here.

dj/service_clients.py
betodealmeida: And for context, the idea with `QueryWithResults` is that it simplifies the query endpoint. You can a single endpoint where you can create queries (doing a `POST`), check for their status (`GET`) and fetch results (also `GET`).

dj/service_clients.py
```diff
@@ -82,3 +84,46 @@ def get_columns_for_table(
...
+            json={
+                "engine": engine,
+                "engine_version": engine_version,
+                "submitted_query": submitted_query,
+                "catalog_name": catalog,
+                "async_": async_,
+            },
+        )
+        if response.status_code >= 400:
+            raise DJException(
```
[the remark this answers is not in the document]
samredai: Great point! We should definitely make sure errors are as clear as possible 💯 . Updated the PR to use a `DJQueryServiceClientException`

dj/service_clients.py
```diff
@@ -82,3 +84,46 @@ def get_columns_for_table(
...
+        query_id: str,
+    ) -> QueryWithResults:
+        """
+        Get a previously submitted query
+        """
+        response = self.requests_session.get(f"/queries/{query_id}/")
+        if response.status_code >= 400:
+            raise DJException(
+                message=f"Error response from query service: {response.text}",
+            )
```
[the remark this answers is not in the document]
samredai: Added a dbapi_exception to the `DJQueryServiceClientException`. Thanks @betodealmeida!
```py
class DJQueryServiceClientException(DJException):
    """
    Exception raised when the query service returns an error
    """

    dbapi_exception: DBAPIExceptions = "InterfaceError"
    http_status_code: int = 500
```

shangyian (approved):

dj/errors.py
```diff
@@ -217,3 +217,12 @@ class DJInternalErrorException(DJException):
 
     dbapi_exception: DBAPIExceptions = "InternalError"
     http_status_code: int = 500
+
+
+class DJQueryServiceClientException(DJException):
```
shangyian, line 222: Nice, glad that we're going for more specific exceptions going forward.

dj/service_clients.py
agorajek: hold-on-hmm

OK, I will not bring this subject again. If I do please add it to my performance review.

betodealmeida (approved):
Thanks for addressing the comments!

Issue #382, opened 2023-03-03
Rethinking the Database <-> Engine <-> Catalog mapping


shangyian:
I started thinking through the mappings between `Database`, `Engine`, `Catalog` and how they all fit together and drew a quick sketch to solidify things in my head:

<img src="https://user-images.githubusercontent.com/9524628/222763225-6b9d86db-6f12-4933-bc4e-52ca5d992ec8.png" width="550" />

Essentially, there are many different databases, some of which include both data storage + an engine (like Postgres and Druid), others of which only have the data storage component (like the data warehouse). Some of the data stores have concepts of catalogs and schemas (Postgres, data warehouse), but others don't (Druid). 

Perhaps instead of having a `Database` concept, we can have a `DataStore` concept, which more cleanly reflects how some components should be represented in this system. So something like this:

<img src="https://user-images.githubusercontent.com/9524628/222765116-98d22bef-8bbe-499f-9f70-33cd134d43ff.png" width="400" />

* A **DataStore** (example values `druid`, `postgres`, `warehouse`) has one or more catalogs associated with it. 
* A **Catalog** has one or more engines (example values `druid`, `postgres`, `spark`, `trino`) associated with it.

What do you folks think?

samredai:
I think this is great. Thanks for the chat earlier and clearing up my funky understanding of postgres 😂 . Adding the datastore ultimately makes sense to me as the final design, but I think it should be paired with adding support for multi-catalog queries. So maybe we can go with the `Catalog` + `Engine` design for now, and then whenever we're ready to add muti-catalog support we tackle adding the Datastore level.

shangyian:
Definitely! This doesn’t need to be done for the alpha release, just wanted to clear up these concepts first. 

@betodealmeida it’d be great to hear your thoughts as well!

betodealmeida:
This is super interesting... I always like to think about how to apply this to more unconventional databases:

- SQLite: should the catalog be a file (eg, `database.sqlite`)? And the data store be a filesystem?
- Google Sheets: there's no catalog nor schema, so should have `main` for both?
- Druid: similarly, there's no catalog.. should we use `main`? Or `default`?

samredai:
[...]

Issue #121, opened 2022-04-17
Add support for partitions


betodealmeida:

betodealmeida:
This has to be implemented per-DB. For Postgres: https://dba.stackexchange.com/questions/223327/how-to-identify-the-column-used-to-partition-a-table-from-the-postgres-system-ca

betodealmeida:
Some work started here: https://github.com/DataJunction/datajunction/compare/partitions

agorajek:
Let's consider the following use cases for partitions in DJ:

1. Partition as a filter, for efficient data querying. 
    Example: a query referencing DJ nodes is generated by DJ to run on some database. 
    This can apply to both source nodes and transform nodes, that is tables managed by DJ and outside of DJ control.

2. Partition for data management, for efficient data storage and access.
    A DDL/insert statement is generated by DJ to create or load a materialized version of a DJ node.

The 1st use case requires two features:
- ability to recognize partition features in tables corresponding to source nodes
- ability to translate DJ queries into partition filters, they should not be different than regular column filter

The 2nd use case is much more involved:
- ability to let the users add a partitioning scheme at the DJ repo level and at the DJ node level.
- ability to create proper DDL and insert/load statements with the correct partition values

Lets consider the following partition spec for DJ world:
```
partitions:
  - name: <string>
    type: [date, str, int]
    value | interval: day, month, year, etc (optional) 
    tz: ISO time zone (optional)
    format: YYYYMMDD (optional)
```

PR #397, opened 2023-03-10
Complete migration over to REST API and Catalog/Engine + Build changes

samredai:
### Summary

This PR completes the migration to the REST API, DJ AST, and Catalog/Engine concept. Since this also required a lot of changes to the build (thanks @CircArgs for all the help!) it was hard to keep the build in a transition state where it worked along smaller piecemeal changes, hence the bigger PR. Here's a list of the changes:

- Fully migrated all yaml based logic over to the REST api
- Consolidated all node creation in tests to a single fixture function load_examples that makes API calls to a test client and can optionally be used in any given test. (Also I create all of the examples there that where in YAML before)
- Changes all Database logic to Catalog+Engine logic.
- Adds common catalog validation through change management, allowing the build to assume a common catalog for all node paths and also materializations. (**Way** faster build, linear time complexity, no more build planning)
- Changes to the build to use availabilities posted by a materialization service and instrumented logic to automatically check a build criteria method. I haven’t added the analysis yet, that logic should be discussed during the Materialization service development. (An example of the build using availability has been added to the roads example notebook)
- Reverted back to external tables being defined directly on a source node instead of as a separate entity (buyer’s remorse). Also source nodes now require a catalog and that catalog is automatically propagated down to downstream nodes.
- Transforms/dimension nodes that do multi-catalog joins are now caught during node creation (also catches cube node creation with elements from multiple dimensions). Later when we add support for multi-catalog joins we can have a setting to enable it just like postgres.
- Removed the watchdog container and repository logic. As discussed a few syncs back, we can move these to a client library for users to utilize in a distributed way (many different repos) which all post back to a central DJ server. We can also add the YAML parsing logic there that was removed here for users who want to use that pattern. (Probably a good time to start on the client library)

### Test Plan

Create and build all examples through the API (basic, dbt, and roads), run all notebook examples, run `make check` and `make test`.

- [x] PR has an associated issue: #382 #298 #271 #255 #121 #305 #301
- [x] `make check` passes

### Deployment Plan

The only remaining API endpoints planned are the ones around registering functions. Once that's added, I think we may be ready to stabilize the REST API spec (which we should keep stable within major DJ releases).


CircArgs (approved):

Dockerfile
```diff
@@ -6,5 +6,7 @@ RUN pip install -e .
 RUN pip install https://github.com/rogerbinns/apsw/releases/download/3.38.1-r1/apsw-3.38.1-r1.zip \
   --global-option=fetch --global-option=--version --global-option=3.38.1 --global-option=--all \
   --global-option=build --global-option=--enable-all-extensions
-CMD ["uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "--reload"]
+
+RUN pip install jupyter nb-black
```
shangyian, line 10: Do we want to do this here or keep track of these in a separate requirements file?

dj/models/table.py
```diff
@@ -53,21 +53,21 @@ class TableBase(BaseSQLModel):
...
-        default=None,
-        foreign_key="table.id",
-        primary_key=True,
-    )
-    node_revision_id: Optional[int] = Field(
-        default=None,
-        foreign_key="noderevision.id",
-        primary_key=True,
-    )
+# class TableNodeRevision(BaseSQLModel, table=True):  # type: ignore
```
shangyian, line 56: Just remove this?

docker-compose.yml
```diff
@@ -22,26 +22,11 @@ services:
       - .:/code
     ports:
       - "8000:8000"
+      - "8001:8001"
     depends_on:
       - db_migration
       - postgres_examples
-      - redis
-    extra_hosts:
```
shangyian, line 29: Won't removing `extra_hosts` cause issues with running the query service?

dj/api/graphql/column.py
```diff
@@ -1,38 +0,0 @@
-"""
```
shangyian, line 1: These don't get added back right? Is the idea that we can add the graphql API later, after the existing rest API is more settled?

dj/api/nodes.py
```diff
@@ -515,79 +545,6 @@ def add_dimension_to_node(
...
-            )
-
-    # When no columns are provided, attempt to find actual table columns
-    # if a query service is set
-    columns = [
-        Column(name=column.name, type=ColumnType(column.type))
-        for column in data.columns
-    ]
-    if not columns:
-        if not query_service_client:
```
shangyian, line 553: Not a blocker, but just flagging that we should add this again so that we can infer columns for loaded tables (in a follow-up PR)

shangyian (approved):
Changes looks good! I left a few minor comments - mostly about follow-ups, nothing blocking.

dj/api/graphql/column.py
samredai: Yeah we needed to thoroughly comb through the graphql API anyway so it seemed much easier to add this back and confirm parity when we're locked into an openapi spec for an entire major version (cc: @CircArgs).

shangyian: Makes sense!

Dockerfile
samredai: Let me try moving it to the test requirements. I believe the only place it's used is in the pre-commit runs to make sure the example notebooks are all formatted nicely.

dj/api/nodes.py
samredai: Another great catch! Let me move this over to the create node logic for source nodes since the table info is back on their definitions now.

Dockerfile
samredai: Actually, pre-commit handles it's own requirements, I just went ahead and removed this.

dj/models/table.py
samredai: Done

dj/api/helpers.py
```diff
@@ -125,53 +111,41 @@ def get_catalog(session: Session, name: str) -> Catalog:
...
     """
     metric = get_node_by_name(session=session, name=metric, node_type=NodeType.METRIC)
-    database_id = (
-        get_database_by_name(session=session, name=database_name).id
-        if database_name
-        else None
-    )
-    query_ast, optimal_database = await build_node_for_database(
+
+    query_ast = build_node(
```
shangyian, line 125: Does this need to also include the engine? As in, now instead of building the query for a database, we're building the query for an engine. Or is the idea that this is just a generic query AST, which would then need to be translated to engine-specific SQL by the query engine?

dj/construction/utils.py
```diff
@@ -105,4 +105,4 @@ def amenable_name(name: str) -> str:
             ret.append(LOOKUP_CHARS.get(char, "UNK"))
             cont = []
 
-    return "_".join(ret) + "_" + "".join(cont)
+    return ("_".join(ret) + "_" + "".join(cont)).strip("_")
```
agorajek, line 108: If someone wanted to have `_foo_` we would make it `foo` ?

dj/models/node.py
```diff
@@ -245,6 +255,16 @@ class AvailabilityState(AvailabilityStateBase, table=True):  # type: ignore
...
     )
 
+    def is_available(
+        self,
+        criteria: Optional[BuildCriteria] = None,  # pylint: disable=unused-argument
+    ) -> bool:  # pragma: no cover
+        """
+        Determine whether an availability state is useable given criteria
+        """
+        # Criteria to determine if an availability state should be used needs to be added
```
agorajek, line 265: You don't like the `# TODO: ...` marker?

tests/conftest.py
```diff
@@ -1,9 +1,8 @@
 """
 Fixtures for testing.
 """
-# pylint: disable=redefined-outer-name, invalid-name, W0611
+# pylint: disable=redefined-outer-name, invalid-name, W0611, too-many-lines
```
agorajek, line 4: lol, @samredai please start writing smaller PRs. I bet in Java this lint error would come up in the Hello World app.

tests/conftest.py
```diff
@@ -103,689 +86,958 @@ def get_settings_override() -> Settings:
...
+def load_examples():
+    """
+    Fixture function for loading all node examples to a client
+    """
+
+    def load_examples_fn(client: TestClient):
+        """
+        Load all node examples to a client
+        """
+        for endpoint, json in (
```
agorajek, line 113: YAML to JSON... yuck! :)

tests/construction/dj_query_test.py
```diff
@@ -28,45 +27,44 @@ async def test_build_dj_metric_query(mocker, request):
...
-        b.user_metadata AS b_DOT_user_metadata,
-        b.age AS b_DOT_age,
-        b.country AS b_DOT_country,
-        b.gender AS b_DOT_gender,
-        b.preferred_language AS b_DOT_preferred_language,
-        b.secret_number AS b_DOT_secret_number
- FROM (SELECT  basic.comments.country,
-        COUNT(DISTINCT basic.comments.id) AS num_users
- FROM basic.comments
+        basic_DOT_source_DOT_users.id AS b_DOT_id,
```
agorajek, line 30: If we show these `_DOT_`s to some users, they'll think we are insane ;)

tests/errors_test.py
```diff
@@ -42,3 +42,5 @@ def test_dj_exception() -> None:
 - Error 1 (error code: 0)
 - Error 2 (error code: 0)"""
     )
+
+    assert DJException("Message") == DJException("Message")
```
agorajek, line 46: This looks like `assert foo == foo` to me. What am I missing?

tests/sql/parsing/queries/case_when_null.sql
```diff
@@ -3,7 +3,7 @@ SELECT CASE
          ELSE store.item_sk
        END              item_sk,
        CASE
-         WHEN web.d_date IS NOT NULL THEN web.d_date
+         WHEN NOT web.d_date IS NULL THEN web.d_date
```
agorajek, line 6: `IS NOT NULL` is more SQL-ish

tests/sql/parsing/test_ast.py
```diff
@@ -497,9 +500,9 @@ def test_map_subscripts():
     assert compare_query_strings('SELECT some_map["x"] FROM a', str(query))
 
 
-def test_query_to_select(cte_query):
+def test_queryto_select(cte_query):
```
agorajek, line 503: Looks like collateral damage from a `codemod "_to_select" "to_select"` ?

agorajek (approved):
This PR feels like a new era for DJ. As much as I would like to hold on to my precious little things, I know DJ is becoming and adult and needs to move on ;)

Great work @samredai !

Just a few comments in line, some of which you may want to address before merge.

alembic/versions/2023_03_10_0331-143a068da0ed_initial_migration.py
```diff
@@ -196,20 +235,97 @@ def upgrade():
...
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_columnattribute")),
+        sa.UniqueConstraint(
+            "attribute_type_id",
+            "column_id",
+            name=op.f("uq_columnattribute_attribute_type_id"),
+        ),
+    )
+    op.create_table(
+        "cuberelationship",
```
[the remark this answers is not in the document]
samredai: This is actually autogenerated by alembic from the sqlmodel `CubeRelationship` class. Since cube definitions are just node definitions, they share the `node` table and this is just a two column lookup table used as a "link model". Would a better name be something like "CubeLookup" or "CubeLink"?

dj/models/node.py
samredai: I do but we have a pre-commit check which will fail if it finds any to-do's. I know this is basically an attempt at circumventing that but I'm expecting to follow-up pretty quickly with an implementation of this function. (famous last words, I know...)

alembic/versions/2023_03_10_0331-143a068da0ed_initial_migration.py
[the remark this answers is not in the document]
betodealmeida: @samredai you can override the default table names generated for models using the `__tablename__` attribute (something like that).

tests/conftest.py
samredai: Haha, I did feel awkward putting that there. It's because of the long list of examples I added to the conftest. Thanks for keeping me honest here, I moved them to an examples.py file and now both files are under the limit, allowing me to remove this. 🥳

tests/conftest.py
samredai: Don't worry, the DJ python client will have the greatest YAML to JSON converter the world has ever known! 😂

tests/errors_test.py
samredai: This is to check that the `__eq__` method is working properly...and to get coverage on those lines :D

dj/api/nodes.py
```diff
@@ -400,6 +405,12 @@ def create_cube_node_revision(
             message=("At least one dimension is required to create a cube node"),
             http_status_code=http.client.UNPROCESSABLE_ENTITY,
         )
+    if len(set(catalogs)) != 1:
```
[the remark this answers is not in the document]
samredai: Good idea! The requirements now are that each node have a catalog but you're right, it's good to explicitly to catch any strange behaviors. Added a `< 1` check.

dj/api/nodes.py
```diff
@@ -492,7 +516,13 @@ def add_dimension_to_node(
         name=dimension,
         node_type=NodeType.DIMENSION,
     )
-
+    if node.current.catalog.name != dimension_node.current.catalog.name:
+        raise DJException(
+            message=(
+                "Cannot add dimension to column, catalogs do not match: "
```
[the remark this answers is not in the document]
samredai: Fixed 👍

tests/sql/parsing/test_ast.py
samredai: You're exactly right, that was the exact find and replace that I did. 😅 (fixed it!)

tests/sql/parsing/queries/case_when_null.sql
samredai: I think this was due to a small bug in the build. I believe @CircArgs is planning a fast follow-up to fix this.

tests/construction/dj_query_test.py
samredai: Agreed, I'll add this to the sync to discuss the best syntax for these system generated aliases.

alembic/versions/2023_03_10_0331-143a068da0ed_initial_migration.py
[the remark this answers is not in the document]
samredai: Nice! Alright I'll keep the classname the same and change the table name to "cube"

dj/construction/utils.py
samredai: Hmm...good point. I think you're right, this would be a problem. @CircArgs can you confirm?

dj/api/data.py
```diff
@@ -27,59 +27,57 @@ def add_availability(
...
+        if node_revision.schema_ != data.schema_ or node_revision.table != data.table:
             raise DJException(
                 message=(
                     "Cannot set availability state, "
                     "source nodes require availability "
-                    "states match an existing table: "
-                    f"{new_availability.catalog}."
-                    f"{new_availability.schema_}."
-                    f"{new_availability.table}"
+                    "states match the set table: "
```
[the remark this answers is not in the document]
samredai: Fixed

alembic/versions/2023_03_10_0331-143a068da0ed_initial_migration.py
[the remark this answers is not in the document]
samredai: That worked nicely, thanks @betodealmeida!

alembic/versions/2023_03_12_0158-e3fc8f2f003f_initial_migration.py
```diff
@@ -14,13 +14,29 @@
...
 # revision identifiers, used by Alembic.
-revision = "b9f370ef912a"
+revision = "e3fc8f2f003f"
 down_revision = None
 branch_labels = None
 depends_on = None
 
 
 def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
betodealmeida, line 24: Nit: can we remove these so we know the auto-generated migration was revised? Thanks!

alembic/versions/2023_03_12_0158-e3fc8f2f003f_initial_migration.py
```diff
@@ -303,43 +415,31 @@ def upgrade():
...
-            "node_revision_id",
-            name=op.f("pk_tablenoderevision"),
-        ),
-    )
+    # ### end Alembic commands ###
 
 
 def downgrade():
-    op.drop_table("tablenoderevision")
+    # ### commands auto generated by Alembic - please adjust! ###
```
betodealmeida, line 422: Ditto.

dj/construction/utils.py
samredai: Since this is the only remaining comment, I'm going to merge this to unblock the work in a few people's forks so that they can rebase.

dj/models/node.py
```diff
@@ -501,6 +492,9 @@ class SourceNodeFields(BaseSQLModel):
     Source node fields that can be changed.
     """
 
+    catalog: Optional[str] = ""
+    schema_: Optional[str] = ""
+    table: Optional[str] = ""
```
betodealmeida, line 497: If these are optional should we make the default `None`?

dj/models/table.py
betodealmeida: It's still here? 😛

tests/api/data_test.py
```diff
@@ -17,83 +15,12 @@ class TestAvailabilityState:  # pylint: disable=too-many-public-methods
...
-                Column(name="customer_id", type=ColumnType.INT),
-                Column(name="account_type", type=ColumnType.STR),
-            ],
-        )
-        session.add(node_rev1)
-        session.add(node_rev2)
-        session.add(node_rev3)
-        session.commit()
-        return session
+        load_examples(client)
```
betodealmeida, line 22: This is a bit confusing, since in other tests we need to explicitly call `load_examples`, but here the `client` fixture calls it automatically. It would be better to be consistent in the fixtures so we don't have surprises. Maybe call this fixture `client_with_examples`?

betodealmeida (approved):
This is great! I left a comment on the `client` fixture being inconsistent, otherwise looks good.

dj/models/table.py
samredai: Thanks @betodealmeida!
image

alembic/versions/2023_03_12_0158-e3fc8f2f003f_initial_migration.py
samredai: Done!

dj/models/node.py
samredai: Good idea, updated 👍

dj/models/table.py
samredai: Ok, now it's really really removed. lol

tests/api/data_test.py
samredai: This is a great catch, I like the `client_with_examples`, updated this pattern to use that.

Issue #402, opened 2023-03-12
Add separate creation endpoints for each node type


shangyian:
We should have separate creation endpoints for each of the different node types. The rationale behind this is:
* Creating each node type requires different input parameters. For someone using the API, it's hard to tell which parameters need to be provided for which node type. It's better for us to be super clear in the API spec about the expected inputs rather than returning errors post-processing.
* It's hard to wire pydantic models, as we don't have much control over which model in the `Union` type each of the different expected creation inputs will be parsed into.
* The logic for node creation is already highly separated and any reuse can be shared behind the scenes.

[mostly pasting from slack]

PR #403, opened 2023-03-12
Add separate creation endpoints and rewire query-service for source node schema auto-population

shangyian:
### Summary

This PR does two main things:

1. Adds separate creation endpoints for each of the different node types. This allows us to be much crisper about the different pydantic models used for creation. These are the new endpoints:
  * `POST /nodes/source/` 
  * `POST /nodes/dimension/`
  * `POST /nodes/transform/`
  * `POST /nodes/metric/`
  * `POST /nodes/cube/`
2. Adds back the query service client used for source node schema auto-population.

### Test Plan

Modified some unit tests to conform to the new endpoints and added an additional test for the query service's schema population.

- [X] PR has an associated issue: #402
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -31,12 +31,13 @@ from dj.sql.parsing import ast
 from dj.sql.parsing.backends.exceptions import DJParseException
 
 
-def get_node_by_name(
+def get_node_by_name(  # pylint: disable=too-many-arguments
     session: Session,
     name: str,
     node_type: Optional[NodeType] = None,
     with_current: bool = False,
     raise_if_not_exists: bool = True,
+    raise_if_exists: bool = False,
 ) -> Node:
     """
     Get a node by name
@@ -59,6 +60,12 @@ def get_node_by_name(
                     f"node with name `{name}` does not exist."
                 ),
                 http_status_code=404,
+            )
+    if raise_if_exists:
+        if node:
+            raise DJException(
+                message=f"A node with name `{name}` already exists.",
+                http_status_code=HTTPStatus.CONFLICT,
             )
     return node
 
@@ -311,7 +318,7 @@ def resolve_downstream_references(
                 .unique()
                 .one()
             )
-            downstream_node_revision.parents.append(node_revision)
+            downstream_node_revision.parents.append(node_revision.node)
             downstream_node_revision.missing_parents.remove(missing_parent)
             (
                 _,
--- a/dj/api/helpers.py
+++ b/dj/api/helpers.py
@@ -37,7 +37,6 @@ def get_node_by_name(  # pylint: disable=too-many-arguments
     node_type: Optional[NodeType] = None,
     with_current: bool = False,
     raise_if_not_exists: bool = True,
-    raise_if_exists: bool = False,
 ) -> Node:
     """
     Get a node by name
@@ -50,8 +49,6 @@ def get_node_by_name(  # pylint: disable=too-many-arguments
         node = session.exec(statement).unique().one_or_none()
     else:
         node = session.exec(statement).one_or_none()
-
-    # Only raise an error for non-existent nodes if this flag is set
     if raise_if_not_exists:
         if not node:
             raise DJException(
@@ -61,13 +58,19 @@ def get_node_by_name(  # pylint: disable=too-many-arguments
                 ),
                 http_status_code=404,
             )
-    if raise_if_exists:
-        if node:
-            raise DJException(
-                message=f"A node with name `{name}` already exists.",
-                http_status_code=HTTPStatus.CONFLICT,
-            )
     return node
+
+
+def raise_if_node_exists(session: Session, name: str) -> None:
+    """
+    Raise an error if the node with the given name already exists.
+    """
+    node = get_node_by_name(session, name, raise_if_not_exists=False)
+    if node:
+        raise DJException(
+            message=f"A node with name `{name}` already exists.",
+            http_status_code=HTTPStatus.CONFLICT,
+        )
 
 
 def get_column(node: NodeRevision, column_name: str) -> Column:
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -50,8 +50,9 @@ from dj.models.node import (
     UpdateNode,
     UpsertMaterializationConfig,
 )
+from dj.service_clients import QueryServiceClient
 from dj.sql.parsing import parse
-from dj.utils import Version, VersionUpgrade, get_session
+from dj.utils import Version, VersionUpgrade, get_session, get_query_service_client
 
 _logger = logging.getLogger(__name__)
 router = APIRouter()
@@ -323,7 +324,16 @@ def create_node_revision(
     """
     Create a non-source node revision.
     """
-    node_revision = NodeRevision.parse_obj(data)
+    node_revision = NodeRevision(
+        name=data.name,
+        display_name=data.display_name
+        if data.display_name
+        else generate_display_name(data.name),
+        description=data.description,
+        type=data.type,
+        status=NodeStatus.VALID,
+        query=data.query,
+    )
     (
         validated_node,
         dependencies_map,
@@ -427,6 +437,7 @@ def create_cube_node_revision(
 def create_node(
     data: Union[CreateSourceNode, CreateCubeNode, CreateNode],
     session: Session = Depends(get_session),
+    query_service_client: QueryServiceClient = Depends(get_query_service_client),
 ) -> NodeOutput:
     """
     Create a node.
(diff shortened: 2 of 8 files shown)


dj/api/helpers.py
```diff
@@ -31,12 +31,13 @@
...
 
 
-def get_node_by_name(
+def get_node_by_name(  # pylint: disable=too-many-arguments
     session: Session,
     name: str,
     node_type: Optional[NodeType] = None,
     with_current: bool = False,
     raise_if_not_exists: bool = True,
+    raise_if_exists: bool = False,
```
samredai, line 40: nit and style opinion: I think we should avoid these raise_ args and let the calling logic catch the exception. Even when this is set to not raise we still have to check something like `if not foo:`. I realize it's painful to have to do try/excepts everywhere and feels duplicative since DJExceptions raise nice HTTP response, so maybe an alternative could be to just have separate functions for these different flows such as `raise_if_node_exists(...)`.

dj/api/nodes.py
```diff
@@ -343,7 +340,7 @@ def create_node_revision(
         raise DJException(
             f"Cannot create nodes with multi-catalog dependencies: {set(catalog_ids)}",
         )
-    catalog_id = next(iter(catalog_ids), "draft")
+    catalog_id = next(iter(catalog_ids), 0)
```
samredai, line 343: I was wondering how to handle this in a previous PR and curious to hear your thoughts. When a node is created with no missing parent nodes, everything works nicely and it inherits the catalog. But when a draft node is added with missing parent nodes, there's no way to know the catalog until those parents are added. So it felt like a choice between making catalogs optional on the model or defaulting them to something. `"default"` didn't feel right because that's an actual catalog that we would offer "out of the box" and could get confusing if for some reason when the missing parent nodes get added their catalog value isn't propagated properly. Just thinking out loud here.

samredai (approved):
Awesome PR 🚀 ! The swagger docs look so much clearer now. I just had a few open ended comments with thoughts that came to mind as I was reviewing. Nothing blocking. I'll follow-up with a PR to update the requests from the example notebooks.

dj/api/nodes.py
betodealmeida: @samredai I'm confused here, why does a `NodeRevision` have a catalog (and schema/table)? The catalog/schema/table should be a query concept, and not part of the node itself.

dj/api/nodes.py
```diff
@@ -418,85 +410,150 @@ def create_cube_node_revision(
...
+        name=data.name,
+        display_name=data.display_name
+        if data.display_name
+        else generate_display_name(data.name),
+        description=data.description,
+        type=NodeType.SOURCE,
+        status=NodeStatus.VALID,
+        catalog_id=catalog.id,
+        schema_=data.schema_,
+        table=data.table,
```
betodealmeida, line 505: @shangyian I think I missed this when we introduced the node revisions. What if I want to create a source node that points to multiple tables, is that no longer possible?

dj/api/nodes.py
samredai: The catalog is just a logical concept for the nodes so that the "common catalog" validation can be done through change management. That way the build doesn't have to do that validation on the fly. Catalogs can have multiple "engines" though and that's all handled in the query service so it's still a query concept if I understand what you meant by that. For example you can have a postgres, trino, and druid engine defined in the query service that all run the query against different locations.

betodealmeida: But with this we lost support for source nodes that are materialized in different catalogs, which is a huge feature in the long run. In addition, if we later move the data to a different catalog, we won't be able to compute an older node version in the new catalog, since the version is tied to a specific catalog.

dj/api/nodes.py
samredai: Hmm, if they're truly different tables with different schema/table names, I think you're right that it's no longer possible for the system to automatically swap them out without the user explicitly doing that. These are sort of alt-paths down the DAG right? Is there a way we can have these other tables defined on separate nodes and tied together through some other construct?

dj/api/nodes.py
samredai: I might not fully understand the feature but we've talked about supporting multi-catalog joins in the future, would this feature be part of that? I added this as a topic for the next sync.

betodealmeida: Oh no, I'm not talking about multi-catalog joins. I'm talking about having multiple catalogs where a given node can be computed.

I'm trying to understand where this change (introduction of revisions?) happened, but originally any node could have multiple representations. In the original examples some source nodes pointed to tables in Druid/Postgres/GSheets, some pointed to tables in in Druid/Postgres, and so on. The query engine would evaluate a request and figure out the best database to fulfill the request. Depending on the columns in the request it would use Druid, or Postgres, or GSheets.

Some requests couldn't be fulfilled, because there was no common database (or catalog, today). For cases like that in the future we could implement multi-catalog joins. But even without implementing the multi-catalog joins **it's important to decouple the concept of a node from where it's materialized (or can be computed). So nodes should not have catalogs, schemas, tables associated with them.**

dj/api/nodes.py
shangyian: @betodealmeida hmm, that change you're describing wasn't introduced with the node revisions, but rather with the recent changes to move away from the `database` concept and over to `catalog` instead.

shangyian: But agreed that we should discuss this on Wednesday. If a source node has multiple tables, it will also need to manage any schema discrepancies between the nodes.

dj/api/nodes.py
samredai: I see, I think this change is from the recent big PR that got merged yesterday (as @shangyian suggested).

> Some requests couldn't be fulfilled, because there was no common database (or catalog, today). For cases like that in the future we could implement multi-catalog joins. But even without implementing the multi-catalog joins it's important to decouple the concept of a node from where it's materialized (or can be computed). So nodes should not have catalogs, schemas, nor tables associated with them.

So if I understand correctly, this complexity lived in the build logic where it had to account for the variable locations where data for a node may actually be available, also considering certain limitations such as data in druid not being able to join to google sheets? Can you give some use cases where the same exact data would exist in different tables in different systems? Are these multiple representations of the data equivalent?

dj/api/nodes.py
betodealmeida: Ah, thanks @shangyian!

> But agreed that we should discuss this on Wednesday. If a source node has multiple tables, it will also need to manage any schema discrepancies between the nodes.

Yeah, we had that in place — the schema of a source node was the superset of the schemas of its representations. One of the source nodes in the original examples had a column that only existed in GSheets (slow), I think it was called `secret_column`. Any query that used that column would hit GSheets, instead of a faster database, since it was required to fulfill the request.

dj/api/nodes.py
betodealmeida: > So if I understand correctly, this complexity lived in the build logic where it had to account for the variable locations where data for a node may actually be available, also considering certain limitations such as data in druid not being able to join to google sheets?

Right, when fulfilling a request the query builder would figure out which database could execute the query, taking into consideration where the nodes are materialized and the schemas of the tables. If you had (brackets are databases where the node is materialized):

```
A[db1] B[db1, db2]
 \    /
  C[db3]
```

Then the node `C` could be computed in either `db1` or `db3`, but not `db2` (replace database with catalog for the new design). Furthermore, it also accounted for heterogeneous schemas and the cost of each database/table when doing that evaluation. Eg, in the original examples the dimension node had a `secret_column` that existed only in GSheets (a slow DB). Queries that requested that column would then be evaluated in GSheets, all other would use a faster database.

> Can you give some use cases where the same exact data would exist in different tables in different systems?

At Facebook we had data arriving in Kafka-like streams, and they would be stored in Hive. A single source node would point to both, and we could generate streaming SQL or Presto SQL when computing metrics, using the former if needed more recent data and the latter for historical queries.

There's also the use case where a source node can have **multiple tables** in the same schema. At Facebook we had `dim_all_users` (a user dimension table that had 100+ columns) and `dim_fast_users` (a version with fewer columns, only the ones that could be loaded quickly in the daily job). The `users` node pointed to both, and DJ would use the fastest table if it could, unless it needed columns from the slow table.

> Are these multiple representations of the data equivalent?

Yes, they're conceptually the same, and should be consistent, even if they have different schemas or different time ranges of data.

dj/api/helpers.py
shangyian: I moved these to a separate `raise_if_node_exists` function. I can do a separate follow-up that changes this setup for all the `get_X_by_name` helpers.

dj/api/nodes.py
agorajek: +1 on discussing this early on. We can start on Wed but if we need more time we should schedule a separate chat.

One thing to add to the discussion from me would be checking/enforcing the grain (aka PK) in the source nodes that point to multiple tables. I know having separate columns available in different datasets is ok, but I think accounting for different grain would be tough.

dj/api/nodes.py
agorajek: I really like the engine/catalog separation. It makes a better match to real the data stack. But whether the catalog should be part of the node itself or part of the node materialization attributes that's a big question.

PR #409, opened 2023-03-13
Adding endpoints for retrieving data and SQL for a node

shangyian:
### Summary

This PR adds two endpoints:
* `/data/{node_name}` to retrieve data for a node
* `/sql/{node_name}` to retrieve SQL for a node

These endpoints work for sources, transforms, metrics and dimensions. In order to support data/sql generation for source and transform nodes, I removed the check for a `query` on the node and included a stage that builds the AST for a source node. This also required an additional check to make sure that we don't include any fields in the `SELECT` clause that aren't in the `GROUP BY` clause, when the latter exists. 

I'll work on support for cube nodes in a separate PR.

Other changes:
* Bugfix for generated SQL: When users provide a filter, it shouldn't be included in the `SELECT` clause by default. In addition to not making sense, including it in `SELECT` will also generate a query that doesn't run for metrics or any other node with an aggregation (fails in Postgres with `must appear in the GROUP BY clause or be used in an aggregate function`). 
* Removes `query` table as that is now part of djqs.
* Moved the availability state posting from `/data/availability/{node_name}/` to `/data/{node_name}/availability/` to be consistent with the URL order of `/data/{node_name}/...`
* Added docs to `CONTRIBUTING.rst`

These changes require a minor companion change to DJQS: https://github.com/DataJunction/djqs/pull/10

### Test Plan

Added some tests with a mocked query service client. I also ran this with (a slightly modified) DJQS against the DJ Roads tables.

- [X] PR has an associated issue: #342
- [X] `make check` passes

### DIFF
--- a/dj/api/data.py
+++ b/dj/api/data.py
@@ -3,21 +3,25 @@ Data related APIs.
 """
 
 import logging
+from typing import List
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, Query
 from fastapi.responses import JSONResponse
 from sqlmodel import Session
 
-from dj.api.helpers import get_node_by_name
-from dj.errors import DJException
+from dj.api.helpers import get_node_by_name, get_query
+from dj.errors import DJException, DJInvalidInputException
+from dj.models.metric import TranslatedSQL
 from dj.models.node import AvailabilityState, AvailabilityStateBase, NodeType
-from dj.utils import get_session
+from dj.models.query import QueryWithResults
+from dj.service_clients import QueryServiceClient
+from dj.utils import get_query_service_client, get_session
 
 _logger = logging.getLogger(__name__)
 router = APIRouter()
 
 
-@router.post("/data/availability/{node_name}/")
+@router.post("/data/{node_name}/availability/")
 def add_availability(
     node_name: str,
     data: AvailabilityStateBase,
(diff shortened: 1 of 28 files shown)


dj/api/data.py
```diff
@@ -83,3 +87,46 @@ def add_availability(
...
+) -> QueryWithResults:
+    """
+    Gets data for a node
+    """
+    node = get_node_by_name(session, node_name)
+    if node.type not in (NodeType.METRIC, NodeType.CUBE, NodeType.DIMENSION):
+        raise DJException(message=f"Can't get data for node type {node.type}!")
+
+    if node.type == NodeType.DIMENSION:
+        if dimensions or filters:
```
agorajek, line 110: Why not for Dimension node data with filters?

dj/api/data.py
```diff
@@ -83,3 +87,46 @@ def add_availability(
...
+    filters: List[str] = Query([]),
+    async_: bool = False,
+    session: Session = Depends(get_session),
+    query_service_client: QueryServiceClient = Depends(get_query_service_client),
+) -> QueryWithResults:
+    """
+    Gets data for a node
+    """
+    node = get_node_by_name(session, node_name)
+    if node.type not in (NodeType.METRIC, NodeType.CUBE, NodeType.DIMENSION):
```
agorajek, line 106: @shangyian I am trying to come up with a good reason for not allowing users to query data for a source or transform node but I can't. We can always add a default row limit, on any node type. Just trying to understand this restriction better.

agorajek (commented):
Nice little api. But really wondering why the restriction?

dj/api/data.py
shangyian: Good point, in thinking about this more, we can open this up to filters.

dj/api/data.py
shangyian: I guess I just feel like it's a bit unusual for people to try to access sources or transforms through DJ, but I can also see how if someone is in the system and exploring, it might be useful. But in my view the most powerful part of DJ is its ability to generate queries, and for source nodes that's not going to be applicable. I could see being used for transform nodes though.

agorajek: > But in my view the most powerful part of DJ is its ability to generate queries

Fully agree ^, but in this case I think we don't need to tell people that they can't see data for a source or transform node if they just wanna have a peek at `SELECT * FROM any_node LIMIT 10` or something. It just feels like an artificial restriction, even if this is not a common use case.

agorajek:
Two in-line comments, but please adjust at your conveniance.

agorajek (approved):

dj/api/data.py
shangyian: Done! I've added support for dimensions with filters.

dj/api/data.py
shangyian: Makes sense! I just added some changes to support querying both source and transform nodes. It turns out the existing `get_query` doesn't support nodes with no queries, so I added some modifications to generate a select star query AST.

shangyian:
@agorajek I added support for sources + transforms. I'll leave the cube nodes to another PR... Also added a /sql endpoint, which serves as a parallel to /data but just returns the SQL without running it.

agorajek:
Thanks for the extras @shangyian . Adding query for any node sounds great.

Issue #411, opened 2023-03-20
Support dimensions discovery & joining beyond first order dimensions


shangyian:
We currently only support discovery and query-building for first-order dimensions of a given node. This means that we're only getting the dimensions defined immediately on the node and one layer beyond that. However, we should be able to parse through the entire DAG and support building queries based on second-order dimensions (or beyond). 

For a more concrete example, if we look at the `avg_repair_order_discounts` metric in the DJ Roads database, calling:
```python
GET /sql/avg_repair_order_discounts/?dimensions=hard_hat.state
```
should yield the two levels of joins necessary to bring in the `hard_hat` dimension:
```sql
    SELECT
      avg(repair_order_details.price * repair_order_details.discount) AS avg_repair_order_discount,
	  hard_hat.state 
    FROM "roads"."repair_order_details" AS repair_order_details
    LEFT JOIN (
      SELECT
        repair_orders.dispatched_date,
	repair_orders.dispatcher_id,
	repair_orders.hard_hat_id,
	repair_orders.municipality_id,
    	repair_orders.order_date,
        repair_orders.repair_order_id,
        repair_orders.required_date 
      FROM "roads"."repair_orders" AS repair_orders
    ) AS repair_order
        ON repair_order_details.repair_order_id = repair_order.repair_order_id
        LEFT JOIN (
          SELECT
            hard_hats.address,
	    hard_hats.birth_date,
	    hard_hats.city,
	    hard_hats.contractor_id,
            hard_hats.country,
            hard_hats.first_name,
            hard_hats.hard_hat_id,
            hard_hats.hire_date,
            hard_hats.last_name,
            hard_hats.manager,
            hard_hats.postal_code,
            hard_hats.state,
            hard_hats.title 
          FROM "roads"."hard_hats" AS hard_hats
        ) AS hard_hat
            ON repair_orders.hard_hat_id = hard_hat.hard_hat_id 
        GROUP BY  hard_hat.state
```

shangyian:
@agorajek To make sure I'm understanding this right -- what about the current node's columns + their linked dimensions nodes' parents? An example is `avg_repair_price`, which references the `repair_order_details` transform. `repair_order_details` is linked to the `repair_order` dimension, which references the `repair_orders` source. `repair_orders` then has a bunch of linked dimension nodes on its columns, all of which are ideally included in the the current node's available dimensions.

agorajek:
@shangyian let me use an example.

[...]

PR #412, opened 2023-03-21
Listing dimensions for a node should yield all possible dimensions

shangyian:
### Summary

**Nth Order Dimensions:** We currently only display the first-order set of dimensions for a node. However, dimensions can be found on any column in the existing node or on any column in the node's parents, or the parents' parents etc, so we should walk through the parents + dimension links DAG in order to populate the complete set of a node's dimensions.

**Dimension Attributes:** The current setup will yield all columns on a node as available dimensions. However, we should only include them if they've been marked with the `dimension` column attribute.

Note: This doesn't make the necessary changes to the compile + build stages to support query generation for these Nth-order dimensions. A preview of that setup for the old AST [can be found here](https://github.com/DataJunction/dj/compare/main...shangyian:dj:add-multi-level-dims?expand=1), but I'll work on supporting this with the new AST when that's in.

### Test Plan

Ran through some of the existing DJ Roads node examples.

- [X] PR has an associated issue: #411
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/dag.py
+++ b/dj/sql/dag.py
@@ -1,9 +1,10 @@
 """
 DAG related functions.
 """
+import collections
 from typing import List
 
-from dj.models.node import Node, NodeRevision
+from dj.models.node import Node
 from dj.utils import get_settings
 
 settings = get_settings()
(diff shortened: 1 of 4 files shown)


samredai (approved):
🚀

dj/sql/dag.py
```diff
@@ -1,27 +1,35 @@
...
+    while to_process:
+        current_node = to_process.popleft()
+        for column in current_node.current.columns:
+            if any(
+                attr.attribute_type.name == "dimension" for attr in column.attributes
+            ):
+                dimensions.append(f"{current_node.name}.{column.name}")
+            if column.dimension and column.dimension not in processed:
+                processed.add(column.dimension)
+                to_process.extend(column.dimension.current.parents)
```
agorajek, line 30: Based on our conversation [here](https://github.com/DataJunction/dj/issues/411), we may want to skip this part and stick to the closest parents only. Adding dimensions from all upstream parents may be hard/impossible.

shangyian: Yep, makes sense. And to support this correctly, we'll also need to add the dimension attribute "push-down" functionality -- as in, when someone links a column to a dimension node, DJ needs to push that link down to its children, where possible.

agorajek: Yes, as long as the column (with attributes) passed down is not altered any way. But its better to do this check during node building than during dimension discovery imo.

shangyian: Yeah I agree -- it shouldn't be during dimension discovery, but I was thinking it could happen at the dimensions linking stage. When a user links a column to a dimension node, that's when we push the dimension down to downstream nodes, where possible.

shangyian: @agorajek I removed this layer of adding the dimensions' upstreams (and simplified the function). I'll have a separate PR handle the attribute push-down, since it's closely linked to the build functionality.

Issue #351, opened 2023-02-23
Generate Python clients


shangyian:
Use https://github.com/OpenAPITools/openapi-generator to generate a python client based on the OpenAPI spec.

samredai:
Hoping we can re-use this issue to also discuss the user-facing python client (`djclient`) that will wrap the OpenAPI generated client. Here are the actions I can think of that we need to support from `djclient`.
- [ ] Create Catalogs
- [ ] Create Engines
- [ ] Attach Engines to Catalogs
- [ ] Create/Validate Source nodes
- [ ] Create/Validate Transform nodes
- [ ] Create/Validate Dimension nodes
- [ ] Create/Validate Metric nodes
- [ ] Create/Validate Cube nodes
- [ ] View a history summary for a node (versions and creation dates/times)
- [ ] Set a materialization configuration on a node
- [ ] View availability state for a node
- [ ] View a history summary for a node's availability state
- [ ] Roll back a node to a specified version
- [ ] Roll back a node's availability state
- [ ] List Dimensions for a set of Metrics
- [ ] List Metrics for a set of Dimensions
- [ ] List Dimensions for a Cube (in other words, for the Metrics in the Cube)
- [ ] List Metrics for a Cube (in other words, for the Dimensions in the Cube)
- [ ] Get data for a set of compatible Metrics and Dimensions
- [ ] Get data for a Cube (a named set of compatible Metrics and Dimensions)

samredai:
_notes from the last sync_
```py
from djclient import DJClient, Source, Transform, Dimension, Metric

# Creating namespaces as first class objects as opposed to implied namespaces in node name
# Sub-namespaces cannot have the same name as parent

class DJClient:
    def __init__(self, host):
        self.namespace = "foo"
        self.client = openapiClient("http://localhost:8000")


dj = DJClient(“http://localhost:8000/foo)
dj.sources()                                             # lists sources
dj.transforms()
dj.dimensions()
dj.metrics()

# Row limit 200,001

repair_orders_external_table = Source(...)   # instantiate a source
repair_orders = Dimension(...)
avg_repair_price = Metric(...)
avg_time_to_dispatch = Metric(...)

foo = Transform(...)

# Validategit
avg_time_to_dispatch.validate()

# Publish individual node
repair_orders_external_table.publish()

# Publish to the server
dj.publish(nodes=[...]) # publish mulitple nodes at the same time

# List dimensions for one or more metrics
[...]

PR #413, opened 2023-03-23
poetry config for openapi client

samredai:
### Summary

This adds a poetry config for the openapi generated client. As currently configured (suggestions welcome) it publishes the client to pypi as `djclient`. I don't expect the openapi generator to change dependencies often so I did `cat requirements.txt | xargs -I % sh -c 'poetry add "%"'` and don't expect we'll have to do that again any time soon.

Thoughts welcome on what our [PEP 440](https://peps.python.org/pep-0440/) strategy should be. For a first publish I have it as `0.1.0a1` as an alpha. We could probably script auto increments of the alpha tag and have that release automatically, then manually cut an `rc` tag when we're ready to test it out before a final release.

Deploying this to PyPi is as simple as:

1. Update the version in the `[tool.poetry]` section in `openapi/python/pyproject.toml`
2. Run `poetry build` (creates the wheel and zipped source file in a `dist` directory)
3. Run `poetry publish` (publishes the version to pypi)

**note**: I also moved the generated openapi clients from the top-level `clients` directory to a top level `openapi` directory. Since we're going to eventually have a user-facing python client that wraps the openapi client, I didn't want the `client` directory to be confusing.

### Test Plan

N/A

- [x] PR has an associated issue: #351 
- [x] `make check` passes

### Deployment Plan

N/A


agorajek:
This LGTM. I don't have any strong opinions on the versioning or dependency patterns.

agorajek (approved):

shangyian (approved):
Nice! In terms of versioning, are we going with the strategy of releasing a library for the openapi client and a separate library that includes this auto-generated library as a dependency for the user-friendly client? I agree with the approach of auto-releasing alphas. We can cut releases on both the openapi client and the user-friendly client each time we update the latter to include the latest features from the generated client.

samredai:
@shangyian Yeah that's exactly what I was picturing. Ideally the REST API endpoints stabilize and we can keep them consistent across an entire major version of DJ (making room for bug-fixes of course). In that case, we can stay pinned to a specific openapi client version for a while, even through many versions of the user-friendly client.

PR #414, opened 2023-03-23
rename djclient to djopenapi

samredai:
### Summary

This renames the openapi generated python client to `djopenapi`, as to preserve the `djclient` library name for the user-friendly client that will wrap this.

### Test Plan

Will manually run the workflow that generates the client.

- [x] PR has an associated issue: #351 
- [ ] `make check` passes

### Deployment Plan

N/A


.github/workflows/generate-clients.yml
```diff
@@ -5,6 +5,7 @@ on:
       - main
     paths:
       - dj/**
+  workflow_dispatch:
```
samredai, line 8: I added this so that we can also manually trigger the workflow to generate the client. I'll do that after this PR is merged so that it can regenerate with the new name `djopenapi`.

shangyian (approved):

PR #425, opened 2023-04-06
Add docker page

samredai:
### Summary

Fills in the Getting Started/Docker page that covers starting up DJ + DJQS and trying out the example notebooks.

### Test Plan

Ran `hugo serve --contentDir=content/0.1.0/` and navigated to [http://localhost:1313/](http://localhost:1313/).

- [x] `make check` passes

### Deployment Plan

N/A


docs/content/0.1.0/docs/getting-started/docker.md
```diff
@@ -0,0 +1,80 @@
+---
+weight: 10
+---
+
+# Docker
+
+The easiest way to try out DJ is to use the docker setups that are available in
```
shangyian, line 7: Can we users to use `dj-demo`? That will start all the necessary services in conjunction (including the jupyter notebook server) so that people can see how it all works together.

samredai: Ok I updated to open with a `DataJunction Demo Docker Compose Setup` section that specifies that it's the easiest way to try it out. I was going to remove the guide for starting the individual services, but I figured it may actually be useful, even if just to describe how these individual services connect to each other. I also figured it may be useful to describe starting up just the notebook server and updating the DJ server info at the top of the notebook to connect to a remote DJ server.

shangyian: Yeah agreed that it's still useful!

shangyian (approved):

PR #427, opened 2023-04-06
Fix source node schema inference

shangyian:
### Summary

Upon source node creation, we infer the schema of the node based on the table schema, but this breaks with the new Spark-backed DJQS, as it doesn't specify which engine to use. This change passes in the right engine.

A similar change is also needed in DJRS. Companion to https://github.com/DataJunction/djqs/pull/13 

### Test Plan

Ran it end-to-end. We can actually modify the DJ Roads notebook to not include the column schemas upon source node creation, but I've left that in there for reference purposes. 

- [X] PR has an associated issue: #
- [X] `make check` passes

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -494,6 +494,7 @@ def create_source_node(
             data.catalog,
             data.schema_,  # type: ignore
             data.table,
+            catalog.engines[0] if len(catalog.engines) >= 1 else None,
         )
 
     node_revision = NodeRevision(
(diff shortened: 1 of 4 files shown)


samredai (approved):
Looks good, thanks @shangyian! Since we have the [dj-demo](https://github.com/datajunction/dj-demo) repo now, should we remove the notebooks directory in this repo? That way we can maintain the set of example notebooks in a single place.

shangyian:
@samredai Sure! Let me do that in a separate PR

Issue #438, opened 2023-04-08
Invalid query building on dimensions with aliased join keys


shangyian:
Consider the query for the following dimension `us_states`, linked to another dimension `hard_hat` on `us_state.state_short = hard_hat.state`:
```
SELECT
    us_states.state_id,
    us_states.state_name,
    us_states.state_region,
    us_region.us_region_description AS state_region_description,
    us_states.state_abbr AS state_short
FROM roads.us_states AS us_states
LEFT JOIN roads.us_region AS us_region
ON us_states.state_region = us_region.us_region_id
```
Because the above dimension has an aliased join key (`states_abbr AS state_short`), when a query needs to be build that joins together the two dimensions, the generated join `ON` is invalid:
```
LEFT JOIN ...
ON hard_hat.state = us_states.state_abbr AS state_short
```

Instead, the alias should just replace the name:
```
LEFT JOIN ...
ON hard_hat.state = us_states.state_short
```

PR #439, opened 2023-04-08
Fixing invalid query building on dimensions with aliased join keys

shangyian:
### Summary

During query building, joining in dimensions with an aliased join key should use the alias directly, rather than constructing an invalid join `ON` with the `AS` clause. This fix "promotes" the alias name to the actual join column name, which results in a valid query. 

### Test Plan

Wrote unit tests. Also tested on some actual queries running locally and they're generating runnable queries.

- [X] PR has an associated issue: #438 438
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/construction/build.py
+++ b/dj/construction/build.py
@@ -160,15 +160,11 @@ def _build_joins_for_dimension(
                 left_table.add_ref_column(
                     cast(ast.Column, join_left_columns[join_col.name]),
                 )
-                join_right_col = join_right_columns[
-                    join_col.dimension_column or "id"
-                ].copy()
-                join_right_col.name = join_right_col.alias_or_name
-                join_right_col.alias = None
                 join_on.append(
                     ast.BinaryOp.Eq(
                         join_left_columns[join_col.name],
-                        join_right_col,
+                        join_right_columns[join_col.dimension_column or "id"],
+                        use_alias_as_name=True,
                     ),
                 )
             else:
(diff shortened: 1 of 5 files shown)


dj/construction/build.py
```diff
@@ -160,10 +160,15 @@ def _build_joins_for_dimension(
...
+                join_right_col = join_right_columns[
+                    join_col.dimension_column or "id"
+                ].copy()
+                join_right_col.name = join_right_col.alias_or_name
+                join_right_col.alias = None
                 join_on.append(
                     ast.BinaryOp.Eq(
                         join_left_columns[join_col.name],
-                        join_right_columns[join_col.dimension_column or "id"],
+                        join_right_col,
```
samredai, line 171: Instead of making a copy of the column like you do a few lines above, could we just handle it inside of the `BinaryOp.__str__` function?

If it's because we only want to preference the alias in certain situations, maybe we can have a `prefer_alias: Optional[bool] = False` attribute on BinaryOp, and then do something like this in `BinaryOp.__str__` (while using `prefer_alias = True`on this line)
```py
    def __str__(self) -> str:
        if self.prefer_aliases and self.right.alias:
            right = self.right.copy()
            right.name = self.right.alias
            right.alias = None
        else:
            right = self.right
        ret = f"{self.left} {self.op.value} {right}"
        if self.parenthesized:
            return f"({ret})"
        return ret
```
Sort of a separate question: why do we only need to do this to the right column?

shangyian: Yeah, having this live in `__str__` makes more sense to me as well. I moved the alias promotion to a separate function on `Column`, since this can't apply to non-Column generic Expressions. And good call -- I added this check to both the left and the right column.

samredai (approved):
🚀

shangyian:
Oh wait I was going to add more tests... oh well, different PR.

samredai:
Whoops 😅

PR #443, opened 2023-04-09
Basic Python Client

shangyian:
### Summary

This change adds some basic functionality to the Python client.
* List catalogs
* List engines
* List nodes, including by the different node types
* Create source nodes

At some point we may want to export the shared models from DJ core as a separate package so that they can be used in client libraries like this one. I didn't want to make this dependent on DJ core, as that would mean users of the client would need to install a lot of unnecessary dependencies. Also used pyscaffold to generate the scaffolding for `setup.py`.

### Test Plan

- [X] `make check` passes

### Deployment Plan

N/A


samredai:
This is a great start! I realized I didn't upload the notes/snippets from the last sync so I've added them to the issue [here](https://github.com/DataJunction/dj/issues/351#issuecomment-1501175563).

> At some point we may want to export the shared models from DJ core as a separate package so that they can be used in client libraries like this one.

This makes sense to me although I think we'll extend the classes quite a bit to add a bunch of client side convenience features (like calling `.metrics()` on a catalog instance to see the metrics defined in a given catalog). Assuming we keep these models mostly stable, maybe it's not so bad that we define them in both places and not have to worry about having to pin the extra dependency version for something like a `djmodels` library.

samredai (approved):

client/python/README.md
agorajek, line 1: Love it!

agorajek:
Thank you @shangyian !

PR #447, opened 2023-04-10
Additional functionality for DJ Client

shangyian:
### Summary

This adds additional functionality for the DJ client:
* Linking dimensions to a node
* Retrieving dimensions for a metric
* Getting SQL for a node given dimensions and filters
* Retrieving data for a node (for now this only supports retrieval as a pandas dataframe, but we can add additional options later)

### Test Plan

Running locally in a notebook

- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/client/python/README.md
+++ b/client/python/README.md
@@ -9,31 +9,107 @@ pip install djclient
 ## Examples
 
 To initialize the client:
-```
-from djclient import DJClient
+```python
+from djclient.dj import DJClient, Source, Dimension, Transform, Metric
 client = DJClient("http://dj-endpoint:8000")
 ```
 
+### Catalogs and Engines
+
 To list available catalogs for the DJ host:
-```
+```python
 client.catalogs()
 ```
 
 To list available engines for the DJ host:
-```
+```python
 client.engines()
 ```
 
-ALl nodes can be found with:
+To create a catalog:
+```python
+catalog = Catalog(
+    name="prod"
+)
+catalog.publish()
 ```
+
+To create an engine:
+```python
+engine = Engine(
+    name="spark",
+    version="3.2.2",
+    uri="..."
+)
+engine.publish()
+```
+
+To attach an engine to a catalog:
+```python
+catalog.add_engine(engine)
+```
+
+### Nodes
+
+All nodes can be found with:
+```python
 client.nodes()
 ```
 
 Specific node types can be retrieved with:
-```
+```python
 client.sources()
 client.dimensions()
 client.metrics()
 client.transforms()
 client.cubes()
+```
+
+To create a source node:
+```python
+repair_orders = Source(
+    name="repair_orders",
+    display_name="Repair Orders",
+    description="Repair orders",
+    catalog="dj",
+    schema_="roads",
+    table="repair_orders",
+)
+repair_orders.publish()
+```
+
+Nodes can also be created as drafts with:
+```python
+repair_orders.draft()
+```
+
+To create a dimension node:
+```python
+repair_order = Dimension(
+    name="repair_order",
+    query="""
+    SELECT
+      repair_order_id,
+      municipality_id,
+      hard_hat_id,
+      dispatcher_id
+    FROM repair_orders
+    """,
+    description="Repair order dimension",
+)
+repair_order.publish()
+```
+
+To create a metric:
+```python
+num_repair_orders = Metric(
+    name="num_repair_orders",
+    query="""
+    SELECT
+      count(repair_order_id) AS num_repair_orders
+    FROM repair_orders
+    """,
+    description="Number of repair orders",
+)
+num_repair_orders.publish()
 ```
--- a/client/python/README.md
+++ b/client/python/README.md
@@ -104,6 +104,25 @@ repair_order = Dimension(
 repair_order.publish()
 ```
 
+To create a transform node:
+```python
+from djclient import Transform
+large_revenue_payments_only = Transform(
+    name="large_revenue_payments_only",
+    query="""
+    SELECT
+      payment_id,
+      payment_amount,
+      customer_id,
+      account_type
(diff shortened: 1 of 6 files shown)


client/python/README.md
```diff
@@ -9,31 +9,107 @@ pip install djclient
 ## Examples
 
 To initialize the client:
-```
-from djclient import DJClient
+```python
+from djclient.dj import DJClient, Source, Dimension, Transform, Metric
```
samredai, line 13: I see why the `.dj` was added, but is it worth using the `__all__` trick in `__init__.py` for these main classes so we can keep it `from djclient`?

client/python/README.md
```diff
@@ -9,31 +9,107 @@ pip install djclient
...
 client = DJClient("http://dj-endpoint:8000")
 ```
 
+### Catalogs and Engines
+
 To list available catalogs for the DJ host:
-```
+```python
 client.catalogs()
```
samredai, line 21: I'd find it clearer if we instantiate the client as `dj = DJClient("http://dj-endpoint:8000")`, then this line would become `dj.catalogs()`. What do you think?

client/python/README.md
```diff
@@ -9,31 +9,107 @@ pip install djclient
...
+```
+
+To create an engine:
+```python
+engine = Engine(
+    name="spark",
+    version="3.2.2",
+    uri="..."
+)
+engine.publish()
```
samredai, line 46: Is the reason for the `REGISTRY` map in order to support this `.publish()` syntax that can find the client in the global namespace? I know this is mainly opinion here but I'd find it a bit clearer to use the dj client directly, something like `dj.publish(engine)`. Plus it's only 2 more characters. :D

client/python/README.md
shangyian: Oh definitely -- forgot to add this

client/python/README.md
shangyian: Pretty much. I think some folks might prefer `engine.publish()`, but I also don't think it's a huge hassle to swap to `dj.publish(engine)`. Let me make this switch

shangyian: @samredai Although what about things like `catalog.add_engine(...)`? I think that's cleaner than `dj.add_engines_to_catalog(...)`, but it fits with the current pattern better of `<obj>.<verb>` rather than `<client>.<verb>(<ob>)`

samredai: I see, yeah I agree that's not ideal. And having to pass the client as an argument to these object methods is only slightly better. Maybe we should stick with the original setup here then. @betodealmeida @agorajek @CircArgs any thoughts?

samredai: Also just to clarify, by "passing the client as an argument" I meant `catalog.add_engine(dj, engine)`. Always having to pass in the client is something that's easy to forget though...

Another alternative I can think of is that new objects are created by methods on the client and the client sets itself as an attribute on the object.
```py
from djclient import DJClient

dj = DJClient("http://dj-endpoint:8000")

catalog = dj.new_catalog(
    name="prod"
)
catalog.publish()

engine = dj.new_engine(
    name="spark",
    version="3.2.2",
    uri="..."
)
engine.publish()

catalog.add_engine(engine)
```

shangyian: Another thing for catalogs: I think in the `add_catalog` endpoint, the DJ core service should also call DJQS and add the catalogs + engines there. I was considering wrapping this functionality into the client, but I think having DJ call DJQS is cleaner since then the client wouldn't need to provide/know about a DJQS endpoint.

samredai: Having to add it in both places does seem redundant but I don't think it's too bad since these are more server administrative tasks right? I'm wondering now if we should even allow the user-facing client to create catalogs and engines. I'd expect that's configured once and rarely updated.

shangyian: Yeah, perhaps we can just remove these from the user-facing client, since I'm not sure users will create catalogs correctly anyway 😆 But I'd still prefer it if DJ core could keep DJQS in sync instead of making the server admin make both calls.

agorajek: Agree that managing catalogs and engines can be less user friendly for the sake of keeping things simple. I don't have much preferences around `object.publish()` over `dj.publish(object)`.

client/python/README.md
agorajek, line 1: Do we show examples of creating Transform nodes yet?

And do we let the user specify the SQL flavor they give us for the Transform node? I understand we can have some default, but either way we should store it on the node after we take it in right?

agorajek (approved):
LGTM, but I left few comments inline.

client/python/README.md
samredai: I was discussing this earlier with @shangyian. We can add an `engine` attribute to nodes but I think an important behavior that we'll want is to automatically set it for the user if there's only a single engine available for a catalog. This way they'd only have to specify it if there are multiple engines available, i.e. Spark or Trino. Does that sound reasonable?

shangyian: @agorajek good point, let me add some examples of Transforms. 

As for the dialect, I agree that we can use engine as a substitute for dialect. Arguably we can let users create SQL in any dialect they want, and we can do the translation into a dialect compatible with the available engine, but that's not something we want to deal with at the moment I think.

shangyian: Also @samredai I'll put the client format discussion (i.e., `dj.publish(...)` vs `node.publish()` as a topic for tomorrow. I'm not that strongly opinionated so we can change it if folks feel strongly one way or the other. But for the time being I'll just merge this so that we can build on top of it.

agorajek: Sounds good. I agree we should not have to deal with dialects for now, but just having it stamped on the transform node will help us keep things clean imo. Engine seems like an ok proxy.

samredai (approved):

PR #449, opened 2023-04-10
Add node deletion

shangyian:
### Summary

Adds node deletion. Also provides Python client support for deletion.

### Test Plan

Ran locally through the client and through the REST API directly. Also testing just from regular use of this functionality as I create broken nodes 😆 

- [x] PR has an associated issue: #448
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/client/python/djclient/dj.py
+++ b/client/python/djclient/dj.py
@@ -244,7 +244,7 @@ class Node(BaseModel):
 
     def delete(self):
         """
-        Sets the node's mode to DRAFT and pushes it to the server.
+        Deletes the node
         """
         session = self._get_initialized_client()
         response = session.delete_node(self)
--- a/client/python/djclient/dj.py
+++ b/client/python/djclient/dj.py
@@ -137,6 +137,13 @@ class DJClient:
         """
         return self._nodes_by_type(type_="cube", names_only=names_only)
 
+    def delete_node(self, node: "Node"):
+        """
+        Delete this node
+        """
+        response = self._session.delete(f"/nodes/{node.name}/", timeout=self._timeout)
+        return response
+
     def create_node(self, node: "Node", mode: "NodeMode"):
         """
         Helper function to create a node.
@@ -234,6 +241,15 @@ class Node(BaseModel):
         """
         session = self._get_initialized_client()
         session.create_node(self, NodeMode.DRAFT)
+
+    def delete(self):
+        """
+        Sets the node's mode to DRAFT and pushes it to the server.
+        """
+        session = self._get_initialized_client()
+        response = session.delete_node(self)
+        assert response.status_code == 204
+        return f"Successfully deleted `{self.name}`"
 
 
 class Source(Node):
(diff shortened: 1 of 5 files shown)


dj/api/nodes.py
```diff
@@ -213,6 +213,16 @@ def read_node(name: str, *, session: Session = Depends(get_session)) -> NodeOutp
     return node  # type: ignore
 
 
+@router.delete("/nodes/{name}/", status_code=204)
+def delete_node(name: str, *, session: Session = Depends(get_session)):
+    """
+    Delete the specified node.
+    """
+    node = get_node_by_name(session, name, with_current=True)
+    session.delete(node)
```
samredai, line 222: In this function we need to get all of the downstream nodes and switch their status to invalid. That will probably be a common action so maybe a helper function that just invalidates all downstream nodes will probably be useful.

I'm not sure how to handle if this is a dimension node and there are columns on other nodes linking to it. On the one hand we could simply delete the link as well but that makes deleting a dimension node potentially have a very large blast radius.

This might need some more discussion so feel free to leave these for a follow-up PR.

samredai (approved):

dj/api/nodes.py
shangyian: Good point, I forgot about invalidation! Let me add that.

Although regarding dimension nodes having a large blast radius - I'm not sure that's the case. Wouldn't we just remove the links to that dimension for any nodes that have a field linked to that dim node? The downstream nodes (via dimension links) themselves don't need to be invalidated. However, if it's a case of downstream nodes that reference that dimension node as a child, then they'll need to be invalidated, but I don't think there would be that many of those in the system.

samredai: Yeah that's true, I was thinking of a scenario where someone naively thinks, "oh let me just delete this dimension and recreate it from scratch", not realizing that all of the potential links will be dropped and need to be re-added. Maybe that's not as much as a concern as it seems. 🤔

agorajek: Wow, this a great question... my gut feeling for dimension nodes deletion is that we should not invalidate any of the dim/connected nodes but we may simply consider a missing dimension link as a run time error that we could handle gracefully (or not) depending on the situation. Either way I would treat this linkage loosely so as to not cause a node panic/invalidation for this reason. The other side (parent-child nodes) is clear to follow the valid pattern but perhaps dimension links can be a softer touch, which would allow someone to just drop a dimension node and recreate it. Just a thought. I agree we should run this by others, I can see an argument for keeping it as strict as the parent-child node relationship.

client/python/djclient/dj.py
```diff
@@ -235,6 +242,15 @@ def draft(self):
         session = self._get_initialized_client()
         session.create_node(self, NodeMode.DRAFT)
 
+    def delete(self):
+        """
+        Sets the node's mode to DRAFT and pushes it to the server.
```
agorajek, line 247: Where do you set the mode to DRAFT? I think I may be missing something...

shangyian:
I'm going to merge this for now and add "what to do about dimensions links upon node deletion" as a discussion topic to Wednesday. We can put out a follow-up PR if the conclusions from that are different 🎸

PR #450, opened 2023-04-11
Rename API functions and set OpenAPI workflows to manual dispatch

samredai:
### Summary

This PR does two main non-functional things.

1. Sets the OpenAPI workflows to only run on `workflow_dispatch` and splits them out into a workflow that generates the client and one that only generates the spec (we can use each as needed as part of releases)
2. Renames a number of the API functions so that they have more readable headers in the generated docs--for example `Read Metrics` -> `List Metrics` and `Create Cube Node` -> `Create Cube` (see image below to see the **_before_** names)
Screenshot 2023-04-10 at 8 34 36 PM

### Test Plan

`make check` and `make test`

- [ ] `make check` passes

### Deployment Plan

N/A


dj/api/tags.py
```diff
@@ -104,7 +104,7 @@ def update_tag(
 
 
 @router.get("/tags/{name}/nodes/", response_model=List[str])
-def find_nodes_by_tag(
+def list_tagged_nodes(
```
shangyian, line 107: perhaps `list_nodes_for_tag`?

dj/api/data.py
```diff
@@ -22,7 +22,7 @@
 
 
 @router.post("/data/{node_name}/availability/")
-def add_availability(
+def add_an_availability(
```
shangyian, line 25: what about `add_an_availability_state`?

shangyian (approved):

Issue #454, opened 2023-04-12
Require setting primary key on dimension nodes


shangyian:
When a user creates a dimension node, we should require that they tell us which column or columns are part of the primary key. Then when we're assembling the join to bring in the dimension, we can default to the dimension's primary key when no `dimension_column` is explicitly provided (it currently defaults to `id`, which is not accurate for the majority of use cases).

Issue #446, opened 2023-04-10
Check for compatible types when connecting a column to a dimension


samredai:
Right now we don't check that a type is compatible for joining to the specific column on the dimension node. We should check that both at the time the request comes in to associate the two columns and possibly at build time as well (I feel less certain about the latter).

PR #455, opened 2023-04-12
Require primary key for dimension nodes + type check compatibility during linking

shangyian:
### Summary

When someone creates a dimension node, they should set the column(s) that are part of dimension's primary key. During query building, when we assemble the join to bring in the dimension, we'll default to the dimension's primary key when no `dimension_column` is explicitly provided. The existing setup just uses `id` by default, which is not accurate for the majority of use cases as `id` may or may not exist in the columns in the node.

This change also adds a check to make sure the types of the target and dimension columns are compatible during dimension node linking.

One additional requirement that might be worth adding: making sure that links only happen on the dimension node's primary key columns, rather than allowing arbitrary column linking.

### Test Plan

  - #454
  - #446
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -13,6 +13,7 @@ from fastapi.responses import JSONResponse
 from sqlalchemy.orm import joinedload
 from sqlmodel import Session, select
 from starlette.requests import Request
+from starlette.responses import Response
 
 from dj.api.helpers import (
     get_attribute_type,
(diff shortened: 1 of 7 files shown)


tests/examples.py
```diff
@@ -519,13 +526,6 @@
         ),
         {},
     ),
-    (
```
shangyian, line 522: Turns out this dimension link fails the type compatibility check 😅

shangyian, line 522: Turns out this dimension link fails the type compatibility check 😅

dj/api/nodes.py
```diff
@@ -559,10 +580,37 @@ def create_a_node(
     Create a node.
     """
     node_type = NodeType(os.path.basename(os.path.normpath(request.url.path)))
+
+    if node_type == NodeType.DIMENSION and not data.primary_key:
```
agorajek, line 584: +1

PR #456, opened 2023-04-12
Adding PERCENT_RANK

shangyian:
### Summary

Adds the `PERCENT_RANK` function.

### Test Plan

Added a test to make sure it parses + the type inference is correct.

- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/functions.py
+++ b/dj/sql/functions.py
@@ -758,6 +758,18 @@ def infer_type(  # noqa: F811  # pylint: disable=function-redefined
     return ct.DoubleType()
 
 
+class PercentRank(Function):
+    """
+    Window function: returns the relative rank (i.e. percentile) of rows within a window partition
+    """
+
+    is_aggregation = True
+
+    @staticmethod
+    def infer_type() -> ct.DoubleType:
+        return ct.DoubleType()
+
+
 class Quantile(Function):  # pragma: no cover
     """
     Computes the quantile of a numerical column or expression.
--- a/tests/construction/inference_test.py
+++ b/tests/construction/inference_test.py
@@ -378,7 +378,8 @@ def test_infer_types_min_max_sum_ceil(construction_session: Session):
             (PARTITION BY first_name ORDER BY last_name),
           SUM(id) OVER
             (PARTITION BY first_name ORDER BY last_name),
-          CEIL(id)
+          CEIL(id),
+          PERCENT_RANK() OVER (PARTITION BY id ORDER BY id)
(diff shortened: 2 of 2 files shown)


agorajek (approved):
Love how easy it is  (and was) to add new UDFs. 

But... this PR makes me realize that (at some point) we could use another attribute to represent window-kosher agg functions. Such functions need not only a `(state + element) --> (new_state)` transition but also `(state - element) --> (new_state)`. And this is surely nice to have as we can simply allow the users to write the SQL that's runnable on any given DB. This deeper knowledge of UDFs could potentially throw warnings so people can see, when their numbers don't add up.

shangyian:
> But... this PR makes me realize that (at some point) we could use another attribute to represent window-kosher agg functions. Such functions need not only a `(state + element) --> (new_state)` transition but also `(state - element) --> (new_state)`. And this is surely nice to have as we can simply allow the users to write the SQL that's runnable on any given DB. This deeper knowledge of UDFs could potentially throw warnings so people can see, when their numbers don't add up.

Agreed, this would be very helpful!

dj/sql/functions.py
```diff
@@ -758,6 +758,18 @@ def infer_type(  # noqa: F811  # pylint: disable=function-redefined
...
 
 
+class PercentRank(Function):
+    """
+    Window function: returns the relative rank (i.e. percentile) of rows within a window partition
+    """
+
+    is_aggregation = True
+
+    @staticmethod
```
CircArgs, line 768: this could be 
```python
@PercentRank.register
def infer_type() -> ct.DoubleType:
        return ct.DoubleType()
```

And then calling would use the error from dispatch for a consistent end-user experience with exceptions.

CircArgs: @shangyian @samredai the dispatch stuff was shoved in during the rewrite of compile and such. any thoughts on its usage now that we are adding new functions? I had only converted some functions to use it. You can see some functions are overloaded by using it and so the thinking was that rather than write the type checks in the functions themselves we can just have this dispatcher be aware of our own `ColumnTypes` (whereas regular single dispatch would not manage this)

shangyian: Good point @CircArgs, let me open another PR with the change. I also realized that we didn't check to see if the Raw functionality still works with the new build/compile (in fact, the tests for them are commented out).

In terms of usage, I think the dispatch pattern makes sense, but at some point we should go and convert all the functions to use this pattern so that anyone new coming in and adding a function can just copy and paste the pattern. Well, that and we probably need docs

Issue #435, opened 2023-04-07
First-class support for namespaces


shangyian:
We should support namespaces as first-class citizens in DJ. This means that we'd allow namespace management from an API. Nodes that are created with no specified namespace can be slotted under a `default` namespace.

PR #457, opened 2023-04-13
Node namespaces with a default 'default' namespace

samredai:
### Summary

As discussed in the recent sync, this changes namespaces to be more explicit. This PR adds a `POST /nodes/namespace/{namespace}/` endpoint and a `GET /nodes/namespace/all/` endpoint to create and list node namespaces. It keeps in-tact the payloads used to create nodes, so creating a namespaced node is still as simple as giving it a name `a.b.c` (node named "c" in namespace "a.b"), but what happens immediately server side is the namespace portion is used to lookup if a namespace exists by that name. If no `.` is used in the name, such as `foo`, then it's automatically assigned to a `default` namespace. That does mean that a new DJ server will need to be primed by creating a `default` namespace.

Since names are immutable, I've kept the namespace prefix in the actual node name since that seems much more convenient than having to always disambiguate a node name by fetching the namespace. That means the stored node attributes will roughly look like this:
```json
{
  "name": "a.b.c",
  "namespace": "a.b",
  ...
}
```

I've also added a namespace for listing namespaces to the python client.
```py
from djclient import DJClient

dj = DJClient()
dj.namespaces()
# [{"namespace": "default"}]
```

### Test Plan

N/A

- [x] PR has an associated issue: #435 
- [x] `make check` passes

### Deployment Plan

N/A


dj/api/nodes.py
```diff
@@ -567,6 +590,51 @@ def create_a_source(
     return node  # type: ignore
 
 
+@router.post("/nodes/namespace/{namespace}/", status_code=201)
```
shangyian, line 593: Do we want this under `/nodes` or can it instead be under its own endpoint `/namespaces/`?

dj/api/nodes.py
```diff
@@ -567,6 +590,51 @@ def create_a_source(
...
+def create_a_node_namespace(
+    namespace: str,
+    session: Session = Depends(get_session),
+) -> JSONResponse:
+    """
+    Create a node namespace
+    """
+    if get_node_namespace(
```
shangyian, line 601: Does this part care if you create a namespace where its parent namespace doesn't exist yet? For example can you create `basic.users` without having already created `basic`? And if so, should we just auto-create `basic` in that case as an available namespace?

dj/api/nodes.py
samredai: You're right I like that much better, changed it to `/namespaces/{namespace}/` 🚀

dj/api/nodes.py
samredai: Right now this line will raise a 404 and return a json error if the namespace doesn't exist. We definitely could make it auto-create the namespace but doesn't that bring us back to the original design? I thought we were moving in the other direction to make namespaces more explicit. Some people have even asked if namespace creation can be locked down to a curator/server-admin. 😨 

I'm not convinced which direction is better. Is there any value in keeping it explicit for now and then adding the auto namespace creation later? It could also be something that's instrumented client side and implemented differently by different clients, i.e. maybe the python client automatically creates the namespace when it gets the 404 error but a UI requires an explicit namespace creation by the user.

shangyian: Yeah, that's true.. I guess I view the value of explicit namespaces as largely around being able to track the namespaces in a separate field, which lets us filter by namespace more easily, rather than trying to parse each node's name to extract the implicit namespace. That makes queries like "find all nodes under X namespace" faster to execute. It also opens things up later on, like if we wanted to add ACLs to namespaces or limit access in some other way. But I view creating us auto-creating some namespaces as acceptable, or at least something that could vary depending on the DJ deploy.

shangyian (approved):

dj/api/nodes.py
agorajek: This is a tough call. We want to make it seem like working with DJ is easy, but also we should be aware of the problem of metric proliferation. I think a middle ground could be that by default:
- the top level namespaces would require a direct call
- and inside them we could make the namespace creation automatic, with the ability to lock it down by the top-level namespace owner. 

This could offer some notion of a free but controlled environment.

Issue #458, opened 2023-04-14
Optimize build through column pruning


shangyian:
When we join in a dimension, we're either selecting directly from a materialized node or building from a query if no materialization exists. However, we currently select all fields on the query/table, but we should optimize this stage by pruning columns that aren't required later. In other words, in each subquery, we should select only the columns that are required for the outer query. Depending on the query engine and how it generates a query plan, this will reduce the amount of data that needs to be processed and transferred. 

This is a case where Trino does these optimizations in its query plan (which allows us to be lazy), but Spark does not.

PR #461, opened 2023-04-15
Optimize build + add dialect + `UNNEST` + type compatibility fixes

shangyian:
### Summary

This PR has a number of miscellaneous fixes.

**Build optimization**

This change optimizes the SQL built by pruning unnecessary columns from the select of each joined subquery. It only includes columns that are (a) explicitly used in the outer query's select, where or group by clauses, (b) part of the primary key, or (c) useable as a dimension join column.

**Dialect**

This allows users to optionally set a dialect for each engine. Engines themselves can have arbitrary names, but the dialects are restricted to a set of supported SQL dialects. When users ask for SQL or data, they can either specify a dialect or rely on DJ to determine an appropriate dialect based on the available engines. There is currently no detection of nodes with incompatible dialects.

**`UNNEST`**

Adds the `UNNEST` function to support Trino's version of `EXPLODE`. 

Also added tests for SQL building of metric nodes that bring in dimensions with both `LATERAL VIEW ... EXPLODE` and the equivalent version in `CROSS JOIN ... UNNEST` and make sure that both versions can be built and resolve to runnable queries.

**Type Compatibility + More Types**

The original type compatibility checks when attaching dimensions to columns was just a simple check on column type equality, but this doesn't work because types like `int` and `long` weren't considered compatible, but they should be for the purposes of joining.

This adds a function that checks for common ancestors between two types, up to the top-level ancestors like `ColumnType` or `PrimitiveType`, meaning that that `int` and `long` will both resolve to their common ancestor `IntegerBase`, thus resolving as compatible, whereas `int` and `string` will resolve to `ColumnType` as their common ancestor, thus resolving as incompatible.

### Test Plan

  - #458
  - #446
- [x] `make check` passes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,10 @@
 FROM python:3.10
 
-ARG RELOAD="--reload"
-ENV RELOAD ${RELOAD}
-
 WORKDIR /code
 COPY . /code
 RUN pip install --no-cache-dir --upgrade -r /code/requirements/docker.txt
 RUN pip install -e .
 RUN pip install opentelemetry-distro==0.38b0 opentelemetry-exporter-otlp-proto-grpc==1.17.0
 
-CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "${RELOAD}"]
+CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000"]
 EXPOSE 8000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
 FROM python:3.10
+
+ARG RELOAD="--reload"
+ENV RELOAD ${RELOAD}
+
 WORKDIR /code
 COPY . /code
 RUN pip install --no-cache-dir --upgrade -r /code/requirements/docker.txt
 RUN pip install -e .
 RUN pip install opentelemetry-distro==0.38b0 opentelemetry-exporter-otlp-proto-grpc==1.17.0
 
-CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "--reload"]
+CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "${RELOAD}"]
 EXPOSE 8000
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,5 +5,5 @@ RUN pip install --no-cache-dir --upgrade -r /code/requirements/docker.txt
 RUN pip install -e .
 RUN pip install opentelemetry-distro==0.38b0 opentelemetry-exporter-otlp-proto-grpc==1.17.0
 
-CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000"]
+CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "--reload"]
 EXPOSE 8000
--- a/alembic/versions/2023_04_13_2048-4f6449741463_initial_migration.py
+++ b/alembic/versions/2023_04_13_2048-4f6449741463_initial_migration.py
@@ -81,6 +81,15 @@ def upgrade():
         sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column("version", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column("uri", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column(
+            "dialect",
+            sa.Enum(
+                "SPARK",
+                "TRINO",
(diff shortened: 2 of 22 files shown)


Dockerfile
```diff
@@ -5,5 +5,5 @@ RUN pip install --no-cache-dir --upgrade -r /code/requirements/docker.txt
 RUN pip install -e .
 RUN pip install opentelemetry-distro==0.38b0 opentelemetry-exporter-otlp-proto-grpc==1.17.0
 
-CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000"]
+CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "--reload"]
```
samredai, line 8: The reason I removed this `--reload` is because it interferes with the opentelemetry auto instrumentation. The server still start up fine but in dj-demo none of the traces will make it to the jaeger container. That being said, it is definitely convenient to have the reload on while developing. Maybe we could do something like this:
```docker
ARG RELOAD="--reload"
ENV RELOAD ${RELOAD}
CMD ["opentelemetry-instrument", "uvicorn", "dj.api.main:app", "--host", "0.0.0.0", "--port", "8000", "${RELOAD}"]
```
Then in the [docker-compose.yml](https://github.com/DataJunction/dj-demo/blob/main/docker-compose.yml#L26) file for dj-demo we can do:
```yaml
services:
  dj:
    ...
    build: 
        context: "${CORE_SERVICE}"
        args:
            RELOAD: ""
 ...
```
Definitely not urgent since the tracing is just a "nice to have" part of the demo. Feel free to disregard this as a non-blocker and we can worry about solving this later.

alembic/versions/2023_04_13_2048-4f6449741463_initial_migration.py
```diff
@@ -81,6 +81,15 @@ def upgrade():
         sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column("version", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column("uri", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column(
+            "dialect",
+            sa.Enum(
+                "SPARK",
+                "TRINO",
```
samredai, line 88: I also saw `POSTGRES` defined in the dialect enum. Does it also need to be included here?

dj/models/engine.py
```diff
@@ -1,14 +1,26 @@
...
 
 
+class Dialect(str, enum.Enum):
+    """
+    SQL dialect
+    """
+
+    SPARK = "spark"
+    TRINO = "trino"
+    POSTGRES = "postgres"
```
samredai, line 21: Does this translate to the lowercase or uppercase string when saved to the backend? I noticed the enum defined in the alembic migration uses uppercase so I'm wondering if these need to be `TRINO = "TRINO"` for example.

dj/sql/parsing/types.py
```diff
@@ -107,6 +107,37 @@ def __hash__(self):
         """
         return hash(str(self))
 
+    def is_compatible(self, other: "ColumnType") -> bool:
+        """
+        Returns whether the two types are compatible with each other by
+        checking their ancestors.
+        """
+
+        def has_common_ancestor(type1, type2) -> bool:
```
samredai, line 116: Very cool!

tests/construction/dj_query_test.py
```diff
@@ -74,17 +74,12 @@ async def test_build_dj_metric_query(request):
...
     LEFT OUTER JOIN (
       SELECT
-        basic_DOT_source_DOT_users.age,
         basic_DOT_source_DOT_users.country,
-        basic_DOT_source_DOT_users.full_name,
-        basic_DOT_source_DOT_users.gender,
-        basic_DOT_source_DOT_users.id,
-        basic_DOT_source_DOT_users.preferred_language,
-        basic_DOT_source_DOT_users.secret_number
+        basic_DOT_source_DOT_users.id
```
samredai, line 78: A nice side-effect here is cleaner generated SQL 😄 (besides making the query engine's job easier, of course)

samredai (approved):
This is awesome @shangyian, a few comments and questions but no blockers.

alembic/versions/2023_04_13_2048-4f6449741463_initial_migration.py
shangyian: Hmm, I'm thinking we should remove postgres, actually, since I'm not sure the way we're generating SQL now will actually be compatible with postgres. Let me do that

dj/models/engine.py
shangyian: So I think what happens here is that the database engine translates everything to lowercase (we have the same pattern for NodeType, where the alembic migration uses uppercase but the enum has lowercase). Although perhaps we should just be consistent in the migration and have these be lowercase to make it clearer. Let me regenerate

Dockerfile
shangyian: Good point! I'll make this change

tests/construction/dj_query_test.py
shangyian: Haha, yes I can actually read the queries now 😂

PR #462, opened 2023-04-16
function registry more use of dispatch/correct typings

CircArgs:
### Summary

WIP prelim review

Just adding more use of dispatch to get the type checks in the function registry + correcting typings 

I'm using these as references:

Spark:
https://github.com/apache/spark/tree/74cddcfda3ac4779de80696cdae2ba64d53fc635/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions

https://spark.apache.org/docs/2.3.0/api/sql/index.html

As I've come upon functions that are not in that last link, I am removing them for now 

Java strictmath reference
https://docs.oracle.com/javase/8/docs/api/java/lang/StrictMath.html

Databricks reference:
https://docs.databricks.com/sql/language-manual/sql-ref-functions-builtin-alpha.html

I've also ~~gotten rid of~~ _aliased_ Long in favor of bigint and added smallint

### Test Plan

unit and doctests

- [ ] `make check` passes

### DIFF
--- a/dj/sql/functions.py
+++ b/dj/sql/functions.py
@@ -12,6 +12,9 @@ https://github.com/apache/spark/tree/74cddcfda3ac4779de80696cdae2ba64d53fc635/sq
 
 Java strictmath reference
 https://docs.oracle.com/javase/8/docs/api/java/lang/StrictMath.html
+
+Databricks reference:
+https://docs.databricks.com/sql/language-manual/sql-ref-functions-builtin-alpha.html
 """
 
 import inspect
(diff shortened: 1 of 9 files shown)


shangyian:
will this still parse long successfully if someone casts the type of a field to `long`?

CircArgs:
> will this still parse long successfully if someone casts the type of a field to `long`?

@shangyian I think it would have worked, but I've added `LongType` back and it's more of a genuine alias now rather than a distinct type. This seemed like a nice little "hack" to get them to play nicely the way the code is designed with singletons and such:
https://github.com/DataJunction/dj/blob/620938c36c286058e9376590e86b0a9b78819dcd/dj/sql/parsing/types.py#L575-L595

```python
In [1]: import dj.sql.parsing.types as ct

In [2]: ct.LongType()==ct.BigIntType()
Out[2]: True

In [3]: str(ct.LongType())
Out[3]: 'long'

In [4]: ct.LongType()
Out[4]: LongType()
```

samredai:
@CircArgs @shangyian do we have a longer term strategy for different query engines that have different type systems? We've been running the generated queries against Trino without any issues even though our parser is based on Spark's grammar. Not sure if it will all still work (maybe it will?) for SQL written that will eventually run against snowflake or clickhouse.

I'm wondering how much flexibility we ultimately have here since we really only need to _represent_ these types and aren't doing any validation or processing which we're leaving up to the respective query engine. Is it enough flexibility to have a single large set of type representations that's a superset of types across all engines? Is that also true for the parser, i.e. a permissive grammar that parses elements from a wide set of SQL dialects? (the latter one seems deceptively difficult, but I'm not exactly sure)

CircArgs:
@samredai is that not what the aliasing will do like in that long type vs bigint case?

samredai:
@CircArgs I guess that's what I'm asking. To use the example mentioned by @shangyian, if someone writes `cast(1 as long)`, that works in spark but fails in trino. My understanding is that this is ok since we'll parse `long` as `LongType` and when we eventually render the generated query it will render to the identical `cast(1 as long)` that will be sent to spark. So it sounds like as we come across new types in the wild as this is used with more query engines, we can just keep adding `ColumnType` implementations and extending the grammar and visitors.

It's then implied that even though someone could include `cast(1 as long)` in the query for a DJ node and it will parse properly and be used in generated queries, it's on the user to understand that their query engine is trino and so any generated query using that node will raise a `TrinoException - Unknown type: long`. Of course any implementation could use some client side validation to help the user out. Does this all sound aligned with the direction we're going? I'm asking because I want to capture this somewhere in the docs as I expect people will have questions about this.

shangyian:
> We've been running the generated queries against Trino without any issues

Technically not true, because if someone writes `LATERAL VIEW ... EXPLODE ...` it won't work with Trino and we don't currently do any conversion from that to `CROSS JOIN ... UNNEST ...`. It might not be hard to do this conversion, but I'm also a bit iffy about diving down the path of supporting conversion between SQL dialects when we have a lot of other features we need first 😄

samredai:
@shangyian that all makes sense to me and I agree we shouldn't pre-empt the feature with docs on it.

> I also suspect that most users are going to be using transform nodes that they create, which will remove the painful problem of node A is written for Spark and node B depends on node A but is written for Trino, since users are likely to pick a single dialect for their set of nodes.

> Technically not true, because if someone writes `LATERAL VIEW ... EXPLODE ...` it won't work with Trino and we don't currently do any conversion from that to `CROSS JOIN ... UNNEST ...`.

The above two points you made are good ones and seem very related. I think user feedback here will be useful. As you said, I also suspect that users will be mostly using their own transform nodes but even in situations where they use transform nodes that don't belong to them, it's most likely one created by someone "close" in the org chart, i.e. a teammate. It's reasonable to believe that teams will be aligned on the dialect they use. So if their team has decided to use trino, I'd expect they won't use `LATERAL VIEW ... EXPLODE ...`.

(@CircArgs sorry for hijacking this PR for this discussion)

samredai (approved):

shangyian (approved):

PR #465, opened 2023-04-17
Several namespaces changes  + Dockerfile reload

shangyian:
### Summary

A few changes for namespaces:
* Adding the ability to list nodes for a given namespace
* Fixed a bug where source nodes weren't being assigned namespaces
* Moved namespaces out to their own module (there appears to be enough functionality here)
* Added client support for listing nodes in a namespace.

Added reload as an arg to the Dockerfile in case dj-demo wants to use this argument.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/tests/api/nodes_test.py
+++ b/tests/api/nodes_test.py
@@ -18,7 +18,7 @@ def test_list_all_namespaces(client_with_examples: TestClient) -> None:
     """
     Test ``GET /namespaces/all/``.
     """
-    response = client_with_examples.get("/namespaces/all")
+    response = client_with_examples.get("/namespaces/")
     assert response.ok
     assert response.json() == [
         {"namespace": "default"},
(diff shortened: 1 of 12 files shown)


tests/api/nodes_test.py
```diff
@@ -35,6 +35,36 @@ def test_list_all_namespaces(client_with_examples: TestClient) -> None:
     ]
 
 
+def test_list_nodes_by_namespace(client_with_examples: TestClient) -> None:
+    """
+    Test ``GET /namespaces/all/``.
```
samredai, line 40:
```suggestion
    Test ``GET /namespaces/{namespace}/``.
```
Also these should be moved to a `namespaces_test.py` file since the logic has been moved there.

samredai (approved):

samredai (approved):
LGTM just had one small suggestion for the tests.

tests/api/nodes_test.py
shangyian: Makes sense, done!

PR #466, opened 2023-04-18
Update client to use the factory pattern

shangyian:
### Summary

This change modifies the client to use the factory pattern, which makes it clearer when we're creating objects vs retrieving objects. Also added an optional `requests_session` to the client initializer.

Creating nodes now takes the form of:
```py
from djclient import DJClient, NodeMode

dj = DJClient("http://localhost:8000/")
source = dj.new_source(
    name="repair_orders",
    description="Repair orders",
    catalog="default",
    schema_="roads",
    table="repair_orders",
)
source.save(NodeMode.PUBLISHED)
```

A node can be rehydrated with data from the server using:
```py
node.sync()
```

This also lets us remove the `REGISTRY` global object.

### Test Plan

Tested things end-to-end against a DJ deploy

- [X] `make check` passes

### Deployment Plan

We should be able to release a version of the client once this gets in.

### DIFF
--- a/client/python/djclient/dj.py
+++ b/client/python/djclient/dj.py
@@ -400,8 +400,8 @@ class DJClient:  # pylint: disable=too-many-public-methods
             "cube": Cube,
         }
         return [
-            node_type_to_classes[node["type"]].parse_obj(  # type: ignore
-                {**node, **{"dj_client": self}},
+            node_type_to_classes[node["type"]](  # type: ignore
+                **{**node, **{"dj_client": self}},
             )
             for node in nodes_list
             if not type_ or node["type"] == type_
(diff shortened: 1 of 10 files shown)


client/python/djclient/dj.py
```diff
@@ -63,24 +64,174 @@ def construct_url(self, url):
...
+    def verify_node_exists(self, node_name: str, type_: str) -> Dict[str, Any]:
+        """
+        Retrieves a node and verifies that it exists and has the expected node type.
+        """
+        node = self.get_node(node_name)
+        if "name" not in node:
+            raise DJClientException(f"No node with name {node_name} exists!")
+        if "name" in node and node["type"] != type_:
+            raise DJClientException(
+                f"A node with name {node_name} exists, but it is not a {type_} node!",
```
samredai, line 232: nit: no shouting at the users 😄 
image

samredai (approved):
Looks great, very clean and intuitive. The client is coming along great! 🎉

client/python/djclient/dj.py
shangyian: Hahaha oops... that should be an April Fools release, where all error messages are in all-caps.

PR #467, opened 2023-04-19
Cubes + Materialization + Queries for Multiple Metrics

shangyian:
### Summary

**Cubes + Dimension Attributes**

Cubes should operate on dimension attributes, not nodes. This PR changes `cube_elements` to be of type `Column` rather than `NodeRevision`, which lets it reference either dimension columns or metric columns (of which there's only one per metric node). It also changes the cube creation endpoint to take `metrics`, `dimensions`, and `filters`. 

**Queries for Multiple Metrics**

During cube node creation, we also build and save the query that gets used to assemble the cube (see the new function `build_metric_nodes`), as well as the columns associated with this query, with the dimension attributes propagated. This query can then be used to materialize the cube to some external data warehouse.

Note that this functionality (specifically `build_metric_nodes`) can also be reused for generating queries for multiple metrics. 

**Additional Materialization Config Fields**

This adds an additional `schedule` field (cron) to `MaterializationConfig` and changes `config` from a string to a JSON dict. (Unclear why it was a string before, likely a bug). 

Engine-specific config can be provided in the `config` field, and there is no limitation as to what goes in there. For example, one possible `MaterializationConfig` that supports a Druid materializer could look like this:
```python
MaterializationConfig(
  engine="druid",
  schedule="0 * * * *",
  config={
    "timestamp_column": "some_timestamp_col",  # used for Druid ingestion, which requires a timestamp column
    "intervals": ["2020-01-01/2021-01-01"]  # used for Druid ingestion, which requires an interval for ingestion
  }
)
```

### Test Plan

- [X] PR has an associated issue: #468
- [X] `make check` passes


dj/api/nodes.py
```diff
@@ -421,57 +424,113 @@ def create_node_revision(
...
+    for node_name in data.metrics:
+        metric_node = get_node_by_name(session=session, name=node_name)
+        if metric_node.type != NodeType.METRIC:
             raise DJException(
                 message=(
-                    f"Node {cube_element.name} of type {cube_element.type} "
-                    "cannot be added to a cube"
+                    f"Node {metric_node.name} of type {metric_node.type} "
+                    f"cannot be added to a cube."
+                    + " Did you mean to add a dimension attribute?"
```
samredai, line 448: Nice!

dj/models/node.py
```diff
@@ -461,7 +465,7 @@ def extra_validation(self) -> None:
         """
         Extra validation for node data.
         """
-        if self.type in (NodeType.SOURCE, NodeType.CUBE):
+        if self.type in (NodeType.SOURCE,):
```
samredai, line 468: I'm curious why cubes are removed from this check. They shouldn't have a query attribute right?

samredai (approved):
This is incredible. And thanks for cleaning up the docs! Approving but I left a question about one part.

dj/models/node.py
shangyian: So in this case I've added the generated query for a cube to `query`, which triggers this extra_validation check to fail. But it shouldn't be in the user input, so this check might need to be on the creation input model.

samredai: I see, makes sense!

Issue #470, opened 2023-04-21
Cubes --> Decomposing Metrics


shangyian:
In some sense, cubes in DJ serve as a caching definition -- that is, someone defines a cube and we'll ingest the underlying cube data into Druid/Starrocks/some other OLAP database in order to support low-latency metric queries across all defined dimensions. 

However, this ingestion is not straightforward, as we can't always dump a metric with an arbitrary aggregation into such a database directly. Instead, the metric expression needs to be broken down into its constituent measures, each of which needs to be ingested into separate columns and then combined when the user requests metric data, in the post-aggregation operation stage.

A few examples:

<table>
<tr><td>Metric Expression</td><td>Measures</td><td>Post-Agg Operation</td></tr>
<tr>
<td>

`SUM(x) AS x`
</td>
<td>

```
[
  {
    "name": "x", 
    "agg": "sum", 
    "expr": "x"
  }
]
```
</td>

<td>N/A</td>
</tr>

<tr><td>

`AVG(x) AS x`
</td>
<td>

```
[
  {
    "name": "x_sum", 
    "agg": "sum", 
    "expr": "x"
  },
  {
    "name": "x_count", 
    "agg": "count", 
    "expr": "x"
  }
]
```
</td>
<td>

`x_sum / x_count`
</td></tr>

<tr><td>

```
CAST(
  SUM(
    COALESCE(is_ready, 0.0)
  ) AS DOUBLE
) / COUNT(*)
```
</td>
<td>

```
[
  {
    "agg": "sum",
    "expr": "CAST(COALESCE(is_ready, 0.0) AS DOUBLE)",
    "name": "is_ready_ratio_numerator"
  },
  {
    "agg": "count",
    "expr": "*",
    "name": "is_ready_ratio_denominator"
  }
]
```
</td>
<td>

`is_ready_ratio_numerator / is_ready_ratio_denominator`
</td></tr>
<tr><td>

`SUM(a) + SUM(b)`
</td>
<td>

```
[
  {
    agg: SUM,
    expr: a
    name: metric_left
  },
  {
    agg: SUM,
    expr: b
    name: metric_right
  }
]
```
</td>
<td>

`metric_left + metric_right`
</td></tr>
</table>

samredai:
Being able to decompose the expression into measures makes sense and is probably a useful feature generally (even for display purposes). How does this data structure of a list of maps work out for deeply nested expressions? i.e.
```sql
SELECT (
  CAST(
    SUM(
[...]

PR #471, opened 2023-04-22
Decompose Cube Metrics

shangyian:
### Summary

When we attach a materialization config to a cube node, the implication is that we want to ingest the underlying data into an analytics database. However, many such databases are only configured to work with simple aggregations, so each metric expression needs to be broken down into constituent simple aggregation measures. These needs to be ingested into separate columns and then combined when the user requests metric data, in the post-aggregation operation stage. See https://github.com/DataJunction/dj/issues/470 for a more detailed description.

This PR adds the decomposed query and measures list to the materialization config. Note that the query in the materialization config is different from the one stored on the cube node, as the materialization config's query is meant for building a temporary table in preparation for ingestion into an analytics database like Druid. The query on the cube node is meant for direct querying of the cube without materialization to an OLAP database (i.e., if no materialization config is specified but the user wants to query the cube).

### Test Plan

Some unit tests + end-to-end verification

- [X] PR has an associated issue: #470
- [X] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -311,14 +311,20 @@ def cube_materialization_config(cube_node: NodeRevision, config: Dict):
         **config,
         **{
             "measures": {
-                metric.alias_or_name.name: [
-                    Measure(
-                        name=str(measure.alias_or_name),
-                        agg=str(measure.child.name),
-                        expr=str(measure.child),
-                    ).dict()
-                    for measure in measures
-                ]
+                metric.alias_or_name.name: sorted(
+                    [
+                        measure_obj.dict()
+                        for measure_obj in {
+                            Measure(
+                                name=str(measure.alias_or_name),
+                                agg=str(measure.child.name),
+                                expr=str(measure.child),
+                            )
+                            for measure in measures
+                        }
+                    ],
+                    key=lambda x: x["name"],
+                )
                 for metric, measures in metrics_to_measures.items()
             },
         },
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 """
 Node related APIs.
 """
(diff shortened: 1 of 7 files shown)


dj/api/nodes.py
```diff
@@ -272,6 +275,58 @@ def delete_a_node(name: str, *, session: Session = Depends(get_session)):
...
+    metrics from measures based on this query's output.
+
+    The query directly on the cube node is meant for direct querying of the cube
+    without materialization to an OLAP database.
+    """
+    combined_ast = parse(cube_node.query)
+    dimensions_set = {
+        dim.name for dim in cube_node.columns if dim.has_dimension_attribute()
+    }
+    metrics_to_measures = decompose_metrics(combined_ast, dimensions_set)
```
agorajek, line 299: I feel that if anyone is ever confused about the difference between measures and metrics, this line makes it as clear as it gets :)

tests/api/cubes_test.py
```diff
@@ -283,11 +300,31 @@ def test_cube_sql(client_with_examples: TestClient):
           dispatcher.company_name,
           municipality_dim.local_region
     """
+    print(results["query"])
```
agorajek, line 303: lost from testing?

agorajek (approved):
This is amazing... the only suggestion I would have is to show somewhere in the tests or docs the different approach this code would take for:
```
SUM(col_one * col_two) AS sum_x
SUM(col_one) * SUM(col_two) AS sum_y
```

tests/api/cubes_test.py
shangyian: good point, removing :)

shangyian:
@agorajek Makes sense! I added a test for the `SUM(col_one) * SUM(col_two) AS sum_y` case by creating a bogus `double_total_repair_cost` metric.

Issue #328, opened 2023-02-12
Trim responses to not return too much information


samredai:
For the GET "/nodes/" endpoint, we return a large payload of details for every node. I'm not sure how useful this would be in common usage of the API and it might be better to return less such as a simple List[str] of node names that can be plugged into the GET "/nodes/{name}/" endpoint. Alternatively, we could have the response be a dictionary with a key for each node type and the value a list of names of nodes of that type.

We could always maintain the current route behavior in a GET "/nodes/details/" route or something similar.

agorajek:
+1 for the no-argument option to just return a list of node names, wo details. 

Is there a REST api standard to pass a list of node names and get the them back with details, as a generalization of a single node query?

shangyian:
Realistically, given enough nodes, wouldn't we need to set up pagination? I also think we should add some filters to the GET /nodes/ endpoint -- for example, let users provide a query arg `type` to filter down to nodes of a specific type.

shangyian:
Yeah, that's the idea. Although on second thought, the GraphQL API pretty much does exactly this. In fact, given that we will have a GraphQL API, maybe we shouldn't worry too much about large payloads for the REST API.

samredai:
Coming back to this thread as I've recently been running some performance tests. A relatively small DJ cluster (3 containers, 8cpus, 16g) performs well with 100 simulated users hammering the service at ~50 requests per second. But the biggest impact to the average response time are the endpoints that return long lists of nodes with all of the related metadata, nested all the way into the node revision and the node revision's details. As the number of nodes gets into the thousands, there's significant spikes in latency when all users bum rush the `/nodes/` or `/namespaces/{namespace}/` endpoints. In the chart below you can see that the first user startup requests have very low latency (creating a new namespace, creating 26 nodes, 10 dimension links -> median of ~600ms, average of ~800ms).

But once all 100 users start hitting the listing endpoints, each request returning full details on 2600 nodes (100x26), you can see the significant latency spikes (yellow) even though the total requests per second is dramatically lower (green).

<img width="1703" alt="Screenshot 2023-04-22 at 5 16 23 PM" src="https://user-images.githubusercontent.com/43911210/233806938-05b0bbb7-11c7-400a-b729-1eba9892f2bf.png">

[...]

PR #472, opened 2023-04-24
Improve DJ client tests + remove top-level nodes access

shangyian:
#### DJ client tests

This change improves client tests by using the starlette `TestClient` setup (similar to what we have in the DJ server) to spin up a requests session with actual API, rather than mocking inputs and outputs. This makes it more robust to actually catching errors when changes are made to the DJ server and corresponding changes should be in the client lib. This change also adds a stage to run client tests.

#### Namespace-only nodes access

Users cannot use `client.nodes()` or `client.sources/transforms/metrics/...()` to access all nodes on a DJ server. Instead they can use `client.namespace("some.namespace").nodes()`. See https://github.com/DataJunction/dj/issues/328#issuecomment-1518830614 for more context.

### Test Plan

- [X] PR has an associated issue: #328
- [x] `make check` passes

### Deployment Plan

N/A


samredai (approved):
Overall this looks good but I had one question. I see that we still keep the `names_only: bool = False` argument to the methods that pull nodes from namespaces. Are we making incremental steps towards making that the default behavior of the server with the caveat that we also include type and description? I'm worried that namespaces can grow very large and lead to the same problem as the top level `/nodes/` endpoint.

shangyian:
@samredai Good point, let me make those changes in this PR

PR #473, opened 2023-04-24
bugfix: Node validation check should take into account type inference

shangyian:
### Summary

When a user tries to create a published node that fails type inference, we should raise an exception rather than allowing the creation of a published node with `status=invalid`. I noticed this bug because in the DJ roads example modelling, `avg_length_of_employment` gets created as a published metric with invalid status (due to a typing mismatch between `NOW()` and `hire_date`, which this also fixes).

### Test Plan

Added unit tests

- [x] `make check` passes

### DIFF
--- a/dj/errors.py
+++ b/dj/errors.py
@@ -34,7 +34,8 @@ class ErrorCode(int, Enum):
 
     # SQL Build Error
     COMPOUND_BUILD_EXCEPTION = 300
-    MISSING_PARENT = 201
+    MISSING_PARENT = 301
+    TYPE_INFERENCE = 302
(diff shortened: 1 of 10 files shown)


dj/errors.py
```diff
@@ -34,7 +34,8 @@ class ErrorCode(int, Enum):
 
     # SQL Build Error
     COMPOUND_BUILD_EXCEPTION = 300
-    MISSING_PARENT = 201
```
shangyian, line 37: This was the same error code as something above

samredai (approved):

PR #474, opened 2023-04-25
Align dependency management + workflows for publishing and releasing

shangyian:
#### Align Dependency Management

We currently have two ways of managing dependencies for the DJ server: `poetry` (used to publish the server and client libraries to PyPi) and `pip-compile-multi` (used to generate the requirements used by tests and Docker). The DJ client only uses poetry.

Changes include:
* Updating the DJ server's poetry dependencies (defined in `pyproject.toml`, locked to `poetry.lock`) to actually match what's used by `pip-compile-multi`, allowing us to setup and run both local and CI tests with `poetry install` followed by `poetry run pytest ...`.
* Updating the DJ client's test dependencies so that we can actually run tests in an independent environment from the DJ server. Originally these client tests were just piggy-backing directly off the server's dependencies.

tl;dr is that everything works with Poetry now, so if we wanted to, we can use `poetry export --without-hashes --format=requirements.txt > docker.txt` to export requirements for the Docker image and then remove everything for `pip-compile-multi` and move fully to Poetry.

#### Workflow Updates

These changes include:
* Separated out the client and server tests. There's no reason for them to be run sequentially, especially when the server tests take significantly longer than the client tests.
* Added a release github action, which can be run by specifying the version bump rule and the library to bump (either the server or client). This will then bump the version with poetry, create a tag, create a release in github, and finally publish to PyPi.

### Test Plan

- [X] `make check` passes

### Deployment Plan

N/A


samredai (approved):
This is great, thanks for cleaning this up @shangyian. Looks like there are a couple of repo configs we'll need to update for this.
- Setting the PR merge requirements to use the `build` definition in `test.yml` (instead of looking for `python-package.yml`)
- Add a PYPI token to the repo secrets (whoever will do this can ping me their pypi name and I can set them as the owner for the project in PYPI)

@betodealmeida or @agorajek, when you get a chance can you make these updates?

requirements/test.
agorajek, line 1: Is this a real file or a typo?

agorajek (approved):
LGTM, one comment inline.

requirements/test.
shangyian: haha, definitely a typo! I removed it

shangyian: @agorajek or @betodealmeida In order to merge this we actually need to change the merge requirement status check to use the results of the new matrix (that is, `python-version` + `library`, rather than just `python-version`).

cwegener (commented):
-

samredai:
@cwegener would pdm's scripting ability be a viable replacement for make?

shangyian:
@cwegener Thanks for having a look! These are all great points.

> pytest invocation
Agreed that they should be integrated into a single flow and not copied across these different places. I'm not a huge fan of `tox`, but I'll explore what it might look like with `pdm`. 

> Developer docs
I'll add some docs on this in another PR. I think putting them into the Hugo docs under the developer section makes sense.

> Just confirming that the binary wheel that gets built is platform agnostic?
Yeah, it should be

cwegener:
> @cwegener would pdm's scripting ability be a viable replacement for make?

The two things are very different (`make` and `pdm` user scripts) and not really interchangeable.

One thing that I can say is that in my experience there is a lack of knowledge in maintaining Makefiles in the real world. Not all engineers will be comfortable editing Makefiles, but that should not bar anyone ever from utilising Makefiles in their projects. :grin:


The current Makefile in dj is really very simple and can easily be folded into the existing new python specific toolchain of `pdm` + `hatch`, since the targets in the Makefile are quite simple task runners.

On the other hand though, I don't really see any urgent need to replace the Makefile. Keeping the Makefile doesn't add any new cost to the maintenance of the project (ignoring my comment about lack of experience in maintaining Makefiles)

I think the only thing we should do is to migrate  the developer related target definitions (`test`, `integration`, `check`, `spellcheck`) into pdm user scripts and change the Makefile to invoke those pdm user scripts instead.

And furthermore the release management related targets (`version`, `release`) are now probably 100% dependent onto GH Actions anyway and they should either be removed (therefore leaving GH actions as the only way to make a release) or the GH Actions should call the release management targets in the Makefile, thereby leaving a bit of flexibility for the case should the GH actions ever cause a problem during release management. So should the need for a fallback option ever arise one could craft a release "manually" (or in another CI runner) by simply executing the same Makefile targets without relying on GH actions. (presuming that PyPI upload privileges are available)

PR #476, opened 2023-04-27
Creating namespaces from the client and namespace docs

samredai:
### Summary

This PR does a couple of things in the client and docs.

- Add a `new_namespace` method to create namespaces from the datajunction client
- Add a docs page that explains namespaces and how to create them
- Sets all of the empty docs pages as drafts so they don't show up in the build
- A few various small fixes to docs weights and content

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A


docs/content/0.1.0/docs/getting-started/namespaces.md
```diff
@@ -0,0 +1,41 @@
...
+| Node Name           | Namespace    |
+|---------------------|--------------|
+| roads.demo.repairs  | roads.demo   |
+| finance.revenue     | finance      |
+| hr.people.employees | hr.people    |
+| customer            | default      |
+
+Since namespaces are inferred directly from the node name, creating a node in a particular namespace simply requires prefixing
+the node name with the namespace. In other words, creating a node named `roads.demo.repairs` will automatically create the node in the
+`roads.demo` namespace.
```
shangyian, line 20: Should we tell users here that they can't create nodes in namespaces that don't exist yet?

docs/content/0.1.0/docs/getting-started/namespaces.md
```diff
@@ -0,0 +1,41 @@
...
+All nodes in DataJunction exist within a namespace. Node names are dot separated alpha-numeric elements. The leading elements
+identify the namespace where the node exists. Nodes that do not include any dots in the name are automatically
+defined in the `default` namespace.
+
+| Node Name           | Namespace    |
+|---------------------|--------------|
+| roads.demo.repairs  | roads.demo   |
+| finance.revenue     | finance      |
+| hr.people.employees | hr.people    |
+| customer            | default      |
```
shangyian, line 16: So the `default` case is a special one where the namespace won't be included in the node name? Do you think it's worth being consistent and trying to include it in the name or not?

shangyian (approved):

docs/content/0.1.0/docs/getting-started/namespaces.md
samredai: We could and probably should, but I like that it's easier to go from 0 to 1 for a new user before they learn about namespaces. Where I think we're going is that there could be an equivalent to `USE default` or `USE roads.demo`.

PR #477, opened 2023-04-27
Fix DJ client release 

shangyian:
### Summary

The DJ client release workflow was broken because we didn't have hatch set to use dynamic versions. This change adds that support to the client. It also changes the tagging scheme to be `<new_version>-server` or `<new_version>-client`, making it clear which library we're releasing via the git tag.

It also removes the extraneous workflow `test-client.yml`, as this is already covered by one of the matrix parameters under `test.yml`.

@cwegener it looks like I can't request you as a reviewer, but if you get a chance 😄 

### Test Plan

- [x] `make check` passes

### Deployment Plan

N/A


cwegener (approved):
As mentioned in Slack. Looks good. This PR will indeed fix the client release problem.

samredai (approved):

agorajek (approved):
@shangyian as usual... thank you for figuring this out.

PR #478, opened 2023-04-28
Set /metrics/ return type to List[str] and add it to the client

samredai:
### Summary

This PR makes a couple of small changes across the client, server, and docs site.
- Set the `/metrics/` endpoint to return only metric names (`List[str]`)
- Add a `dj.metrics()` method to the client that calls `/metrics/`
- Fill in the "Listing Metrics" docs page
- Adds an intro illustration to the README and docs main page

### Test Plan

`make check` && `make test` as well as install the client and trying it against a local DJ server.

- [ ] `make check` passes

### Deployment Plan

N/A


shangyian (approved):
Goodbye latency 🤞

docs/static/datajunction-illustration.png
agorajek, line 1: Awesome pic. What do you think about changing the 3 charts on the right hand side to:
- one graph / BI tool
- one Python Jupyter notebook
- one ChatGPT window

To show that there are many ways to consume the DJ metadata?

client/python/datajunction/client.py
```diff
@@ -304,6 +304,12 @@ def engines(self):
         response = self._session.get("/engines/", timeout=self._timeout)
         return response.json()
 
+    def metrics(self):
```
agorajek, line 307: Nice and scary. Wonder when we hit some perf issues when clients start calling this few times a second. Search APIs can surely come later...

agorajek (approved):
Awesome. Just one suggestions on the pic.

docs/static/datajunction-illustration.png
samredai: Woah, love this idea. Going to update it now!

client/python/datajunction/client.py
samredai: From what I can see, this largely depends on the backend database you're using. That being said, I have wondered if we can add some very light in-memory caching of this to at least cache intrasecond requests, but we have to think about the read-your-write behavior there and not make it confusing. I'll run some more performance tests soon and will think about this a bit more.

PR #483, opened 2023-05-01
refactor: make uvicorn+otel optional for server

cwegener:
Choosing the ASGI server is something that users may do themselves. The
uvicorn AGSI server should be an extras requirement in the PEP-621
dependencies metadata.

Uvicorn can still be installed at the same time as datajunction-server
via: `pip install 'datajunction-server[uvicorn]'`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "celery<6.0.0,>=5.2.7",
     "fastapi<0.80.0,>=0.79.0",
     "msgpack<2.0.0,>=1.0.5",
+    "opentelemetry-instrumentation-fastapi==0.38b0",
     "python-dotenv<1.0.0,>=0.19.0",
     "redis<5.0.0,>=4.5.4",
     "requests<3.0.0,>=2.28.2",
@@ -46,7 +47,6 @@ classifiers = [
 
 [project.optional-dependencies]
 uvicorn = [
-    "opentelemetry-instrumentation-fastapi==0.38b0",
     "uvicorn[standard]>=0.21.1",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,6 @@ dependencies = [
     "celery<6.0.0,>=5.2.7",
     "fastapi<0.80.0,>=0.79.0",
     "msgpack<2.0.0,>=1.0.5",
-    "opentelemetry-instrumentation-fastapi==0.38b0",
     "python-dotenv<1.0.0,>=0.19.0",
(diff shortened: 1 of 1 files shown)


cwegener:
**NOTE**: I have not looked at modifying the `Dockerfile` to take advantage of this optional dependency. The `Dockerfile` is currently already doing a `pip install` from the `docker.txt` pdm-exported lockfile, so anyone using that `Dockerfile` at the moment should not be impacted until we actually set up some hooks to auto-generate the pdm lockfile exports into the `requirements` folder.

shangyian (commented):
Hmm, I'm not sure that `opentelemetry-instrumentation-fastapi ` should be removed, since it's specific to fastapi but not uvicorn. I think it's also what's causing the pylint failure in the linter stage (not sure why python 3.10 passed though, some odd behavior with linters).

cwegener:
> Hmm, I'm not sure that `opentelemetry-instrumentation-fastapi ` should be removed, since it's specific to fastapi but not uvicorn. I think it's also what's causing the pylint failure in the linter stage (not sure why python 3.10 passed though, some odd behavior with linters).

Oh. You're right. I'll change that. I was under the impression sure that the OTel Python SDK was actually doing pure automatic instrumentation. I see that there indeed are hooks that wire up the OTel instrumentation. I will create a separate PR for the OTel integration (I think the OTel integration should actually be 100% functioning out of the box with something like an OTel console exporter and maybe some additional libraries can be auto-instrumented as well, e.g. SQLAlchemy)

shangyian (approved):
Nice, LGTM!

samredai:
Sorry, late to the party here but just wanted to add that we could probably get away with making `opentelemetry-instrumentation-fastapi` if we really wanted to. What we ended up doing internally is importing the routers and defining our own app around them including our own otel instrumentation. The instrumentation that I included here in the source code was really just to support the jaeger container in the dj-demo app but maybe we could eventually just do some custom app instantiation logic there (w/ a custom Dockerfile as well) and switch the app logic here in the main repo back to a simpler one.

cwegener:
> The instrumentation that I included here in the source code was really just to support the jaeger container in the dj-demo app
That sounds great. I'll look at the dj-demo app in conjunction with the OTel Auto-instrumentation.
> but maybe we could eventually just do some custom app instantiation logic there (w/ a custom Dockerfile as well) and switch the app logic here in the main repo back to a simpler one.
That was my plan to just rely on the OTel Python SDK Agent functionality and use pure Auto-instrumentation. Without needing neither manual nor programmatic instrumentation in this repo.

PR #484, opened 2023-05-01
npm installable javascript DJ client

samredai:
### Summary

This adds an ES6 javascript DJ client that's npm installable. This PR only includes the get endpoints to pull entities but in a follow-up PR I'll add the ability to create entities as well.

**update:** I've restructured this a bit to use webpack with the type set as universal module definition (UMD) so we can use the single published npm package for both server side and client side javascript. (read more [here](https://webpack.js.org/guides/author-libraries/))

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A


client/javascript/src/httpclient.js
```diff
@@ -0,0 +1,79 @@
...
+      this._headers.Authorization = `Basic ${btoa(`${username}:${password}`)}`;
+      return this;
+    }
+  
+    setBearerAuth(token) {
+      this._headers.Authorization = `Bearer ${token}`;
+      return this;
+    }
+  
+    get(endpoint, options = {}) {
```
shangyian, line 40: Should we just have all of these be async as well? Then it's up to the user of the client to use `await` or indeed run it async if they want to.

shangyian (approved):
Other than the one comment, looks good to me! To get the build-working we'll probably need to set the linters to run only on Python code and then add JS linters to the build in a separate path.

client/javascript/src/httpclient.js
samredai: Good idea, updated!

samredai:
Thanks @shangyian, updated the python builds to ignore the `client/javascript` directory then added an eslint scripts and some workflow steps to build and lint the js client.

@agorajek or @betodealmeida not urgent but we should add the javascript jobs to be required checks for merging PRs. Also we have to setup a release job using an npm access token just like we did with pypi. I went ahead and published the `0.0.1` version for now: https://www.npmjs.com/package/datajunction

PR #487, opened 2023-05-03
fix: new 'uvicorn' group missing from pdm.lock

cwegener:
In commit 4b38435 I of course forgot to refresh the lock file so that it
passes a `pdm lock --check`.

This commit fixes that, AND it also adds the pre-commit hook to catch
this common mistake.

The command to refresh the lockfile is `pdm lock -G :all --refresh`


shangyian (approved):
Good catch!

PR #490, opened 2023-05-04
Python client changes, more docs, fix pdm

samredai:
### Summary

This PR makes a number of minor changes to the python client as well as docs.
- Removed pandas as a requirement for the client and catch+raise when it's not installed (only when the data methods are called). This lets those interested in simply using the CRUD component of the python client to not have to pull in a beefy dependency like pandas.
- Adds a few inspections of the response status and the "state" value to raise some clearer exceptions when the response or query fails.
- Sets `engine_name` and `engine_version` defaults to `None` so the server default is used. I also moved allowed these to be set at client instantiation so users can set it once and have it consistently used through the session. (Users can still override the client setting on individual method calls)
- Adds a client level method for retrieving multiple metrics and dimensions together, `dj.data(metrics=[...], dimensions=[...], filters=[...])`
- Fills in the content for the "Requesting Data" docs page and makes some corrections to the "Requesting SQL" docs page.

### Test Plan

pip installed the client and used it against a local DJ instance.

- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -11,7 +11,10 @@ try:
     import pandas as pd
 except ImportError:
     warnings.warn(
-        "Optional dependency `pandas` not found, data retrieval disabled",
+        (
+            "Optional dependency `pandas` not found, data retrieval"
+            "disabled. You can install pandas by running `pip install pandas`."
+        ),
         ImportWarning,
     )
 import requests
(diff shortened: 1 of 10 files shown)


client/python/datajunction/client.py
```diff
@@ -463,35 +475,89 @@ def sql(  # pylint: disable=too-many-arguments
...
+                "engine_name": engine_name or self.engine_name,
+                "engine_version": engine_version or self.engine_version,
             },
         )
         if response.status_code == 200:
             return response.json()["sql"]
         return response.json()
 
-    def data(  # pylint: disable=too-many-arguments
+    def data_for_single_node(  # pylint: disable=too-many-arguments
```
shangyian, line 486: I'm starting to think that we should actually limit the data retrieval in the client to dimensions, metrics, and cubes. We can perhaps offer a sampling endpoint for other nodes (and of course the API itself still exists). Because otherwise the data retrieved for the intermediate nodes could be very large and I'm not sure we want the burden of operating a generic querying service.

client/python/datajunction/client.py
```diff
@@ -1,11 +1,19 @@
...
+import warnings
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urljoin
 
-import pandas as pd
+try:
+    import pandas as pd
+except ImportError:
+    warnings.warn(
+        "Optional dependency `pandas` not found, data retrieval disabled",
```
shangyian, line 14: Should this include some directions on how to install pandas?

shangyian (approved):
LGTM! Left some comments but only one is specifically about this PR

client/python/datajunction/client.py
samredai: Fully agree, great point. so sounds like I should just remove this method entirely and have data access be through the `DJClient.data` method? I can make that method's args more flexible, for example allowing the `metrics` arg type to be `List[str | Metric]`.

client/python/datajunction/client.py
samredai: That's a good idea, let me update that 👍

samredai:
Huge thanks to this kind soul for sharing the solution as setting `PDM_DEPS='urllib3<2'`: https://github.com/pdm-project/pdm/issues/1879#issuecomment-1533484132

Issue #491, opened 2023-05-04
Type inference fails for struct select


shangyian:
We don't do type inference correctly for struct selections. 

For example, the following query, with `counts` being of type `struct<a string, b bigint>`, fails type inference:
```
SELECT SUM(counts.b) FROM source
```

An example unit test for the failure:

```
def test_struct_select_type_inference(client_with_examples: TestClient):
    """
    Testing struct select type inference
    """

    client_with_examples.post("/nodes/source/",
        json={
            "columns": [
                {"name": "id", "type": "int"},
                {"name": "painter", "type": "string"},
                {
                    "name": "counts",
                    "type": "struct<a string, b bigint>",
                },
            ],
            "description": "Murals",
            "mode": "published",
            "name": "basic.murals2",
            "catalog": "public",
            "schema_": "basic",
            "table": "murals2",
        },
    )

    response = client_with_examples.post("/nodes/metric/",
        json={
            "query": "SELECT SUM(counts.b) FROM basic.murals2",
            "description": "Mural Counts",
            "mode": "published",
            "name": "basic.mural_count",
        },
    )
```

One issue is that `sbp.DereferenceContext` cannot always be interpreted as a `<table>.<column>` and can sometimes be `<column>.<field>`.

CircArgs:
@shangyian taking a look at this. here is a dummy minimal example:

```python
query = parse("SELECT SUM(counts.b) FROM (select cast(_ as struct<a string, b bigint>) as counts) dummy")
```

compiling `query` we get the expected `DJError(code=<ErrorCode.INVALID_COLUMN: 206>, message='Column`counts.b` does not exist on any valid table.', debug=None, context='')` on the exception in the compilation context.

You mention an issue while visiting `sbp.DereferenceContext` but I wonder if we can address this in the ast by maybe in https://github.com/DataJunction/dj/blob/95a2c033074fbb6503d37c58e06068d9054d7e28/dj/sql/parsing/ast.py#L907 making the check more sophisticated when using the namespace columns. 

https://github.com/DataJunction/dj/blob/95a2c033074fbb6503d37c58e06068d9054d7e28/dj/sql/parsing/ast.py#L945-L953

shangyian:
@CircArgs good point, I think something like this should work:

```
                # For struct types we can additionally check if there's a column that matches
                # the search column's namespace and if there's a nested field that matches the
                # search column's name
                if isinstance(col.type, StructType):
[...]

PR #495, opened 2023-05-04
Struct type inference in select

shangyian:
### Summary

We weren't doing type inference correctly for struct selections -- for more details see https://github.com/DataJunction/dj/issues/491.

This fixes it by adding a check in `add_ref_column` to see if the search column is a struct, and if it is, tries to check if there's a column that matches the search column's joined namespace and a nested field that matches the search column's name. If such a column exists, we add the table to the search column and assign the nested field's type as the column's type.

### Test Plan

Unit tests + running tests of some local metric definitions

- [X] PR has an associated issue: #491
- [ ] `make check` passes

### DIFF
--- a/dj/sql/parsing/ast.py
+++ b/dj/sql/parsing/ast.py
@@ -711,8 +711,11 @@ class Column(Aliasable, Named, Expression):
     def is_compiled(self):
         return self._is_compiled or (self.table and self._type)
 
-    # flake8: noqa
-    def find_table_sources(self, ctx: CompileContext) -> List["TableExpression"]:
+    def find_table_sources(
+        self,
+        ctx: CompileContext,
+    ) -> List["TableExpression"]:
+        # flake8: noqa
         """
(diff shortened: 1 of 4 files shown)


samredai (approved):

dj/sql/parsing/ast.py
```diff
@@ -710,6 +711,7 @@ def use_alias_as_name(self) -> "Column":
     def is_compiled(self):
         return self._is_compiled or (self.table and self._type)
 
+    # flake8: noqa
```
samredai, line 714: This should be moved to the end of the next line to just noqa for this function. Otherwise it will noqa for all lines in the file after this line.

CircArgs (approved):

dj/sql/parsing/ast.py
shangyian: Done! (flake8 thinks it's too complex 😭 )

PR #497, opened 2023-05-05
Update OpenAPI Spec - 3564a69

agorajek:

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -7,7 +7,7 @@
             "name": "MIT License",
             "url": "https://mit-license.org/"
         },
-        "version": "0.0.post1.dev792+g76f08b5.d20230417"
+        "version": "unknown"
     },
     "paths": {
         "/catalogs/": {
@@ -287,7 +287,7 @@
                                     "title": "Response List Metrics Metrics  Get",
                                     "type": "array",
                                     "items": {
-                                        "$ref": "#/components/schemas/Metric"
+                                        "type": "string"
                                     }
                                 }
                             }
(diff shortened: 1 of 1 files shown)


samredai (approved):

openapi.json
```diff
@@ -7,7 +7,7 @@
             "name": "MIT License",
             "url": "https://mit-license.org/"
         },
-        "version": "0.0.post1.dev792+g76f08b5.d20230417"
+        "version": "unknown"
```
shangyian, line 10: Hmm why is the version now unknown?

samredai: It's most likely coming from [here](https://github.com/DataJunction/dj/blob/main/dj/__init__.py#L12). Could it be that some of the package manager changes that were recently made is making this logic hit the `PackageNotFoundError` now? I'll look into it--I think we ideally want the generated API spec to be mapped to the major version of the library.

samredai: Oh the comment there actually describes exactly what's happening, lol
```py
# Change here if project is renamed and does not equal the package name
```
This is probably when we changed the package name to `datajunction-server` since `dj` was taken in PyPI (which was grandfathered in I think since they don't allow two character package names anymore). It always feels worth it to ensure the import name always matches the pypi name. Any objections to me changing `from dj import ...` to `from datajunction_server import ...` across the code-base?

agorajek: No objection, although I think this was my wishful thinking. You gotta do what you gotta do.

Issue #488, opened 2023-05-03
Node created without namespace isn't assigned one


shangyian:
If a node is created without a namespace (i.e., the name is just `repair_orders`), we should assign `default` as the namespace and save that to the node name (i.e., the node's name should be `default.repair_orders`). This fits with the downstream logic, which assumes that the node's namespace is always part of the name.

samredai:
@shangyian to avoid confusion I think we should maintain this shorthand when retrieving the nodes as well, do you agree? In other words, if someone requests `repair_orders` then we should automatically find and return `default.repair_orders`. 

Something that I'm worried about though is the situation where someone writes some reasonable logic that requests a node and has follow-up logic that (very reasonably) includes an expectation that the node returned has the same name as the node requested.
```py
name = "repair_orders"
node = dj.node(name)
assert node.name == "repair_orders"  # raises an error since "repair_orders" != "default.repair_orders"
```
Where these thoughts are leading me is that maybe we should abandon all of this default namespace logic and just make a user-defined namespace an absolute requirement by the server. We already allow `DEFAULT_NAMESPACE` to be configurable in clients, maybe we should avoid any special treatment server side and leave it to the clients to make sure a namespace is always included in requests. Am I crazy?

shangyian:
@samredai I agree, let's just make the namespace an absolute requirement so that it's super clear what the API is doing.

CircArgs:
@shangyian so the implication is just raising if there is no namespace right?

shangyian:
@CircArgs yep, it just raises if there's no namespace so nothing is implicit.

PR #506, opened 2023-05-08
Replace automatic "default" namespace attribution with explicit namespacing

samredai:
### Summary

This PR makes namespaces required with no automatic namespace attribution when no namespace is provided. Based on the discussion in #488, explicit namespace requirements should lead to more consistent behavior that should be more intuitive to users and simplifies some of the server side logic.

The source code change is pretty minimal here (barely 3 lines) and the rest of the line changes here were made to tests (we've used the "automatic default namespace assignment" pretty aggressively in tests, part of the reason why the bug in #488 was able to sneak by unnoticed).

### Test Plan

`make check`, `make test`, and a few adhoc testing

- [x] PR has an associated issue: #488 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/client/python/tests/examples.py
+++ b/client/python/tests/examples.py
@@ -50,7 +50,7 @@ EXAMPLES = (  # type: ignore
             ],
             "description": "All repair orders",
             "mode": "published",
-            "name": "repair_orders",
+            "name": "default.repair_orders",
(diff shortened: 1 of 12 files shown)


agorajek (approved):
Explicit is good.

Issue #492, opened 2023-05-04
[bug] Node Deletion+Recreation Bugs


shangyian:
When we delete a node, it removes all parent links from its downstream children, which makes sense. However, there are several bugs:
* The downstream children are not invalid nodes, but while their modes are updated to `INVALID`, their statuses remain `PUBLISHED`. We should update their statuses to `DRAFT`. 
* We add a copy of the deleted node back, we expect the downstream children don't get their upstream links refreshed to point to the recreated node. Instead, they all stay invalid, which doesn't fit with the stated behavior in the docs. 

The second issue is not as easy to fix and we may want to consider soft deletes rather than hard deletes.

samredai:
For point 1, I've thought about this and if possible I hope we can avoid the system flipping between `draft` and `published` mode since that's really meant to display the intent of the user. We do have a requirement that a `draft` node must have a system controlled status of `valid` before they can flip it to `published` but as you pointed out, an already `published` node can be made `invalid` by changes in upstream dependencies.

My take on that situation is that the system's job there is just to highlight those cases. If we leave the mode as `published`, that makes it a more valuable signal to surface all nodes that are `published`+`invalid` as "someone else made a change that broke your definition". It could be an external change to a data warehouse table but if it's an internal change, we should aim to catch all of those destructive changes and require the user to provide some explicit "force" parameter. This is a diagram that I put together some time ago to capture this, the main point being that if we automatically flip broken nodes to draft, we won't be able to differentiate between draft nodes that are just a WIP or nodes that were automatically changed by the system. Curious to hear your thoughts.

<img width="711" alt="Screenshot 2023-05-04 at 3 42 27 PM" src="https://user-images.githubusercontent.com/43911210/236312513-8be97c0d-ea0b-4523-9c7b-9a1ea9c96f9f.png">

For point 2, someone recently described an expectation to me that if a node is deleted and then recreated, they would expect the recreated node to show the history before the deletion up to "deleted at <time>" and then "recreated at <time>". I think this is only possible if we commit to only doing soft deletes. As you said, that also solves the problem of links to downstream children. It's a tricky problem but my gut is leaning towards us not allowing a delete command but instead allowing a deactivate command which is the equivalent of a soft delete.

The next question that opens up is--what happens when a node has a deactivated node as a dependency? If when a node is deactivated, we find all of the effected nodes and mark them as `invalid`, and we also prevent the system from building SQL or requesting data for any `invalid` nodes, that might be enough.

[...]

PR #507, opened 2023-05-09
Soft-delete Nodes

samredai:
### Summary

This changes node deletion to a soft-delete by adding a nullable `Node.deactivated_at` attribute. Deletion involves simply setting a timestamp value. The implication for the rest of the codebase is that wherever nodes are pulled (usually through helper functions), the select has to include `.where(is_(Node.deactivated_at, None))`.

Soft-deletes do not have any effect on the dimension links that may have been set on the node columns which should solve #492; that being said, we have to be extra careful to not display deactivated nodes in various listing operations such as listing nodes, listing metrics, listing dimensions for metrics, listing metrics for a dimensions, etc.

An aside but for consistency, I think we should use the terms "deactivate" and "reactivate" in docs and code comments instead of "delete" or "soft-delete". I actually think it would be useful to show the deactivated nodes in some areas, for example in visual DAGs. Mainly because it will be useful when investigating why a child node has been invalidated.

### Test Plan

N/A

- [x] PR has an associated issue: #492
- [x] `make check` passes

### Deployment Plan

N/A


dj/api/nodes.py
```diff
@@ -259,19 +261,16 @@ def delete_a_node(name: str, *, session: Session = Depends(get_session)):
         downstream.current.status = NodeStatus.INVALID
         session.add(downstream)
 
-    # If the node is a dimension, find all columns that
```
shangyian, line 262: Hmm... don't we still need to do a version of this where we invalidate the downstream nodes? And then for the dimension links, we'd need to invalidate those as well I think. That might need to be stored at the attribute level, not sure.

dj/api/nodes.py
```diff
@@ -249,7 +251,7 @@ def get_a_node(name: str, *, session: Session = Depends(get_session)) -> NodeOut
 @router.delete("/nodes/{name}/", status_code=204)
 def delete_a_node(name: str, *, session: Session = Depends(get_session)):
```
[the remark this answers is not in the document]
samredai: Yep, I was just changing this. Updated the PR 👍

dj/api/nodes.py
samredai: I was thinking about how to handle this and I can't come up with anything that doesn't have some tradeoffs. If deactivating a dimension invalidates everything linked to it, that would actually have a larger blast radius than we need. Take the example of a valid metric node that has access to 5 dimension nodes and all of the upstream nodes are valid. Just because one of those dimension nodes are deactivated, I don't think we should mark the metric node as invalid. It's still largely valid in that you could use it with any of the 4 dimension nodes left and build a cube with other metrics that share one or more dimensions.

On the other hand, we don't want to show a deactivated dimension node anywhere since they shouldn't be allowed for use in building a SQL query or retrieving data. What if we left the nodes using a deactivated dimension as valid but made sure to not display that dimension node in any listings and also made sure to not allow using it to get SQL/data?

samredai:
@shangyian something I've noticed is that when running the client tests locally (and maybe during CI?), it doesn't seem to pick up local changes to the server. Is that something I should expect with the `"dj @ {root:uri}/../.."` defined in pyproject.toml?

dj/api/nodes.py
shangyian: > Just because one of those dimension nodes are deactivated, I don't think we should mark the metric node as invalid.

Yeah, agreed that I don't think the node should be marked as invalid, but the dimension link needs to be marked as invalid (or hidden in the ways you describe). While we could just filter out this invalid dimension link when we return dimensions for metrics, we still store the link on the column attributes -- should this also have a valid/invalid setting to indicate a deactivated vs activated dimension node?

samredai:
Just an FYI I had to pin `urllib3<2` because it was causing some problems with the mock client used in the tests. I'm not sure how soon it will all be resolved but urllib3's latest major version release is wreaking havoc.

dj/api/nodes.py
agorajek: I think it may be helpful to think about this space in the following way:
- data flow graph (DAG): 
   parent node deactivation --> all child nodes become invalid
- dimension attribute graph (not a DAG) 
   dimension deactivation --> children nodes are still valid, just the dimension/column attribute becomes invalid --> should affect only the functionality that depends on the dimension attributes, nothing else.

@shangyian @samredai would you agree? Sorry if this is obvious but I wanted to spell it out.

shangyian: @agorajek That separation makes sense to me! Definitely good to spell it out

samredai: @agorajek agreed! I'll work on adding a user facing description of these behaviors to the docs as well.

samredai:
@agorajek @betodealmeida @CircArgs @shangyian this is now ready to review. 👍

dj/api/nodes.py
```diff
@@ -259,21 +270,55 @@ def delete_a_node(name: str, *, session: Session = Depends(get_session)):
...
+    node.deactivated_at = UTCDatetime(
+        year=now.year,
+        month=now.month,
+        day=now.day,
+        hour=now.hour,
+        minute=now.minute,
+        second=now.second,
+    )
+    session.add(node)
+    session.commit()
```
agorajek, line 283: Does this call make all the child nodes `invalid` or is it somewhere else?

dj/api/nodes.py
```diff
@@ -259,21 +270,55 @@ def delete_a_node(name: str, *, session: Session = Depends(get_session)):
...
+        content={"message": f"Node `{name}` has been successfully deactivated"},
+    )
+
+
+@router.post("/nodes/{name}/activate/", status_code=201)
+def activate_a_node(name: str, *, session: Session = Depends(get_session)):
+    """
+    Activate the specified node.
+    """
+    node = get_node_by_name(session, name, with_current=True, include_deactivated=True)
```
agorajek, line 295: nit: I think I see the following naming pattern:
- action: de/activate
- state: in/active
So perhaps this argument could be called `include_inactive` ?

agorajek (approved):
Looks good. Just 2 nits in line.

dj/api/nodes.py
samredai: That logic already existed and can be found in these two helper functions: https://github.com/DataJunction/dj/blob/main/dj/api/helpers.py#L394-L474

samredai: Oh sorry that was a link to the logic that revalidates the children after a node is re-activated. The logic that makes all the child nodes `invalid` also already exists and is here: https://github.com/DataJunction/dj/blob/main/dj/api/nodes.py#L256-L260

shangyian (approved):

tests/api/nodes_test.py
```diff
@@ -241,27 +241,6 @@ def test_delete_node(
             response = client_with_examples.get(f"/nodes/{downstream}/")
             assert response.json()["status"] == NodeStatus.INVALID
 
-        node_with_link = client_with_examples.get(
```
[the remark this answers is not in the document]
shangyian: These look great, thanks for adding!

PR #508, opened 2023-05-09
Adding setup for materialization of cube nodes

shangyian:
### Summary

One key functionality of a cube node is that it’s meant to be materialized by an engine for low-latency downstream querying. When a user defines a cube node, it should be engine-agnostic, meaning that it’s expected to work with any set of available engines. 

In order to facilitate this, at the cube node’s creation time, regardless of whether the user chooses to configure materialization, we need to store these generic cube attributes (this is done on the `MaterializationConfig`, which can capture arbitrary json):
* dimensions: a list of dimensions that the cube is built for
* measures: the cube’s selected need to be broken down into constituent measures, or simple aggregations
* filters: a list of filters that apply to the cube
* partitions: a list of partition columns, which must be one of the selected dimensions, or we’ll add it to the list of dimensions
* measures query: a query that pulls all the necessary measures, dimensions, and filters for the cube

When the user adds an actual `MaterializationConfig`, they'll specify a specific engine, like Druid, and provide additional metadata needed for the engine. We'll use the generic cube attributes stored from above + the provided engine-specific metadata to generate a full materialization config for that engine. We'll also call the query service to schedule the materialization, if such an endpoint is available.

### Test Plan

Ran locally. 

- [X] PR has an associated issue: #470 
- [X] `make check` passes


agorajek:
Pretty cool so far. Thanks @shangyian !

dj/materialization/jobs/materialization_job.py
```diff
@@ -0,0 +1,69 @@
...
+    Base class for a materialization job
+    """
+
+    dialect: Optional[Dialect] = None
+
+    def __init__(self):
+        ...
+
+    @abc.abstractmethod
+    def execute(
```
samredai, line 23: Would `schedule` be a better name for this method? It sends the instructions to the query service but may not exactly execute it until the next scheduled instance runs right?

dj/service_clients.py
```diff
@@ -130,3 +131,27 @@ def get_query(
...
+        schedule: str,
+        query: str,
+        spark_conf: str,
+        druid_spec: Dict,
+    ):
+        """
+        Kick off scheduling of materialization job for cube nodes
+        """
+        response = self.requests_session.post(  # pragma: no cover
+            "/materialization/druid/",
```
samredai, line 147: Should this be hardcoded to druid? Since the json in the request is pretty permissive (it can contain a spark conf, druid spec, and I'm guessing more as more implementations of `MaterializationJob` are added), I would have thought it would all go to a generic `/materialization/` endpoint on the query service.

samredai (commented):
This overalls looks great but I left a few comments and nits.

Also just to make sure I understand the design I'll try and summarize. The query service will be responsible for the actual act of materializing a cube. That being said, the main DJ server has to have a reliable way to take all of its knowledge of the node, including the materialization config, and generate materialization instructions for a given target technology (Spark, Trino, Druid, etc) to then send over to the query service. It then has faith that the query service will go off and interact with whichever other systems it needs to for performing the actual materialization. Other materialization targets can be added by adding new implementations of the `MaterializationJob` interface that does the conversion of "dj metadata -> technology specific materialization instructions".

As a silly example, adding something that materializes to CSV files would look like:
- Adding a `CSVMaterializationJob` that implements the `MaterializationJob` interface
- Fill in the execute method that builds up a csv materialization spec
- Passes the spec to the `materialize_cube` method on the query service client which forwards the instructions to the query service

Does that all sound right?

dj/api/nodes.py
```diff
@@ -276,7 +290,11 @@ def delete_a_node(name: str, *, session: Session = Depends(get_session)):
     return Response(status_code=HTTPStatus.NO_CONTENT.value)
 
 
-def cube_materialization_config(cube_node: NodeRevision, config: Dict):
+def build_cube_custom_config(
```
agorajek, line 293: nit: do we need the `_custom_` infix here?

dj/materialization/jobs/materialization_job.py
agorajek: +1, see the 1st word in the docstring.

dj/materialization/jobs/cube_materialization.py
```diff
@@ -0,0 +1,118 @@
...
+        druid_spec = {
+            "dataSchema": {
+                "dataSource": druid_datasource_name,
+                "parser": {
+                    "parseSpec": {
+                        "format": cube_config.druid.parse_spec_format or "parquet",  # type: ignore
+                        "dimensionsSpec": {"dimensions": cube_config.dimensions},
+                        "timestampSpec": {
+                            "column": cube_config.druid.timestamp_column,  # type: ignore
+                            "format": "yyyyMMdd",
```
agorajek, line 85: Does this mean the above `timestamp_column` must match this format or will it be adjusted somewhere?

agorajek (approved):
Can't wait to see this in action! Few nits in-line.

dj/materialization/jobs/cube_materialization.py
shangyian: Good call out! I added the ability to adjust this so that users can provide a different timestamp format if they want.

dj/api/nodes.py
shangyian: Nope, I removed it

dj/service_clients.py
shangyian: Yeah, so in this case the function is `materialize_cube` and it's meant to be only for cubes, but it's true that it's not limited to Druid. I went ahead and removed `druid` and also added a `node_type` parameter so that we can tell the query service what the type of materialization is. For example, in the Druid cube materialization, it has both Spark and Druid config because we use Spark to create the table first, then we ingest the table into Druid.

dj/materialization/jobs/materialization_job.py
shangyian: Makes sense! I renamed it to schedule.

dj/service_clients.py
```diff
@@ -130,3 +131,27 @@ def get_query(
...
+    def materialize_cube(  # pylint: disable=too-many-arguments
+        self,
+        node_name: str,
+        schedule: str,
+        query: str,
+        spark_conf: str,
+        druid_spec: Dict,
+    ):
+        """
+        Kick off scheduling of materialization job for cube nodes
```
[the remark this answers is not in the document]
shangyian: Good point, I reworded it to be more precise about what it's doing

shangyian:
> Does that all sound right?

@samredai Yep, this is pretty much it! As an aside, I might change `materialize_cube` to just `materialize`, since the args work for most nodes beyond cube nodes (those nodes would just ignore the `druid_spec` arg, but will reuse everything else). I'll also add some more documentation about the expected implementation of the `/materialization/` endpoint on either a query service or materialization service.

PR #510, opened 2023-05-11
History table to capture activity for different entities

samredai:
### Summary

This PR adds an activity table and sqlmodel that we can use to store user events. For actions that don't rise to the level of requiring versioning, we can use the activity table to capture some information on what was performed and expose an endpoint to allow listing out these events. Essentially, anything changes that don't require "rollback" functionality can be much more easily captured as activity instead of the more complex mechanisms of versioning.

### Test Plan

Tests that perform a series of events and then compare the expected history results to the actual ones.

- [x] `make check` passes

### Deployment Plan

N/A


dj/models/history.py
```diff
@@ -0,0 +1,40 @@
...
+class EntityType(str, Enum):
+    attribute = "attribute"
+    catalog = "catalog"
+    engine = "engine"
+    namespace = "namespace"
+    node = "node"
+    query = "query"
+    tag = "tag"
+
+class Activity(SQLModel, table=True):  # type: ignore
```
samredai, line 24: @shangyian @CircArgs @betodealmeida @agorajek @cwegener let me know what you all think about the attributes on this `Activity` class. Are there any other things we'd want to capture at the top level for individual activity events? I've included a JSON `details` attribute for capturing dynamic pieces of metadata but anything that we'd want to expose a filter on should probably be pulled out as a first class attribute.

I'm thinking the history endpoint would allow filtering on `entity_type`, i.e. `entity_type=EntityType.node` and `entity_name`, i.e. `entity_name=basic.num_users`.

dj/api/nodes.py
```diff
@@ -722,6 +724,14 @@ def create_a_source(
...
     # Point the node to the new node revision.
     save_node(session, node_revision, node, data.mode)
+    make_history(
+        session,
+        History(
+            entity_type=EntityType.NODE,
+            entity_name=data.name,
+            activity_type=ActivityType.CREATE,
+        ),
+    )
```
samredai, line 734: Here's a an example of adding a simple event to the history table that captures a node being created.

dj/models/history.py
shangyian: I think this looks like a good start! Do we want to capture some kind of diff of what changed (perhaps represented by json in `details`) by default? That way we'd be capturing all the potential information we need for the future, in case there's something we later decide is important (and we can pull it out as a first-class attribute then).

dj/models/history.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    An event to store as part of the server's activity history
+    """
+
+    id: Optional[int] = Field(default=None, primary_key=True)
+    entity_type: Optional[EntityType] = Field(default=None)
+    entity_name: Optional[str] = Field(default=None)
+    activity_type: Optional[ActivityType] = Field(default=None)
+    user: Optional[str] = Field(default=None)
+    details: Dict[str, Any] = Field(default={}, sa_column=SqlaColumn(JSON))
```
agorajek, line 54: Wondering how we can represent the `details` for each type of activity. I think it should be easy for most event types, but what about UPDATE?

agorajek (approved):
@samredai this sounds like a great idea. Wondering if there is some SQL alchemu functionality of comparing objects that we could use for the `details`.

dj/models/history.py
samredai: What if we added a `pre: Dict[str, Any]` and `post: Dict[str, Any]` attribute and these can be populated when the history is captured? How the diff is displayed can then be handled client said depending on the environment and context. For example the UI may use something similar to [react-diff-viewer](https://github.com/praneshr/react-diff-viewer/tree/v2.0).

samredai: Added the pre and post attributes.

PR #512, opened 2023-05-14
Update docs snippets and add validation scripts

samredai:
### Summary

This PR fixes a lot of the snippets in the docs so they flow naturally and can be executed in order on a fresh DJ deployment.

**note:** This PR only tackles the curl and js code snippets. I plan to follow up with a similar PR for the python code snippets.

### Test Plan

I've included a shell script (container curl commands) and a nodejs script. Both of these include all of the docs snippets, ran sequentially. It'll take some diligence to make sure we keep these in sync with the snippets we have in the docs, but it's useful in that we can run this script on a new DJ deployment to make sure the code snippets all run successfully.

- [x] `make check` passes

### Deployment Plan

N/A


shangyian (approved):
Nice! Should these be added to CI process so that we can run a docs check before any PRs get merged?

PR #513, opened 2023-05-14
Allow node updates in Python client + update bugs

shangyian:
### Summary

This change adds node update support to the Python client for sources, dimensions, metrics, and transforms. It also fixes some bugs with node updates:
* Dimension links should be propagated during updates where possible
* A node's primary keys should be modifiable, with the node's status changing accordingly

### Test Plan

Locally and added tests

- [x] `make check` passes

### DIFF
--- a/dj/api/nodes.py
+++ b/dj/api/nodes.py
@@ -975,7 +975,12 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals
         and data.columns is not None
         and ({col.identifier() for col in old_revision.columns} != data.columns)
     )
-    major_changes = query_changes or column_changes
+    pk_changes = (
+        data is not None
+        and data.primary_key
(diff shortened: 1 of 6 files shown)


dj/api/nodes.py
```diff
@@ -1019,16 +1025,22 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals
...
-    ):
+    if new_revision.type != NodeType.SOURCE and (query_changes or pk_changes):
         (
             validated_node,
             dependencies_map,
             missing_parents_map,
             type_inference_failed_columns,
         ) = validate_node_data(new_revision, session)
+
+        # Keep the dimension links and attributes on the columns from the node's
```
samredai, line 1036: Great catch here!

samredai (approved):

PR #516, opened 2023-05-15
check metric nodes have the column named after the node name

CircArgs:
### Summary

Metrics should name their expression to be the name as the node name

### Test Plan

- [x] `make check` passes


dj/sql/parse.py
```diff
@@ -2,26 +2,52 @@
...
+            "Invalid Metric. Node query has does not have "
+            "a single expression in the projection.",
+        )
+
+    if (
+        not hasattr(tree.select.projection[0], "alias_or_name")
+        or tree.select.projection[0].alias_or_name.name != node.name  # type: ignore
+    ):
+        raise DJInvalidInputException(
+            "Invalid Metric. The expression in the projection "
```
samredai, line 42: Instead of requiring that the expression has an alias that matches the metric node name, is there a way we can automatically set the alias for the user? I guess there could be some confusion if for example they define a metric `sales.num_orders` as `SELECT count(orders) AS total_order_count FROM ...`. If we replace the metric alias with `sales_DOT_num_orders` the user might be surprised because they were expecting it to be `total_order_count`. Maybe we should not allow any alias to be set by the user in the SQL definition. It would be a very reasonable explanation that the system defines metric expression aliases for consistency.

agorajek: How about this:
- if there is an alias and it does not match the expected alias, raise the error and tell the user
- otherwise if the alias is missing we can auto-add it

samredai: I like that solution @agorajek, if the user already sets the proper alias it would definitely be unnecessary to tell them to remove it so that the system can re-add the exact same one 😅 .

@CircArgs that would mean this PR remains mostly as it is now, just have to break out the `not hasattr(tree.select.projection[0], "alias_or_name")` condition to its own if block and just set the alias for the user.

shangyian: Yeah, I suspect most users won't add aliases to the column when they write their metrics, so the addition of auto-adding aliases will be very useful!

shangyian (approved):

Issue #515, opened 2023-05-15
Allow setting order_by for /sql/ and /data/


samredai:

CircArgs:
On it

PR #520, opened 2023-05-19
add ordering `order by`s in sql and data routes

CircArgs:
### Summary

**_WIP working on coverage_**

Adds order by dimension columns to the `sql` and `data` api routes
Adds filter cols to the projection

#### Bug fixes:
- `Name`s as `Name.name` stripped to be just `str`
- tables referencing nodes can be pre-bound to columns and build will discover and add them as it used to

### Test Plan

Needs changes/additional unit tests

- [x] PR has an associated issue: https://github.com/DataJunction/dj/issues/515
- [x] `make check` passes

### DIFF
--- a/dj/construction/build.py
+++ b/dj/construction/build.py
@@ -4,10 +4,8 @@ import collections
 # pylint: disable=too-many-arguments,too-many-locals,too-many-nested-blocks,too-many-branches,R0401
 from typing import DefaultDict, Deque, Dict, List, Optional, Set, Tuple, Union, cast
 
-from sqlalchemy.sql import operators
-
 from sqlmodel import Session
-import operator
+
 from dj.construction.utils import to_namespaced_name
 from dj.errors import DJException, DJInvalidInputException
 from dj.models.column import Column
@@ -301,7 +299,10 @@ def _build_tables_on_select(
                 ]
             node_ast.compile(context)
 
-            select.replace(tbl, node_ast)#, lambda x, y: type(x)==type(y) and hasattr(x, 'identifier') and hasattr(y, 'identifier') and x.identifier(False) == y.identifier(False))
+            select.replace(
+                tbl,
+                node_ast,
+            )
 
 
 def dimension_columns_mapping(
@@ -341,9 +342,10 @@ def _build_select_ast(
     join_tables_for_dimensions(session, dimension_columns, tables, build_criteria)
     _build_tables_on_select(session, select, tables, build_criteria)
 
+
 def _consolidate_tables(query: ast.Query):
     """
-    Checks all the tables whose nearest query is 
+    Checks all the tables whose nearest query is
     the one given, matching identifiers, and makes them
     all the same `Table` in memory
     """
@@ -355,12 +357,12 @@ def _consolidate_tables(query: ast.Query):
         ident = tbl.identifier(False)
         if isinstance(tbl.parent, ast.Column):
             if ident not in column_tables_map:
-                column_tables_map[ident]=[tbl]
+                column_tables_map[ident] = [tbl]
             else:
                 column_tables_map[ident].append(tbl)
         else:
             if ident not in foundation_tables_map:
-                foundation_tables_map[ident]=[tbl]
+                foundation_tables_map[ident] = [tbl]
             else:
                 foundation_tables_map[ident].append(tbl)
 
@@ -371,12 +373,13 @@ def _consolidate_tables(query: ast.Query):
             for col_tbl in column_tables_map[ident]:
                 col_tbl.swap(tbls[0])
             del column_tables_map[ident]
- 
+
     for tbls in column_tables_map.values():
         for tbl in tbls[1:]:
             tbl.swap(tbls[0])
 
-    
+
+# pylint: disable=R0915
 def add_filters_dimensions_orderby_limit_to_query_ast(
     query: ast.Query,
     dialect: Optional[str] = None,  # pylint: disable=unused-argument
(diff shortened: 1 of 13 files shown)


dj/construction/build.py
```diff
@@ -339,37 +339,65 @@ def _build_select_ast(
...
-            ).select
-            query.select.group_by += temp_select.group_by  # type:ignore
-            projection_addition += list(temp_select.find_all(ast.Column))
+    if not query.organization:
+        query.organization = ast.Organization([])
+
+    orderby_tables = set()
+    if orderbys:
+        for order in orderbys:
+            temp_query = parse(
```
agorajek, line 386: Is this the WIP ?

dj/construction/build.py
```diff
@@ -339,37 +339,65 @@ def _build_select_ast(
...
+                f"select * order by {order}",
+            )
+            orderby_tables |= set(temp_query.find_all(ast.Table))
+            projection_addition |= set(temp_query.find_all(ast.Column))
+            query.organization.order += temp_query.organization.order  # type:ignore
+
+    if diff := orderby_tables - dimension_tables:
+        raise DJInvalidInputException(
+            "Order by columns can only be from "
+            "dimensions used in filters or dimensions arguments"
```
agorajek, line 396: What is `dimensions arguments` ?

agorajek (commented):
`s/orderbys/orderby` ? order_by (similarly to limit) sounds better in singular form.

dj/construction/build.py
CircArgs: very much so. ran up against https://github.com/DataJunction/dj/issues/523

dj/construction/build.py
CircArgs: tried to clarify in the message since there are api parameters `filters` and `dimensions` but of course you are using dimension columns in `filters` as well

dj/construction/build.py
```diff
@@ -339,47 +343,144 @@ def _build_select_ast(
...
+                    projection_addition[ident] = col
+                if col.table:#pragma: no cover
+                    orderby_tables.add(cast(ast.Table, col.table).identifier(False))
+            
+            query.organization.order += temp_query.organization.order  # type:ignore
+
+    if diff := orderby_tables - dimension_tables:
+        raise DJInvalidInputException(
+            "Order by columns can only be from "
+            "dimensions used in filters or dimensions arguments"
```
shangyian, line 458: Should "dimensions arguments" just be called "group bys" instead? (I'm assuming that most of users would understand that better since it's largely a SQL-based system, whereas the distinction between `dimension arguments` and `dimensions` is not super clear)

shangyian, line 458: Should "dimensions arguments" just be called "group bys" instead? (I'm assuming that most of users would understand that better since it's largely a SQL-based system, whereas the distinction between `dimension arguments` and `dimensions` is not super clear)

dj/sql/parsing/ast.py
```diff
@@ -671,6 +699,16 @@ def expression(self) -> Optional[Expression]:
         """
         return self._expression
 
+    def namespace_table(self):
+        """
+        Turns the column's namespace into a table.
+        """
+        if self.name.namespace:
+            self._table = Table(name=self.name.namespace)
```
shangyian, line 707: Does this need to look for an already existing table with the same name on the ast first? Although I suppose it doesn't matter if there's `_consolidate_duplicates`

shangyian (approved):
Nice, thanks for taking care of this! Is my understanding right that this handles https://github.com/DataJunction/dj/issues/523 or do you think for that issue we need to handle expressions in non-`SELECT` clauses in a more streamlined way?

CircArgs:
@shangyian it doesn't currently do anything to directly address that issue. This sidesteps it by taking advantage of adding columns to the projection so when the build touches the table in the projection it also inadvertently changes the one in the ordering. Kinda a hack for now. That issue will need more direct attention I think than I felt could fit in this PR

dj/construction/build.py
```diff
@@ -339,47 +343,144 @@ def _build_select_ast(
     _build_tables_on_select(session, select, tables, build_criteria)
 
 
-def add_filters_dimensions_limit_to_query_ast(
+def _consolidate_tables(query: ast.Query):
```
agorajek, line 346: I am having a hard time understanding how any changes made inside this function make it outside. Is it through the `query` argument?

dj/sql/parsing/ast.py
```diff
@@ -814,34 +858,39 @@ def compile(self, ctx: CompileContext):
...
+                        message=f"Column`{self}` does not exist on any valid table.",
+                    ),
+                )
+                return
+
+            if len(found_sources) > 1:
+                ctx.exception.errors.append(
+                    DJError(
+                        code=ErrorCode.INVALID_COLUMN,
+                        message=f"Column `{self.name.name}` found in multiple tables. Consider namespacing.",
```
agorajek, line 880: `s/Consider namespacing/Consider using fully qualified name`

dj/construction/build.py
CircArgs: a lot of key functions in `build.py` modify in-place. `add_filters_dimensions_limit_to_query_ast` and `_consolidate_tables` both modify in-place by acting on their `query` arguments

dj/construction/build.py
CircArgs: @shangyian are you suggesting to change all the api arguments from `dimensions` to `group_bys` or just the terms in that error

tests/construction/build_test.py
```diff
@@ -157,6 +157,7 @@ async def test_build_metric_with_dimensions_filters(request):
     )
     expected = """
     SELECT
+        basic_DOT_dimension_DOT_users.age,
```
agorajek, line 160: @CircArgs this new line seems incorrect. If there was a group_by on `users.age` defined byt the caller then it would make sense to have it here but then the GROUP_BY clause is missing.

agorajek: Yeah... but if you want that SQL to be valid then any non-agg column you add to SELECT clause you also need to add to GROUP_BY clause.

agorajek: Although... I think that in this case (of filter columns) we don't want to add them to SELECT clause.

CircArgs: shoot!... you're right. I'll fix this.

Issue #521, opened 2023-05-20
Building multiple metrics with different parents should fail


shangyian:
When we build queries for multiple metrics that have different parent nodes, we generate a query that cannot run because the generated query assumes that both parents are already being queried from rather than raising an error. 

It looks like the original intent of the check here (https://github.com/DataJunction/dj/blob/main/dj/construction/build.py#L535-L540) was to raise an error, but it doesn't cover all cases. 

I'll find some time tomorrow to append onto this issue with the exact edge case.

Issue #519, opened 2023-05-19
Filter validation


CircArgs:
Currently we validate that dimensions exist when a call is made from the api, but we do not extract the columns from filters to validate that they are from shared dimensions.

This is where the ordinary groupby dimensions are validated for example:
https://github.com/DataJunction/dj/blob/9424702275d90cfea1e6528dcfe2eac7459e6125/dj/construction/build.py#L487-L493

shangyian:
Good catch!

PR #524, opened 2023-05-21
Fix query building for multiple metrics + add filter validation 

shangyian:
### Summary

We should be able to build queries for any set of metrics, as long as they share the same set of dimensions. This applies even if they have different upstream sources. Given this, the multiple metric query building logic can look like this:

1. Build each metric node query independently
2. Wrap each built metric node query in a `WITH` statement
3. Join the node queries together via the dimension columns
4. In the final select, bring in all metric columns and the coalesced version of all dimension columns

The cube node decomposition also needs to change to accommodate this logic.

### Test Plan

  - #521
  - #519
- [X] `make check` passes

### DIFF
--- a/dj/construction/build.py
+++ b/dj/construction/build.py
@@ -482,8 +482,18 @@ def build_metric_nodes(
 ):
     """
     Build a single query for all metrics in the list, including the
-    specified group bys (dimensions) and filters.
+    specified group bys (dimensions) and filters. As long as all
+    metric nodes share the same set of dimensions, we can:
+    (a) build each metric node query separately
+    (b) wrap each built metric node query in a WITH statement
+    (c) join the node queries together via the dimension columns
     """
+    if any(metric_node.type != NodeType.METRIC for metric_node in metric_nodes):
+        raise DJInvalidInputException(
+            "Cannot build a query for multiple nodes if one or more "
+            f"of them aren't metric nodes.",
+        )
+
     shared_dimensions = get_shared_dimensions(metric_nodes)
     for dimension_attribute in dimensions:
         if dimension_attribute not in shared_dimensions:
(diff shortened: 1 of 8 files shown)


shangyian:
@CircArgs I went ahead and folded the fix for https://github.com/DataJunction/dj/issues/519 into this PR since it's touching the same code

tests/sql/utils.py
```diff
@@ -27,11 +27,23 @@ def compare_query_strings(str1: str, str2: str) -> bool:
...
     query2.select.projection = sorted(
         query2.select.projection,
         key=lambda x: str(x.alias_or_name),  # type: ignore
     )[:]
+    for cte in query2.ctes:
+        cte.select.projection = sorted(
+            query2.select.projection,
+            key=lambda x: str(x.alias_or_name),  # type: ignore
+        )[:]
+
```
CircArgs, line 46: not sure if we could just "bake" the ctes in with the `Query` method and use the original logic here. Doesn't matter though.

CircArgs (approved):

dj/construction/build.py
```diff
@@ -492,52 +502,98 @@ def build_metric_nodes(
...
+            for col in final_select_columns
+            if str(col.alias_or_name) not in dimensions
+        ]
+        dimension_columns = [
+            col for col in final_select_columns if str(col.alias_or_name) in dimensions
+        ]
+        all_dimension_columns += dimension_columns
+        combined_ast.select.projection.extend(metric_columns)
+
+        if not combined_ast.select.from_.relations:  # type: ignore
```
agorajek, line 560: What does this condition mean?

tests/examples.py
```diff
@@ -1686,20 +1686,37 @@
...
-        "default_DOT_repair_orders.hard_hat_id,\n\tdefault_DOT_repair_orders.municipality_id,"
-        "\n\tdefault_DOT_repair_orders.repair_order_id \n FROM roads.repair_orders AS "
-        "default_DOT_repair_orders) AS default_DOT_repair_order ON "
-        "default_DOT_repair_order_details.repair_order_id = "
-        "default_DOT_repair_order.repair_order_id\nLEFT OUTER JOIN (SELECT  "
-        "default_DOT_dispatchers.company_name,\n\tdefault_DOT_dispatchers.dispatcher_id "
-        "\n FROM roads.dispatchers AS default_DOT_dispatchers) AS default_DOT_dispatcher "
-        "ON default_DOT_repair_order.dispatcher_id = default_DOT_dispatcher.dispatcher_id "
-        "\n GROUP BY  default_DOT_dispatcher.company_name\n LIMIT 10"
+        "WITHm0_default_DOT_num_repair_ordersAS(SELECTdefault_DOT_dispatcher.company_name,\t"
```
agorajek, line 1689: Wondering why this text has no spaces in it, e.g.: "WITHm0_default_DOT_num_repair_ordersAS"

agorajek (approved):
Good fix @shangyian . Just two questions in-line.

dj/construction/build.py
shangyian: Ah so these two conditions are trying to build the final select that joins together all the CTEs. This condition is checking if anything is in the select statement's `FROM` clause, and in the very first iteration through the CTEs, there won't be any, so we want to add that as the `FROM ...` relation. Then, the subsequent CTEs need to be joined in via `JOIN <first relation> ON ...`, which is where the `else` comes in.

PR #532, opened 2023-05-24
A few miscellaneous bugfixes

shangyian:
### Summary

Bugfixes include:
* Provide hints when user asks for a dimension that isn't prefixed: https://github.com/DataJunction/dj/commit/dd272bbf5e1d2990a6aecba0feebbe109dd39dbd 
* We should be able to automatically populate the alias of a metric column whenever it is empty, but there was a bug where complex metric column expressions without an alias would lead to DJ raising an error about the column alias rather than just auto-aliasing: https://github.com/DataJunction/dj/commit/51f6a95cd03408a561a2ed49a65a6dd9c7c55c7f
* Fixing release workflow condition check: https://github.com/DataJunction/dj/commit/446c311b04e43896e35d796cecca536a3c8939fc
* Include dimension node link column as a dimensional attribute available when asking for a node's dimensions: https://github.com/DataJunction/dj/commit/49822869f5acaf77637709163c4fa1f480fc7961

### Test Plan

- [x] `make check` passes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Tag release
         run: |
           export NEW_VERSION=v$(hatch version)
-          export LIBRARY=${{ "server" ? steps.changes.outputs.server : "client" }}
+          export LIBRARY=${{ steps.changes.outputs.server ? "server" : "client" }}
           git tag -a $NEW_VERSION-$LIBRARY -m $NEW_VERSION-$LIBRARY
           git push origin $NEW_VERSION-$LIBRARY
         env:
--- a/dj/construction/build.py
+++ b/dj/construction/build.py
@@ -599,12 +599,20 @@ def build_metric_nodes(
     for filter_ in filters:
         temp_select = parse(f"select * where {filter_}").select
         columns_in_filter = temp_select.where.find_all(ast.Column)  # type: ignore
+        dims_without_prefix = {
+            dim.split(".")[-1]: dim
+            for dim in shared_dimensions
+        }
         for col in columns_in_filter:
             if str(col) not in shared_dimensions:
+                potential_dimension_match = (
+                    f"Did you mean `{dims_without_prefix[str(col)]}`?"
+                    if str(col) in dims_without_prefix else ""
+                )
                 raise DJInvalidInputException(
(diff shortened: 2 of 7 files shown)


dj/models/node.py
```diff
@@ -466,24 +466,27 @@ def format_metric_alias(query: str, name: str) -> str:
...
         # otherwise, we will just add the alias we want e.g. the node name
-        expr_name: ast.Name = projection_0.alias_or_name  # type: ignore
-        if expr_name.name != amenable_name(name):
-            if expr_name.parent_key == "alias":
-                raise DJInvalidInputException(
-                    "Invalid Metric. The expression in the projection "
-                    "cannot have alias different from the node name. Got "
-                    f"`{expr_name}` but expected `{amenable_name(name)}`",
-                )
+        expr_name: Optional[ast.Name] = (
```
shangyian, line 474: @CircArgs I changed the logic for this so that any metric expression, regardless of whether it's a binaryop or a single column, as long as it doesn't have an alias, will get an alias set with this method. Otherwise I think users will be confused as to why their complex metric expression doesn't get auto-assigned an alias.

.github/workflows/release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Tag release
         run: |
           export NEW_VERSION=v$(hatch version)
-          export LIBRARY=${{ "server" ? steps.changes.outputs.server : "client" }}
+          export LIBRARY=${{ steps.changes.outputs.server ? "server" : "client" }}
```
shangyian, line 58: @agorajek This is the switch in the conditions 😄

dj/construction/build.py
```diff
@@ -599,12 +599,18 @@ def build_metric_nodes(
     for filter_ in filters:
         temp_select = parse(f"select * where {filter_}").select
         columns_in_filter = temp_select.where.find_all(ast.Column)  # type: ignore
+        dims_without_prefix = {dim.split(".")[-1]: dim for dim in shared_dimensions}
         for col in columns_in_filter:
             if str(col) not in shared_dimensions:
+                potential_dimension_match = (
```
agorajek, line 605: great idea

agorajek (approved):
Awesome!

PR #535, opened 2023-05-25
Data retrieval polling via DJ client + progress bar 

shangyian:
#### Async Queries

This PR changes the way that data retrieval via the Python client works. In the existing approach, we call the `/data/` endpoint synchronously and then block until the query finishes running and we have data. This is unnecessary (and not super scaleable), as it will always consume one gunicorn worker every time someone uses the client to request data, even if it does nothing beyond wait for query results to come back.

We can instead issue the data request with `async_=true` and then poll (with backoff) to wait for the query to finish running so that we can retrieve results.

#### Progress Bar

Since data retrieval queries may take some time to run, this change also adds in a progress bar (using [alive-progress](https://github.com/rsalmei/alive-progress)) to indicate to users that the query is running:

out

There is also additional summary statistics provided after the progress bar is complete:

Screenshot 2023-05-25 at 1 47 23 PM

#### Models

This also moves all the pydantic models into `models.py`. pylint reminded me that this was needed since `client.py` is getting absurdly long.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -549,7 +549,7 @@ class DJClient:  # pylint: disable=too-many-public-methods
             length=20,
             bar="smooth",
             force_tty=True,
-            calibration=5e40,
+            calibrate=5e40,
         ) as progress_bar:
             poll_interval = 1  # Initial polling interval in seconds
             job_state = models.QueryState.UNKNOWN
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -2,9 +2,9 @@
 import abc
 
 # pylint: disable=redefined-outer-name, import-outside-toplevel
-import enum
 import logging
 import platform
+import time
 import warnings
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urljoin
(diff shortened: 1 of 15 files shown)


client/python/datajunction/client.py
```diff
@@ -583,33 +543,68 @@ def data(  # pylint: disable=too-many-arguments
...
-            raise DJClientException("No data returned for requested set")
-        columns = results["results"][0]["columns"]
-        rows = results["results"][0]["rows"]
-        return pd.DataFrame(rows, columns=[col["name"] for col in columns])
+
+        with alive_bar(
+            title="Processing",
+            length=20,
+            bar="smooth",
+            force_tty=True,
```
agorajek, line 551: I see this in their docs:
```
The values accepted are:

force_tty=True -> always enables animations, and auto-detects Jupyter Notebooks!
force_tty=False -> always disables animations, keeping only the final receipt
force_tty=None (default) -> auto detect, according to the terminal's tty state
```

Do you think True is safer enough for our most common uses? We can go with True and see how it goes.

agorajek (approved):
This is super awesome! Thanks @shangyian . Just one question in line.

client/python/datajunction/client.py
shangyian: Yeah, I agree that `True` should be enough for most of our use cases. I don't think many folks will be using the DJ Python client from the Python interactive interpreter, whereas Jupyter is the most likely candidate. But we can always change it if that turns out to be wrong 😄

PR #538, opened 2023-05-26
pdm exports + clean up pip-compile-multi artifacts

shangyian:
### Summary

This adds two pdm hooks to export into `requirements/docker.txt` and `requirements/test.txt`, which we use in the Dockerfile downstream. This setup also provides maximum flexibility for downstream users, in case they prefer to use pip to install the requirements rather than pdm.

The change also removes all the `pip-compile-multi` artifacts from `requirements/`, like the `*.in` files and the `base` setup.

### Test Plan

- [X] `make check` passes


.pre-commit-config.yaml
```diff
@@ -106,3 +106,9 @@ repos:
   rev: 2.5.3
   hooks:
     - id: pdm-lock-check
+#    - id: pdm-export
```
shangyian, line 109: @cwegener Have you seen this issue before where the pre-commit hook doesn't appear to use the right version of pdm and then fails?

agorajek (approved):

.pre-commit-config.yaml
cwegener: Interesting. No. I haven't come across this before. I'll keep an eye out for it.

.pre-commit-config.yaml
```diff
@@ -103,6 +103,12 @@ repos:
...
-  rev: 2.5.3
+  rev: 2.6.1
   hooks:
     - id: pdm-lock-check
+    - id: pdm-export
+      args: ['-o', 'requirements/docker.txt', '--without-hashes', '--prod']
+      files: ^pdm.lock$
+#    - id: pdm-export
+#      args: ['-o', 'requirements/test.txt', '--without-hashes', '--dev']
+#      files: ^pdm.lock$
```
cwegener, line 114: I was thinking that the export should actually run automatically after a `pdm lock` when a dev changes dependencies.
The workflow being:
1. Dev uses `pdm add` or `pdm update ...` to modify dependencies, each of those two commands trigger a lock / re-lock
2. Therefore, it should be possible to add a command into pdm's `post_lock` hook that runs the exports

The benefit of using pdm hooks being that they happen before pre-commit in the dev cycle and the extra step of re-adding the requirements file after pre-commit runs the export would be avoided.

The pdm-lock-check on the other hand is still a last barrier to catch any discrepancies between pyproject.toml and pdm.lock.

Makefile
```diff
@@ -31,6 +31,8 @@ spellcheck:
 
 check:
 	pdm run pre-commit run --all-files
+	pdm export -o requirements/docker.txt --without-hashes --prod
+	pdm export -o requirements/test.txt --without-hashes --dev
```
cwegener, line 35: These would not be needed if we can get a pdm `post_lock` hook to run the exports automatically.

cwegener (commented):
Left some comments about `pdm export` approaches.

cwegener:
> Left some comments about `pdm export` approaches.

I think that the `pdm-autoexport` plugin does what I described by hooking into the `post_lock` hook .... https://github.com/pdm-project/pdm-autoexport

.pre-commit-config.yaml
cwegener: > Have you seen this issue before where the pre-commit hook doesn't appear to use the right version of pdm and then fails?
Where are you seeing theses issues? Running pre-commit locally or in the GH actions?

cwegener: I just noticed something really strange ... the pdm.lock file schema version is 3.1 .... Which is the lockfile schema from pdm version 2.0 ... there have been multiple lockfile verison upgrades since then.

https://github.com/DataJunction/dj/blob/49ac7b00015d3b9d4abea508d64f02a9da82d48d/pdm.lock#L818

We'll need to fix whatever is broken there first, because somehow that pdm.lock file got created/edited with an old pdm 2.0 version ...

cwegener: @shangyian Somehow you downgraded pdm to 2.0 last week ... :joy: https://github.com/DataJunction/dj/commit/afb1b09#diff-4dc59eeadc4eee1785e5a8461355721763a4ab9a32b8e3b88def661da9cb497cR818

cwegener: I made #539

.pre-commit-config.yaml
shangyian: Perfect! That approach makes a lot of sense to me. I'll look into `pdm-autoexport`.

.pre-commit-config.yaml
shangyian: Thanks for the fix! As for the pre-commit hook, I was seeing this only in GH actions, but not while running it locally. In any case it shouldn't be needed with the post hook.

shangyian:
@cwegener I've updated the PR with pdm-autoexport. It seems to work nicely.

PR #539, opened 2023-05-27
fix: upgrade pdm back to a recent version

cwegener:
afb1b09 downgraded pdm to version 2.0.x and an old lockfile format (3.1),
causing tooling issues.

Upgrading pdm back to the newest version again.


shangyian (approved):
Oops! Thanks for fixing this 😄

PR #542, opened 2023-05-31
Authentication + authorization dependency injection setup

samredai:
### Summary

This PR is an attempt at adding some OAuth2 scaffolding, utilizing fastapi's dependency injection to allow injecting some custom auth logic. I've been trying to research the best way to do this and it definitely seems like there are a lot of different designs out there so any feedback on this draft would be helpful. At a high level:

- DJ determines the resource identifier
  - Since our API pretty neatly maps to a resource authorization scheme, I figured we may be able to get away with using the HTTP verb and url path as the resource identifier. `f"{self.request.method}:{self.request['path']}"`, i.e. `GET:/nodes/default.foo`
- Dependency injection determines the authentication and authorization
  - `get_current_user` - Performs whatever logic necessary to determine the current user. Right now it's a silly password bearer setup where there's only a single hard-coded username "dj" with a password "dj". This dependency can be overriden with any oauth2 scheme (or maybe even any scheme in general?...)
  - `PermissionsChecker` - This is a dependency that can be added to any route where it will then be initialized by fastapi with a `permission` attribute that contains the resource identifier such as `GET:/nodes/default.foo`. The `__call__` logic then takes a user and does whatever custom logic may be required to check if that user can access that resource, i.e. check roles defined in some RBAC server. This allows us to determine in OSS where the checks are performed without necessarily dictating how they are performed.

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/api/auth.py
+++ b/dj/api/auth.py
@@ -30,17 +30,16 @@ class PermissionsChecker:  # pylint: disable=too-few-public-methods
     """
 
     def __init__(self, request: Request):
-        self.request = request
+        self.permission = f"{request.method}:{request['path']}"
 
     def __call__(self, user: User):
         """
         A placeholder method for verifying if a user has the required permissions
         """
-        permission = f"{self.request.method}:{self.request['path']}"
         _logger.error(
             "Checking if user %s has permission %s",
             user.username,
-            permission,
+            self.permission,
         )
 
 
--- a/dj/api/auth.py
+++ b/dj/api/auth.py
@@ -1,0 +1,54 @@
+"""
+Auth dependencies
+"""
+import logging
+from typing import Annotated
+
+from fastapi import APIRouter, Depends, Request
+from fastapi.security import OAuth2PasswordBearer, OAuth2PasswordRequestForm
+
+from dj.models.user import Token, User
+
+_logger = logging.getLogger(__name__)
+router = APIRouter()
+
+oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
+
+
+async def get_current_user(
+    token: Annotated[str, Depends(oauth2_scheme)],  # pylint: disable=unused-argument
+) -> User:  # pylint: disable=unused-argument
+    """
+    Overrideable placeholder function for getting a user from a token
+    """
+    return User(username="unknown")
+
+
+class PermissionsChecker:  # pylint: disable=too-few-public-methods
+    """
+    Overrideable placeholder for checking if a user has required permissions
+    """
+
+    def __init__(self, request: Request):
+        self.request = request
+
+    def __call__(self, user: User):
+        """
+        A placeholder method for verifying if a user has the required permissions
+        """
+        permission = f"{self.request.method}:{self.request['path']}"
+        _logger.error(
+            "Checking if user %s has permission %s",
+            user.username,
+            permission,
+        )
+
+
+@router.post("/token", response_model=Token)
+async def generate_an_auth_token(
+    form_data: OAuth2PasswordRequestForm = Depends(),  # pylint: disable=unused-argument
+):
+    """
+    Overrideable placeholder endpoint for generating a bearer token
+    """
+    return {"access_token": "unsecured", "token_type": "bearer"}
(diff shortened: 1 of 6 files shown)


dj/api/auth.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    return User(username="admin")
+
+
+class PermissionsChecker:  # pylint: disable=too-few-public-methods
+    """
+    Overrideable placeholder for checking if a user has required permissions
+    """
+
+    def __init__(self, request: Request):
```
CircArgs, line 32: Summarizing here from the conversation. Maybe some more dedicated parameters like the verb, the base route, the nodes - more targeted arguments overall to avoid any complexity of handling parsing stuff from the url

shangyian (commented):
Also adding this note from our conversation earlier: the query service client will also need to be modified to generate a new requests session every time a request is issued, with caching removed, so that it's not sharing auth across sessions.

dj/api/auth.py
samredai: Good point. We definitely need to land on a design where a custom implementation of a permissions checker doesn't need to have a deep understanding on how to parse a request but I'm wondering if we should go even further than just providing the targeted arguments (since they'll still need to make sense of how to compose the arguments together).

DJ can define the `__init__` logic here in a universal way that doesn't need to be replaced and then a custom implementation just needs to worry about the `__call__` logic. Right now DJ will automatically inject the `self.permission` attribute which would be something like `GET:/nodes/default.foo` and the permissions checker could just use that as a standard permission string. You brought up a great point earlier that for the more complicated sql and data endpoint it's not as straightforward. For example `/sql/?metrics=default.foo&metrics=default.bar&dimensions=default.baz&dimensions=default.qux` doesn't make sense as a single resource identifier.

So maybe we leave it fully in the hands of the OSS logic to determine the array of resource identifiers explicitly for every permission check. For the sql endpoint example, DJ would provide an array that looks like `["GET:/nodes/default.foo", "GET:/nodes/default.bar", "GET:/nodes/default.baz", "GET:/nodes/default.qux"]`. The implementer won't need to know about how those resource identifiers are constructed, just that the user needs to have all of those permissions.

shangyian: Yeah, I'm in favor of that logic living in OSS, and the only part that's specific to each internal implementation is just the part that checks a specific grant and returns `access granted` or `access denied`. This also lets us layer grants better - for example, we could have something like:
1. Collect all nodes and see if they have grants. Apply each node's grants and deny access to the endpoint if any of them fail.
2. If any node doesn't have a grant, we default to the node's namespace's grant and apply that if it exists. If nothing exists, we default to granting access (or denying access).

CircArgs:
One question I have is about doing auth eagerly or not. Do we have some builder styled `AuthContext` that we build up all the permissions we come across like when compiling we could stumble upon dozens of nodes. For each node would be eagerly ask "can I use this" or do we build everything up in such a context and say "can I use all of these".

Things that I can currently think of for this are:

- In the eager case you may make many separate auth requests

- In the lazy case, DJ might end up doing a bunch of worthless work if in the end it doesn't have access but build everything just to aggregate all permissions.

samredai:
@CircArgs I like this idea of having a builder where we accumulate the list of permissions required and pass them all to the permission checker abstraction as one context object.  As for eager vs. lazy, if only considering the issues you listed I'd rather go the lazy route. With the eager route I feel like we'd still end up lazily accumulating the auth results for the many separate requests since we want to give users the full picture. In other words "Access denied because you don't have this set of permissions".

But digging in there a bit more, I'm wondering if we would do auth checks during compile at all. Should requesting a metric/dimension node require that the user have read access to all of the upstream transform/source nodes used to build them, or is it enough that they have read access to the metric/dimension nodes they explicitly requested?

agorajek:
> But digging in there a bit more, I'm wondering if we would do auth checks during compile at all. Should requesting a metric/dimension node require that the user have read access to all of the upstream transform/source nodes used to build them, or is it enough that they have read access to the metric/dimension nodes they explicitly requested?

I think it will be easier to manage if:
1. We check the Node owner permissions on all upstream nodes during a node create/update operation.
2. Then we should only need to check the permissions of requesting user on the node being requested.

Any node traversal for auth purpose during runtime seems harder to manage.

Issue #544, opened 2023-05-31
Missing spark functions in DJ


shangyian:
From https://spark.apache.org/docs/latest/api/sql/index.html, here are all the functions we're currently missing:
- [ ] `abs`
- [ ] `acos`
- [ ] `acosh`
- [ ] `add_months`
- [ ] `aes_decrypt`
- [ ] `aes_encrypt`
- [ ] `aggregate`
- [ ] `any`
- [ ] `any_value`
- [ ] `approx_count_distinct`
- [ ] `approx_percentile`
- [ ] `array_agg`
- [ ]  `array_append`
- [ ] `array_compact`
- [ ] `array_contains`
- [ ] `array_distinct`
- [ ] `array_except`
- [ ] `array_insert`
- [ ] `array_intersect`
- [ ] `array_join`
- [ ] `array_max`
- [ ] `array_min`
- [ ] `array_position`
- [ ] `array_remove`
- [ ] `array_repeat`
- [ ] `array_size`
- [ ] `array_sort`
- [ ] `array_union`
- [ ] `arrays_overlap`

I'll continue to udpate this list.

Issue #543, opened 2023-05-31
Support lambda functions in node query


shangyian:
We should be able to use lambda functions when writing node queries. An example would be:

```
SELECT aggregate(array(1, 2, 3), 0, (acc, next) -> acc + next)
> 6
```

CircArgs:
@shangyian we can already parse lambdas with the new parser I believe. 
<img width="668" alt="Screenshot 2023-05-31 at 11 50 54 AM" src="https://github.com/DataJunction/dj/assets/35903677/2939af77-c51a-4758-8a7c-702d4e16bdfb">

shangyian:
@CircArgs So the parsing works, but then the type inference fails, which then causes node creation to fail. I think we need a custom compile function for `Lambda` that adds the types to `acc` and `next`, and this would have to implemented for each custom function (i.e., `aggregate` would be different from `array_sort` etc).

PR #545, opened 2023-05-31
Support lambda functions (ex: `aggregate`) and add several missing Spark functions

shangyian:
### Summary

This change adds the following Spark functions:
* `element_at`
* `split`
* `aggregate`
* `first`
* `cardinality`

Note that `aggregate` is a bit of a special case function that relates to https://github.com/DataJunction/dj/issues/543, since it takes in a lambda function as an argument. In order to support other functions that take lambdas (ex: `map_zip_with`, `map_filter`), we modify the way that `ast.Function` is compiled:
1. Compile the arguments of the function
2. Compile the lambda function expression. This is done with a custom method called `compile_lambda` that is defined on each such function's class. The main purpose of this function is to add types to the input arguments of the lambda expression so that we can infer the return type of the lambda by evaluating the expression.
3. Compile all other child elements of the function, if any.

### Test Plan

Added some unit tests. Tested with DJ server locally for a complex query with all of these functions used.

- [X] PR has an associated issue: #543, #544
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/tests/sql/functions_test.py
+++ b/tests/sql/functions_test.py
@@ -4,10 +4,12 @@ Tests for ``dj.sql.functions``.
 # pylint: disable=line-too-long
 
 import pytest
+from sqlmodel import Session
 
 import dj.sql.functions as F
 import dj.sql.parsing.types as ct
-from dj.errors import DJNotImplementedException
+from dj.sql.parsing.backends.antlr4 import parse
+from dj.errors import DJNotImplementedException, DJException
 from dj.sql.functions import (
     Avg,
     Coalesce,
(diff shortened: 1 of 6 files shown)


tests/sql/functions_test.py
```diff
@@ -256,3 +258,70 @@ def test_floor(types, expected) -> None:
...
+        when acc = '' then element_at(split(x, '::'), 1)
+        when acc = 'a' then acc
+        else element_at(split(x, '::'), 1) end)) as item
+    from (
+      select 1 as id, ARRAY('b', 'c', 'a', 'x', 'g', 'z') AS items
+    )
+    """,
+    )
+    exc = DJException()
+    ctx = ast.CompileContext(session=session, exception=exc)
```
samredai, line 280: Should these tests also include an assert that the `exc` variable doesn't contain any errors?

samredai (approved):
LGTM, just one comment about the tests.

agorajek (approved):
Nice one @shangyian !

tests/sql/functions_test.py
shangyian: Good point, added!

PR #546, opened 2023-06-01
orderby can be used on metrics

CircArgs:
### Summary

Currently, `orderby` from the api only supports dimension columns. This adds support to orderby any selected metrics as well.

### Test Plan

Add to a sql test some metrics in different positions in the `orderby`

- [x] `make check` passes

### DIFF
--- a/dj/construction/build.py
+++ b/dj/construction/build.py
@@ -757,18 +757,12 @@ def build_metric_nodes(
     combined_ast.select.projection.extend(dimension_columns)
 
     # go through the orderby items and make sure we put them in the order the user requested them in
-    orderby_sort_items_resorted = orderby_sort_items + [  # type: ignore
-        None,  # type: ignore
-    ] * len(orderby_mapping)
+
     for idx, sort_item in enumerate(orderby_mapping.values()):
         if isinstance(sort_item, ast.SortItem):
-            orderby_sort_items_resorted.insert(idx, sort_item)
-    orderby_sort_items_resorted = [
-        sort_item for sort_item in orderby_sort_items_resorted if sort_item is not None
-    ]
+            orderby_sort_items.insert(idx, sort_item)
 
-    combined_ast.organization = ast.Organization([])
-    combined_ast.organization.order = orderby_sort_items_resorted
+    combined_ast.organization = ast.Organization(orderby_sort_items)
 
     if limit is not None:
         combined_ast.limit = ast.Number(limit)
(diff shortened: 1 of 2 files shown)


dj/construction/build.py
```diff
@@ -708,10 +753,22 @@ def build_metric_nodes(
         else columns[0]
         for col_name, columns in dimension_grouping.items()
     ]
+
     combined_ast.select.projection.extend(dimension_columns)
 
-    if organization is not None:
-        combined_ast.organization = organization
+    # go through the orderby items and make sure we put them in the order the user requested them in
+    orderby_sort_items_resorted = orderby_sort_items + [  # type: ignore
```
shangyian, line 760: I might be missing something but why is `[None] * len(orderby_mapping)` needed? Won't the `orderby_sort_items_resorted.insert` further down take care of putting the element in the right location? And then the filtering out of `None` can also be removed

shangyian, line 760: I might be missing something but why is `[None] * len(orderby_mapping)` needed? Won't the `orderby_sort_items_resorted.insert` further down take care of putting the element in the right location? And then the filtering out of None can also be removed

dj/construction/build.py
```diff
@@ -708,10 +753,22 @@ def build_metric_nodes(
...
+        None,  # type: ignore
+    ] * len(orderby_mapping)
+    for idx, sort_item in enumerate(orderby_mapping.values()):
+        if isinstance(sort_item, ast.SortItem):
+            orderby_sort_items_resorted.insert(idx, sort_item)
+    orderby_sort_items_resorted = [
+        sort_item for sort_item in orderby_sort_items_resorted if sort_item is not None
+    ]
+
+    combined_ast.organization = ast.Organization([])
```
shangyian, line 770: could this just be a single line
```python
combined_ast.organization = ast.Organization(order=orderby_sort_items_resorted)
```

shangyian (approved):
Thanks for adding this!

dj/construction/build.py
CircArgs: Great point this looks really stupid! 🤣 For some reason I assumed that `insert` would raise if the index were out of bounds. Fixing it

CircArgs: Great point this looks really stupid! 🤣 For some reason I assumed that `insert` would raise if the index were out of bounds. Fixing it

samredai (approved):

agorajek (approved):
Nice. Pretty useful for metric ranked SQL.

PR #548, opened 2023-06-03
array_agg function

samredai:
### Summary

This adds type inference for the array_agg spark function.

```sql
SELECT array_agg(col) as x FROM (select 1 as col)
/* type of column x is inferred as ListType(IntegerType()) */
```
```sql
SELECT array_agg(col) as x FROM (select 'foo' as col)
/* type of column x is inferred as ListType(StringType()) */
```

### Test Plan

Added a `test_array_agg` test.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/functions.py
+++ b/dj/sql/functions.py
@@ -1280,13 +1280,13 @@ def infer_type(  # noqa: F811
     return ct.ListType(element_type=element_type)
 
 
-class Array_Agg(Function):  # pylint: disable=abstract-method,disable=invalid-name
+class ArrayAgg(Function):  # pylint: disable=abstract-method,disable=invalid-name
     """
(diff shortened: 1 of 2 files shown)


dj/sql/functions.py
```diff
@@ -1277,6 +1277,25 @@ def infer_type(  # noqa: F811
     return ct.ListType(element_type=element_type)
 
 
+class Array_Agg(Function):  # pylint: disable=abstract-method,disable=invalid-name
```
shangyian, line 1280: We can probably just use `ArrayAgg` for consistency (or switch everything else to be snake rather than camel case).

shangyian (approved):

PR #549, opened 2023-06-03
add array_append function

samredai:
### Summary

This adds type inference for the `array_append` function.

```sql
SELECT array_append(array('b', 'd', 'c', 'a'), 'd')
/* inferred as ListType(StringType()) */
```

```sql
SELECT array_append(array(1, 2, 3, 4), 5)
/* inferred as ListType(IntegerType()) */
```

```sql
SELECT array_append(array(true, false, true, true), false)
/* inferred as ListType(BooleanType()) */
```

### Test Plan

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/functions.py
+++ b/dj/sql/functions.py
@@ -1299,6 +1299,20 @@ def infer_type(  # noqa: F811
     return ct.ListType(element_type=element_type)
 
 
+class Array_Append(Function):  # pylint: disable=abstract-method,disable=invalid-name
+    """
+    Add the element at the end of the array passed as first argument
+    """
+
+
+@Array_Append.register  # type: ignore
+def infer_type(  # noqa: F811
+    array: ct.ListType,
+    item: ct.ColumnType,
+) -> ct.ListType:
+    return ct.ListType(element_type=item.type)
+
+
 class Map(Function):  # pylint: disable=abstract-method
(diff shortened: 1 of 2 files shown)


dj/sql/functions.py
```diff
@@ -1277,6 +1277,20 @@ def infer_type(  # noqa: F811
...
+    Add the element at the end of the array passed as first argument
+    """
+
+
+@Array_Append.register  # type: ignore
+def infer_type(  # noqa: F811
+    array: ct.ListType,
+    item: ct.ColumnType,
+) -> ct.ListType:
+    return ct.ListType(element_type=item.type)
```
CircArgs, line 1291: Is it up to the engine to raise if the array and item have different types

CircArgs (approved):
Just a question

dj/sql/functions.py
samredai: That's what I was thinking. I started to validate that here but realized it would be duplicating what the engine will do anyway. If this is used with an engine that supports multi-type arrays, that would raise a different issue with the type inference but I can't think of an engine that does allow that. Maybe that's just be a bridge we cross when we get to it.

shangyian (approved):

PR #550, opened 2023-06-03
Fixing class name for array_contains implementation

samredai:
### Summary

I noticed that this function was already implemented but the class was named `ArrayContains` instead of `Array_Contains` so it wasn't getting picked up by the function registry logic. This PR fixes that and tweaks the type inference logic a bit, plus adds a test for it.

### Test Plan

Added a `test_array_contains` test.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/dj/sql/functions.py
+++ b/dj/sql/functions.py
@@ -1206,7 +1206,9 @@ def infer_type(  # noqa: F811  # pragma: no cover
     return ct.ListType(element_type=arg.type)  # pragma: no cover
 
 
-class ArrayContains(Function):  # pragma: no cover  # pylint: disable=abstract-method
+class ArrayContains(
+    Function,
+):  # pragma: no cover  # pylint: disable=abstract-method
     """
     array_contains(array, value) - Returns true if the array contains the value.
     """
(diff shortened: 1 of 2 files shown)


CircArgs (approved):

dj/sql/functions.py
```diff
@@ -1206,15 +1206,18 @@ def infer_type(  # noqa: F811  # pragma: no cover
     return ct.ListType(element_type=arg.type)  # pragma: no cover
 
 
-class ArrayContains(Function):  # pragma: no cover  # pylint: disable=abstract-method
+class Array_Contains(
```
shangyian, line 1209: @samredai actually we already handle the snake case to underscores with this function at the end of `functions.py`:
```
function_registry = FunctionRegistryDict()
for cls in Function.__subclasses__():
    snake_cased = re.sub(r"(?<!^)(?=[A-Z])", "_", cls.__name__)
    function_registry[cls.__name__.upper()] = cls
    function_registry[snake_cased.upper()] = cls
```
I think we can either keep this logic and keep the class name `ArrayContains` or switch all functions class names match the function name exactly. The former looks nicer from a Python class naming perspective I think.

samredai: Oh nice! I must have just confused myself while I was running tests. Definitely think we should stick to the pythonic naming so I'll change this back.

samredai: Alright, fixed 👍

Issue #527, opened 2023-05-23
Add Parameters to Nodes


shangyian:
Per discussion in https://github.com/DataJunction/dj/discussions/421, we can add some support for basic parameters:
* Parameters should be defined at the node level
* Each parameter should be based on a single partition column on the node
* Start off by only allowing simple parameter conditions like `=`, `IN (<list>)` or `BETWEEN <a> AND <b>`. These would be specified by the user when defining a materialization config.

PR #555, opened 2023-06-04
Add partitions + parameters to materialization

shangyian:
### Summary

This change adds support for partitions and parameters in a node's materialization config, based on what was [discussed here](https://github.com/DataJunction/dj/discussions/421). 

For more context, when a user wants to schedule node materialization, they will add a materialization config to the node. However, users may only wish to materialize specific partitions of the node, so we add the concept of a `partition` to `MaterializationConfig` (functions equivalently to the previously discussed `parameters`):
- `name`: the column name,
- `type_`: either `temporal` or `categorical`
- `range`: a range of values like `[20200101, 20200201]`
- `values`: a list of values like ['a', 'b', 'c']
- `expression`: an expression temporal partitions that gets evaluated for scheduled jobs

This means that using the client to add a materialization config with partitions specified looks like:

```python
dimension_node.add_materialization_config(
    MaterializationConfig(
        engine=Engine(name="spark", version="3.3"),
        schedule="0 * * * *",
        config={
            "spark": {
                <spark conf>
            },
            "partitions": [
                {
                    "name": "date_int",
                    "type_": "temporal",
                    "range": [20200101, 20200201],
                    "values": [20210101, 20210204],
                },
                {
                    "name": "country",
                    "type_": "categorical",
                    "values": ["DE", "MY"],
                },
            ]
        }
    )
)
```

After the above call, DJ will save the config and ask the query service to schedule the necessary materialization jobs. The query service will be expected to schedule one or more jobs that do the following:
* **adhoc**: handles adhoc jobs for a configurable set of partitions
* **scheduled**: a reoccurring job that will handle ongoing materialization for the node

Both types of jobs, adhoc and scheduled, will respect the `categorical` partition values specified. In the example above, the `dimension_node` will only be materialized for `country` partitions `DE` and `MY` for both types of jobs. Then, for backfills, it will backfill the range specified by the `temporal` partition `date_int`, but ongoing jobs will continue to processing more recent temporal partitions.

### Test Plan

Ran things locally with a query service

- [X] PR has an associated issue: #527
- [X] `make check` passes


agorajek:
@shangyian thanks for starting this PR. Two clarifying questions for now (sorry I haven't looked at all the code yet):
1. I think that a partition inside the MatConfig should only have values or range defined. I know I specified all in https://github.com/DataJunction/dj/discussions/421 but I meant just one of them. What do you think?
2. For the ongoing job, we should have a way to define the current value that will go into any given partition, right?

shangyian:
> 1. I think that a partition inside the MatConfig should only have values or range defined. I know I specified all in https://github.com/DataJunction/dj/discussions/421 but I meant just one of them. What do you think?

@agorajek agreed, I think either-or makes sense. Although do you think `range` needs to be changed to allow for multiple ranges, like `[[20210101, 20210201], [20210501, 20210601]]`, or is the single range enough for now?

> 2. For the ongoing job, we should have a way to define the current value that will go into any given partition, right?

Yeah, we should. I'm not sure what the best way to do that would be -- currently I'm just relying on the internal implementation of the query (eventually materialization) service to decide the current value, which it derives based on the scheduler. I'll put out the WIP PR so you can take a look at that as well.

agorajek:
@shangyian based on our chat today there should be two implicit MaterializationConf types: **adhoc** and **scheduled** one.
- **scheduled** type must have a `schedule` defined and single `value/expression` for all temporal partitions, other partitions are optional
- **adhoc** type does not have a `schedule` defined and the rest of params optional

dj/api/nodes.py
```diff
@@ -413,19 +416,25 @@ def build_cube_config(
     }
     if config.get("druid"):
         return DruidCubeConfig(
-            node_name=cube_node.name,
             query=str(combined_ast),
             dimensions=list(dimensions_set),
             measures=measures,
             spark=SparkConf(__root__=config.get("spark", {})),
             druid=DruidConf(**config["druid"]),
+            upstream_tables=[
```
agorajek, line 424: Why do we need upstream tables in DruidCubeConfig?

dj/api/nodes.py
```diff
@@ -493,6 +520,34 @@ def upsert_a_materialization_config(
...
+                else []
+            ),
+            dimensions=[],
+            orderby=[],
+        )
+        generic_config = GenericMaterializationConfig(
+            query=str(materialization_ast),
+            spark=data.config["spark"] if "spark" in data.config else {},
+            partitions=data.config["partitions"] if "partitions" in data.config else [],
+            upstream_tables=[
```
agorajek, line 544: Oh, I think I understand. Perhaps we can use this to setup the dependencies for the jobs?

dj/models/node.py
```diff
@@ -722,6 +725,61 @@ class SparkConf(BaseSQLModel):
...
 
+class PartitionType(str, enum.Enum):
+    """
+    Partition type.
+
+    A partition can be temporal or categorical
+    """
+
+    TEMPORAL = "temporal"
+    CATEGORICAL = "categorical"
```
agorajek, line 736: +1

dj/models/node.py
```diff
@@ -722,6 +725,61 @@ class SparkConf(BaseSQLModel):
...
+          Partition(name=“dateint”, type="temporal", values=[20230601], range=(20220101, 20230101))
+
+        For categorical partition types, the ongoing materialization job will *only* operate on the
+        specified partition values in `values` and `range`:
+            Partition(name=“group_id”, type="categorical", values=["a", "b", "c"], range=())
+    """
+
+    name: str
+    values: Optional[List]
+    range: Optional[Tuple]
```
agorajek, line 764: `values` and `range` make sense... and perhaps for the single value that we gonna need for the temporal partition in **scheduled** runs we could add an `expression: Optional[str]` here?

dj/api/nodes.py
```diff
@@ -446,6 +455,24 @@ def materialization_job_from_engine(engine: Engine) -> MaterializationJob:
     return engine_to_job_mapping[engine.dialect]  # type: ignore
 
 
+def filters_from_partitions(partitions: List[Partition]):
+    """
+    Derive filters needed from partitions spec.
+    """
+    filters = []
+    for partition in partitions:
+        if partition.type_ != PartitionType.TEMPORAL:
```
agorajek, line 464: Just realized that we probably need to include this filter for temporal partitions as well. Say, if we have the query:
```
SELECT dateint, foo FROM bar
```
then we probably need to add this filter when we materialize:
```
SELECT dateint, foo FROM bar WHERE dateint = {expression}
```
This should be sufficient to set make query right even if other tables need to reference the same (current) date value, as long as we join them to the partition column.

dj/api/nodes.py
shangyian: Yep, it's mainly so that if the materialization service wants to have dependencies on the upstream tables, it can use this setting.

dj/api/nodes.py
shangyian: Agreed that this needs to be done at some point. 

Do you think it's reasonable if this is done at the materialization/query service level, where it can add on the temporal filter after it also has the parameterized version of `dateint`? That's what I have set up at the moment. 

The reason I ask is that if we generate this with the filter appended, it won't be able to be parameterized based on the scheduled run parameters for `expression`. Or were you thinking that it would be a parameterized value that the materialization service can then replace? (I tried that at one point but DJ's AST won't allow for generating of templatized queries).

dj/models/node.py
shangyian: Yep, that makes sense! I'll add that

dj/api/nodes.py
agorajek: > Do you think it's reasonable if this is done at the materialization/query service level, where it can add on the temporal filter after it also has the parameterized version of dateint?

Yes, I think we are on the same page. 

Wrt to materialization (or some generic query) parameters, I think we can try to get by w/o for the time being. Meaning that the special filters for temporal partitions will be tacked on during the materialization phase only. This would make the DJ queries cleaner imo. Ideally I'd like to stay away from some crazy SQL templating based on user notion of the "current" ot "execution" date. These concepts, if defined poorly, may lead to insane setup and misunderstanding of the run-date vs data-date. If we can handle that automatically it should be easier for the AE-users that DJ is targeted to.

PR #561, opened 2023-06-11
Merge DJQS and commit history into main DJ repo

samredai:
### Summary

This merges [djqs](https://github.com/DataJunction/djqs) into the main dj repo. This is part of a series of PRs to bring the OSS implementations of the customizable services into the main repo. All of the content as well as the commit history have been moved into a `/djqs` directory.


CircArgs (approved):

agorajek (approved):
I like this idea. Thanks for the execution.

shangyian:
Thanks for doing this @samredai! I've been thinking about this more and unfortunately we'll still have to have separate pydantic models unless we publish a shared library for models that can be imported by both dj core and djqs. This is an interesting setup I've been looking into: https://sqr-075.lsst.io/ Also cc @agorajek in case you have opinions

samredai:
@shangyian no problem! I also noticed that it's not as straightforward to combine the pydantic model definitions. I don't mind as much because it's nice to have DJQS as a self-contained representation of the query service API built from the ground up to match the DJQS spec. Having it use a shared model library would allow us to remove some duplication but at the cost of some of the query service code seeming "magical" and not obvious for someone inspecting it as an example, possibly to write a query service in an entirely different language. And if a custom query service isn't written in python, that also means the shared library isn't useful.

That article you linked is super interesting, especially this part:

Screenshot 2023-06-11 at 3 47 09 PM

We're not far off from that design except for how it describes the interface models for requests/response being in the client, and the server having the client as a dependency to then extend those models with any additional server side requirements. I've definitely been concerned about having the dj server and all of its dependencies dragged into the python client, but the other direction actually doesn't sound that bad--adding the python client as a dependency for the server. 🤔

PR #564, opened 2023-06-11
Output dimension attributes with type

shangyian:
### Summary

When we output a list of dimension attributes for a metric, we should also output their types. This provides better metadata for folks using the dimension attributes list, with information that we already maintain.

#### Example

Instead of returning a list of dimensions like:
```
[
    "default.dispatcher.company_name",
    "default.dispatcher.dispatcher_id",
    "default.dispatcher.phone",
    "default.hard_hat.address",
    "default.hard_hat.birth_date"
]
```
We return:
```
[
    {"name": "default.dispatcher.company_name", "type": "string"},
    {"name": "default.dispatcher.dispatcher_id", "type": "int"},
    {"name": "default.dispatcher.phone", "type": "string"},
    {"name": "default.hard_hat.address", "type": "string"},
    {"name": "default.hard_hat.birth_date", "type": "timestamp"}
]
```

cc @matthewwardrop

### Test Plan

Tested with the Python client and unit tests

- [X] PR has an associated issue: https://github.com/DataJunction/dj/issues/558
- [x] `make check` passes


samredai (approved):
Awesome, this is definitely super helpful.

PR #572, opened 2023-06-12
simplify PR template

samredai:
### Summary

I feel like most times I don't have anything meaningful to add to the `Test Plan` and `Deployment Plan` sections. Also, having to check the boxes that make check passes and make test passes seems redundant since GitHub already shows the green checkmark or red x on the PR page or in the PR list. What does everyone think of removing those parts in favor of a simpler PR template that asks for a summary and a link to an associated issue?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,14 +2,4 @@
 
 <!-- What's this change about? -->
 
-### Test Plan
-
-<!-- How did you test your change? -->
-
 - [ ] PR has an associated issue: #
-- [ ] `make check` passes
-- [ ] `make test` shows 100% unit test coverage
-
-### Deployment Plan
-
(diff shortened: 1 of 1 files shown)


agorajek:
I feel like checking checkboxes manually is silly, but both test and deployment plans are good parts of the template. You don't need to fill them in, but seeing them makes you think about them, which in turn will make you remember something that otherwise you might have forgotten.

shangyian:
I think the checkboxes aren't as relevant for those of us who have been contributing to the repo for a while, since we already know to run `make check` and `make test` and it's just part of the development flow. But for someone who is new, it's a nice reminder of what you need to do, so it kind of depends on who we're optimizing the PR template for. We can probably just ignore those checkboxes if we find them annoying...

samredai:
Alright, enough good points raised, closing this PR 😅

PR #573, opened 2023-06-13
Move `dj` to its own directory and rename

shangyian:
### Summary

What do we think about the following changes for better directory structure?
* Move `dj` to its own directory under this repo, so that it fits the same pattern of a typical monorepo (and is parallel in setup to the query and reflection services)
* Move each of these services into directories with names that match the Python package, so `dj` -> `datajunction-server`, `djqs` -> `datajunction-query`, and `djrs` -> `datajunction-reflection`

The directory structure would look like this: 
```
├── datajunction-server
│   ├── dj
│   │   ├── ...
│   ├── alembic
│   ├── tests
│   ├── pyproject.toml
├── datajunction-query
│   ├── djqs
│   │   ├── ...
│   ├── alembic
│   ├── tests
│   ├── pyproject.toml
├── datajunction-reflection
│   ├── djrs
│   │   ├── ...
│   ├── alembic
│   ├── tests
│   ├── pyproject.toml
```

### Test Plan

- [ ] `make check` passes


dj.db
samredai, line 1: Can we leave in the dj.db and djqs.db sqlite files? They're mainly meant to make it easier to try out DJ with a naive docker compose up by having the roads examples preloaded. I'm not sure if these would need to get moved into the subdirectories. 🤔

samredai (commented):
Hmm, I'm 50/50 on moving the main dj library into a subdirectory but maybe slightly leaning towards that? Something related I've been thinking about...I've always had a pet peeve with packages where the pypi package name was significantly different from the import name. If I `pip install foo` I want to `import foo`, or if there's a hyphen like `pip install foo-bar` I expect to use `import foo_bar`.

I know `from dj import x` is attractive, but since two letter packages aren't allowed in pypi, I'm starting to feel like we should just switch to `from datajunction_server import x` and be consistent with that pattern across `datajunction-query` and `datajunction-reflection` as well. The python client is already aligned with this--`pip install datajunction` -> `import datajunction`.

dj.db
shangyian: > Hmm, I'm 50/50 on moving the main dj library into a subdirectory but maybe slightly leaning towards that?
So part of the reason for this is that the structure makes more sense if it's a clear mapping between <folder> -- <package>, otherwise it seems like `djqs` and `djrs` are included in `datajunction-server`. The other thing I wanted to change is that the alembic migrations aren't included as a part of the library, meaning that running migrations cannot be done from just installing the library, so it would be nice to package that in.
 
> switch to from datajunction_server import x

Yeah, I think this is a good idea -- it's definitely confusing otherwise.

CircArgs:
This makes sense to me. What will this ultimately look like from a `pip install` perspective?

shangyian:
@CircArgs it would still just be `pip install datajunction-server`, and then for each of the other services, `pip install datajunction-reflection` and `pip install datajunction-query`. Only the directory structure changes

dj.db
samredai: Great points, I'm even more convince now that this is a better structure. 👍

samredai (approved):

PR #574, opened 2023-06-13
DJ UI: Top level page for building SQL and viewing sample data

samredai:
### Summary

This adds a "Cube Builder" page (all suggestions for a better name are welcome!). It's a top level page where you can select metrics and dimensions and see the SQL get generated. As metrics are selected, it uses the `/metrics/common/dimensions/` endpoint to filter the dimension list down to those shared by the metrics.

I've also added a `Run Query` button that will let you run the generated query and provides the data in a table below. I don't think we need to go all out and build this as an analytics BI tool (there are plenty out there that are feature rich and that DJ should eventually integrate with), but it's a good demo/dev mechanism to testing much of the functionality end-to-end. I'm planning to put a very low hard cap on the data table rendered, maybe 100 records, so it will serve as more of a mechanism for viewing a "sample" of the results of the query.

Curious to hear everyone's thoughts.

https://github.com/DataJunction/dj/assets/43911210/132fcf73-088c-4ac3-9c7f-2d6d1f80e568

### Test Plan

- [ ] `make check` passes


shangyian (approved):
This looks great! Thanks for adding

I do wonder if at some point we want to move towards using the JS client, since otherwise we'll just have a lot of parallel functionality in `DJService` 😅 No need to do anything for this PR obviously

I also think this will be useful, even if we're just using this to test DJ and plan on integrating a different UI for data exploration later.

samredai:
> I do wonder if at some point we want to move towards using the JS client, since otherwise we'll just have a lot of parallel functionality in `DJService` 😅 No need to do anything for this PR obviously

I kept thinking a lot about this and I totally agree. I was going to work on this but I figured it may be good to get the UI to a point where it's touching a larger surface area of the GET portion of API first (which is where I think we are now with this PR and your materialization one). My plan now is to match up the direct API calls to the javascript client and implement any missing features into the javascript client. Then I'll do an in-place swap and make sure there aren't any regressions.

shangyian (commented):
@samredai What do you think about using something like https://react-select.com/ for the selection of dimensions in this view? Just wondering because that's what the current node SQL tab uses and I think there would be less styling involved for that.

samredai:
@shangyian sounds good! I'll update this PR to use that 👍

samredai:
@shangyian I updated the PR to use react-form and updated the video in the PR description. After rebasing I also worked in some various cleanups including fixing an incorrect join in the demo notebook, renaming the djui directory to be datajunction-ui for consistency, and adding the docker compose `dj.demo.db -> dj.db` copy hack that I did for djqs.

Issue #528, opened 2023-05-23
Support for Materialization


shangyian:
Checklist:
- [x] Add `name` field to materializations. Based on discussion [here](https://github.com/DataJunction/dj/discussions/421#discussioncomment-5975195). A sensible default can be generated based on the config's schedule so that it's an optional field for users when creating.
- [x] Rename materialization models to be more intuitive
- [x] Allow updating of a materialization. Users can do so by specifying a node and the name of a materialization. Once updated, we'll push a new scheduled job using the materialization service. 
- [x] Allow deletion of a materialization
- [X] Every time a node is updated, we need to update all materializations and call the query service  
- [x] It should be possible to generate Druid spec with just partitions info (and thus not require it)
- [x] If a user modifies only the temporal partition range on an existing materialization, just update existing materialization and do not create a new one 
- [x] Fix availability state to handle partitions
- [ ] Docs

shangyian:
Added in https://github.com/DataJunction/dj/pull/555. Still need to do some more verification around updates.

PR #576, opened 2023-06-14
Add list materializations endpoint and refactor of models

shangyian:
### Summary

* Added an endpoint `GET /nodes/{node_name}/materializations/` that lists materializations and scheduling info (like urls, output tables etc) for a given node
* Renamed `MaterializationConfig` to `Materialization`. The former ended up being quite confusing since there is also a `config` field under the entity, and the latter is shorter as well.
* Moved all the models relevant to materialization to a separate `models/materialization.py` module, as `models/node.py` was getting very long

### Test Plan

- [X] PR has an associated issue: https://github.com/DataJunction/dj/issues/528
- [x] `make check` passes

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -630,7 +630,7 @@ class DJClient:  # pylint: disable=too-many-public-methods
                 f"Error retrieving data: {response.text}",
             )
 
-    def upsert_materialization_config(
+    def upsert_materialization(
         self,
         node_name: str,
         config: models.MaterializationConfig,
@@ -675,7 +675,7 @@ class Node(ClientEntity):
     availability: Optional[Dict]
     tags: Optional[List[models.Tag]]
     primary_key: Optional[List[str]]
-    materialization_configs: Optional[List[Dict[str, Any]]]
+    materializations: Optional[List[Dict[str, Any]]]
     version: Optional[str]
 
     def save(self, mode: models.NodeMode = models.NodeMode.PUBLISHED):
@@ -726,12 +726,12 @@ class Node(ClientEntity):
         self.sync()
         return link_response
 
-    def add_materialization_config(self, config: models.MaterializationConfig):
+    def add_materialization(self, config: models.MaterializationConfig):
         """
-        Adds a materialization config for the node. This will not work for source nodes
+        Adds a materialization for the node. This will not work for source nodes
         as they don't need to be materialized.
         """
-        upsert_response = self.dj_client.upsert_materialization_config(
+        upsert_response = self.dj_client.upsert_materialization(
             self.name,
             config,
         )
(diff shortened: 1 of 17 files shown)


dj/models/materialization.py
```diff
@@ -29,10 +41,198 @@ class DruidMaterializationInput(GenericMaterializationInput):
...
+    """
+
+    name: Optional[str]
+    engine: EngineInfo
+    config: Dict
+    schedule: str
+    job: str
+
+
+class MaterializationConfigInfoUnified(
```
samredai, line 66: nit: This class name was a bit scary when I saw it further up but looking at the definition it makes sense. Is there a more intuitive naming we can use here? I'm thinking `s/MaterializationInfo/MaterializationsReported` and `s/MaterializationConfigInfoUnified/MaterializationInfo`? My logic is that the first model represents which materializations are reported by the query service (which tables) and the second model represents all info we have for a materialization (it's config as well as any reported materializations).

dj/service_clients.py
```diff
@@ -143,24 +142,33 @@ def materialize(  # pylint: disable=too-many-arguments
...
-        Gets a list of materializations for the node and materialization config name.
+        Gets materialization info for the node and materialization config name.
         """
         response = self.requests_session.get(
             f"/materialization/{node_name}/{materialization_name}/",
+            timeout=3,
         )
-        return response.json()
+        if not response.ok:
+            return MaterializationInfo(output_tables=[], urls=[])
```
samredai, line 173: What does it mean when the other service reports multiple output tables? Is the idea that it will include all previous/expired materializations, for example previous days for a materialization that's recreated daily? Then one of those tables would be set as the `AvailabilityState` for a given node which would then be used during the SQL build/compile, is that all right?

shangyian: So the multiple output tables is actually to handle the Druid use case, where we would have both the output of the Spark query and the Druid datasource after it's ingested there. Arguably we don't need the former and can just have a single output, but I think it can be nice to have the list as a "we produced these artifacts" type of record. But yes, only one will get set as the availability state.

dj/models/materialization.py
shangyian: Good call on the naming! Will update once my aircraft gets wifi 🙃 I'll also go through this PR again and see if I can correct confusing bits of naming.

dj/service_clients.py
samredai: I see, that makes sense although I think it could be confusing when someone sets a druid materialization and sees both the name of a druid datasource and a spark table. Would there even be a way to tell that one is an intermediate spark table? As I'm writing this I'm starting to think the spark table is just an implementation detail of how the data gets to to druid which can vary wildly across different query services.

Maybe we should stick to maintaining the assumption that an output table for a given materialization config is always a table for the given engine+catalog combinations, the engine being selected in the materialization config and the catalog being the one for the node. That sounds like a guarantee that's up to the query service to provide so I don't think I'm suggesting any changes to this PR 😄

samredai (approved):

PR #578, opened 2023-06-14
add duckdb-engine to djqs

samredai:
### Summary

The reflection logic in djqs raises some errors without the `duckdb-engine` package. This PR adds it to the dependencies list.

This also includes a small fix to how the djqs/djqs.db metadata file is committed. There's no way to gitignore further modifications to an existing file. That means whenever you run a couple of samples, the demo djqs.db file gets updated and shows as an unstaged change in your local git repo. During dev that can be quite annoying since you'll have to always make sure you exclude this. So I moved djqs.db to a djqs.demo.db file that gets copied over to djqs.db during the docker compose startup. In summary, djqs.demo.db is never changed by the runtime so it won't be unintentionally placed into the unstaged git area, and djqs.db will be excluded and gitignored.

### Test Plan

`docker compose up --build`

- [ ] `make check` passes


shangyian (approved):
LGTM! Should the same thing be done for dj.db?

samredai:
I think we should but I figured I should wait for #573 to get in first.

PR #581, opened 2023-06-18
Fix orderby building

CircArgs:
### Summary

Fixes a bug where multiple metric builds were not building join key column tables.
Changes the AST so SetOp is now a linked list bound to a select. This also correct a hidden type error from the parser visitors.

### Test Plan

New multi metric test. All tests passing

- [x] `make check` passes


samredai:
very-nice

samredai (approved):

PR #582, opened 2023-06-18
Updating a node's query should trigger node materializations updates

shangyian:
### Summary

When a node is updated and its query changes, we should update its configured materializations and call the query service to reschedule all corresponding new materialization jobs. The existing behavior did not handle this case, meaning that updating a node's query would additionally require calling the upsert materialization endpoint in order to reschedule updated materialization jobs.

This PR also rolls in a few other more minor changes:
* Add Pydantic schema on the `config` field in a `Materialization`, so that it's not just an arbitrary dict but actually has a set of expected schemas (these differ depending on the materialization type)
* Remove the abillity to set materializations on cube node creation. This was originally added in only for cube nodes, which is inconsistent with the creation behavior for all nodes. It's also unwieldy having the materialization creation logic be available through multiple paths.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -65,8 +65,8 @@ from datajunction_server.models.materialization import (
     Partition,
     PartitionType,
     SparkConf,
-    UpsertCubeMaterializationConfig,
-    UpsertMaterializationConfig,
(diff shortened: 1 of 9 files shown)


samredai (approved):
> It's also unwieldy having the materialization creation logic be available through multiple paths.

Totally agree 🚀

Issue #583, opened 2023-06-19
Retrieving query results from the datajunction-query service without including the data


samredai:
Currently when you submit a DJ query (metrics+dimensions+filters/order by/limit), once the query is completed it always includes the data. Even if you submit it async, as you continuously poll, the response will eventually come back with a `"status": "FINISHED"` and include the data.

In some cases it would be useful to submit the query (either sync or async) and simply show the status and various details of the query execution. Once the query is completed, an explicit request (i.e. a click on a UI) to view the data can then re-request the query with the data included. There are a few situations but the most obvious one is probably when a user wants to select a number of different metric+dimension combinations and quickly submit queries for all of them. Instead of returning all of the data for all of the queries at once, it would be more useful if they can see the status of all of them together. As they complete, the user selects something like "View Results" for each query to then pull the results for them to see.

samredai:
My proposal is to add an `include_data=true|false` parameter to the `/data/` endpoint. That's then passed to the query service which will appropriately return the query information with or without including the data.

samredai:
Instead of including this flag, the UI is set to display 10 records by default with a selector to increase that to 100 or 1000. Also query information is now displayed after a query runs so links to another system/UI can be displayed there.

PR #585, opened 2023-06-19
show query info in UI and allow settings display rows

samredai:
### Summary

I've updated this PR after some feedback. Now the changes are only located in the UI:

**datajunction-ui**
- The SQL page recently added in #574 now also includes query information after clicking the "Run Query" button
- The results show 10 records by default with a dropdown that lets you increase that to 100 or 1000 (the total number of rows is included in the query information displayed)
- Instead of re-generating the SQL on every select, it now stages the metrics and dimensions as the drop-down is open and selections are being made. When `onMenuClose` is triggered, the SQL is generated once as opposed to on every individual selection.

https://github.com/DataJunction/dj/assets/43911210/c183df72-9069-4b80-bdbf-11a94251096d

### Test Plan

- [x] PR has an associated issue: #583 
- [x] `make check` passes

### DIFF
--- a/datajunction-ui/webpack.config.js
+++ b/datajunction-ui/webpack.config.js
@@ -1,7 +1,9 @@
 const webpack = require('webpack');
+const dotenv = require('dotenv').config()
 const path = require('path');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 // const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+require('dotenv').config({ path: './.env' }); 
 
 var babelOptions = {
   presets: ['@babel/preset-react'],
@@ -100,7 +102,7 @@ module.exports = {
       template: path.resolve(__dirname, 'public', 'index.html'),
     }),
     new webpack.DefinePlugin({
-      REACT_APP_DJ_URL: 'http://localhost:8000',
+      "process.env": JSON.stringify(process.env)
     }),
     // new MiniCssExtractPlugin({
     //   filename: './styles/index.css',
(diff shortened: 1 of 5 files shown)


samredai:
Got some good feedback offline. Instead of allowing enabling/disabling data viewing, I'll simply enforce a UI view limit of a 100 record sample. The query service can of course enforce any kind of limits. Over time we can increase this view limit as we add things to the UI like pagination, column filtering, sorting, etc.

samredai:
Updated the PR, much smaller now 😄 . @shangyian and @agorajek would you mind taking another look when you have a chance?

datajunction-ui/webpack.config.js
```diff
@@ -100,7 +102,7 @@ module.exports = {
       template: path.resolve(__dirname, 'public', 'index.html'),
     }),
     new webpack.DefinePlugin({
-      REACT_APP_DJ_URL: 'http://localhost:8000',
+      "process.env": JSON.stringify(process.env)
     }),
     // new MiniCssExtractPlugin({
```
agorajek, line 107: nit: why so many commented out lines in this file?

agorajek (approved):
Code makes sense to me but I didn't try to locally.

The PR's title should be updated now and I have one nit in-line.

shangyian (approved):
LGTM!

datajunction-ui/webpack.config.js
samredai: Updated this file to remove the commented lines. @shangyian I'll follow your lead on the UI testing suite. I started to get familiar with the component testing setup that you have there so maybe I'll try adding those for these new pages 👀

PR #586, opened 2023-06-20
Add support for partition-level availability states + merging states

shangyian:
### Summary

This PR has two main changes:
1. Adds support for partition-level availability states. This is done through an additional `partitions` field on `AvailabilityState`, which will track the available temporal min-max ranges for categorical partitions. The top-level `min_partition` and `max_partition` will be used to track the availability state of all partitions for the node.
2. Adds support for merging partition-level availability states. When someone adds a new availability state for a node, we check against the existing availability state and merge via this logic:

    * Merge any partition-level availabilities so that the range for each partition is the widest possible
    * Merge any node-level availabilities so that the range is the widest possible
    * Prune the partition-level availabilities so that only the ones still relevant in addition to the node-level availabilities are kept

### Test Plan

- [X] PR has an associated issue: #528 
- [X] `make check` passes


datajunction-server/datajunction_server/api/data.py
```diff
@@ -78,23 +78,12 @@ def add_an_availability_state(
...
-            (
-                node_revision.availability.min_partition,
-                data.min_partition,
-            ),
-        )
+        data.merge(node_revision.availability)
 
-    db_new_availability = AvailabilityState.from_orm(data)
-    node_revision.availability = db_new_availability
+        # Update the node with the new availability state
```
agorajek, line 83: The indent of this comment looks off.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -234,8 +282,74 @@ class AvailabilityStateBase(BaseSQLModel):
...
     valid_through_ts: int
-    max_partition: List[str] = Field(sa_column=SqlaColumn(JSON))
-    min_partition: List[str] = Field(sa_column=SqlaColumn(JSON))
+
+    # Node-level temporal ranges
+    min_partition: Optional[List[str]] = Field(sa_column=SqlaColumn(JSON))
+    max_partition: Optional[List[str]] = Field(sa_column=SqlaColumn(JSON))
+
+    # Specific partitions and their temporal ranges.
+    partitions: Optional[List[AvailablePartition]] = Field(
```
agorajek, line 291: I think you are assuming that any given node can only have a single non-temporal partition key. Is that correct or am I missing something?

agorajek (commented):
One big question in-line wrt to how many non-temporal partitions we want to allow on nodes.

shangyian (commented):
@agorajek and I chatted offline about this and settled on the following approach.

At the node level, we'll store the following optional fields:
* `temporal_partitions`: an ordered list of temporal partitions like `["date", "hour"]` or `["date"]`
* `categorical_partitions`: an ordered list of categorical partitions like `["country", "group_id"]` or `["region_id", "age_group"]`

Assuming node-level values of `temporal_partitions=["date"]` and `categorical_partitions=["country", "group_id"]`, an `AvailabilityState` object will look like this:
```
{
    "catalog": "test",
    "schema": "dj",
    "table": "materialized_table",
    "valid_through_ts": 123,
    "min_partition": ["20230101"],
    "max_partition": ["20230110"],
    "partitions": [
        {
            "value": ["DE", null],
            "min_partition": ["20230101"],
            "max_partition": ["20230101"],
            "valid_through_ts": 123
        }
    ]
}
```

Each entry in `partitions` will have a list of partition values under the `value` field. This list maps to the ordered list of categorical partitions at the node level. So in the example above: `["DE", null] <-> ["country", "group_id"]`. In any case where the partition value is `null`, the `null` acts like a wildcard and matches all partitions. The `min_partition` and `max_partition` are also a list of partition values, and their ordering maps to the ordered list of node-level temporal partitions.

The logic to merge will include the ability to remove entries in `partitions` that are encompassed by other entries, essentially merging them to the smallest list of entries that covers the widest range.

shangyian:
Cases:
| Merger | Mergee | Result | Explanation | 
|--------|--------|--------|-------------|
| <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table>  |  <table>  <tr>  <th>Value</th>  <td>`["MY", "abc-123"]`</td> </tr> <tr><th>Min</th>  <td>2021-02-01</td></tr>  <tr>  <th>Max</th><td>2021-02-10</td>  </tr> </table> | <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table>  <table>  <tr>  <th>Value</th>  <td>`["MY", "abc-123"]`</td> </tr> <tr><th>Min</th>  <td>2021-02-01</td></tr>  <tr>  <th>Max</th><td>2021-02-10</td>  </tr> </table> | No overlap between two entries, so keep both |
| <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table> | <table>  <tr>  <th>Value</th>  <td>`["DE", "abc-123"]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table> | <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table> | Former's min/max range is wider than (or equal to) the latter, so keep the former. |
| <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-03</td></tr>  <tr>  <th>Max</th><td>2021-01-04</td>  </tr> </table> | <table>  <tr>  <th>Value</th>  <td>`["DE", "abc-123"]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table> | <table>  <tr>  <th>Value</th>  <td>`["DE", null]`</td> </tr> <tr><th>Min</th>  <td>2021-01-03</td></tr>  <tr>  <th>Max</th><td>2021-01-04</td>  </tr> </table> <table>  <tr>  <th>Value</th>  <td>`["DE", "abc-123"]`</td> </tr> <tr><th>Min</th>  <td>2021-01-01</td></tr>  <tr>  <th>Max</th><td>2021-01-10</td>  </tr> </table> | Former's min/max range is smaller than the latter, so keep both. |

The logic for merging partition availability:
* Check if the partition availabilities overlap using the items in the `value` field.
* If they overlap, we check `min_partition` and `max_partition` to see which partition availability is wider.
* If the broader categorical partition availability (i.e., the one with `null` vs specific values) has a wider temporal range than the narrower categorical one, just keep the broader one.
* If the broader categorical partition availability has a smaller temporal range than the narrower categorical one, keep both.

I will probably refine this as I write code...

agorajek:
@shangyian this is great, full agreement on the logic. And thanks for the examples.

One comment on naming. Your AvailabilityState example looks good but I wonder if it would be even better if we were more explicit. What do you (and others) think about this:

```
{
    "catalog": "test",
    "schema": "dj",
    "table": "materialized_table",
    "valid_through_ts": 123,
    "min_temporal_partition": ["20230101"],
    "max_temporal_partition": ["20230110"],
    "categorical_partitions": [
        {
            "value": ["DE", null],
            "min_temporal_partition": ["20230101"],
            "max_temporal_partition": ["20230101"],
            "valid_through_ts": 123
        }
    ]
}
```

Proposed changes:
- min|max_partition --> min|max_temporal_partition
- partitions --> categorical_partitions
- also... wondering if we would make things clearer if we said `valid_through_unixtime` instead of `valid_through_ts`.

shangyian:
@agorajek Those naming changes make sense to me! I like the idea of being more explicit.

shangyian:
@agorajek I've updated the PR to switch to the approach that we discussed. One deviation is that on the naming front I kept `partitions`, because I added `categorical_partitions` and `temporal_partitions` as fields that contain a list of the actual partition fields, i.e., `["country", "group_id"]` or `["date", "hour"]`.

The full logic for availability state merging is a bit complicated - I ended up building a trie to keep track of the different partition levels and then merged where appropriate. The bulk of that logic is in the `AvailabilityTracker`.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -234,8 +360,96 @@ class AvailabilityStateBase(BaseSQLModel):
...
+        for partition in all_partitions:
+            tracker.insert(partition)
+        final_partitions = tracker.get_partition_range()
+
+        self.partitions = [
+            partition
+            for partition in final_partitions
+            if not all(val is None for val in partition.value)
+        ]
+        merged_top_level = [
```
agorajek, line 437: Will this work as well if we only have temporal partitions on a nod and no categorical ones?

agorajek (approved):
This makes sense @shangyian . Thank you for this PR. Just one question in line.

docker-compose.yml
```diff
@@ -12,6 +12,8 @@ services:
   dj:
     container_name: dj
     stdin_open: true
+    env_file:
+      - ./datajunction-server/.docker-env/.env
```
samredai, line 16: @shangyian this .env file was moved up one directory to `./datajunction-server/.env`. Also since it's in the root of the build context directory it should just get picked up automatically without this line.

shangyian: Oh good point! Let me remove this in a new PR

Issue #592, opened 2023-06-22
Listing available dimensions should include information on how the dimension is linked in


shangyian:
Let's assume the following dimensions:
```python
{
    "name": "default.date_dim",
    "description": "Date dimension",
    "query": """
        SELECT
            dateint,
            month,
            year,
            day
        FROM default.date
    """,
    "mode": "published",
    "primary_key": ["dateint"]
}
```

```python
{
    "name": "default.hard_hat",
    "description": "Hard hat dimension",
    "query": """
                SELECT
                  hard_hat_id,
                  last_name,
                  first_name,
                  title,
                  birth_date,
                  hire_date,
                  ...
                FROM default.hard_hats
            """,
    "mode": "published",
    "primary_key": ["hard_hat_id"]
}
```

We can link both `birth_date` and `hire_date` on `default.hard_hat` to the `default.date_dim` dimension. When we retrieve dimensions on a downstream metric, however, we need a way to distinguish between the `birth_date` `default.date_dim` and the `hire_date` `default.date_dim`. One possibility is that in the list of dimensions returned, in cases of multiple column <-> dimension links on a node, we should prefix the dimension with the column: `birth_date.default.date_dim` and `hire_date.default.date_dim`.

PR #594, opened 2023-06-22
Include link column to distinguish cases where dimension is linked to multiple columns on a node

shangyian:
### Summary

When the same dimension is linked to multiple columns on a node (for example, when a `date` dimension is linked to both `birth_date` and `hire_date`), we should be able to distinguish between the two types of links using the column link when we retrieve dimension attributes for the node or any downstream nodes. 

This PR modifies the output of `get_dimensions(...)` to include a `path` field on the dimension attribute output. This field contains the linked columns in the join path between the dimension attribute and the node, like this:
```
[
    {
        "name": "default.date_dim.dateint",
        "path": ["default.user_dim.birth_country", "default.special_country_dim.formation_date"],
        "type": "int",
    },
    {
        "name": "default.date_dim.dateint",
        "path": ["default.user_dim.birth_country", "default.special_country_dim.last_election_date"],
        "type": "int",
    },
    {
        "name": "default.date_dim.dateint",
        "path": ["default.user_dim.residence_country", "default.special_country_dim.formation_date"],
        "type": "int",
    },
    {
        "name": "default.date_dim.dateint",
        "path": ["default.user_dim.residence_country", "default.special_country_dim.last_election_date"],
        "type": "int",
    },
    ...
]
```

It also adds the path to the various `show dimensions` dropdowns in the UI:

Screenshot 2023-06-23 at 8 29 31 AM

### Test Plan

Unit tests

- [X] PR has an associated issue: https://github.com/DataJunction/dj/issues/592
- [x] `make check` passes


agorajek (approved):
This is great! I think it will provide a lot of clarity to the users. 

Nit: `link` is pretty vague so I'd suggest using `link_column` or `linked_column` for the new DimensionAttributeOutput attribute.

samredai (approved):
+1 to the comment by @agorajek, I'm wondering if we should include the word link at all in the naming since the higher level construct is itself a "dimension link". Is there another name that works? Maybe something like "root" where we would say the higher level thing is a dimension link and the root is the column used to join to the dimension node? Not sure if that's intuitive...

But yeah this PR is great, it looks awesome in the UI!

shangyian:
@agorajek and @samredai Hmm, what about `role` or `parent` for the name?

agorajek:
I think `link(ed)_column` is the best and it fits the context of a Dimension Link imo.

samredai:
> @agorajek and @samredai Hmm, what about `role` or `parent` for the name? 

Both of those sounds good! Maybe role is better since parent is used elsewhere to describe the DJ DAG. I can see role being eventually used in auth stuff as well but hopefully they never overlap in the source code. Another term I think I've heard is "path" which does nicely match the fact that we're describing a specific join path to the dimension node. Any of those sound pretty good to me. 👍

shangyian:
Going with `path` as there are some scenarios where we may need to provide a join path. For example, there could be a second-order dimension that also has two columns linked to the same node. Then in order to disambiguate, we'd need the path.

shangyian:
Had some more offline discussion with @agorajek and @samredai. We've concluded that `path` needs to be a list of column links to get to a specific dimension attribute.

Here's a specific example of a scenario with tracking such a join path may be relevant.

**Dimensions:**

| Name | Type | Columns | Primary Key |
|--------|--------| ------ | ------------ |
| **`country_dim`** | dimension | <table><tr><td>`country_code`</td></tr> <tr><td>`name`</td></tr> <tr><td>`formation_date`</td></tr> <tr><td>`last_election_date`</td></tr></table> | `country_code` |
| **`date_dim`** | dimension | <table><tr><td>`date`</td></tr> <tr><td>`year`</td></tr> <tr><td>`month`</td></tr> <tr><td>`day`</td></tr></table> | `date` |
| **`users_dim`** | dimension | <table><tr><td>`user_id`</td></tr> <tr><td>`birth_country`</td></tr> <tr><td>`residence_country`</td></tr> <tr><td>`age`</td></tr></table> | `user_id` |

**Metrics:**

| Name | Type | Query | 
|--------|--------| ------ | 
| **`avg_user_age`** | metric | `SELECT AVG(age) FROM users_dim` |

**Dimension Links**

* On `users_dim`, link `birth_country` to `country_dim`
* On `users_dim`, link `residence_country` to `country_dim`
* On `country_dim`, link `last_election_date` to `date_dim`
* On `country_dim`, link `formation_date` to `date_dim`

**Dimensions (with paths)**

When we ask for all available dimension attributes for the `avg_user_age` metric, they should include the following `date_dim` attributes, which can only be distinguished from each other if the join path is provided:
* `date_dim.date` <--> `country_dim.formation_date` <--> `users_dim.birth_country`
* `date_dim.date` <--> `country_dim.last_election_date` <--> `users_dim.birth_country`
* `date_dim.date` <--> `country_dim.formation_date` <--> `users_dim.residence_country`
* `date_dim.date` <--> `country_dim.last_election_date` <--> `users_dim.residence_country`
* `date_dim.year` <--> `country_dim.formation_date` <--> `users_dim.birth_country`
* `date_dim.year` <--> `country_dim.last_election_date` <--> `users_dim.birth_country`
* `date_dim.year` <--> `country_dim.formation_date` <--> `users_dim.residence_country`
* `date_dim.year` <--> `country_dim.last_election_date` <--> `users_dim.residence_country`
* `date_dim.month` ... etc

agorajek:
I am glad we resolved it @shangyian . This looks good now.

PR #595, opened 2023-06-23
Fix OpenAPI spec generation action

shangyian:
### Summary

Now that we've moved the relevant script to live under `datajunction-server/scripts/generate-openapi.py`, we need to change this action to reference the right location to run the script.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,8 +12,6 @@ services:
   dj:
     container_name: dj
     stdin_open: true
-    env_file:
-      - ./datajunction-server/.docker-env/.env
     tty: true
     networks:
       - core
@@ -35,8 +33,6 @@ services:
 
   db_migration:
     container_name: db_migration
-    env_file:
-      - ./datajunction-server/.docker-env/.env
     networks:
       - core
     build:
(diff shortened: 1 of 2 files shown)


samredai:
Thanks for catching this. Would it be better to make the change below so that the spec is still placed in the root directory?

```diff
      - name: Generate OpenAPI Spec
-        run: ./scripts/generate-openapi.py -o openapi.json
+        run: ./scripts/generate-openapi.py -o ../openapi.json
```

samredai (approved):

shangyian:
@samredai I noticed that this is used later to generate the markdown file in the docs. I guess that could make sense as a part of this step too so that it's entirely automated.
> `widdershins openapi.json -o docs/content/0.1.0/docs/developers/the-datajunction-api-specification.md --code=true --omitBody=true --summary=true`

samredai:
> @samredai I noticed that this is used later to generate the markdown file in the docs. I guess that could make sense as a part of this step too so that it's entirely automated. What do you think?
> 
> > `widdershins openapi.json -o docs/content/0.1.0/docs/developers/the-datajunction-api-specification.md --code=true --omitBody=true --summary=true`

Great idea, that makes sense to me!

docker-compose.yml
```diff
@@ -12,8 +12,6 @@ services:
   dj:
     container_name: dj
     stdin_open: true
-    env_file:
```
shangyian, line 15: @samredai removing it here

PR #597, opened 2023-06-25
Add dimension links to node graph UI

shangyian:
### Summary

This PR adds dimensions links to the node graph tab in the UI:
* In this iteration, the dimension link is more precise -- we now point an arrow from a specific column to a dimension, instead of just linking at the node level. This makes the functionality of a dimension attribute's `path` very clear visually.
* Changed `Show Columns` toggle for non-metric nodes into `More Columns`. By default we display at most 10 columns per node, and the remaining can be requested via this toggle.
* The list of dimension attributes for metric nodes is now grouped by dimension node. Hopefully this makes easier to navigate, as instead of a long list of dimensions, there will be some organization. Let me know if this is a good idea!
* Added a `/nodes/{node_name}/dag` endpoint to the backend, which allows the frontend to make a single API call that retrieves all of a node's upstreams, downstreams, and dimensions.
* Also did some refactoring to try to split up the graph node renderer + clean up the styling

https://github.com/DataJunction/dj/assets/9524628/d41f0e2d-263f-4bbc-8e5e-d098d1cbb1d0

### Test Plan

- [ ] `make check` passes


samredai (approved):
Woah, I feel like this PR just takes the UI to a whole new level. 🚀  (all in under 500 lines 🤯 )

One minor feedback is that the dimension column name is a bit hard to see in on the metric node. Maybe the text needs to be a bit darker or the background a bit lighter?
Screenshot 2023-06-26 at 12 08 13 AM

shangyian:
@samredai I'm trying to make a series of changes that will make the demo this week easier 😄 

Good point about the dimension column display -- just updated that to:

Screenshot 2023-06-26 at 7 34 39 AM

PR #598, opened 2023-06-26
On the node page for cubes, display the cube elements

shangyian:
### Summary

This change displays the cube elements (metrics + dimensions) on the cube node info page and hides the node query (since it's a generated query and mostly irrelevant to users).

Screenshot 2023-06-26 at 8 56 03 AM

### Test Plan

- [x] `make check` passes


samredai (approved):
Looks great, although if we could maybe have the compiled SQL somewhere inconspicuous like behind a toggle or collapsed section, that might be better. I can imagine people who know a little bit more about the internals will want to get a peek at the generated SQL for the cube. 🔍  (not urgent enough to update this PR so feel free to worry about that in a follow-up later)

shangyian:
@samredai good point! I might save that for later and figure out if there's a good way of display the auto-determined measures in each cube as well.

PR #599, opened 2023-06-26
Client code popover for node creation, dimension linking, and add materialization

shangyian:
### Summary

This change adds a popover (controlled via a toggle) that displays the Python client code for various actions:
* Creating a node
* Linking a dimension to a column on a node
* Adding materialization to a node

This may be useful in several ways:
* Users can use the client code as a template to perform one of these actions (stand in for before we have an actual creation/modification UI)
* Users can copy a particular node, make minor changes, and create a new node
* Demos (so that we don't have to move to a separate notebook window)

https://github.com/DataJunction/dj/assets/9524628/fa879cb4-b5ad-45ac-bf15-837baa7dbe4e

I'm not 100% sure about having the client code generation live in the backend, but it does work for now.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/client.py
+++ b/datajunction-server/datajunction_server/api/client.py
@@ -1,0 +1,168 @@
+"""
+APIs related to generating client code used for performing various actions in DJ.
+"""
+
+import json
+import logging
+
+from fastapi import APIRouter, Depends
+from sqlmodel import Session
+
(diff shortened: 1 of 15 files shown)


agorajek (approved):
@shangyian I think this addition is great and I don't think we need to worry about its testing for now. I understand the concern of making sure the code works but I think the value of seeing anything like hat is much bigger than a potential api change or typo.

PR #601, opened 2023-06-30
remove duplicates when getting common dimensions

shangyian:
### Summary

This PR fixes a bug where we get duplicates for every dimension attribute when retrieving common dimensions across multiple metric nodes. It was introduced at some point with https://github.com/DataJunction/dj/pull/594

### Test Plan

- [x] `make check` passes


agorajek (approved):
Everything looks good but how did the deduplication cause `test_client.py` to change is a bit confusing.

shangyian:
@agorajek That threw me for a loop as well... let me go into more detail.

This bug exposed something else about the data modeling we have around the Roads database that's slightly off (or it could be that the way we're representing `paths` isn't enough for dimension links, and we do need something akin to `roles`). It's a little bit confusing because you would think that all nodes in the Roads example should share all dimensions, but when we dig into the way they're linked together, it's actually not the case. For example:

The `default.num_repair_orders` metric has this query:
```
SELECT count(repair_order_id) FROM default.repair_orders
```
The upstream node it queries from, `default.repair_orders`, is linked to several dimensions, including `default.municipality_dim`, `default.hard_hat`, and `default.dispatcher` by the respective `ids` on `default.repair_orders`.

The `default.avg_repair_price` metric has this query:
```
SELECT avg(price) as default_DOT_avg_repair_price 
FROM default.repair_order_details
```
The upstream node it queries from, `default.repair_order_details`, is linked to the `default.repair_order` dimension, which *in turn* is linked to the `default.dispatcher`, `default.hard_hat`, and `default.municipality_dim` dimensions. However, because the path to each of those dimensions is different (i.e., for `default.avg_repair_price` they come from the path `default.repair_order_details` -> `default.repair_order`, whereas for `default.num_repair_orders` they come from the path `default.repair_orders`), they aren't considered to share dimensions. 

This doesn't really make sense for the nodes we're trying to model (since intuitively they should share all dimensions), but I think that's a problem with the way we're modeling the data and not a problem with the proposed logic around paths. For example, if we changed `default.num_repair_orders` to query the `default.repair_order` dimension rather than the `default.repair_orders` source node, they would share all dimensions. What do you think?

agorajek:
Thank you @shangyian. Your explanation does make sense. And I think that the concept of a `role` in the form of additional attribute on the tagged column makes sense. I think we still need to review this design, especially when wrt rules of the role assignment on nodes connected to the dimension node through different paths. But let's talk about this separately.

As far as the question of shared dimensions VS their paths. I have a feeling that (if there were no roles) we could simply consider to different dimension to be shared if their paths overlap at the end of the "road". Say, if metric `m1` connects with dimension `d` through nodes: `m1 -> t1 -> t2 -> d` and metric `m2` connects to the same dimension over `m2 -> t2 -> d` then I think (by default) we should considered dimensions `d` shared among metrics `m1` and `m2`. I think this shared end of the path simply means that we have a way to join these two metrics together over that dimension. 

And wrt the dimensions roles, I think that (if defined) they could simply be an additional required attribute to match on the shared dimensions.

shangyian:
@agorajek That's a good point about the overlap of the tail end of the dimension paths. I haven't thought of any edge cases that doesn't cover so I'll go ahead and change the logic over to that! 

And agreed that we need to flesh out dimension roles more before any implementation

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -80,29 +80,60 @@ def get_dimensions(
     return sorted(list(processed), key=lambda x: x.name)
 
 
-def get_shared_dimensions(
-    metric_nodes: List[Node],
-) -> List[DimensionAttributeOutput]:
+def check_convergence(path1: List[str], path2: List[str]) -> bool:
     """
-    Return a list of dimensions that are common between the nodes.
+    Determines whether two join paths converge before we reach the
```
agorajek, line 85: Yeah, I think this is great for now. I can see some folks asking us why two metrics can't be pulled together if their dim paths don't overlap, but I think we can think of that use case later.

datajunction-server/tests/examples.py
agorajek, line 1: Some leftover comments in this file.

datajunction-server/datajunction_server/sql/dag.py
shangyian: Yeah, we can add some clearer data modeling examples in the docs, which will also help us clarify the cases we need to support. I'll try to do that at some point

PR #603, opened 2023-07-03
Add Common Issues to the contribution guide.

agorajek:
### Summary

What the title says.

### Test Plan

In the wash. Not sure how to render these `.rst` files locally.

- [ ] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -391,3 +391,35 @@ Creating a Diagram from the Grammar
 Use https://bottlecaps.de/convert/ to go from ANTLR4 -> EBNF
 
 Input the EBNF into https://bottlecaps.de/rr/ui
+
+Common Issues
+===================
+
+Alembic migration error
+----------------------------------------
+
+If during development your alembic migrations get into a spot where the automatic upgrade (or downgrade) is stuck you may see something
+similar to the following output in your `db_migration` agent log:
+
+.. code-block:: sh
+
+  db_migration  | sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) duplicate column name: categorical_partitions
+  db_migration  | [SQL: ALTER TABLE availabilitystate ADD COLUMN categorical_partitions JSON]
+  db_migration  | (Background on this error at: https://sqlalche.me/e/14/e3q8)
+  db_migration exited with code 1
+
+The easiest way to fix it is to reset your database state using these commands (in another terminal session):
+
+.. code-block:: sh
+
+  $ docker exec -it dj bash
+  root@...:/code#
+
+  root@...:/code# alembic downgrade base
+  ...
+  INFO  [alembic.runtime.migration] Running downgrade e41c021c19a6 -> , Initial migration
+
+  root@...:/code# alembic upgrade head
+  ...
+
+After this, the `docker compose up` command should start the db_migration agent without problems.


shangyian (approved):
Nice!

CONTRIBUTING.rst
```diff
@@ -391,3 +391,39 @@ Creating a Diagram from the Grammar
...
+
+Alembic migration error
+----------------------------------------
+
+If during development your alembic migrations get into a spot where the automatic upgrade (or downgrade) is stuck you may see something
+similar in your `db_migration` agent log:
+
+.. code-block:: sh
+
+  db_migration  | sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) duplicate column name: categorical_partitions
```
shangyian, line 406: Hmm, this also makes me wonder if we should do another alembic migrations squash...

CONTRIBUTING.rst
```diff
@@ -391,3 +391,39 @@ Creating a Diagram from the Grammar
...
 Input the EBNF into https://bottlecaps.de/rr/ui
+
+Common Issues
+===================
+
+Alembic migration error
+----------------------------------------
+
+If during development your alembic migrations get into a spot where the automatic upgrade (or downgrade) is stuck you may see something
+similar in your `db_migration` agent log:
```
samredai, line 402:
```suggestion
similar to the following in your `db_migration` agent log:
```

CONTRIBUTING.rst
```diff
@@ -391,3 +391,39 @@ Creating a Diagram from the Grammar
...
+  INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
+  ...
+  INFO  [alembic.runtime.migration] Running downgrade e41c021c19a6 -> , Initial migration
+
+  root@...:/code# alembic upgrade head
+  INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
+  INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
+  ...
+
+After this the `docker compose up` should start db_migration agent without problems.
```
samredai, line 429:
```suggestion
After this, the `docker compose up` command should start the db_migration agent without problems.
```

samredai (approved):
Thanks @agorajek!

CONTRIBUTING.rst
samredai: There may be a few ways to end up in this situation, right? In some cases, the fact that we're squashing is what would get someone into this situation. I wonder how much value it is in practice for us to have a series of alembic commits in the OSS repo--is that the normal practice? As opposed to just always having a single squashed alembic commit.

To elaborate, I can only really see it being used by the demo environment. As a hypothetical, let's say someone has a datajunction-service instance in production running v0.0.1 and they're upgrading to v0.1.0. Wouldn't they use alembic themselves to create a diff to apply so that it's in-line with their production deployments? Let me know if I'm missing something.

agorajek: Great points. I agree that:
1. For demo purposes one alembic file is good, maybe best cause restart can be easily coordinated.
2. It would be nice (or rather critical) to have a separation of DB migration between versions of DJ. Then there is a chance that if users didn't touch their DJ tables the alembic migration will actually work. But practice is often different here, that's where consultants/support comes in.

samredai: Agreed, all good points. Another thing I just realized is this error is probably caused by the demo sqlite databases not being updated after recent changes to the metadata model. I'll open a PR to update those but it sounds like something we need to make a note of somewhere--that is, updating the demo sqlite databases when things change.

shangyian: @samredai my most recent PR should've updated the demo db (this is the one that fixed the dimensions logic)

agorajek: @shangyian yes, I got the error before your PR. I think the problem was on my side when the DB already existed from my previous work. This is not the first time I saw this issue but it took me few minutes to figure out the solution, hence this PR so maybe others will benefit.

samredai: Makes sense!

PR #605, opened 2023-07-03
UI enhancements

shangyian:
### Summary

Various UI enhancements that I added for the presentation... which I have now modified further.

#### New Exploration Tab

The main change here is integrating the namespaces page with the node listing pages into a single exploration tab. The namespaces are arranged hierarchically in the left-hand panel, with the nodes from the selected namespace displayed on the right.

https://github.com/DataJunction/dj/assets/9524628/c9747468-aaf5-4388-af38-f58828b26bea

#### Node revisions in history tab

Screenshot 2023-07-03 at 9 51 15 AM

#### Node info tab changes

* Added table to a source node's info tab
* Added primary key to the node info tab
* Fixed display of cube elements

Screenshot 2023-07-03 at 9 56 37 AM

### Test Plan

- [ ] `make check` passes


datajunction-ui/src/app/services/DJService.js
```diff
@@ -146,7 +153,7 @@ export const DataJunctionAPI = {
     metricSelection.map(metric => params.append('metrics', metric));
     dimensionSelection.map(dimension => params.append('dimensions', dimension));
     const data = await (
-      await fetch(DJ_URL + '/data/?' + params + '&limit=100')
+      await fetch(DJ_URL + '/data/?' + params + '&limit=100&async_=true')
```
samredai, line 156: I see this was switched to async but I don't see polling anywhere. Isn't some kind of polling mechanism required to eventually display the data after the "Run Query" button is clicked on the SQL page?

agorajek:
I really like all the new features. @betodealmeida check out this PR!

I didn't look much into the code. I don't think I can add value at this point :)

agorajek (approved):

datajunction-ui/src/app/services/DJService.js
shangyian: Good point! Let me fix this

shangyian:
@agorajek and @samredai: I made some more changes wrt to the hierarchical namespaces -- let me know what you think about this new design based on the screen recording in the PR description. It basically merges the list-namespaces page with the list-nodes-per-namespace page into a single "Explore" page. 

Also @samredai I might just get rid of the async=true part and I'll make that update in a separate PR, as I think there are already enough changes in this PR, haha

datajunction-ui/src/app/icons/CollapsedIcon.jsx
shangyian, line 1: Moved all the icons to `app/icons/*.jsx` so that we don't have large svg blobs littered in the code

samredai:
@shangyian I think the hierarchical namespace explorer looks fantastic and it looks much more intuitive having the explorer and node list share the same page. I pulled this PR down to try it out and for some reason it's not showing up the same way as your video. I ran the integration tests which created some nodes in the `integration.javascript.elegant_lamport3` namespace, but each namespace level is just showing flat. Selecting `integration` shows all of the nodes and selecting `javascript` or `elegant_lamport3` show an empty list. Let me know if you think this might be something to do with the state of my local dev environment.

Screenshot 2023-07-05 at 12 19 39 AM

shangyian:
@samredai good catch! It wasn't handling the case where the created namespace doesn't have all of its subnamespaces in the namespace list (i.e., a namespace like `integration.python.amazing_liskov` would need to have `integration` and `integration.python` as well). I fixed this and now see: 

Screenshot 2023-07-04 at 10 32 35 PM

agorajek:
@shangyian the new expanded tree of namespaces looks awesome! Thank you!

Issue #579, opened 2023-06-15
Multi-language client integration tests


samredai:
We should have a github actions workflow that launches the demo environment (dj + djqs + djrs) and tests each of the language clients against the server.

PR #606, opened 2023-07-04
Client integration tests

samredai:
### Summary

This adds integration tests for both the python and javascript client as well as adds an integration test workflow that runs daily (or can be manually dispatched). It also includes some small tweaks to the clients where I found missing or unexpected behavior.

**example client integration workflow ran in my fork**: https://github.com/samredai/dj/actions/runs/5451795214

The integration test workflow builds and launches the demo DJ environment using `docker compose --profile demo up`. It then runs only the integration tests for the clients which simulate a sequence of various actions such as creating nodes of all types, activating/deactivating nodes, generating SQL, running queries to retrieve data, finding common dimensions for a set of metrics, creating draft nodes, etc...

As clients are added in the future in different languages, this pattern should be re-usable to include integration tests. I've seen some more clever things done where a matrix of different language clients are "paired" up to test things like creating a node using the python client and then reading that node using the javascript client. We can go down that road in the future if we find it useful.

### Test Plan

Ran the python and javascript integration tests, also tested the github workflow in my fork.

- [x] PR has an associated issue: #579 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -615,7 +615,7 @@ class DJClient:  # pylint: disable=too-many-public-methods
                     rows,
                     columns=[col["name"] for col in columns],
                 )
-            except NameError:
+            except NameError:  # pragma: no cover
                 return Results(
                     data=rows,
                     columns=tuple(col["name"] for col in columns),  # type: ignore
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -1,24 +1,13 @@
 """DataJunction client setup."""
 import abc
 
-# pylint: disable=redefined-outer-name, import-outside-toplevel
+# pylint: disable=redefined-outer-name, import-outside-toplevel, too-many-lines
 import logging
 import platform
 import time
-import warnings
-from typing import Any, Dict, List, Optional, Union
-from urllib.parse import urljoin
+from typing import Any, Dict, List, Optional, Tuple, TypedDict, Union
+from urllib.parse import urlencode, urljoin
 
-try:
-    import pandas as pd
-except ImportError:  # pragma: no cover
-    warnings.warn(
-        (
-            "Optional dependency `pandas` not found, data retrieval"
-            "disabled. You can install pandas by running `pip install pandas`."
-        ),
-        ImportWarning,
-    )
 import requests
 from alive_progress import alive_bar
 from pydantic import BaseModel, Field, validator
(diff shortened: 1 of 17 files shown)


client/python/datajunction/client.py
```diff
@@ -547,16 +592,16 @@ def sql(  # pylint: disable=too-many-arguments
         return response.json()
 
     @staticmethod
-    def results_to_dataframe(results) -> pd.DataFrame:
+    def process_results(results) -> Results:
```
samredai, line 595: @shangyian I changed this to just return the data and want to get your thoughts on it. The two main benefits I see here is that the datajunction client would no longer have to include pandas as a dependency and we can avoid an assumption that the user's end goal is a pandas dataframe. Here are some examples of one-liners that can get from this `Results` data model to a number of other in-memory data formats including pandas.

_directly to a pandas dataframe_
```py
import pandas as pd
df = pd.DataFrame(**results)
```
_directly to a pyarrow table_
```py
import pyarrow as pa
table = pa.Table.from_arrays(list(map(list, zip(*results['data']))), names=results['columns'])
```
_directly to a partitioned dask dataframe_
```py
from dask import dataframe as dd
ddf = dd.from_dict({k:v for k, v in zip(results['columns'], list(map(list, zip(*results['data']))))}, npartitions=4)
```
_directly to a distributed modin dataframe (using their drop-in replacement for pandas)_
```py
import modin.pandas as pd
mdf = pd.DataFrame(**results)
```

shangyian: I think my main concern here is that we should try to give users exactly what they want, whether that's pandas, pyarrow or something else, rather than our `Results` object. We can certainly have that as a part of an intermediary stage on the client, but we should also include the conversion functions you have above and provide all of them to the user, and then have several versions of the client like `datajunction[pandas]`, `datajunction[pyarrow]`, each of which would include the dependencies to support those specific libraries.

samredai: That's a fair compromise, I'll update this with your proposal. 👍

client/python/datajunction/client.py
```diff
@@ -547,17 +603,24 @@ def sql(  # pylint: disable=too-many-arguments
...
         """
         if "results" in results and results["results"]:
             columns = results["results"][0]["columns"]
             rows = results["results"][0]["rows"]
-            return pd.DataFrame(
-                rows,
-                columns=[col["name"] for col in columns],
-            )
+            try:
+                return pd.DataFrame(
```
samredai, line 614: @shangyian I added back the import error catch logic you had and then did a try/except here. Pandas is an optional dependency now so as you described, it will work either if someone already has pandas in their environment or if they installed the dj client using `pip install datajunction[pandas]`

client/python/tests/test_client.py
shangyian, line 1: Presumably with these changes, this entire set of tests could be moved into the integration testing suite, and we can remove the dependency on `datajunction-server` (although it is only a test dependency, so it's not a big deal)? No need to do that here of course!

shangyian (approved):
Thanks for adding this! The changes look good

agorajek (approved):
Love this idea. Thanks for putting this together @samredai .

client/python/tests/test_client.py
samredai: Agreed, that's a great idea

PR #608, opened 2023-07-06
Build SQL for materialized cubes

shangyian:
### Summary

This PR adds functionality to build queries for materialized cube nodes. This was also in the presentation, but I've gone through this a second time to handle various edge cases.

The main changes include:
* When a cube is created, we store some additional info needed for building materialized cube queries into the default "materialization" config, namely a `combiner` expression that combines the measures back into the metric.
* When building SQL for multiple metrics, we add a check to see if those metrics + dimensions can be found in an existing materialized cube
* If there is a materialized cube that encompasses the metrics + dimensions, we build SQL that queries the cube directly by combining the measures back into the selected metrics 
* Refactor so that we have a single function `build_sql_for_multiple_metrics` that generates SQL for both the `/sql` and `/data` endpoints

#### Example

Let's say we have a cube named `default.repairs_cube` that contains the following metrics and dimensions:
* `default.discounted_orders_rate` (metric)
* `default.num_repair_orders` (metric)
* `default.avg_repair_price` (metric)
* `default.hard_hat.country` (dim)
* `default.hard_hat.city` (dim)

After we post an availability state for this cube, whenever we query metrics or dimensions that are in the cube, we'll generate SQL that queries the materialized cube directly. For example, if we ask for `default.discounted_orders_rate`, `default.num_repair_orders` and `default.avg_repair_price` grouped by `default.hard_hat.country`, we'll get the following query:

```sql
SELECT
  sum(discount_sum) / count(placeholder_count) default_DOT_discounted_orders_rate,
  count(repair_order_id_count) default_DOT_num_repair_orders,
  sum(price_sum) / count(price_count) default_DOT_avg_repair_price,
  country
FROM repairs_cube
GROUP BY
  country
```

### Test Plan

- [x] PR has an associated issue: #470 
- [x] `make check` passes


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -616,3 +622,114 @@ def validate_orderby(
...
+    statement = select(NodeRevision)
+    for name in element_names:
+        statement = statement.filter(
+            NodeRevision.cube_elements.any(Column.name == name),  # type: ignore  # pylint: disable=no-member
+        )
+
+    existing_cubes = session.exec(statement).unique().all()
+    for cube in existing_cubes:
+        return cube
+    return None
```
agorajek, line 645: Not necessary, `None` is the default return value.

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -616,3 +622,114 @@ def validate_orderby(
...
+        dimensions,
+    )
+    leading_metric_node = get_node_by_name(session, metrics[0])
+    available_engines = leading_metric_node.current.catalog.engines
+
+    # Try to find a built cube that already has the given metrics and dimensions
+    # The cube needs to have a materialization configured and an availability state
+    # posted in order for us to use the materialized datasource directly
+    cube = find_existing_cube(session, metric_columns, dimension_columns)
+    if cube and cube.materializations and cube.availability:
```
agorajek, line 678: nit: I wonder if it makes sense to add this condition to the `find_existing_cube()` method itself... perhaps not, but then I feel like we could add a wrapper on that one called `find_materialized_cube()` and have this condition in there for reuse.

agorajek (approved):
This is good. Two small comments in line.

datajunction-server/datajunction_server/api/helpers.py
shangyian: Agreed, I think it would be a good idea to add an option to look for materialized cubes. I added a parameter to `find_existing_cube` called `materialized` so that we can filter for materialized cubes if desired.

PR #613, opened 2023-07-07
squash alembic migrations

shangyian:
### Summary

This PR squashes the database migrations into a single migration and updates `dj.demo.db` with the changes.

The main reason behind doing this now is that our existing migrations don't work for CockroachDB, which doesn't allow us to alter the primary key of a table in the same transaction as making another change to the table. This is allowed for sqlite, which is why the migrations were working fine with our demo docker compose setup.

### Test Plan

Ran locally with the updated `dj.demo.db` + tests.

- [x] PR has an associated issue: #301 
- [x] `make check` passes

### Deployment Plan

We should think about how database migration maintenance should work long-term, but that's a separate discussion.


samredai (approved):
🚀

PR #614, opened 2023-07-07
Add specialized exception to Python client.

agorajek:
### Summary

When trying Python client I wanted to have this code be idempotent:
```
dj.new_namespace(namespace="foo.bar")
```
, but it is not because we get
```
DJClientException: Node namespace `foo.bar` already exists
```

With this addition of specialized exceptions it will be easier to catch the exceptions that are safe to pass. And instead of doing something like:
```
try:
    dj.new_namespace(namespace="foo.bar")
except DJClientException as e:
    if "already exists" not in str(e):
        raise e
```
I'll be able to do:
```
try:
    dj.new_namespace(namespace="foo.bar")
except DJNamespaceAlreadyExists:
   pass
```

Also adjusted the contributing guide.

### Test Plan

- [ ] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -5,13 +5,14 @@ Contributing
 Pre-requisites
 ==============
 
-DataJunction (DJ) is currently supported in Python 3.8, 3.9, and 3.10. It's recommended to use ``pyenv`` to create a virtual environment called "dj", so you can use the included ``Makefile``:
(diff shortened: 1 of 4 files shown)


shangyian (approved):
Looks great! Thanks for updating the docs too

shangyian:
You probably have to run `make check` to get the linters going.

samredai (approved):
LGTM, thanks @agorajek! You'll also probably need to add a single test using pytest.raises to get the required coverage.

PR #615, opened 2023-07-09
Pin pydantic<2 for the Python client

shangyian:
### Summary

Pin pydantic<2 for the Python client as we're not compatible with the latest release.

Found this issue from running the client. Since we don't pin an upper boundary for Pydantic, client users can find themselves using pydantic>=2, which isn't compatible with our models.

### Test Plan

- [x] `make check` passes


agorajek (approved):
Yes please. And thank you @shangyian .

samredai (approved):

Issue #618, opened 2023-07-11
source nodes become invalid


shangyian:
There is an odd bug where source nodes become invalid after creating a metric that sources from it.

PR #619, opened 2023-07-11
Add a `POST /nodes/{name}/validate/` endpoint for revalidating already existing nodes by name

samredai:
### Summary

Currently DJ attempts to validate nodes as any live change is being made to the system. When things are missed or preconditions are fulfilled after the fact, this PR provides a mechanism to request the server to revalidate an existing node, requiring that you simply provide the node name. Over time, we should catch these scenarios more and more so that system status is more accurately captured in realtime and revalidation is never required.

Here's a video example of how this was added to the UI. I manually set nodes in the metadata that I knew to be valid as invalid, and clicking the icon revalidated them and updated the status. If they were invalid, the status icon would have stayed as a red x.

https://github.com/DataJunction/dj/assets/43911210/3416b0ef-4822-4379-954a-f2c8f3eb885d

### Test Plan

`docker compose --profile demo up` and tested it in the UI. Also tested curl commands directly to the dj server.

- [ ] PR has an associated issue: Related to but does not solve #618 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -136,6 +136,54 @@ def validate_a_node(
     )
 
 
+@router.post("/nodes/{name}/validate/", response_model=NodeValidation)
+def revalidate_a_node(
+    name: str,
+    session: Session = Depends(get_session),
+) -> NodeValidation:
+    """
+    Revalidate a single existing node and update its status appropriately
+    """
+    node = get_node_by_name(session, name)
+    current_node_revision = node.current
+    if current_node_revision.type == NodeType.SOURCE:
+        current_node_revision.status = NodeStatus.VALID
+        session.add(current_node_revision)
+        session.commit()
+        return JSONResponse(
+            status_code=HTTPStatus.OK,
+            content={
+                "message": f"source node `{current_node_revision.name}` has been set to valid",
+                "status": NodeStatus.VALID,
+            },
+        )
+
+    (
+        _,
+        _,
+        missing_parents_map,
+        type_inference_failed_columns,
+    ) = validate_node_data(current_node_revision, session)
+    if missing_parents_map or type_inference_failed_columns:
+        status = NodeStatus.INVALID  # pragma: no cover
+    else:
+        status = NodeStatus.VALID
+
+    if current_node_revision.status != status:  # pragma: no cover
+        session.add(current_node_revision)
+        session.commit()
+    return JSONResponse(
+        status_code=HTTPStatus.OK,
+        content={
+            "message": (
+                f"{current_node_revision.type} node `{current_node_revision.name}` "
+                "has been set to {status}"
+            ),
+            "status": status,
+        },
+    )
+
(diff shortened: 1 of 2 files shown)


shangyian (approved):
Looks useful! 

Once we've ironed out all the bugs, do you think we'll still want this endpoint though?

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -136,6 +136,54 @@ def validate_a_node(
...
+    name: str,
+    session: Session = Depends(get_session),
+) -> NodeValidation:
+    """
+    Revalidate a single existing node and update its status appropriately
+    """
+    node = get_node_by_name(session, name)
+    current_node_revision = node.current
+    if current_node_revision.type == NodeType.SOURCE:
+        current_node_revision.status = NodeStatus.VALID
```
agorajek, line 150: This is dangerous, if the source node does not match the underlying table anymore. Right?

agorajek (commented):
Agree with @shangyian in that I think this endpoint should be __internal__ only. Meaning if a user ever need to use it then we have a bigger problem that should be addressed.

2 questions in-line.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -136,6 +136,54 @@ def validate_a_node(
...
+        _,
+        missing_parents_map,
+        type_inference_failed_columns,
+    ) = validate_node_data(current_node_revision, session)
+    if missing_parents_map or type_inference_failed_columns:
+        status = NodeStatus.INVALID  # pragma: no cover
+    else:
+        status = NodeStatus.VALID
+
+    current_node_revision.status = status
```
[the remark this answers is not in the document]
samredai: This doesn't create a new revision, it just updates the system status value on the existing revision. But good point that there's no reason to even do a db commit at all so I'll fix this.

datajunction-server/datajunction_server/api/nodes.py
samredai: You're right, I hadn't considered that reflection could be setting a source node as invalid. Good catch!

samredai:
Makes sense! I'll update this PR to remove the UI changes then and we can just call the endpoint when needed.

datajunction-server/datajunction_server/api/nodes.py
samredai: @agorajek thinking about this a bit more, since we're going to leave this as an admin endpoint, is it valuable to leave this here as an escape hatch for when a reflection service is acting up?

agorajek: Yes agree.

datajunction-server/datajunction_server/api/nodes.py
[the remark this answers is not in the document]
samredai: Updated to check `if current_node_revision.status != status:`

agorajek (approved):
Looks good.

Issue #617, opened 2023-07-11
Bug: when node is deleted we don't clean up downstream links


agorajek:
Say you have a source node S1 and metric M1 that is reading from S1. If S1 is inactivated, the metric properly does not see S1 anymore but there remains a connection from S1 to M1 that is causing troubles in the big picture.

PR #621, opened 2023-07-11
Fix node update status bug

shangyian:
### Summary

There is a bug when we update any non-source node, where if we make a minor change that doesn't involve the node's query, it will not assign a status to the new node revision, and just defaults to invalid. This fixes it so that it assigns it to the old node revision's status and then updates it if anything has changed that warrants updating.

### Test Plan

- [ ] `make check` passes


samredai (approved):
Yes! Awesome find 🎉 

feliz-pops-maellard

agorajek:
Nice! 

Do you think this will also solve the issue we found in https://github.com/DataJunction/dj/issues/617.

shangyian:
@agorajek It's possible! Let me also test that case.

PR #622, opened 2023-07-12
bound dimensions

CircArgs:
### Summary

Adds an attribute to Metric Nodes that allow the author to specify that when the metric is used, some columns from the sourced node must be grouped by.

### Test Plan

Unit test

- [ ] PR has an associated issue: https://github.com/DataJunction/dj/issues/611
- [ ] `make check` passes


datajunction-server/alembic/versions/2023_07_12_0008-25637bbf2d52_add_bound_dimensions_attribute_to_.py
samredai, line 1: Looks like you ran the revision twice, you should be able to just delete this file.

samredai:
@CircArgs VERY cool! It's awesome that the columns objects are actually linked in the metadata so we could probably very easily add in a check in the node update logic to prevent dropping a column in a transform if it's being used as a required dimension by any metrics.

@agorajek can you take a look at this when you get a chance and see how this works for some of the use cases you had in mind?

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -489,6 +489,10 @@ def build_node(  # pylint: disable=too-many-arguments
             ),
         )
 
+    # get dimension columns which are required
```
shangyian, line 492: Does this need to additionally check if the user-provided dimensions overlap with the `bound_dimensions` set on the node?

shangyian (commented):
Thanks for adding! I think the data models here all make sense.

Will this additionally need to modify the `POST /nodes/metric` endpoint so that someone creating a metric can set the bound dimensions of the metric? And presumably that would be where you do the check for whether their provided bound_dimensions are actually valid.

CircArgs:
@shangyian 
> Will this additionally need to modify the `POST /nodes/metric` endpoint so that someone creating a metric can set the bound dimensions of the metric? And presumably that would be where you do the check for whether their provided bound_dimensions are actually valid.

Yeah I think that's a good place to do that validation. Can it be done in a follow-up PR?

shangyian:
@CircArgs follow-up PR sounds good!

shangyian (approved):

samredai (approved):

Issue #620, opened 2023-07-11
ability to delete linked dimensions


shangyian:
Sometimes a dimension was linked in error and needs to be removed.

samredai:
Phew, I feel like my subconscious has been wanting this endpoint for a while now...haha

PR #623, opened 2023-07-12
Add ability to remove dimension links

shangyian:
### Summary

This adds a new API endpoint that allows users to remove a dimension linked to a node's column: 
```
DELETE /nodes/{name}/columns/{column}
```

It also adds the corresponding Python client call:
```python
source.unlink_dimension("payment_type", "default.payment_type", "payment_type_id")
```

### Test Plan

Ran locally. Added tests.

- [X] PR has an associated issue: #620 
- [x] `make check` passes

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -74,8 +74,6 @@ class RequestsSessionWithEndpoint(requests.Session):  # pragma: no cover
             return response
         except requests.exceptions.RequestException as exc:
             error_message = None
-            if not exc.response:
-                raise DJClientException(exc) from exc
             if exc.response.headers.get("Content-Type") == "application/json":
                 error_message = exc.response.json().get("message")
             if not error_message:
@@ -400,6 +398,7 @@ class DJClient:  # pylint: disable=too-many-public-methods
             timeout=self._timeout,
         )
         json_response = response.json()
+        print(json_response)
         if response.status_code == 409:
             raise DJNamespaceAlreadyExists(json_response["message"])
         return json_response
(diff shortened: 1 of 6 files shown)


shangyian:
@samredai That's a good point. It's definitely possible to add the check, although it fits into the "which nodes are available for this dimension" question, which would require us to traverse the entire graph. But I think it's worth adding since we need to be able to keep the valid/invalid state up-to-date. 

If it turns out to be too slow, another option is that we can have the reflection service come in and refresh the validity of each node, which would be a non-blocking process that doesn't affect users.

client/python/datajunction/client.py
```diff
@@ -531,6 +531,23 @@ def link_dimension_to_node(
         )
         return response.json()
 
+    def unlink_dimension_to_node(
```
agorajek, line 534: I think `unlink_dimension_from_node` makes a bit more sense, no?

Also if this is a helper function should we start adding some `_` in front for good measure?

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1332,6 +1332,52 @@ def link_a_dimension(
...
+@router.delete("/nodes/{name}/columns/{column}/", status_code=201)
+def delete_dimension_link(
+    name: str,
+    column: str,
+    dimension: str,
+    dimension_column: Optional[str] = None,
+    session: Session = Depends(get_session),
+) -> JSONResponse:
+    """
+    Add information to a node column
```
agorajek, line 1344: Comment seems off.

agorajek (approved):
Cool. Thanks @shangyian . Two small naming questions inline.

client/python/datajunction/client.py
shangyian: Good point, I changed the name and prefixed it with `_`

samredai:
> If it turns out to be too slow, another option is that we can have the reflection service come in and refresh the validity of each node, which would be a non-blocking process that doesn't affect users.

I'm starting to lean more and more towards the latter option--having something else do a retrospective validation. In the future we can consider making dimension links bidirectional but for the time being I think that's too strong of a connection.

samredai (approved):
🚀

shangyian:
@samredai I think that function I have actually does work (just did some more testing), but I'll go ahead and add it in a separate PR because it pairs well with some additional endpoints that expose that logic (#321).

samredai:
@shangyian nice!!

PR #625, opened 2023-07-12
fix: reenable pdm lock check hook for all packages

cwegener:
pdm lock check doesn't run for any of the packages after the monorepo
move

only the root level pdm lock check (which only has pylint and pre-commit
in it) actually runs.

keeping only one single top-level pdm.lock file at the repo root level
may work for simple development tasks (e.g. pdm sync from repo root),
but any pdm functions that depend on more granular pdm.lock input (e.g.
pdm export) will break when run from an individual package folder if the
individual pdm.lock files in each package folder is not kept in sync as
well.

adding all packages back into the .pre-commit config fixes this

a new dedicated Github Action step to force a pdm.lock check on every PR
also has been added.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-files: (^datajunction-server/)
+files: ^datajunction-(server|query|reflection)/
 exclude: (^docs/|^openapi/|^client/python/|^client/javascript/|^datajunction-server/dj/sql/parsing/backends/grammar/generated|^README.md)
 
 repos:
@@ -107,3 +107,27 @@ repos:
   rev: 2.6.1
   hooks:
     - id: pdm-lock-check
+      name: pdm-lock-check-root
+      entry: pdm lock --check --project .
+      files: ^pyproject.toml$
+- repo: https://github.com/pdm-project/pdm
+  rev: 2.6.1
+  hooks:
+    - id: pdm-lock-check
+      name: pdm-lock-check-server
+      entry: pdm lock --check --project datajunction-server
+      files: ^datajunction-server/pyproject.toml$
+# - repo: https://github.com/pdm-project/pdm
+#   rev: 2.6.1
+#   hooks:
+#     - id: pdm-lock-check
+#       name: pdm-lock-check-query
+#       entry: pdm lock --check --project datajunction-query
+#       files: ^datajunction-query/pyproject.toml$
+- repo: https://github.com/pdm-project/pdm
+  rev: 2.6.1
+  hooks:
+    - id: pdm-lock-check
+      name: pdm-lock-check-reflection
+      entry: pdm lock --check --project datajunction-reflection
+      files: ^datajunction-reflection/pyproject.toml$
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -107,3 +107,27 @@ repos:
...
+    - id: pdm-lock-check
+      name: pdm-lock-check-server
+      entry: pdm lock --check --project datajunction-server
+      files: ^datajunction-server/pyproject.toml$
+# - repo: https://github.com/pdm-project/pdm
+#   rev: 2.6.1
+#   hooks:
+#     - id: pdm-lock-check
+#       name: pdm-lock-check-query
+#       entry: pdm lock --check --project datajunction-query
```
samredai, line 125: Is there a reason why you kept the datajunction-query one commented out?

samredai (approved):
LGTM thanks @cwegener! Just one comment. Also just to confirm, if I wanted to locally run these checks to make sure the CI will pass, that would just require changing into the `datajunction-(server|query|reflection)/` directory and running:
```sh
pdm lock --check --project datajunction-(server|query|reflection)
```
Is that right? Might be useful to add to the contributing guide.

shangyian (approved):
Looks good, although I think the datajunction-query block needs to be uncommented

.pre-commit-config.yaml
cwegener: It will not succeed as the pdm.lock file in that package is out of sync. I wanted to keep the fix for that in a separate change. I want to focus on the fixing the tooling first.

cwegener:
> Looks good, although I think the datajunction-query block needs to be uncommented. And good catch!

The pdm lock check for datajunction-query will be enabled in a separate change.

cwegener:
> LGTM thanks @cwegener! Just one comment. Also just to confirm, if I wanted to locally run these checks to make sure the CI will pass, that would just require changing into the `datajunction-(server|query|reflection)/` directory and running:
> 
> ```shell
> pdm lock --check --project datajunction-(server|query|reflection)
> ```
> 
> Is that right? Might be useful to add to the contributing guide.

That's a great question.

For now, the most straight-forward answer would be for a developer to perform the exact same test that the dedicated GitHub action runs, which is:

```
pre-commit run pdm-lock-check --all-files
```

The next most important thing would be to strongly encourage all new developers to make sure that the pre-commit hook is installed in their local `dj` repo by running `pre-commit install`, which can only be done through proper documentation and education.

On a bigger picture, there is the overall question of how the tooling should integrate across the different execution layers (`pdm`, `git hooks`, `Github runners`, generic task runners e.g. `make`)

Trying to simplify all these different execution layers and make the configuration of them easy to maintain is still a bit of a challenge and I don't have a good answer at the moment. Using `make` as a generic task runner for abstraction is not really my favourite option (as previously mentioned) since designing and maintaining a set of fully featured makefiles with all the bells and whistles needed is not easily achieved. I'm still investigating and trialing some other options as a generic task runner, mainly https://taskfile.dev

Once I have a proven baseline for a generic task runner that I'd be comfortable with recommending, I'll be making that recommendation. :smile:

Issue #624, opened 2023-07-12
Additional history tracking


shangyian:
We should add these to the history tracker (there's probably more):

**Nodes**
- [ ] Node creation (at the moment the tracker only works for source nodes, but we should track creation for all nodes)
  - `entity_type=NODE`
  - `activity_type=CREATE`
- [ ] Updating node:
  - `entity_type=NODE`
  - `activity_type=UPDATE`
- [ ] Deactivating node
  - `entity_type=NODE`
  - `activity_type=DEACTIVATE`
- [ ] Activating node
  - `entity_type=NODE`
  - `activity_type=ACTIVATE`
- [ ] Tagging node with tag
  - `entity_type=NODE`
  - `activity_type=TAG`
- [ ] Linking dimensions:
  - `entity_type=NODE`
  - `activity_type=LINK`
- [ ] Unlinking dimensions:
  - `entity_type=NODE`
  - `activity_type=UNLINK`
- [ ] Setting column attributes:
  - `entity_type=NODE`
  - `activity_type=SET_ATTRIBUTE`
  
**Namespaces**
- [ ] Creating a namespace:
  - `entity_type=NAMESPACE`
  - `activity_type=CREATE`
  
**Tags**
- [ ] Creating a tag:
  - `entity_type=TAG`
  - `activity_type=CREATE`
- [ ] Updating a tag:
  - `entity_type=TAG`
  - `activity_type=UPDATE`

samredai:
Thanks for creating this, I added two to the end of the nodes section. cc: @betodealmeida @cwegener @CircArgs @agorajek , please add any other events you can think of that we'd want to track in the history table.

PR #626, opened 2023-07-12
Add history tracking for various activities

shangyian:
### Summary

This PR adds history tracking for the following activities:

* Creating a node (for non-source nodes)
* Updating a node
* Posting availability state
* Creating namespace
* Setting column attribute on a node
* Deleting a node
* Restoring a node
* Adding/updating a materialization
* Creating a dimension link
* Creating a tag
* Updating a tag
* Tagging a node

It also adds a `node` field to `History`, which will help us track and return activity at the node level, rather than just at an individual entity level.

Screenshot 2023-07-15 at 8 53 51 AM

### Test Plan

Added tests for all of the above. Planning on recreating `dj.demo.db` with the above history tracking in place.

- [X] PR has an associated issue: #624 
- [X] `make check` passes


datajunction-server/datajunction_server/models/history.py
```diff
@@ -18,14 +18,26 @@ class ActivityType(str, Enum):
...
 
+    AVAILABILITY = "availability"
     CREATE = "create"
     DEACTIVATE = "deactivate"
     ACTIVATE = "activate"
     UPDATE = "modify"
+
+    # Link columns to dimensions
     LINK = "link"
+    UNLINK = "unlink"
```
samredai, line 29: I know `LINK` was already under activity type but looking at it now I think it should be an entity type and we should have a `DELETE` activity type. Then the history could be captured as `activity_type=ActivityType.[DELETE|CREATE], entity_type=EntityType.LINK`?

datajunction-server/datajunction_server/models/history.py
```diff
@@ -18,14 +18,26 @@ class ActivityType(str, Enum):
...
+
     TAG = "tag"
+
+    # Set column attributes
     SET_ATTRIBUTE = "set_attribute"
 
+    # Node materialization
+    ADD_MATERIALIZATION = "add_materialization"
+    UPDATE_MATERIALIZATION = "update_materialization"
+    DELETE_MATERIALIZATION = "delete_materialization"
```
samredai, line 39: Similar to the other comment, I think there should be an entity type `MATERIALIZATION` and then it can be combined accordingly with activity types of `CREATE`, `UPDATE`, or `DELETE`. Also similarly for `AVAILABILITY`

datajunction-server/datajunction_server/models/history.py
shangyian: Yeah, I thought about structuring it the way you described, but then I realized that when we ask for the history of an entity like `node`, we won't get a lot of the other activities that are closely related to nodes like linking a dimension or adding a materialization in a single query. I guess this is thinking of the EntityType as the level at which we want to retrieve user activity, rather than a direct mapping to all of the entities we actually store in the backend. Otherwise we'd take a performance hit if we want to actually display all activity related to a node, as we'd need to hit the backend 3+ times to get node-level, link-level, and materialization-level activity. But maybe that doesn't matter too much?

samredai: That's a good point, I don't think the performance is as much of an issue since the selects should remain very simple, but we do want to make it as convenient as possible to query. Maybe we can add a "context_node" attribute that always includes the relevant node name? That should easily support these query patterns:
- Get all activity for a node
- Get all of a type of activity (i.e. show me global create activity)
- Get all activity for an entity type (i.e. show me global activity for materializations)

What do you think?

agorajek: This is a tough one for me too but I am leaning towards what @samredai is saying and I think a cleaner model will be better in the long term than the ability to query "everything" in one swoop. I think we'll be ok with the ability to query widely "by even type" and just add some consolidation at the places where we need, say if we ever wanted to do "get_everything_about_a_node", but one node, not all of them.

shangyian: Yeah, good point, we can probably rework the history endpoint if needed to provide node-level activity, which would include all the links, materializations etc) and not store the `context_node` at all.

shangyian:
@samredai and @agorajek I've updated this so that `EntityType` maps onto what we actually store (so `LINK`, `ATTRIBUTE`, `MATERIALIZATION` etc), with `ActivityType` being straighforward actions like `CREATE`, `UPDATE`, 
 and `DELETE`. I also went through the checklist in https://github.com/DataJunction/dj/issues/624 and added a bunch of them

agorajek:
@shangyian I am reading through your PR and I feel like the `context` keyword is really not needed anywhere in this new code. Am I missing something or have you arrived at a conclusion that `history by node context` means much more than `history (or activity) by node` ?

agorajek (approved):
Thanks @shangyian !. I think this is awesome. I just dropped one last comment about the `context` keyword, but up to you if you want to act on it.

shangyian:
@agorajek That's a good point -- I renamed it to `node` 😄

PR #631, opened 2023-07-12
Add endpoint to list all nodes that share a given dimension

shangyian:
### Summary

This PR adds endpoint that will find all nodes that share a given dimension:
```
GET /dimensions/{name}/nodes/
```

The same functionality that underlies that endpoint can be used to find and invalidate cube nodes after a relevant dimension link is removed (that was started in #623). 

Also added a tab to display these nodes in the dimension node tab UI. I changed the tabs setup in the UI to hide certain tabs (SQL, materializations) from nodes where the tab isn't relevant (i.e., source nodes).

Screenshot 2023-07-13 at 11 00 49 PM

### Test Plan

- [X] PR has an associated issue: #321 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -171,6 +171,13 @@ def get_nodes_with_dimension(
             statement = (
                 select(NodeRevision)
                 .join(
+                    Node,
+                    onclause=(
+                        (NodeRevision.node_id == Node.id)
+                        & (Node.current_version == NodeRevision.version)
+                    ),  # pylint: disable=superfluous-parens
+                )
+                .join(
                     NodeColumns,
                     onclause=(
                         NodeRevision.id == NodeColumns.node_id
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -3,10 +3,18 @@ DAG related functions.
 """
 import collections
 import itertools
-from typing import Deque, Dict, List, Set, Tuple, Union
+from typing import Deque, Dict, List, Optional, Set, Tuple, Union
+
+from sqlmodel import Session, select
 
 from datajunction_server.models import Column
-from datajunction_server.models.node import DimensionAttributeOutput, Node, NodeType
+from datajunction_server.models.base import NodeColumns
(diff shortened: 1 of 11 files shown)


datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -140,3 +148,55 @@ def get_shared_dimensions(
...
+    final_set: Set[NodeRevision] = set()
+    while to_process:
+        current_node = to_process.pop()
+        processed.add(current_node.name)
+
+        # Dimension nodes are used to expand the searchable graph by finding
+        # the next layer of nodes that are linked to this dimension
+        if current_node.type == NodeType.DIMENSION:
+            statement = (
+                select(NodeRevision)
```
samredai, line 172: Since this selects `NodeRevision`, does that mean it will include older revisions? I think we should only be considering the latest (or currently active?) node revision for each node name.

samredai (approved):
Just left one question but...this is such a game changer!

office-crying

datajunction-server/datajunction_server/sql/dag.py
shangyian: Yep agreed! Just modified this to only take into account the latest revision

shangyian:
@agorajek Great ideas! Haha, I was struggling with naming, so I'm glad you had a suggestion. I can definitely make 1 and 2 happen with ease. 

For 3, do you think it'd be useful to include the path as well, or just the distance? That could be under a `Via` column in the table. I'm trying to be careful about having information overload, but also don't want to hide potentially useful info. Or alternatively we can provide the path in the API response but just show the distance in the UI.

agorajek:
@shangyian 

> For 3, do you think it'd be useful to include the path as well, or just the distance? That could be under a `Via` column in the table. I'm trying to be careful about having information overload, but also don't want to hide potentially useful info. Or alternatively we can provide the path in the API response but just show the distance in the UI.

In a tabular form like this I think showing a distance would be best option. Because sorting by distance could be useful here. I think the paths to all the linked nodes would be best presented in a diagram form, rather then a list, so the Graph tab seems more relevant for this kind of info.

> edit: going to merge this now and I'll make the changes for Distance in a separate PR
Of course!

PR #632, opened 2023-07-13
Bound dimensions api & validation

CircArgs:
### Summary

Adds to the api ability to add bound dimensions as introduced in https://github.com/DataJunction/dj/pull/622

### Test Plan

Unit tests extended

- [x] PR has an associated issue: # See https://github.com/DataJunction/dj/pull/622
- [x] `make check` passes


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -358,19 +358,16 @@ def validate_node_data(
...
-                    DJError(
-                        code=ErrorCode.MISSING_PARENT,
-                        message="Node definition contains references to nodes that do not exist",
-                        debug={"missing_parents": list(missing_parents_map.keys())},
-                    ),
-                ],
-            )
+    if missing_parents_map and validated_node.mode != NodeMode.DRAFT:
+        raise DJException(
+            errors=[
```
samredai, line 363: I know this wasn't created in this PR but this exception should include `http_status_code=HTTPStatus.BAD_REQUEST`

datajunction-server/tests/api/nodes_test.py
```diff
@@ -555,6 +555,65 @@ def test_deactivating_linked_dimension(
...
+            "message": (
+                "Node definition contains references to columns as "
+                "bound dimensions that are not on parent nodes."
+            ),
+            "errors": [
+                {
+                    "code": 206,
+                    "message": (
+                        "Node definition contains references to columns as "
+                        "bound dimensions that are not on parent nodes."
```
shangyian, line 608: Can we call these "required dimensions" rather than "bound" in the error messages? Mainly because we'll probably just surface this error directly to users, and I think the former is clearer naming-wise.

shangyian (approved):
Thanks for adding! Changes LGTM, just added one comment on naming

PR #633, opened 2023-07-13
Validate that a metric only selects from a single source/transform

samredai:
### Summary

Although we validate that a metric has a single expression that includes an aggregation, we don't validate that it only selects from a single table (no joins). This adds that validation.

### Test Plan

Added a metric test `test_raise_on_joins`.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/models/node.py
+++ b/datajunction-server/datajunction_server/models/node.py
@@ -764,8 +764,8 @@ class NodeRevision(NodeRevisionBase, table=True):  # type: ignore
         ):
             raise DJInvalidInputException(
                 http_status_code=HTTPStatus.BAD_REQUEST,
-                message=f"Metric {self.name} selects from more than one table, "
-                "should only select from a single table",
+                message=f"Metric {self.name} selects from more than one node, "
+                "should only select from a single node",
             )
 
     def extra_validation(self) -> None:
--- a/datajunction-server/datajunction_server/models/node.py
+++ b/datajunction-server/datajunction_server/models/node.py
@@ -7,6 +7,7 @@ import sys
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from functools import partial
+from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/models/node.py
```diff
@@ -739,20 +740,34 @@ def check_metric(self):
...
 
+        # must query from a single table
+        if (
+            len(tree.select.from_.relations) != 1
+            or tree.select.from_.relations[0].extensions
+        ):
+            raise DJInvalidInputException(
+                http_status_code=HTTPStatus.BAD_REQUEST,
+                message=f"Metric {self.name} selects from more than one table, "
+                "should only select from a single table",
```
shangyian, line 768: Should this message use the `table` terminology or would `node` be clearer? (Not sure as I'm not even sure users like `node`, but it fits well with what we call them)

datajunction-server/tests/api/metrics_test.py
```diff
@@ -1073,7 +1073,69 @@ def test_raise_on_multiple_expressions(client_with_examples: TestClient):
...
+    response = client_with_examples.post(
+        "/nodes/source/",
+        json={
+            "columns": [
+                {
+                    "name": "counts",
+                    "type": "struct<a string, b bigint>",
+                },
+            ],
+            "description": "Collection of day dreams",
```
shangyian, line 1114: hahah

shangyian (approved):

datajunction-server/datajunction_server/models/node.py
agorajek: Agree with @shangyian . I think there is a chance we can win this "node vs table" so I would push on that concept where it fits well. The other DJ dag terms we can still keep under covers, like "bound dimensions" ;)

samredai: Good catch, we should consistently use node so I'll update this.

PR #634, opened 2023-07-13
fix: refresh pdm.lock, enable pdm-lock-check hook

cwegener:
datajunction-query package ended up with an out-of-sync pdm.lock file
after the monorepo change.

~~fixing up the existing pdm.lock for datajunction-query without updating
any version pins in the lockfile by running:~~

```
pdm lock -d -G :all --refresh --project datajunction-query
```

EDIT: a simple `pdm lock --refresh` was not sufficient to fix the lock
file. Somehow, the `uvicorn` package was completely missing from the
lock file. To re-add the `uvicorn` extra requirement, I ran:

```
pdm remove --project datajunction-query -G uvicorn uvicorn
```

And then re-added it again:

```
pdm add --project datajunction-query -G uvicorn \
'uvicorn[standard]>=0.21.1'
```

Also enable the pre-commit hook for pdm-lock-check-query since that hook
is now completing successfully again after fixing up the lock file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -117,13 +117,13 @@ repos:
       name: pdm-lock-check-server
       entry: pdm lock --check --project datajunction-server
       files: ^datajunction-server/pyproject.toml$
-# - repo: https://github.com/pdm-project/pdm
-#   rev: 2.6.1
-#   hooks:
-#     - id: pdm-lock-check
-#       name: pdm-lock-check-query
-#       entry: pdm lock --check --project datajunction-query
-#       files: ^datajunction-query/pyproject.toml$
+- repo: https://github.com/pdm-project/pdm
+  rev: 2.6.1
+  hooks:
+    - id: pdm-lock-check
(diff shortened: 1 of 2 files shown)


agorajek:
@cwegener thank you for taking care of this. This is still a black box to me.

cwegener:
> @cwegener thank you for taking care of this. This is still a black box to me.

Right now, I am basically pretending to be a new Datajunction contributor as much as a I can. As part of that journey, I'm fixing the main things in the tooling that are broken as a result from the monorepo move.

Ultimately (as Sam has alluded to earlier), there will need to be an update to a Contributing Guide in the Developer docs that makes it smooth and easy for new developers to get productive in the quickest way possible.

agorajek:
> Right now, I am basically pretending to be a new Datajunction contributor as much as a I can. As part of that journey, I'm fixing the main things in the tooling that are broken as a result from the monorepo move

@cwegener

shangyian (approved):
Thank you!! This is awesome

Issue #630, opened 2023-07-12
Cleanup Python client classes and methods.


agorajek:
Here is the current state:

- **_internal.py** - all the stuff that users should never see
- **exceptions.py, models.py** - self-explanatory
- **client.py**
    - DJClient class (basic DJ access)
        - list_[namespaces | dimensions | metrics | cubes] 
           ... list_[sources | transforms | nodes(type)] (**TODO low-pri**)
        - common_[dimensions | metrics]
        - sql() | data()
    - DJBuilder class (building DJ dag)
        - list|get|add|delete : engine|catalogs (add|delete **TODO**)
        - namespace & create|delete|restore_namespace 
        - source|transform|metric|dimension|cube --> return an existing Node
        - create_[source|...] --> create a new Node (in DJ) and return it
        - delete_node(name) --> delete a node by name
        - restore_node(name) --> restore a node by name
- **node.py**
    - Namespace class
        - nodes|sources|transforms|cubes --> get nodes from a namespace, not sure if we need this
    - Node class
        - save() --> creates or updates an existing Node
        - refresh() --> refreshes current object from DJ
        - [un]link_dimensions()
        - add_materialization(), add_availability(), set_column_attributes()
        - list_revisions(), 
    - some other classes...

And few things are left to do:
- [x] Add list_dimension() method for all dimensions.
- [x] Adjust the Node [de]activate to delete/restore (https://github.com/DataJunction/dj/issues/674)
- [x] Add delete/restore methods to Namespace 
    - server side: https://github.com/DataJunction/dj/issues/682
    - client side: COMING...
- [x] Adjust client README to the new code.
- [x] Adjust Demo notebook the the new code.
- [x] Review remaining TODOs

Maybe also:
- [ ] Hide all-the-protected-methods in _internal.DJClient better, cause now they are still visible in the user client classes.

agorajek:
Couple things to do:

1. Split the **DJClient** class into:
  - base/internal.py : _DJClient
  - client.py : DJClient(_DJClient) for DJ user/consumption access
  - client.py : DJBuilderClient(DJClient) for DJ developer/contribution access

2. In **DJClient**: keep things simple:
  - sql(), data()
  - dd a generic `list_<something>` methods for namespaces, dimensions and metrics, to search and list them
  - common_dimensions(), common_metrics()

3. In **DJBuilderClient** also keep things simple :)
  - list_catalogs|engines|namespaces|nodes
[...]

PR #635, opened 2023-07-14
[WIP] Cleanup Python client and set an example for other clients.

agorajek:
### Summary
This is a start to cleanup the client classes and methods according to the plan in #630. Most work is still todo, but just wanted to kick off a PR for a quick review.

### Test Plan
make check + make tests + more

- [ ] PR has an associated issue: #630 
- [ ] `make check` passes

### Deployment Plan
n/a


client/python/datajunction/base.py
```diff
@@ -0,0 +1,1135 @@
...
+        """
+        Allow arbitrary types to support _DJClient but exclude
+        it from the output.
+        """
+
+        arbitrary_types_allowed = True
+        exclude = {"dj_client"}
+
+
+class Node(ClientEntity):
```
samredai, line 774: This class and all of the classes after this (Source, Transform, Dimension, etc..) can all start with an underscore, right? The user should never import them and instantiate them directly. They should always go through the client methods.

```py
dj.source(name: str) -> _Source
```

client/python/datajunction/client.py
```diff
@@ -1,1135 +1,15 @@
...
-            catalog=self.catalog,
-            schema_=self.schema_,
-            table=self.table,
-            columns=self.columns,
-        )
-        return self.dj_client.update_node(self.name, update_node)
-
 
-class NodeWithQuery(Node):
+class DJBuilderClient(DJClient):
```
shangyian, line 12: Other naming options:
- `DJDeveloperClient`
- `DJModelerClient`
- `DJWriteClient`

samredai: Looking at that, I wonder if we could drop the DJ prefix since the import makes it clear where it's coming from. That might make it a little cleaner, i.e. `from datajunction import ModelerClient`

client/python/datajunction/base.py
agorajek: Yes, I was considering that and now that you said I'll just do it. I may also rename the `base.py` to `internal.py`. What do you all think?

client/python/datajunction/client.py
agorajek: @shangyian :
- DJDeveloperClient  --> I considered that but it seems too far on the SWE side
- DJModelerClient --> yes, that's a good alternative to Builder
- DJWriteClient --> not sure if you meant Writer, but either way it feels "not enough"

@samredai, yes I considered doing that but, I think the prefix does not hurt, and it make the imported module "speak for itself". I think that a DJClient module/class or DJModelerClient maintain the context better.

Actually, what do you all think about dropping the Client suffix:
- **DJClient** for the consumption access
- **DJBuilder** or **DJModeler** for the contribution side

Alternatively: **DJReader** and **DJWriter** - hard to go wrong on this one. ;)

shangyian: Hmm, yeah I like `DJReader` and `DJWriter`!

samredai: +1 to `DJReader` and `DJWriter`

shangyian (approved):
Thanks for revamping the client! 🥳 

Just left a few comments but overall it LGTM

client/python/datajunction/client.py
```diff
@@ -570,81 +67,63 @@ def common_dimensions(
...
-                "metrics": metrics,
-                "dimensions": dimensions,
-                "filters": filters,
-                "engine_name": engine_name or self.engine_name,
-                "engine_version": engine_version or self.engine_version,
-            },
-        )
+        Builds SQL for one (or multiple) metrics with the provided dimensions and filters.
+        """
+        if len(metrics) == 1:
```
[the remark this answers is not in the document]
agorajek: Agree. Will fix.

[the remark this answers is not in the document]
samredai: We don't use single metric build anywhere right? Can we just remove that endpoint and always use the endpoint also capable of multi-metric build?

client/python/datajunction/__init__.py
```diff
@@ -3,16 +3,17 @@
 """
 from importlib.metadata import PackageNotFoundError, version  # pragma: no cover
 
-from datajunction.client import (
+from datajunction._internal import (
```
samredai, line 6: The way this import is will unhide all of the internal classes here (like in the below example). It's probably a pain to underscore all of the classes so I wonder if just doing `from datajunction import _internal` and then using `_internal.Cube` in this file will keep them hidden.
```py
import datajunction
for i in dir(datajunction):
  print(i)
```
_output_
```
AvailabilityState
ColumnAttribute
Cube
DIST_NAME
DJReader
DJWriter
Dimension
Engine
MaterializationConfig
Metric
Namespace
Node
NodeMode
Source
Transform
_DJClient
__all__
__builtins__
__cached__
__doc__
__file__
__loader__
__name__
__package__
__path__
__spec__
__version__
_internal
client
exceptions
models
```

client/python/datajunction/_internal.py
```diff
@@ -0,0 +1,916 @@
...
+        namespace: str,
+    ) -> "Namespace":
+        """
+        Creates a new namespace on the server
+        """
+        _namespace = Namespace(dj_client=self, namespace=namespace)
+        self.create_namespace(_namespace)
+        return _namespace
+
+    def verify_node_exists(self, node_name: str, type_: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
samredai: Oh I see, this is really just an internal helper method used by the other methods. So in that case this should have a leading underscore so it's clear users shouldn't use this directly.

client/python/datajunction/_internal.py
```diff
@@ -0,0 +1,916 @@
...
+        node_copy = node.dict().copy()
+        node_copy["mode"] = models.NodeMode.PUBLISHED
+        response = self._session.post(
+            "/nodes/validate/",
+            json=node_copy,
+            timeout=self._timeout,
+        )
+        return response.json()
+
+    def create_node(self, node: "Node", mode: models.NodeMode):
```
samredai, line 424: This should also be `_create_node`, otherwise it looks very usable :)

samredai (approved):
Thanks Olek! This looks great. A lot of comments but they're mostly nits and almost entire just pointing out methods that should have underscores. I remembered midway that you're doing write in a follow-up PR so feel free to ignore those. The general pattern I noticed though is that there are internal methods that take a `node_name` argument on the client but then are called from a method on one of the Node class definitions. All of the versions of those methods on the client should probably start with an underscore and the user path should be to get an instantiated node object and use the methods on it.

client/python/datajunction/__init__.py
agorajek: Yeah, I haven't gotten to this point yet. I only cleaned up the names I moved to the new client.py so far. But I like your suggestion on `from datajunction import _internal`. Will consider that!

client/python/datajunction/client.py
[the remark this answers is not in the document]
agorajek: > Can we just remove that endpoint and always use the endpoint also capable of multi-metric build?

It crossed my mind but then I thought that endpoint is really nice if you just want to look at one metric w/o dimensions and/or filters.

client/python/datajunction/client.py
```diff
@@ -570,81 +67,63 @@ def common_dimensions(
...
-                "filters": filters,
-                "engine_name": engine_name or self.engine_name,
-                "engine_version": engine_version or self.engine_version,
-            },
-        )
-        if response.status_code == 200:
-            return response.json()["sql"]
-        return response.json()
+        # TODO  # pylint: disable=W0511
+        # Add this to the server APIs first.
```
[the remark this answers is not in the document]
shangyian: @agorajek added here: https://github.com/DataJunction/dj/pull/657

PR #636, opened 2023-07-16
Making errors check include a null-check

samredai:
### Summary

Small change that checks if errors is null when the query info is returned from a query service. Otherwise this would error our if a successful query service response doesn't send a placeholder empty errors list.

Also updated the duckdb demo database and added some more example data.

### Test Plan

`docker compose --profile demo up` and ran some queries in the UI.

- [ ] `make check` passes

### Deployment Plan

N/A


CircArgs (approved):

samredai:
(sneaking in an update to the duckdb demo database)

PR #638, opened 2023-07-17
Add ability to refresh source nodes with the latest columns from the …

shangyian:
### Summary

Users should be able to refresh a source node with the latest columns via an API call, rather than waiting for the reflection service to do so. This PR adds the functionality to the `PATCH /nodes/{name}/` endpoint, where passing in `refresh_if_source=true` will refresh the source node with up-to-date table columns from the query service.

This also fixes a two bugs with source nodes:
* dimension links weren't being propagated during node updates
* if a primary key was set during node update, it wasn't reflected in the result

### Test Plan

Will add a few more tests.

- [ ] PR has an associated issue: 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -1493,7 +1493,6 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
     node: Node,
     data: UpdateNode = None,
     version_upgrade: VersionUpgrade = None,
-    refresh: bool = True,
 ) -> Optional[NodeRevision]:
     """
     Creates a new revision from an existing node revision.
@@ -1518,7 +1517,6 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
         data is not None
         and data.columns is not None
         and ({col.identifier() for col in old_revision.columns} != data.columns)
-        or refresh
     )
     pk_changes = (
         data is not None
@@ -1568,8 +1566,8 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
         if data and data.columns
         else old_revision.columns,
         catalog=old_revision.catalog,
-        schema_=old_revision.schema_,
-        table=old_revision.table,
+        schema_=data.schema_ or old_revision.schema_,  # type: ignore
+        table=data.table or old_revision.table,  # type: ignore
         parents=[],
         mode=data.mode if data and data.mode else old_revision.mode,
         materializations=[],
@@ -1614,17 +1612,15 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
 
     # Refresh the columns of the source node using the query service if no
     # columns were specified in the update object
-    if (
-        refresh
-        and old_revision.type == NodeType.SOURCE
-        and (data and not data.columns or not data)
+    if old_revision.type == NodeType.SOURCE and (
+        not data or (data and not data.columns)
     ):
         new_revision.columns = [
             Column(name=col.name, type=col.type)
             for col in query_service_client.get_columns_for_table(
-                old_revision.catalog.name,
-                old_revision.schema_,  # type: ignore
-                old_revision.table,  # type: ignore
+                new_revision.catalog.name,  # pylint: disable=no-member
+                new_revision.schema_,  # type: ignore
+                new_revision.table,  # type: ignore
             )
         ]
 
@@ -1674,7 +1670,6 @@ def update_a_node(
     name: str,
     data: UpdateNode,
     *,
-    refresh: bool = False,
     session: Session = Depends(get_session),
     query_service_client: QueryServiceClient = Depends(get_query_service_client),
 ) -> NodeOutput:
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1486,13 +1486,14 @@ def tag_a_node(
...
 
-def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-many-arguments
+def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-many-arguments,too-many-branches
     session: Session,
     query_service_client: QueryServiceClient,
     old_revision: NodeRevision,
     node: Node,
     data: UpdateNode = None,
     version_upgrade: VersionUpgrade = None,
+    refresh_if_source: bool = True,
```
samredai, line 1496: Since there are longer term plans for reflection to do other things like update materialization information, valid through timestamps, and maybe distinct values for dimensions, would this arg be better as just `refresh: bool`?

shangyian: Makes sense. Let me update this.

shangyian: @agorajek Yeah, good point. I think the only complications are around conflict resolution between the custom columns and the actual columns. The logic here for source nodes would be:
* Retrieve latest columns from query service
* If there are any custom columns passed in, filter down to only those custom columns or raise an error if the custom ones don't match the ones from the query service
* If there are no custom columns, just update the source node's columns to the latest from the query service
* Propagate any dimension links defined on the old source node revision -> it looks like the custom columns allow for adding dimension links, but I don't think we should allow that because it adds to the confusion
* Propagate any primary keys defined on the old source node revision OR update to new primary keys if provided in the update config

shangyian:
@agorajek and @samredai Should we just always assume that a query service is configured? I'm trying to think through this scenario from the user perspective:
* A query service is configured:
   * The user would just ask the query service for columns on the source.
   * If the user needs to select specific columns for a source node, they can just provide a list of column names. There's no reason to include the column types.
* A query service is not configured:
   * The user always needs to provide the column names + types when creating a source node. 
   * A lot of crucial DJ functionality is disabled, as users can't retrieve data, keep the system up-to-date (via reflection) or materialize anything, which makes me question if we should allow this...

If we always assume that a query service is configured (and raise an error if not), a lot of the complications around keeping the source node update config in sync with the latest columns from the query service goes away.

datajunction-server/datajunction_server/api/nodes.py
agorajek: @shangyian I am a bit puzzled by the "custom columns" term you used. Maybe we can try to chat briefly today to clarify?

shangyian: @agorajek I meant that when you create a source node, you can either provide the column names + types or rely on the query service to tell you what they are

agorajek: I see, yeah, I don't see much value in the user telling the system how the source nodes look like.

shangyian:
@agorajek I've updated the PR to always refresh source nodes with the latest columns from the query service during node updates. If we're all aligned on the source nodes always sourcing their columns from the query service, I can make that change in a separate PR. Most of the logic that's in this one will remain intact.

samredai:
> @shangyian 100% in my opinion. I don't see much value in the platform if we are doing this in separation from a DW. Access to a DW is critical imo, and hence the query service (and reflection one, if separate from query).

I think I disagree. It's not completely unreasonable to find value in just the SQL generation capabilities, right? Assuming someone is willing to execute the SQL "on their own" without the assistance of a DJ query service. To ask the question another way, would we raise an error when someone requests SQL for a set of metrics and dimensions just because they don't have a query service configured?

agorajek:
> I think I disagree. It's not completely unreasonable to find value in just the SQL generation capabilities, right? Assuming someone is willing to execute the SQL "on their own" without the assistance of a DJ query service. To ask the question another way, would we raise an error when someone requests SQL for a set of metrics and dimensions just because they don't have a query service configured?

@samredai once the reflection service is a separate thing, I think we could consider a global DJ setting e.g. SOURCE_NODE_MONITORING=on|off, and base off this switch we could add a skip to that auto-refresh, but I still think the auto-refresh mechanism makes sense as a default.

samredai:
Still a draft but I think we should look at this PR alongside PR #645

agorajek:
@shangyian I think the functionality of this PR is good, but from what we talked about recently we probably just want:
- for the managed sources nodes to be monitored by the reflection system
- for the manually create source nodes to not have any sync with reality and no need to refresh their schema

shangyian:
> for the managed sources nodes to be monitored by the reflection system

@agorajek agreed, but what do you think about having a `POST /nodes/{name}/refresh` endpoint that the reflection service can call which will update the managed source nodes to the latest columns? I think that'll be cleaner than having the reflection service post updated columns directly to a source node.

agorajek:
> Instead of doing that, I think we can have the reflection service call: POST /nodes/{name}/refresh. This way, the interface is cleaner and someone can perform "manual" reflection by calling the same endpoint

Agree.

shangyian:
Closing in favor of https://github.com/DataJunction/dj/pull/659

Issue #643, opened 2023-07-18
Reactivation bug for nodes with cubes downstream


shangyian:
To reproduce this bug:
* Create a cube node
* Deactivate any upstream node that this cube depends on
* Reactivate the same node

1. This reactivation will fail because it's unable to revalidate the state of the cube
2. Even if this reactivation succeeds, the cube's elements are modified such that the metric columns no longer reference the right node revisions.

PR #644, opened 2023-07-18
Fix node reactivation bug (where downstream is a cube)

shangyian:
### Summary

* Let's say a metric node that gets referenced by a cube is deactivated. When the node is reactivated, we call `validate_node_data`, which creates a new column (for the metric's single column reference) and saves the node referencing that new column. However, the way that cubes keep track of metrics and dimensions is through the `Column` reference, and now this old column reference will no longer exist, which breaks the cube.
* A cube node should be considered valid when the nodes referenced by its cube elements (all metrics and dimensions) are valid. This modifies the revalidation of cubes to do that.

### Test Plan

- [ ] PR has an associated issue: #643 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -373,16 +373,22 @@ def validate_node_data(  # pylint: disable=too-many-locals,too-many-branches
     # Add aliases for any unnamed columns and confirm that all column types can be inferred
     query_ast.select.add_aliases_to_unnamed_columns()
 
-    validated_node.columns = []
+    new_columns = []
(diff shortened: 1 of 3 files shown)


agorajek (commented):
I am having a hard time

shangyian:
@agorajek So the core of this PR is that I don't think we should be creating new columns when we reactivate a node. It should just be changing the status of the node and reusing the old columns. (That was also what caused the bug)

shangyian:
@samredai I've updated this so that it's also does the type inference check + update when we're trying to revalidate a node.

samredai (approved):
Thanks for the fix! 🎉

PR #645, opened 2023-07-19
global setting for source_node_namespace and /register/table/... endpoint

samredai:
### Summary

It's been pretty confusing to differentiate between manually created source nodes (i.e., user may be manually modeling a table that hasn't been created in the warehouse yet) and source nodes that should be under constant monitoring from the query service. This adds two things to make this a bit easier.

#### `SOURCE_NODE_NAMESPACE` server setting
This setting defaults to `"sources"` and determines the namespace where all reflection managed source nodes will be created. Users can still manually create source nodes in any other namespace in the system, but those won't be monitored via reflection.

#### `POST /register/table/{catalog}/{schema_}/{table}/` endpoint
This endpoint is meant to be an alternative to manually creating a source node through `POST /nodes/source/`. Unlike the manual endpoint, this endpoint does not allow you to choose the node name, namespace, description or provide the columns. It will instead create a source node named `f"{settings.source_node_namespace}.{catalog}.{schema_}.{table}"` and the schema will automatically be set through reflection and only works when a query service is configured.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -1123,7 +1123,6 @@ def save_node(
 def create_a_source(
     data: CreateSourceNode,
     session: Session = Depends(get_session),
-    query_service_client: QueryServiceClient = Depends(get_query_service_client),
 ) -> NodeOutput:
     """
     Create a source node. If columns are not provided, the source node's schema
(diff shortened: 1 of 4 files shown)


agorajek (approved):
I like this PR. I'd go with `source` as the default namespace but the rest looks good.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1169,7 +1172,7 @@ def create_a_source(
         if not query_service_client:
             raise DJException(
                 message="No table columns were provided and no query "
-                "service is configured for table columns inference!",
+                "service is configured for table columns inference",
```
shangyian, line 1175: Should the `create_a_source` endpoint be modified to only work for source nodes where the columns are manually managed? Then it's very clear that in one case the columns are managed via a reflection/query service, but in the other case the columns are specified manually and there's no need for the server to attempt to resolve between the two cases.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1301,6 +1304,45 @@ def create_a_cube(
...
+        )
+    namespace = f"{settings.source_node_namespace}.{catalog}.{schema_}"
+    name = f"{namespace}.{table}"
+    create_a_node_namespace(namespace=namespace, session=session)
+    return create_a_source(
+        data=CreateSourceNode(
+            catalog=catalog,
+            schema_=schema_,
+            table=table,
+            name=name,
```
shangyian, line 1337: Should the "display name" also be explicitly set to `f"{catalog}.{schema_}.{table}"`?

samredai: I like that, it'll make things a lot more clear. I'll update this

samredai: Working on this now, would that be misleading since users have to use the source node namespace prefix? In other words, should display name be the same as name--`f"{settings.source_node_namespace}.{catalog}.{schema_}.{table}"`?

agorajek: Yes, I think that the default display name should basically be the full node name. But I am not sure if we need to populate it in the DB or just pull it in the UI from the name, when display_name is empty.

shangyian: Yeah +1 to the display name being the full node name. I think it's better for the database to store the display name, since the overhead is quite small and otherwise we'll rely too much on that logic being there in the UI.

samredai: Ok I updated this so the name and display will be the same for nodes created through the register table endpoint.

datajunction-server/datajunction_server/api/nodes.py
samredai: Updated this and also made columns required for the manual source node creation path through `/nodes/source/`. It seems pretty clear and intuitive now. Something we might want to eventually add is a check across all node creation to make sure nobody creates nodes in the `SOURCE_NODE_NAMESPACE`. I doubt this will be that big of a problem but it'll lead to cleaner reflection logic in the query service.

PR #646, opened 2023-07-19
Add history for node status changes and dimension link removal

shangyian:
### Summary

More history tracking:
* Node status switched to valid/invalid by the system (there are quite a few of these, but hopefully with this we'll be able to track where some of the node invalidation is occurring)
* Dimension link removed

### Test Plan

- [x] PR has an associated issue: #624
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -380,6 +380,7 @@ def deactivate_a_node(name: str, *, session: Session = Depends(get_session)):
                     downstream.current,
                     NodeStatus.VALID,
                     NodeStatus.INVALID,
+                    parent_node=node.name,
                 ),
             )
             session.add(downstream)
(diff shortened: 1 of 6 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -170,14 +183,23 @@ def revalidate_a_node(
...
+            ),
+        )
         session.commit()
     return JSONResponse(
         status_code=HTTPStatus.OK,
         content={
             "message": (
                 f"{current_node_revision.type} node `{current_node_revision.name}` "
-                "has been set to {status}"
+                f"has been set to {status}"
```
samredai, line 202: Thanks for catching this!

samredai (approved):
This is going to make it waaay easier to understand changes, thanks @shangyian 🚀

agorajek (approved):
Oh, that's cool. Yeah, I have experienced that as well.

PR #647, opened 2023-07-19
feat:allow queryservice be hosted at non-root path

cwegener:
The RequestsSessionWithEndpoint builds a URI from a base URL + a path.

Currently in all cases of the RequestsSessionWithEndpoint being used in
the QueryServiceClient, the paths that are being passed in are all
absolute paths.

By passing in absolute paths, it makes it impossible to use a base URL
that also includes a path (e.g. http://localhost:8000/qs/), because
urllib.parse.urljoin will create the new full URI by substituting the
path of the base URL (e.g. /qs/) with the full absolute path that is
passed in to the RequestsSessionWithEndpoint constructor.

By passing in relative paths to the RequestsSessionWithEndpoint
constructor, the ability to host the query service on a base URL with a
path is now possible.

### DIFF
--- a/datajunction-server/datajunction_server/api/authentication/basic.py
+++ b/datajunction-server/datajunction_server/api/authentication/basic.py
@@ -72,11 +72,17 @@ async def login(
     response.set_cookie(
         DJ_AUTH_COOKIE,
         create_token({"username": user.username}, expires_delta=timedelta(days=365)),
+        domain="djui.fly.dev",
+        secure=True,
+        samesite="none",
         httponly=True,
     )
     response.set_cookie(
         DJ_LOGGED_IN_FLAG_COOKIE,
         "true",
+        domain="djui.fly.dev",
+        secure=True,
+        samesite="none",
     )
     return response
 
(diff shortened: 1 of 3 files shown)


samredai:
Ah nice catch, thanks @cwegener!

shangyian (approved):
Nice! Thanks for fixing

cwegener:
> Ah nice catch, thanks @cwegener! Looks like just a few paths in some tests need to be updated and this should be good to go. rocket

Yeah. I was trying to understand what actual code paths the tests were meant to be testing. :sweat_smile: 
I could just do the same change to test class to make the tests pass again, I guess. But I feel there is a larger issue at foot.

The same RequestsSessionWithEndpoint code is also copied in the python client, but with an even more complex implementation. I'll ponder this one a little bit and update shortly.

agorajek:
@cwegener it looks like your PR has been in draft mode for a while. Are you planning to work on it?

cwegener:
Yeah. Let me make the tests succeed again and then it should be good to be used.

Issue #639, opened 2023-07-17
Server-Sent Events for running DJ queries


samredai:
I want to add a way to query DJ for data that uses server sent events but I want to hear thoughts on if this should be optional and if so, how should it be turned on/off.

Query running in DJ has always had an async option but there’s no way to take advantage of it in a UI layer without doing some form of polling for a terminal state (canceled, finished, failed). With an SSE endpoint that’s mono-directional, events are sent to the client over http whenever the query state changes, meaning a UI could just make one request and process the state changes as they stream over (accepted, scheduled, running, etc).

Main questions are:
1. Any reason why we shouldn’t do this?
2. Should it be optional?
3. What should the endpoint be called? (`/stream/` is a common one)
4. How do we handle this in the OSS UI?

For 4 regarding the OSS UI, I think using standard requests works nicely for small data that comes back in less than a few seconds. So I think that should be the default, essentially with async disabled completely. Then I was thinking maybe we can have a global UI setting such as `USE_SSE` which would switch the UI to using the stream endpoint and display the changes to the query state as they’re sent over.

shangyian:
I think we should definitely do this! I delayed adding polling because I suspected we would want this at some point.

I do wonder if instead of having `USE_SSE` as a global setting, this should be enabled on a case-by-case basis for each engine type. For example, the Druid or Postgres engines might not use this, but the Trino or Spark ones would. This can be optionally configured when the DJ server admin adds engines to the system, with some sort of sensible default.

Of course, all of that depends on the actual latency overhead of just using SSE for all requests. If it's not particularly high, then it wouldn't need to be optional and we can just enable it for all engines.

agorajek:
+1 on making this query pattern a setting on the engine. This makes a lot of sense also because the query service setup will vary between user installations.

betodealmeida:
I agree, for context in Superset we have a per-DB async/sync setting. For sub-second DBs like Druid the async overhead is usually bigger than the query response time, so running queries in sync mode is actually faster.

How is the support for SSE these days? I remember a few years ago we wanted an async solution for Preset and we investigated websockets vs SSE, and because SSE requires HTTP 2.0 it wasn't supported in all proxies (IIRC). We ended up using websockets,  but that brings other problems, and SSE seems simpler.

cwegener:
> How is the support for SSE these days?

[...]

PR #649, opened 2023-07-20
Add server sent event /stream/ endpoint and use in UI

samredai:
### Summary

Currently, we have two options when submitting queries to the DJ backend. They can either be submitted sync or async. If they're submitted sync, the UI works fine, showing a loading experience and then shows the results after the query request is fully complete. We can also switch to async, however the UI will only show the initial state of the query ("ACCEPTED") since we haven't implemented any sort of polling to get updates as the query state changes. This PR adds a third option--a `/stream/` endpoint that's complementary to the `/data/` endpoint and streams server sent events (SSE) to the client.

The UI has also been updated to process these events as they're received. The result is that you get a realtime feel when running a query, seeing the query info box change as the query state changes. If any links are included in the query state, you can see those links immediately, even when the query might be in the "RUNNING" state for a long time.

The change is not too obvious in the video below since the queries against the duckdb demo database complete very quickly, but you can see momentarily that it starts in the "ACCEPTED" state and updates as the server sends subsequent events. This would be more obvious in a longer running query.

https://github.com/DataJunction/dj/assets/43911210/09b4a13c-97b1-4e14-95fc-e3ec6e135fd2

### Test Plan

- [x] PR has an associated issue: #639 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -60,8 +60,6 @@ from datajunction_server.typing import END_JOB_STATES
 
 _logger = logging.getLogger(__name__)
 
-logger = logging.getLogger(__name__)
-
 
 def get_node_namespace(  # pylint: disable=too-many-arguments
     session: Session,
@@ -403,7 +401,7 @@ def validate_node_data(  # pylint: disable=too-many-locals
 
     # Only raise on missing parents if the node mode is set to published
     if missing_parents_map and validated_node.mode != NodeMode.DRAFT:
-        logger.error(
+        _logger.error(
             "Node %s missing parents %s",
             validated_node.name,
             list(missing_parents_map.keys()),
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -11,7 +11,7 @@ import uuid
 from http import HTTPStatus
 from typing import Dict, List, Optional, Set, Tuple, Union
 
-from fastapi import HTTPException, Request
+from fastapi import HTTPException
 from sqlalchemy.exc import NoResultFound
 from sqlalchemy.orm import joinedload
 from sqlalchemy.sql.operators import is_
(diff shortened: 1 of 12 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -821,3 +827,65 @@ def build_sql_for_multiple_metrics(  # pylint: disable=too-many-arguments,too-ma
...
+                query_next.results.__root__[0].columns = columns or []
+            yield {
+                "event": "message",
+                "id": uuid.uuid4(),
+                "retry": retry_timeout,
+                "data": json.dumps(query_next.json()),
+            }
+            _logger.info("connection closed by the server")
+            break
+        if query_prev != query_next:  # pragma: no cover
```
shangyian, line 879: What's an example of a case where it reaches this part? Is this for handling when the query's state changes but it hasn't reached one of the end states?

shangyian (approved):
Exciting, thanks for adding this change! 🎉

agorajek (approved):
dog_science
But I like the results!

datajunction-ui/src/app/pages/SQLBuilderPage/index.jsx
```diff
@@ -46,20 +45,39 @@ export function SQLBuilderPage() {
...
+        );
+        setQueryInfo(response);
+        if (response.results) {
+          setLoadingData(false);
+          setData(response.results);
+          response.numRows = response.results?.length
+            ? response.results[0].rows.length
+            : [];
+          setViewData(true);
+          setShowNumRows(10);
```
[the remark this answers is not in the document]
samredai: Updated this to 100 and created a `DEFAULT_NUM_ROWS` constant at the top since it's set in 3 places. The reason this has to be set separately from the useState hook default is so that the number of rows can be reset after a new query is ran.

datajunction-server/datajunction_server/api/helpers.py
samredai: Yeah that's exactly right. The server constantly checks on the query state on an interval defined by `stream_delay` (defaults to half a second). But this prevents the server from sending an event to the client when the data is exactly the same as the last event. The usual flow of a successful query would then be 3 events of `ACCEPTED`, `SCHEDULED`, and `FINISHED`.

Issue #652, opened 2023-07-20
More context around type inference errors 


shangyian:
Currently the context is just a single column name, but in a longer and more involved query, it can be difficult to tell what part of the query it's referring to. We should provide a larger snippet when giving the user this error message, perhaps by including the column's parent expression.

PR #653, opened 2023-07-20
More context for type inference failure

shangyian:
### Summary

Adds more context around a column's type inference failure by pulling the parent node's expression.

### Test Plan

- [X] PR has an associated issue: #652 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -684,7 +684,7 @@ class Column(Aliasable, Named, Expression):
             self.add_type(self.expression.type)
             return self.expression.type
 
-        raise DJParseException(f"Cannot resolve type of column {self}.")
+        raise DJParseException(f"Cannot resolve type of column {self} in {self.parent}")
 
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -684,7 +684,7 @@ def type(self):
             self.add_type(self.expression.type)
             return self.expression.type
 
-        raise DJParseException(f"Cannot resolve type of column {self}.")
+        raise DJParseException(f"Cannot resolve type of column {self} in {self.parent}")
```
samredai, line 687: What about something like:
```py
parent = f"in {self.parent}" if self.parent else "that has no parent"
raise DJParseException(f"Cannot resolve type of column {self} {parent}")
```
Then the example you updated in the test would be "Cannot resolve type of column status that has no parent".

samredai (approved):
Looks good! Just left one suggestion

datajunction-server/datajunction_server/sql/parsing/ast.py
shangyian: Good point, just updated!

PR #657, opened 2023-07-21
Add functionality to get nodes with a set of common dimensions

shangyian:
### Summary

This adds functionality to get all nodes that have a set of common dimensions, which was referenced recently in https://github.com/DataJunction/dj/pull/635#discussion_r1269497074. It expands on the `get_nodes_with_dimension` function, which works for a single dimension, to a set of dimensions. I added a new endpoint at `GET /dimensions/common` to expose this feature.

### Test Plan

- [x] PR has an associated issue: #321 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -134,6 +134,8 @@ def get_shared_dimensions(
         # Merge each set of dimensions based on the name and path
         to_delete = set()
         common_dim_keys = common.keys() & list(node_dimensions.keys())
+        if not common_dim_keys:
+            return []
         for common_dim in common_dim_keys:
             for existing_attr in common[common_dim]:
                 for new_attr in node_dimensions[common_dim]:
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -207,3 +207,25 @@ def get_nodes_with_dimension(
...
+    first = True
+    for dimension in common_dimensions:
+        new_nodes = get_nodes_with_dimension(session, dimension, node_types)
+        if first:
+            nodes_that_share_dimensions = set(new_nodes)
+            first = False
+        else:
+            nodes_that_share_dimensions = nodes_that_share_dimensions.intersection(
+                set(new_nodes),
+            )
```
agorajek, line 230: I think that we could exit this loop at the end of any iteration if the `nodes_that_share_dimensions` is empty. Otherwise we may be processing more nodes w/o any chance for success.

agorajek (approved):
Thanks @shangyian . I will add it to the Python client side today. 

Just left one comment in-line. A similar optimization could also benefit on the dimensions with common metrics side.

PR #658, opened 2023-07-22
Add common_metrics() to Python client.

agorajek:
### Summary

And while working on this I found some issues with the demo DB. E.g. `s/metric_bound_dimensions/metric_required_dimensions` needed to be fixed.

Also python integration test was failing so wanted to show you this and make sure I am not missing anything else.

### Test Plan

- [ ] `make check` passes

### Deployment Plan

n/a


client/python/datajunction/client.py
```diff
@@ -54,7 +54,7 @@ def list_metrics(self, namespace: Optional[str] = None) -> List[str]:
     def common_dimensions(
         self,
         metrics: List[str],
-    ):  # pragma: no cover # Tested in integration tests
+    ) -> List:  # pragma: no cover # Tested in integration tests
```
samredai, line 57: Should this be `List[str]`?

client/python/datajunction/client.py
```diff
@@ -69,12 +69,17 @@ def common_dimensions(
...
         """
-        # TODO  # pylint: disable=W0511
-        # Add this to the server APIs first.
+        query_params = [("node_type", models.NodeType.METRIC.value)]
+        for dim in dimensions:
+            query_params.append(("dimension", dim))
+        response = self._session.get(
+            f"/dimensions/common/?{urlencode(query_params)}",
+        )
+        return [metric["name"] for metric in response.json()]
```
samredai, line 82: Is this another case where we need to update the server response type to just be a list of strings?

client/python/datajunction/client.py
```diff
@@ -69,12 +69,17 @@ def common_dimensions(
...
         """
         Return common metrics for a set of dimensions.
         """
-        # TODO  # pylint: disable=W0511
-        # Add this to the server APIs first.
+        query_params = [("node_type", models.NodeType.METRIC.value)]
+        for dim in dimensions:
+            query_params.append(("dimension", dim))
+        response = self._session.get(
+            f"/dimensions/common/?{urlencode(query_params)}",
```
shangyian, line 80: if you want you can actually just do this here:
```
self._session.get("/dimensions/common/", params=query_params)
```

client/python/datajunction/models.py
```diff
@@ -42,6 +42,14 @@ class NodeStatus(str, enum.Enum):
     INVALID = "invalid"
 
 
+class NodeType(str, enum.Enum):
+    """
+    DJ node types
+    """
+
+    METRIC = "metric"
```
shangyian, line 50: Hmm, does this need all the other node types too?

samredai, line 50: For completeness I think we should add all node types to this enum. In the function where `models.NodeType.METRIC.value` there's also `"dimension"` there that could be changed to `models.NodeType.DIMENSION.value` for consistency. I also think it's reasonable to just hardcode the values and get rid of this enum all together, only because the client code is pretty thin.

shangyian (approved):
Yay, thanks for adding this! 🎉

samredai (approved):
Looks good, thanks @agorajek! Left a couple of comments/opinions.

client/python/datajunction/client.py
shangyian: I was considering this, but then it'll blow up our calls to the backend from the UI when we call this endpoint to display linked nodes on a dimension. Right now we just do this single call, which will return all the info it needs (i.e., display name, node type etc), but otherwise we'd have to make one call per node that it gets back from this endpoint. Of course, if we lean into GraphQL this problem goes away...

samredai: I see, makes sense. Maybe we can put the graphql stuff back but for a subset of the endpoints, for example only for the GET endpoints.

client/python/datajunction/client.py
agorajek: I thought so initially but then I realized we are returning dimension names with paths so this really is a `List[Dict]` right now.

client/python/datajunction/client.py
agorajek: Sorry @shangyian I don't understand your comment.

client/python/datajunction/client.py
agorajek: Yes, I was having the same conversation in my head as you two just above. My opinion would be to:
- on all APIs that may return all metrics, limit the type to list of names
- use/add extra APIs to retrieve the details of a metric/node if the client needs it
- for GraphQL possible have a separate set of APIs, caveat: I am not an expert in these (graphql) standards.

client/python/datajunction/models.py
agorajek: Yes, of course. I just missed to fill the rest... Will fix.

client/python/datajunction/client.py
shangyian: @agorajek I just meant that you could replace 
```
response = self._session.get(f"/dimensions/common/?{urlencode(query_params)}")
```
with
```
response = self._session.get("/dimensions/common/", params=query_params)
```
Since using the `params` arg that's provided by the requests library is just another way of doing this vs `urlencode`. But totally optional!

client/python/datajunction/client.py
samredai: I like that pattern and agree we should keep it as consistent as we can.

client/python/datajunction/client.py
agorajek: Thanks @shangyian! I like the `params` arg better, but I'll adjust it next time.

Issue #293, opened 2023-02-03
Add a mechanism for a reflection service to plug in


samredai:
We've discussed the idea of an interface where a reflection service can plug in and ensure that source nodes (which reference external tables) are all kept up to date with tables schemas and partition information that reflect the reality of the external tables they represent.

samredai:
An open item is if this will be a push or a pull model (from the perspective of the DJ server). As I've started to look at this, I don't think a pull model is the right choice. A push model may be better but I want to propose an option that is sort of both.

### An Option

**Push**: Using just table schema as an example ("table schema" could be replaced with "partition information"), a reflection service would query the DJ server to get information on all source nodes. It would then validate all table schemas against the actual schemas that exist in the warehouse. Some schemas may not match and need to be updated. The reflection service pushes a batch of updates to source nodes defined in the DJ server (similar to how posting `AvailabilityState`s work)

**Pull**: Additionally, there are changes that are occurring in the DJ server. Source nodes are being created and updated. If the reflection service's URL is included in the DJ config, then during those creations and updates, the DJ server would explicitly ping the reflection service to pull the table schema for a given table. That would allow features such as:

- When creating a source node, just include the table(s) and we'll automatically populate the schema for the node with information from the reflection service (a feature that currently exists for tables from databases with sqlalchemy uris)
- When updating a source node to change a schema, we can include an error/warning if that doesn't reflect the current state of the real table. If the user wants to do it anyway (for example they want to migrate the source node definition first before they modify the table), we can allow it to be forced and possibly a way to disable reflection)

PR #659, opened 2023-07-22
Add endpoint to refresh source nodes

shangyian:
### Summary

This PR adds an endpoint to refresh source nodes at `POST /nodes/{name}/refresh`. It also updates the reflection service to call this endpoint every time it needs to refresh a managed source node. Users can also refresh a source node "manually" via this API call, rather than waiting for the reflection service to do so.

Related to #638, which I'll close in favor of this.

### Test Plan

- [ ] PR has an associated issue: #293 #624 
- [ ] `make check` passes


datajunction-reflection/datajunction_reflection/worker/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+@shared_task(
+    queue="celery",
+    name="datajunction_reflection.worker.tasks.reflect_source",
+    base=ReflectionServiceTask,
+)
+def reflect_source(
+    node_name: str,
+):
+    """
+    This reflects the state of the node's associated table, whether
```
agorajek, line 62: Nice! Thanks for starting this!

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1754,6 +1754,76 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
...
+        Column(name=column.name, type=column.type, node_revisions=[new_revision])
+        for column in columns
+    ]
+
+    # Keep the dimension links and attributes on the columns from the node's
+    # last revision if any existed
+    new_revision.copy_dimension_links_from_revision(current_revision)
+
+    # Point the source node to the new revision
+    source_node.current_version = new_revision.version
```
agorajek, line 1805: I wonder when we'll get the first user request like: "hey, is it possible to like, set my source node to temporarily not be validated by DJ and just stay as is until I finish some work on my table." But I am sure if we really need to worry about delivering this feature right away.

datajunction-reflection/datajunction_reflection/worker/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    logger.info(f"Refreshing source node={node_name} in DJ core")
+    settings = get_settings()
+
+    # Call the source node's refresh endpoint
+    response = requests.post(
+        f"{settings.core_service}/nodes/{node_name}/refresh/",
+        timeout=30,
+    )
+
```
agorajek, line 73: Depending on the response, this is also probably a place for an email notification to the node owner.

agorajek: Or maybe inside the API... let's think about it in the context of a more generic email notification method that we will likely need to have in a few places in DJ.

agorajek (approved):
Love it. Just added some food for though in line.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Good point - maybe we can add a `refresh: bool` setting to source nodes that defaults to true but can be set to false by the source node's owner. That way they can control if they want the node to be actioned on by the reflection service.

datajunction-reflection/datajunction_reflection/worker/tasks.py
shangyian: Agreed, I suspect email notification will be cleaner in the API, since at some point we'll have knowledge about user identity + node ownership, and then we can send emails out using that info.

Issue #502, opened 2023-05-05
Create a Java Client


samredai:

Geo881:
Hi there, I'm interested in working on the Java Client.

samredai:
@Geo881 that would be great! I added you to our weekly sync and we can discuss more on Wednesday if you're able to join.

PR #660, opened 2023-07-24
adding java client folder structure

Geo881:
### Summary

adding folder structure for java client

### Test Plan

no test is needed since its just folder structure

- [x] PR has an associated issue: #502
- [x] `make check` passes

### Deployment Plan

n/a


samredai (approved):
Thanks @Geo881! This looks good, I just left one suggestion about the package name.

EDIT: I just realized my suggestion about the package name was never posted so I left it again, haha

shangyian (approved):
Looks great, thanks for starting this effort!

client/java/settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = 'DataJunctionClient'
```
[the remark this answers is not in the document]
agorajek: +1

client/java/src/main/java/org/datajunction/client/Main.java
```diff
@@ -0,0 +1,7 @@
+package org.datajunction.client;
```
samredai, line 1: Since the site is [datajunction.io](https://datajunction.io/), this should be `io.datajunction.client`. The directory needs to be changed as well and also the group in the build.gradle would need to be `io.datajunction`.

samredai, line 1: Since the site is [datajunction.io](https://datajunction.io/), this should be `io.datajunction.client`. The directory needs to be changed as well and also the group in the build.gradle would need to be `io.datajunction`.

client/java/settings.gradle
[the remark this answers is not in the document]
Geo881: Thank you guys for the feedback

client/java/src/main/java/org/datajunction/client/Main.java
Geo881: Thanks for the feedback

Issue #640, opened 2023-07-17
Add custom function that gets substituted with the current temporal partition


shangyian:
A transform materialization can either happen incrementally or with full replacement. If it gets materialized incrementally, we need the ability to inject the current temporal partition that's being processed into the query. This is useful both in terms of clarity for users as they define their transforms and for cases like the following, where the incrementally processed node requires access to a window defined based on the current temporal partition:
```sql
SELECT
  two.user_id
  SUM(two.clicks) AS clicks_sum
FROM default.some_fact_source two
WHERE two.my_date BETWEEN date_add(<current temporal partition>, -3) AND date_add(<current temporal partition>, 1) 
GROUP BY two.user_id
```

However, instead of using variable substitution, we can add a custom function `CURRENT_TEMPORAL_PARTITION()`. Then the above query would be defined as:
```sql
SELECT
  two.user_id
  SUM(two.clicks) AS clicks_sum
FROM default.some_fact_source two
WHERE two.my_date BETWEEN date_add(CURRENT_TEMPORAL_PARTITION(), -3) AND date_add(CURRENT_TEMPORAL_PARTITION(), 1) 
GROUP BY two.user_id
```

We would substitute a different value for `current_temporal_partition` for each partition that gets processed.

agorajek:
@shangyian let's take this problem apart. The are few ways to achieve this and I think we should have a chat to pick the best one:
1. No special DJ gears: the user uses common DATE_TO_STR/INT udfs and materialization cron settings.
2. Special DJ attributes inside the materialization config. This would mean the DJ modifies the materialization query based on the attributes.
3. Special DJ udfs to make the query "identical" to the materialization query executed by DJ.

shangyian:
A few things we discussed today:

#### The Special UDF
* The special function can be split into two functions, one for date and one for timestamp. We didn't decide on an exact name, but something like `CURRENT_DATE | RUN_DATE | DESIRED_DATE | INTENDED_DATE` could work (and likewise `CURRENT_TS | ...`). Users can use this like `date_add(CURRENT_DATE(), -3)`, which would translate to `date_add(cast(20230706 as date), -3)`.
* Alternatively we can have a setup like `CURRENT_TIME('%Y-%m-%d')`, which would include a format for how the timestamp is intended to be formatted when it's substituted into the query, but this puts more "magic" into the DJ parsing side rather than relying on the user's knowledge of SQL.

#### Conditions on Transforms
[...]

PR #661, opened 2023-07-24
Add DJ_LOGICAL_TIMESTAMP function

shangyian:
### Summary

This PR adds support for a custom function `DJ_LOGICAL_TIMESTAMP()` that is meant to be substituted by an external timestamp variable when the node is being incrementally materialized. It also adds the following config to the server's global settings:
* `dj_logical_timestamp_format`: This specifies what the macro should be replaced with. Some examples include:
  * `${dj_logical_timestamp}`
  * {{ dj_logical_timestamp }}`
  * `$dj_logical_timestamp`

When this function is used in a node's query, the following will happen:
* Attempting to build SQL down to that node without having materialized the node will fail.
* When materialization is configured for the node, we'll generate a query with `DJ_LOGICAL_TIMESTAMP()` substituted based on the global settings above. 

An aside: also adds type inference for another Spark SQL date function: `date_format`.

### Test Plan

- [x] PR has an associated issue: #640 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -281,26 +281,6 @@ def _build_tables_on_select(
         )  # got a materialization
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
-
-            dj_logical_timestamp = list(
-                node_query.filter(
-                    lambda x: isinstance(x, ast.Function)
-                    and x.name.name.lower() == "dj_logical_timestamp",
-                ),
-            )
-            if (
-                dj_logical_timestamp
-                and not node.has_available_materialization(
-                    build_criteria,  # type: ignore
-                )
-                and (build_criteria and not build_criteria.for_materialization)
-            ):
-                raise DJException(
-                    message=f"One of the parent nodes {node.name} has a query that uses "
-                    "dj_logical_timestamp(), which is only meant to be used for materialization."
-                    f" {node.name} must be successfully materialized before it can be used.",
-                )
-
             node_table = build_ast(  # type: ignore
                 session,
                 node_query,
@@ -587,21 +567,6 @@ def build_metric_nodes(
                     f"`{col}`, which is not available on every"
                     f" metric and thus cannot be included.{potential_dimension_match}",
                 )
-
-    # Set the dialect by finding available engines for this node, or default to Spark
-    if not build_criteria:
-        build_criteria = BuildCriteria(
-            dialect=(
-                metric_nodes[0].current.catalog.engines[0].dialect
-                if metric_nodes
-                and metric_nodes[0].current
-                and metric_nodes[0].current.catalog
-                and metric_nodes[0].current.catalog.engines
-                and metric_nodes[0].current.catalog.engines[0].dialect
-                else Dialect.SPARK
-            ),
-            for_materialization=False,
-        )
 
     combined_ast: ast.Query = ast.Query(
         select=ast.Select(from_=ast.From(relations=[])),
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -282,20 +282,19 @@ def _build_tables_on_select(
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
 
-            dj_current_timestamp = [
-                expr
-                for expr in node_query.filter(
+            dj_current_timestamp = list(
+                node_query.filter(
                     lambda x: isinstance(x, ast.Function)
-                    and x.name.name == "dj_current_timestamp",
-                )
-            ]
+                    and x.name.name.lower() == "dj_current_timestamp",
+                ),
+            )
             if dj_current_timestamp and not node.has_available_materialization(
-                build_criteria,
+                build_criteria,  # type: ignore
             ):
                 raise DJException(
-                    message=f"One of the parent nodes {node.name} has a query that uses DJ_CURRENT_TIMESTAMP(),"
-                    f"which is only meant to be used for materialization. {node.name} must be "
-                    "successfully materialized before it can be used.",
+                    message=f"One of the parent nodes {node.name} has a query that uses "
+                    "DJ_CURRENT_TIMESTAMP(), which is only meant to be used for materialization."
+                    f" {node.name} must be successfully materialized before it can be used.",
                 )
 
             node_table = build_ast(  # type: ignore
@@ -525,9 +524,6 @@ def build_node(  # pylint: disable=too-many-arguments
         query = parse(node.query)
     else:
         query = build_source_node_query(node)
-
-    if node.has_available_materialization(build_criteria):
-        node.has_available_materialization(node.query)
 
     add_filters_dimensions_orderby_limit_to_query_ast(
         query,
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -282,18 +282,22 @@ def _build_tables_on_select(
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
 
-            dj_current_timestamp = list(
+            dj_logical_timestamp = list(
                 node_query.filter(
                     lambda x: isinstance(x, ast.Function)
-                    and x.name.name.lower() == "dj_current_timestamp",
+                    and x.name.name.lower() == "dj_logical_timestamp",
                 ),
             )
-            if dj_current_timestamp and not node.has_available_materialization(
-                build_criteria,  # type: ignore
+            if (
+                dj_logical_timestamp
+                and not node.has_available_materialization(
+                    build_criteria,  # type: ignore
+                )
+                and (build_criteria and not build_criteria.for_materialization)
             ):
                 raise DJException(
                     message=f"One of the parent nodes {node.name} has a query that uses "
-                    "DJ_CURRENT_TIMESTAMP(), which is only meant to be used for materialization."
+                    "dj_logical_timestamp(), which is only meant to be used for materialization."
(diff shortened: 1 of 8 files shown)


datajunction-server/datajunction_server/construction/build.py
```diff
@@ -281,6 +281,22 @@ def _build_tables_on_select(
...
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
+
+            dj_current_timestamp = list(
+                node_query.filter(
+                    lambda x: isinstance(x, ast.Function)
+                    and x.name.name.lower() == "dj_current_timestamp",
+                ),
+            )
+            if dj_current_timestamp and not node.has_available_materialization(
```
agorajek, line 291: Please remind me how would we render the query with the `${dj_current_timestamp}` variable (for scheduling purposes) if the build code blocks it with the check.

agorajek (approved):
This is good, but I think one day we'll hit a use-case when some will have a working "virtual" transform node, and they will want to add materialization w/o affecting the current pass-through users. Then we may need to add the partition filter to the materialization itself.

I can see both solutions being valid and coexisting because:
- This one (this PR) is better for performance. The user may know better where to stick the filter so the query runs better than if DJ added that filter to the outer query.
- And the "filter as part of materialization config" solution would be easy to add after and assumes little to none performance problems with the node.

datajunction-server/datajunction_server/construction/build.py
CircArgs: That syntax doesn't look correct right? This is going to be a function executed in the DJ Runtime `DJ_CURRENT_TIMESTAMP()` which makes sense to me. is `${...}` Something else?

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -281,6 +281,22 @@ def _build_tables_on_select(
         )  # got a materialization
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
+
+            dj_current_timestamp = list(
```
CircArgs, line 285: In Slack, I had asked if this should be more generic where there is a function registry dedicated to dj runtime functions. Then this would just happen upon rendering the ast via `Function.__str__`

datajunction-server/datajunction_server/construction/build.py
shangyian: @CircArgs and @agorajek So this is the part that's going to get a bit confusing and I was going to punt implementation to a separate PR 😄 

But the gist of it is this:
* At Netflix we wouldn't want DJ to actually generate any formatted timestamp strings, because we'll have a plugin in our internally deployed DJ server that just replaces the `DJ_CURRENT_TIMESTAMP()` function with something like `${dj_current_timestamp}`. The latter is a variable that gets substituted with the datetime parameters by our internal scheduler. 
* However, in open source DJ needs to actually execute the function to generate the right query for materialization. This means that for each partition that gets backfilled, DJ will need to execute `DJ_CURRENT_TIMESTAMP()` to get the right value for that partition and then kick off the materialization job.
* I'm a little bit concerned about getting the OSS implementation right, since I think the path would be clearest if we were actively building out an integration with a scheduler like Airflow.

agorajek: @shangyian just to show everyone a quick Airflow reference. Found this at https://docs.astronomer.io/learn/airflow-sql#example-2-execute-a-query-with-parameters:
```
opr_param_query = SnowflakeOperator(
    task_id="param_query",
    snowflake_conn_id="snowflake",
    sql="param-query.sql",
    params={"date":mydatevariable}
)
```
and the param-query.sql is:
```
SELECT *
FROM STATE_DATA
WHERE date = {{ params.date }}
```

So notice that we need to generate the SQL with a template that then will be understood by a given scheduler. But whatever the string replacement patter is we will need to generate that SQL... so maybe this code (your PR) is good, but when we generate SQL for materialization purpose we could pass in a special flag to not fail here.

shangyian: @agorajek Thanks for digging into this! It looks like actually all we need to do is replace `DJ_CURRENT_TIMESTAMP()` with `{{ params.date }}` for Airflow, since that SQL would just be written as a part of the transform's query. Also in looking at those docs, what do you think about naming this function `dj_logical date()` instead?

datajunction-server/datajunction_server/construction/build.py
shangyian: Good point! This was your comment on Slack right?

> So, built-in macros or functions that run at the time of rendering a query? Look up a function, see it is a runtime one, when the query is rendered the builtin gives a string

I like this... let me see if I can make this setup more generic.

datajunction-server/datajunction_server/construction/build.py
samredai: +1 to using `DJ_LOGICAL_TIMESTAMP()` as the name.

> At Netflix we wouldn't want DJ to actually generate any formatted timestamp strings, because we'll have a plugin in our internally deployed DJ server that just replaces the DJ_CURRENT_TIMESTAMP() function with something like ${dj_current_timestamp}

Is this something that could reasonably be a global server setting? In other words, on any given deployment of DJ, either you want the function to resolve to a formatted date/time or you want it to resolve to some generic string like `${dj_current_timestamp}` or `{{ params.date }}`. If that's so, maybe we can have a global setting `TIMESTAMP_PARAM=${dj_current_timestamp}` and in the processing logic for `DJ_CURRENT_TIMESTAMP()` we do something along the lines of:
```py
if settings.timestamp_param:
    # replace DJ_CURRENT_TIMESTAMP() with settings.timestamp_param
else:
    # actually resolve to a formatted timestamp strings
```

agorajek: > Is this something that could reasonably be a global server setting?

Yes, and I think we may just hardcode few of these methods as options:
```
DJ_DEFAULT_TIMESTAMP_FORMAT: double_curly | dollar_with_curly | value_as_str | value_as_int
```
Remember that the above setting would only be used if the user does not specify format string as part of the DJ_LOGICAL_TIMESTAMP itself.

shangyian: Yeah, given that DJ is not doing the scheduling, I'm leaning towards:
* We have a server-side global config `DJ_DEFAULT_TIMESTAMP_FORMAT` that's one of the options @agorajek listed above.
* We don't bother implementing any sort of actual formatted timestamp (i.e., parsing a timestamp with the format string `%Y-%m-%d`), since if we integrate with an external scheduler (with the example above being Airflow), they will typically provide that generation for us and we just need to abide by their syntax.
* If DJ ever builds its own scheduler, then we should think about supporting a formatted timestamp.

samredai: Can you help me understand why the global setting value has to be one of an "approved" set of variable formats and then having the user separately pass in the variable name? Wouldn't it be the same variable name that's always used across a single server, such as ${date_partition}?

shangyian: @samredai the global setting isn't about the variable name but rather how the variable substitution format should look like. Options include:
* `{{ date }}`
* `${date}`
* `$date`
etc

samredai: I was actually wondering if the global setting should be about the variable which is why I asked that question. If the behavior desired is that DJ_LOGICAL_TIMESTAMP() always resolves to a parameter, why don't we just let the user directly define that parameter in the global settings, formatting included.

shangyian: Yep, that's what the current implementation looks like -- there's just a single global parameter that replaces DJ_LOGICAL_TIMESTAMP() across the board. That way we can be more constrained or else someone is going to use this to do some odd variable substitution that's not timestamp-based at all... (I just pushed the changes + updated the PR description on it)

CircArgs: @shangyian I'm not saying this is some better alternative, but if we had the concept of variables would that be going too far? 

```sql
Node(
...
query= SELECT *
FROM STATE_DATA
WHERE date = my_date
...
variables = ['my_date']
```

at runtime

`GET dj.server/.../...&my_date='{{ date}}'`

so the user could fill in the variable knowing there is some downstream thing looking to dump into a templated part or they could specify an actual date in this case.

CircArgs: The netflix specific runtime behavior of dumping a different format string as well as having this logic in the build for one case feels odd to me

samredai: @CircArgs yeah I do think variables would be going a bit too far. Arbitrarily templated queries would create a huge amount of ambiguity across the DJ DAG

shangyian: @CircArgs So I'm not sure that's Netflix specific -- if we integrate in OSS with Airflow, we'll need something similar for incremental materialization of transforms. It's just that the substituted string would be different, and that's configurable via settings. But I agree that it's better if it's not in the build

CircArgs: I didn't quite mean that that specific formatting method is Netflix specific, just questioning whether there should be a fixed set of options on how to format or if it should be something generic to my point about these being user-defined runtime functions

shangyian: > if we had the concept of variables would that be going too far?

@CircArgs heh, I'm afraid that would give too much power to the user. I think we should keep this functionality as locked down as possible (like right now it's basically limited to a single, temporal parameter), or at least until we have a clear use case. Otherwise we'll start going down the path of supporting jinja templates... 😄 

Also I removed the function-specific check from the build logic -- I realized that we don't need to stop the user from generating SQL for a node just because an upstream uses `dj_logical_timestamp()`. If their query engine doesn't support that, it'll just fail when they try to run the query, but we might as well let them generate it.

samredai (approved):

PR #665, opened 2023-07-26
Python Client cleanup part 2

agorajek:
### Summary

I know we have been going back and forth on the DJ Client class name but after spending some time in this I would liek to re-suggest this picture and chat about it if needed:

- **_internal.py** - all the stuff that users should never see
- **exceptions.py, models.py** - self-explanatory
- **client.py**
    - DJClient class (basic DJ access)
        - list_[namespaces | dimensions | metrics]
        - common_[dimensions | metrics]
        - sql() | data()
    - DJBuilder class (building DJ dag)
        - list|create|delete : engine|catalogs (some TODO)
        - list|create|delete namespace (some TODO)
        - source|transform|metric|dimension|cube --> returns an existing Node
        - create_[source|...] --> create a new Node (in DJ) and returns it
- **node.py**
    - Namespace
        - nodes|sources|transforms|cubes --> get nodes from a namespace, not sure if we need this
    - Node
        - save() --> creates or updates an existing Node
        - refresh() --> refreshes current object from DJ
        - [un]link_dimensions()
        - [de]activate() -- ?
        - add_materialization(), add_availability(), set_column_attributes()
        - list_revisions(), 
    - some other classes

Some open questions:
1. Is there a reason our `client` directory does not follow the `datajunction-<service>` pattern?
2. What is node activate/deactivate for?

TODO (maybe follow-up PRs):
1. Complete all TODOs in the code. Few methods are missing.
2. Adjust client README to the new code.
3. Adjust Demo notebook the the new code.
4. Hide all-the-protected-methods in _internal.DJClient better, cause now they are still visible in the user client classes.

### Test Plan

make check + make test

### Deployment Plan

n/a

### DIFF
--- a/client/python/datajunction/client.py
+++ b/client/python/datajunction/client.py
@@ -7,16 +7,17 @@ from urllib.parse import urlencode
 from alive_progress import alive_bar
 
 from datajunction import _internal, models
-from datajunction.exceptions import DJClientException
+from datajunction.exceptions import DJClientException, DJNamespaceAlreadyExists
+from datajunction.nodes import Cube, Dimension, Metric, Namespace, Source, Transform
 
 
-class DJReader(_internal.DJClient):
+class DJClient(_internal.DJClient):
     """
-    Client class to consume basic DJ services: metrics and dimensions.
+    Client class for basic DJ dag and data access.
     """
 
     #
-    # List namespace, metrics and dimensions
+    # List
     #
     def list_namespaces(self, prefix: Optional[str] = None) -> List[str]:
         """
@@ -32,7 +33,7 @@ class DJReader(_internal.DJClient):
         List dimension nodes under given namespace.
         TODO: make namespace optional and return all dimension nodes, similar to get("/metrics/").
         """
-        return self.get_nodes_in_namespace(
+        return self._get_nodes_in_namespace(
             namespace=namespace,
             type_=models.NodeType.DIMENSION.value,
         )
@@ -42,7 +43,7 @@ class DJReader(_internal.DJClient):
         List metric nodes for given namespace or all.
         """
         if namespace:
-            return self.get_nodes_in_namespace(
+            return self._get_nodes_in_namespace(
                 namespace=namespace,
                 type_=models.NodeType.METRIC.value,
             )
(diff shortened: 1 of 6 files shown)


client/python/datajunction/client.py
```diff
@@ -215,7 +216,281 @@ def data(  # pylint: disable=too-many-arguments,too-many-locals
...
+    # Catalogs and Engines
+    #
+    def list_catalogs(self):
+        """
+        Gets all catalogs.
+        """
+        response = self._session.get("/catalogs/", timeout=self._timeout)
+        return response.json()
+
+    def list_engines(self):
```
samredai, line 234: Is listing engines on its own any value to client users? If we keep it, it might be better as an underscore method. Only because what users really care about is "what engine can I use for a specific catalog?" So they would list catalogs -> pull a catalog -> then look at the engines defined on it.

client/python/datajunction/client.py
```diff
@@ -215,7 +216,281 @@ def data(  # pylint: disable=too-many-arguments,too-many-locals
...
+        """
+        Returns the specified node namespace.
+        """
+        try:
+            return self.create_namespace(namespace=namespace)
+        except DJNamespaceAlreadyExists:
+            pass
+        return Namespace(namespace=namespace, dj_client=self)
+
+    def create_namespace(self, namespace: str) -> "Namespace":
```
samredai, line 255: The create prefix implies that the this method would actually create it, but it won't since the user still has to call `.save()`. That's why we went with the new prefix but maybe we were overthinking. @shangyian what do you think?

client/python/datajunction/nodes.py
```diff
@@ -0,0 +1,331 @@
...
+        as they don't need to be materialized.
+        """
+        upsert_response = self.dj_client._upsert_materialization(
+            self.name,
+            config,
+        )
+        self.refresh()
+        return upsert_response
+
+    def deactivate(self) -> str:
```
samredai, line 163: To your question in the PR description, deactivate and activate is meant to act as a soft delete. This should appear to the user at the surface as a normal delete so maybe on the client side of things we scrap the deactivate/activate terminology for delete/restore.

samredai (approved):
This looks good, left a few comments. I like the DJClient and DJBuilder class names.

client/python/datajunction/client.py
shangyian: I'd say it is useful because realistically most DJ server instances will be set up for one catalog, so this is like a convenience method where someone can say "give me all the engines for the primary catalog". Or maybe we can even have a "catalog" set when someone initializes the client, and just list all the engines from that catalog.

client/python/datajunction/nodes.py
```diff
@@ -0,0 +1,331 @@
...
+        Deactivates the node
+        """
+        response = self.dj_client._deactivate_node(self)
+        if not response.ok:  # pragma: no cover
+            raise DJClientException(
+                f"Error deactivating node `{self.name}`: {response.text}",
+            )
+        return f"Successfully deactivated `{self.name}`"
+
+    def activate(self) -> str:
```
shangyian, line 174: +1 to "delete" and "restore" from the user perspective

client/python/datajunction/client.py
```diff
@@ -215,7 +216,281 @@ def data(  # pylint: disable=too-many-arguments,too-many-locals
...
+        Gets all engines.
+        """
+        response = self._session.get("/engines/", timeout=self._timeout)
+        return response.json()
+
+    #
+    # Namespace
+    #
+    # TODO: add delete namespace (with warnings) # pylint: disable=fixme
+    def namespace(self, namespace: str) -> "Namespace":
```
shangyian, line 245: Should this actually call `create_namespace`? I would've thought that not prefixing something with `create_` would imply that nothing changes server-side, but it will just return an error if the namespace doesn't exist.

shangyian (approved):
Thanks for the changes - looks great! 

> Is there a reason our client directory does not follow the datajunction-<service> pattern?

We could rename to `datajunction-clients`, but this directory also doesn't quite follow the pattern of the other `datajunction-<service>` dirs since those are all python packages, but this one has `client/python`, `client/javascript`, `client/java` etc under it.

client/python/datajunction/client.py
agorajek: Yes, I changed that pattern from `new + save` to `create`.

client/python/datajunction/nodes.py
agorajek: Thanks, I'll do these changes in my next PR.

PR #667, opened 2023-07-26
Add register_table functionality to the DJ python client

shangyian:
### Summary

This PR adds `register_table` to the client. It can be used with `dj.register_table(<catalog>, <schema>, <table>)`. 

Also adds a small check to table registration endpoint to abort if a source node with the same name already exists. This saves us a good chunk of time since we don't have to go out to the query service and pull columns if there's already a source registered.

Additionally adds a `debug` mode to the client, where setting it to False (default) hides the traceback when using the client from a jupyter notebook.

### Test Plan

Tested locally

- [x] `make check` passes

### DIFF
--- a/client/python/datajunction/_internal.py
+++ b/client/python/datajunction/_internal.py
@@ -34,6 +34,19 @@ _logger = logging.getLogger(__name__)
 #
 # Helpers
 #
+
+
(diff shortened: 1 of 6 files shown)


samredai (approved):

agorajek:
Thanks @shangyian !

PR #668, opened 2023-07-27
Improve error message when node creation fails due to type inference

shangyian:
### Summary

When node creation fails from a type inference error, the current server-side error message just returns `Unable to infer type for some columns on node <name>`. This doesn't tell users what part of the query they should fix.

Since we already have a `type_inference_failures` object that contains a map of all the columns to their failures, this PR exposes all of that directly in the error message as a list.

An example error message would look like:
```
Unable to infer type for some columns on node `example.node_name`.

	* Cannot resolve type of column x.first_name in SELECT  x.name
	* Cannot resolve type of column last_name in x.last_name
```

### Test Plan

I'll add a better test shortly

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -287,7 +287,11 @@ def get_downstream_nodes(
         )
         .where(NodeRelationship.parent_id == node.id)
         .join(NodeRevision, NodeRelationship.child_id == NodeRevision.id)
-        .join(Node, Node.id == NodeRevision.node_id)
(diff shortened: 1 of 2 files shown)


agorajek (approved):
Looks good!

shangyian:
@agorajek I think the downstreams bugfix here should handle #617. I assume when you saw this bug it was in the UI and/or in the `/node/{name}/downstreams` endpoint, where it was continuing to display old node revisions' downstreams. However, all the database links were actually right, it just needed to be surfaced differently.

PR #669, opened 2023-07-27
Mark `Query` as compiled after compilation finishes to reduce compile speed 

shangyian:
### Summary

On large queries with many columns and subqueries, table compilation takes a long time (>5min). Did some digging and noticed that part of the time sink is from us not marking a `Query` object that's already gone through the compilation process as compiled.

This means that if we're working with a query that has many subqueries, those subqueries will get reprocessed over and over when we want to assign table sources to columns in the outer query. This issue doesn't manifest if we have a simple query without subqueries.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -446,6 +446,7 @@ class Node(ABC):
             if not child.is_compiled():
                 child.compile(ctx)
                 child._is_compiled = True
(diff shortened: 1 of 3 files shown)


agorajek:
Looks correct, but the fact that you found 2 places missing this may suggest that there may be more? What was the speed-up after the change?

agorajek (approved):
approve_agent_cooper

samredai (approved):

PR #670, opened 2023-07-27
Rename /client to /datajunction-clients.

agorajek:
### Summary

Just the top level rename.

### Test Plan

Also started `docker compose --profile demo up` and made sure the UI and Jupyter work.

- [x] `make check` passes

### Deployment Plan

n/a


samredai:
Shouldn't it be plural, i.e. "clients"?

agorajek:
@samredai good point.

datajunction-server/datajunction_server/api/client.py
```diff
@@ -86,7 +86,7 @@ def client_code_for_creating_node(
 
 
 @router.get(
-    "/client/python/add_materialization/{node_name}/{materialization_name}",
+    "/datajunction-clients/python/add_materialization/{node_name}/{materialization_name}",
```
shangyian, line 89: To clarify, I don't think this route should change, I meant that the contents instead of (for example) `client_code_for_adding_materialization` will need to change, since we've updated the client code significantly, like the split between `DJClient` and `DJBuilder`.

shangyian (approved):
LGTM! Only one minor comment on the path rename, but feel free to keep that change if you think it's more consistent.

datajunction-server/datajunction_server/api/client.py
agorajek: @shangyian but then what is this reference to `client/python/...` ?

shangyian: Ah, I guess I didn't think of it as a reference to the directory path, but yeah, interpreting it that way makes sense.

samredai: @agorajek this is just the REST API endpoint that the UI calls to return the example snippets it displays. I don't think it has anything to do with the directory path, it was just a coincidence that there was a `/client/python/` directory and this REST endpoint was `GET /client/python/...`.

agorajek: Ah I see now. Ok... I'll merge this and fix this later, if needed.

PR #671, opened 2023-07-28
wip: memoization for build speed

shangyian:
### Summary

This change adds memoization to the query building stage so that we're not recompiling and rebuilding the same node queries over and over. This optimization applies to cases where a node references the same parent nodes in different subqueries -- we'll only compile the parent node once, store it in the memo dict with the query hash as the key, and retrieve it if available. 

This reduced SQL building time for the node with many nested subqueries from ~50s to ~25s. Of course, that's still not particularly fast, and for that we might want to investigate pickling the AST at node creation and upon any node query updates. I haven't managed to successfully do that yet because the AST classes require a number of changes to pickle. 

Also added a `logging.conf` to configure logging for the app.

Pasting some profiling results here:
```
 | Total time: 31.0071 s
 | File: /code/datajunction_server/construction/build.py
 | Function: build_ast at line 818
 | 
 | Line #      Hits         Time  Per Hit   % Time  Line Contents
 | ==============================================================
 |    818                                           @profile
 |    819                                           def build_ast(  # pylint: disable=too-many-arguments
 |    820                                               session: Session,
 |    821                                               query: ast.Query,
 |    822                                               memoized_queries: Dict[int, ast.Query] = None,
 |    823                                               build_criteria: Optional[BuildCriteria] = None,
 |    824                                           ) -> ast.Query:
 |    825                                               """
 |    826                                               Determines the optimal way to build the query AST and does so
 |    827                                               """
 |    828        18       9000.0    500.0      0.0      memoized_queries = memoized_queries or {}
 |    829                                           
 |    830        18      22624.0   1256.9      0.0      start = time.time()
 |    831        18     161041.0   8946.7      0.0      context = CompileContext(session=session, exception=DJException())
 |    832        18  106973165.0 5942953.6      0.3      if hash(query) in memoized_queries:
 |    833                                                   query = memoized_queries[hash(query)]  # pragma: no cover
 |    834                                               else:
 |    835        18 26894907430.0 1494161523.9     86.7          query.compile(context)
 |    836        18   94670710.0 5259483.9      0.3          memoized_queries[hash(query)] = query
 |    837        18      27499.0   1527.7      0.0      end = time.time()
 |    838        18   14499083.0 805504.6      0.0      _logger.info("Finished compiling query %s in %s", str(query)[-100:], end - start)
 |    839                                           
 |    840        18       9166.0    509.2      0.0      start = time.time()
 |    841        11 3893805461.0 353982314.6     12.6      query.build(session, memoized_queries, build_criteria)
 |    842        18      29002.0   1611.2      0.0      end = time.time()
 |    843        18    2020541.0 112252.3      0.0      _logger.info("Finished building query in %s", end - start)
 |    844        18       3165.0    175.8      0.0      return query
```

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/main.py
+++ b/datajunction-server/datajunction_server/api/main.py
@@ -8,6 +8,8 @@ Main DJ server app.
 # pylint: disable=unused-import
 
 import logging
+from logging import config
+from os import path
 from typing import TYPE_CHECKING, Optional
 
 from fastapi import Depends, FastAPI, Request
@@ -40,7 +42,7 @@ from datajunction_server.models.column import Column
 from datajunction_server.models.engine import Engine
 from datajunction_server.models.node import NodeRevision
 from datajunction_server.models.table import Table
-from datajunction_server.utils import get_settings, setup_logging
+from datajunction_server.utils import get_settings
 
 if TYPE_CHECKING:  # pragma: no cover
     from opentelemetry import trace
@@ -55,7 +57,10 @@ def get_dj_app(
     Get the DJ FastAPI app and optionally inject an OpenTelemetry tracer provider
     """
     settings = get_settings()
-    setup_logging(settings.loglevel)
+    config.fileConfig(
+        path.join(path.dirname(path.abspath(__file__)), "logging.conf"),
+        disable_existing_loggers=False,
+    )
     application = FastAPI(
         title=settings.name,
         description=settings.description,
--- a/datajunction-server/datajunction_server/api/main.py
+++ b/datajunction-server/datajunction_server/api/main.py
@@ -40,7 +40,7 @@ from datajunction_server.models.column import Column
 from datajunction_server.models.engine import Engine
 from datajunction_server.models.node import NodeRevision
 from datajunction_server.models.table import Table
-from datajunction_server.utils import get_settings
+from datajunction_server.utils import get_settings, setup_logging
 
 if TYPE_CHECKING:  # pragma: no cover
     from opentelemetry import trace
@@ -55,6 +55,7 @@ def get_dj_app(
     Get the DJ FastAPI app and optionally inject an OpenTelemetry tracer provider
     """
     settings = get_settings()
+    setup_logging(settings.loglevel)
     application = FastAPI(
         title=settings.name,
         description=settings.description,
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -286,7 +286,7 @@ def _build_tables_on_select(
         )  # got a materialization
         if node_table is None:  # no materialization - recurse to node first
             node_query = parse(cast(str, node.query))
-            if hash(node_query) in memoized_queries:
+            if hash(node_query) in memoized_queries:  # pragma: no cover
                 node_table = memoized_queries[hash(node_query)].select  # type: ignore
             else:
                 query_ast = build_ast(  # type: ignore
@@ -530,7 +530,10 @@ def build_node(  # pylint: disable=too-many-arguments
         limit,
     )
     memoized_queries: Dict[int, ast.Query] = {}
-    return build_ast(session, query, memoized_queries, build_criteria)
+    _logger.info("Calling build_ast on %s", node.name)
+    astt = build_ast(session, query, memoized_queries, build_criteria)
+    _logger.info("Finished build_ast on %s", node.name)
+    return astt
 
 
 def build_metric_nodes(
@@ -821,12 +824,12 @@ def build_ast(  # pylint: disable=too-many-arguments
     start = time.time()
     context = CompileContext(session=session, exception=DJException())
     if hash(query) in memoized_queries:
-        query = memoized_queries[hash(query)]
+        query = memoized_queries[hash(query)]  # pragma: no cover
     else:
         query.compile(context)
         memoized_queries[hash(query)] = query
     end = time.time()
-    _logger.info("Finished compiling query in %s", end - start)
+    _logger.info("Finished compiling query %s in %s", str(query)[-100:], end - start)
 
     start = time.time()
     query.build(session, memoized_queries, build_criteria)
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -1,5 +1,7 @@
 """Functions to add to an ast DJ node queries"""
 import collections
+import logging
+import time
 
 # pylint: disable=too-many-arguments,too-many-locals,too-many-nested-blocks,too-many-branches,R0401
 from typing import DefaultDict, Deque, Dict, List, Optional, Set, Tuple, Union, cast
@@ -17,6 +19,8 @@ from datajunction_server.sql.parsing.ast import CompileContext
 from datajunction_server.sql.parsing.backends.antlr4 import ast, parse
 from datajunction_server.sql.parsing.types import ColumnType
 from datajunction_server.utils import amenable_name
+
+_logger = logging.getLogger(__name__)
 
 
 def _get_tables_from_select(
@@ -269,6 +273,7 @@ def _build_tables_on_select(
     session: Session,
     select: ast.SelectExpression,
     tables: Dict[NodeRevision, List[ast.Table]],
+    memoized_queries: Dict[int, ast.Query],
     build_criteria: Optional[BuildCriteria] = None,
 ):
     """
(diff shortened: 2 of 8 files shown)


shangyian:
@agorajek This reduced SQL building time for the node with many nested subqueries from ~50s to ~25s. Of course, that's still not particularly fast, and for that we might want to investigate pickling the AST at node creation and upon any node query updates. I haven't managed to successfully do that yet because the AST classes require a number of changes to pickle.

datajunction-server/datajunction_server/api/main.py
```diff
@@ -55,6 +55,7 @@ def get_dj_app(
     Get the DJ FastAPI app and optionally inject an OpenTelemetry tracer provider
     """
     settings = get_settings()
+    setup_logging(settings.loglevel)
```
samredai, line 58: Instead of this setup logging function I think it'd be cleaner if we use a [config](https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig) file that the standard logger supports. That will allow for complete configuration of the logger without changing any source code. I can do that in a follow-up PR if you'd like.

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -796,12 +809,27 @@ def build_source_node_query(node: NodeRevision):
...
     session: Session,
     query: ast.Query,
+    memoized_queries: Dict[int, ast.Query] = None,
     build_criteria: Optional[BuildCriteria] = None,
 ) -> ast.Query:
     """
     Determines the optimal way to build the query AST and does so
     """
+    if not memoized_queries:
+        memoized_queries = {}
```
samredai, line 822: This is equivalent to setting the default in the function signature--`memoized_queries: Dict[int, ast.Query] = {}`

samredai (approved):
This is great, thanks @shangyian. Left a couple of comments. The idea of a follow-up PR that pickles queries at node creation/update time makes sense to me, but it might be better if we can make all query classes json serializable. Let me know what you think but I'm imagining that should make it easier to store and could come in useful if we want to add a visual explorer of the query ast in the future for debugging. None of the classes are too complex so we can do it manually, or alternatively try something like [jsonpickle](https://github.com/jsonpickle/jsonpickle).

datajunction-server/datajunction_server/api/main.py
shangyian: Good point! I added a `logging.conf` file and then modified this part to configure logging based on that file

datajunction-server/datajunction_server/construction/build.py
shangyian: Actually I think doing that will cause other side effects: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

agorajek:
@shangyian thank you, this is a good start. I wasn't actually aware of our approach till now.

> we might want to investigate pickling the AST at node creation and upon any node query updates
100%. We can experiment with different serialization methods, but we need to do the AST building on change only.

I am pretty sure that once we scale the node numbers the underlying sqlalchemy queries would kill our performance, unless we only use it to walk the DJ dag rather than for getting all the node pieces from various tables. But I am glad we caught it sooner than later.

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -281,12 +286,18 @@ def _build_tables_on_select(
...
+                node_table = memoized_queries[hash(node_query)].select  # type: ignore
+            else:
+                query_ast = build_ast(  # type: ignore
+                    session,
+                    node_query,
+                    memoized_queries,
+                    build_criteria,
+                )
+                node_table = query_ast.select  # type: ignore
+                node_table.parenthesized = True  # type: ignore
```
agorajek, line 299: What does this mean?

agorajek (approved):
lgtm

datajunction-server/datajunction_server/construction/build.py
samredai: Woah, I learned something new. 🧐

datajunction-server/datajunction_server/construction/build.py
samredai: This is a flag that lets the renderer know if this table object needs to be surrounded by parentheses. Which is the case when a from table is a subquery.

CircArgs:
@shangyian I forget if it came up during discussion about the C++ parser, but maybe we should get some concrete timings from a profiler. "This time for parse, this time for building the DJ AST, this time for compile, for build, calls to db..."

Also, I don't think its necessarily a reasonable solution, but Dill might work for our ast.

When I started working on the AST I considered serializing manually to Json and storing that to the db. I don't think it was a priority though.

If we get profile results maybe we can memoize more specific things?

datajunction-server/datajunction_server/construction/build.py
CircArgs: References!

CircArgs: Do people generally like this instead of `memoized_queries = memoized_queries or {}`? I've also wondered why this common pattern hasn't risen to the level of the case that inspired walrus notation in python `:=`

shangyian: Ah yeah, that one-liner is cleaner! Let me update

shangyian:
> once we scale the node numbers the underlying sqlalchemy queries would kill our performance

@agorajek It's certainly possible -- in the example queries we were trying, the largest referenced 49 upstream nodes, so that's 49 queries to the database as we build one metric node. 😱 

One thing I did realize though is that technically this query is not meant to be executed / built down to the original sources; it's supposed to only execute on the last materialized transform, because building it down to the original sources would yield a Spark SQL query that isn't likely to run for performance reasons. But that's not a pattern that holds true for all large queries.

> we can make all query classes json serializable

@samredai Agreed that making the AST json serializable is a good option! See below -- I think we need to make similar changes to the AST to ready it for json serialization.

> Also, I don't think its necessarily a reasonable solution, but [Dill](https://pypi.org/project/dill/) might work for our ast.

@CircArgs I tried dill already, but I was running into issues. I think part of it is that we're passing around a database session buried in the `NodeRevision` reference on several AST objects, which can't be pickled. If we can change it so that it's not passing a `NodeRevision` but rather just a set of pydantic NodeRev-like objects, then it should work.

shangyian:
@CircArgs some more profiling via line_profile:

```
dj                 | Total time: 31.0071 s
dj                 | File: /code/datajunction_server/construction/build.py
dj                 | Function: build_ast at line 818
dj                 | 
dj                 | Line #      Hits         Time  Per Hit   % Time  Line Contents
dj                 | ==============================================================
dj                 |    818                                           @profile
dj                 |    819                                           def build_ast(  # pylint: disable=too-many-arguments
dj                 |    820                                               session: Session,
dj                 |    821                                               query: ast.Query,
dj                 |    822                                               memoized_queries: Dict[int, ast.Query] = None,
dj                 |    823                                               build_criteria: Optional[BuildCriteria] = None,
dj                 |    824                                           ) -> ast.Query:
dj                 |    825                                               """
dj                 |    826                                               Determines the optimal way to build the query AST and does so
dj                 |    827                                               """
dj                 |    828        18       9000.0    500.0      0.0      memoized_queries = memoized_queries or {}
dj                 |    829                                           
dj                 |    830        18      22624.0   1256.9      0.0      start = time.time()
dj                 |    831        18     161041.0   8946.7      0.0      context = CompileContext(session=session, exception=DJException())
dj                 |    832        18  106973165.0 5942953.6      0.3      if hash(query) in memoized_queries:
dj                 |    833                                                   query = memoized_queries[hash(query)]  # pragma: no cover
dj                 |    834                                               else:
dj                 |    835        18 26894907430.0 1494161523.9     86.7          query.compile(context)
dj                 |    836        18   94670710.0 5259483.9      0.3          memoized_queries[hash(query)] = query
dj                 |    837        18      27499.0   1527.7      0.0      end = time.time()
dj                 |    838        18   14499083.0 805504.6      0.0      _logger.info("Finished compiling query %s in %s", str(query)[-100:], end - start)
dj                 |    839                                           
dj                 |    840        18       9166.0    509.2      0.0      start = time.time()
dj                 |    841        11 3893805461.0 353982314.6     12.6      query.build(session, memoized_queries, build_criteria)
dj                 |    842        18      29002.0   1611.2      0.0      end = time.time()
dj                 |    843        18    2020541.0 112252.3      0.0      _logger.info("Finished building query in %s", end - start)
dj                 |    844        18       3165.0    175.8      0.0      return query
```

Issue #674, opened 2023-07-28
Cleanup Node creation/restoring.


agorajek:
And also adjust the Node [de]activate to delete/restore.

Nodes:
 - delete() - deactivates the node (we need to hide it in the UI)
 - restore() - brings it back to the world
 - create() - works fine if the node never existed or if it is deactivated, so people don't need to worry about the state of deleted nodes
 
Namespaces: --> moved to another issue

If that's ok then we need:
 - an "active" flag at the Node level, I thin we have it already
 - and a similar flag at the Namespace level.

samredai:
This all makes sense to me. For some context, the way we determine the active/inactive state for a node is using the [deactivated_at](https://github.com/DataJunction/dj/blob/main/datajunction-clients/python/datajunction/nodes.py#L73) timestamp attribute. So a node is considered active if that value is null. We should use a similar pattern for namespaces to keep it consistent.

agorajek:
@samredai agreed. Ideally I would like to rename everything from "[de]activated" to "deleted|restored" but I wonder if I should touch that attribute at this point.

samredai:
Oh I wasn't suggesting to change the attribute, I was just saying we should use the same kind of timestamp attribute for activating/deactivating a namespace, as opposed to using something like a boolean.

PR #680, opened 2023-08-01
Allow creating a new node whether it used to exist or not.

agorajek:
### Summary

What the title says, plus a lot of other things especially around error handling from validate_node_data().

### Test Plan

- [x] PR has an associated issue: #674
- [x] `make check` passes

### Deployment Plan

n/a


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -437,14 +436,22 @@ def activate_a_node(name: str, *, session: Session = Depends(get_session)):
...
-                missing_parents_map,
-                type_inference_failed_columns,
-            ) = validate_node_data(downstream.current, session)
-            if missing_parents_map or type_inference_failed_columns:
-                downstream.current.status = NodeStatus.INVALID
+            # We should not failed node restoration just because of some nodes
+            # that have been invalid already and stay that way.
+            missing_parents_map = None
+            type_inference_failed_columns = None
+            try:
```
shangyian, line 443: Nice catch! It seems like a good setup is for `validate_node_data` to return a list of validation errors, and then for any downstream function that calls it to decide whether to raise or skip enforcement of those errors.

agorajek: Yes, I think that's the right pattern. Let me try and rewrite this...

samredai: Should we just raise inside the `validate_node_data` the same way, regardless of if it's a draft or published node? That can be caught outside of the function as @shangyian described.

agorajek: One minute... let me show you what I have for this.

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -381,9 +364,12 @@ def validate_node_data(  # pylint: disable=too-many-locals
     Dict[NodeRevision, List[ast.Table]],
     Dict[str, List[ast.Table]],
     List[str],
+    List[DJError],
 ]:
     """
-    Validate a node.
+    Validate a node. This function should never raise any errors.
```
shangyian, line 370: 🎉 yay, glad this change came in

shangyian (approved):

datajunction-server/datajunction_server/api/namespaces.py
```diff
@@ -21,7 +21,7 @@
 
 
 @router.post("/namespaces/{namespace}/", status_code=201)
-def create_a_node_namespace(
```
samredai, line 24: nit: I would keep this `create_a_node_namespace` because this will get automatically translated in the API documentation to capital case and spaces. So "Create a Node Namespace" looks better than "Create Node Namespace".

datajunction-server/tests/api/namespaces_test.py
```diff
@@ -10,7 +10,7 @@ def test_list_all_namespaces(client_with_examples: TestClient) -> None:
     """
     response = client_with_examples.get("/namespaces/")
     assert response.ok
-    assert response.json() == [
+    assert set(response.json()) == {
```
samredai, line 13: Was the `GET /namespaces/` endpoint returning duplicates? If so that should be fixed server side.

datajunction-clients/python/datajunction/nodes.py
```diff
@@ -18,6 +18,9 @@ class Namespace(ClientEntity):  # pylint: disable=protected-access
 
     namespace: str
 
+    #
+    # List
+    #
     def nodes(self):
```
samredai, line 24: I know it's not part of this PR but I'm wondering if we should make this `list_nodes` to be consistent with the other client methods for getting a list of things.

datajunction-server/tests/api/namespaces_test.py
shangyian: I suspect this is because the endpoint doesn't return them in any particular order. It shouldn't be returning duplicates since it's just a simple select all from the namespaces table and there's a unique constraint.

datajunction-server/datajunction_server/api/helpers.py
agorajek: Thanks, I think I missed one little Exception in there but we can address it later.

datajunction-server/datajunction_server/api/namespaces.py
agorajek: omg, I now understand why this strange naming convention. So... `_a_` infix would only be needed on functions that are exposed in router? I am not sure it is worth the trouble. But if you both think it is then ¯\_(ツ)_/¯

datajunction-server/tests/api/namespaces_test.py
agorajek: Yeah, no duplicates, just a different order.

datajunction-clients/python/datajunction/nodes.py
agorajek: Agree. I'll walk through the Python client one more time this week.

datajunction-server/datajunction_server/api/namespaces.py
shangyian: @agorajek and @samredai Turns out there's actually another way of doing this with FastAPI:
```
@router.post("/namespaces/{namespace}/", status_code=201, name="Create a Node Namespace")
```
Using `name` lets us customize what's displayed in the Swagger docs

agorajek: I do prefer this way, cause it is more customizable and has less awkward function names.

agorajek: If you are ok, I'll merge this PR and address this separately.

PR #684, opened 2023-08-02
djql queries

CircArgs:
### Summary

Revisits https://github.com/DataJunction/dj/pull/336

Currently only supports `SELECT`. In the future maybe support `CREATE NODE`, `SHOW ...`

Examples:

```sql
SELECT 
  Sum(avg_repair_price),
  city
FROM 
  (
    SELECT 
      default.avg_repair_price avg_repair_price, 
      default.hard_hat.country, 
      default.hard_hat.city 
    FROM 
      metrics 
    GROUP BY 
      default.hard_hat.country, 
      default.hard_hat.city 
    LIMIT 
      5
  ) 
GROUP BY city
```

```sql
SELECT default.hard_hat.country, default.hard_hat.city  FROM default.hard_hat
```

```sql
SELECT 
  default.avg_repair_price avg_repair_price, 
  default.hard_hat.country, 
  default.hard_hat.city 
FROM 
  metrics 
GROUP BY 
  default.hard_hat.country, 
  default.hard_hat.city 
ORDER BY       
  default.hard_hat.country, 
  default.hard_hat.city 
```

### Test Plan

So far, built and run queries manually. Tests wip

- [x] `make check` passes


samredai:
So much potential with this feature, both to provide an exploratory SQL interface directly to the server but also to build SQL connectors using other protocols/SDKs out there. I know this is a WIP but I have a question about this example:
```sql
SELECT 
  default.avg_repair_price avg_repair_price, 
  default.hard_hat.country, 
  default.hard_hat.city 
FROM 
  metrics 
GROUP BY 
  default.hard_hat.country, 
  default.hard_hat.city 
ORDER BY       
  default.hard_hat.country, 
  default.hard_hat.city 
```
Why would the `GROUP BY` need to be provided here? Shouldn't we be able to detect which items in the projection are metric nodes and which are dimension nodes? Then we just check if they're compatible and the existing server side logic will group appropriately, right?

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -871,6 +856,49 @@ async def query_event_stream(  # pylint: disable=too-many-arguments
...
+    """
+    Build SQL for multiple metrics. Used by both /sql and /data endpoints
+    """
+
+    query_ast, metrics = build_dj_query(session, query)
+
+    leading_metric_node = metrics[0]
+    available_engines = leading_metric_node.current.catalog.engines
+
+    # Check if selected engine is available, or if none is provided, select the fastest
```
shangyian, line 874: I guess technically this comment isn't true, right? We don't actually rank the engines by speed, we just select the first one available. At some point we might want to include a ranking of engines that users can provide manually.

datajunction-server/datajunction_server/construction/dj_query.py
```diff
@@ -1,206 +1,268 @@
...
+        metric_nodes = []
+        if id(col) in touched_nodes:
+            continue
+        ident = col.identifier(False)
+        if metric_node := try_get_dj_node(session, ident, {NodeType.METRIC}):
+            curr_cols.append((True, col))
+            # if we found a metric node we need to check where it came from
+            parent_select = col.get_nearest_parent_of_type(ast.Select)
+            if not parent_select or not selects_only_metrics(parent_select):
+                raise ast.DJParseException(
```
shangyian, line 65: Should this error be more explicit and just say that we can only support queries that reference metrics? In order words, you can't use this query engine to ask DJ for data for a transform node (or am I misinterpreting the logic here?)

CircArgs:
@samredai the idea was to mimic the api and I figured it is intuitive to provide the group by for the dimensions, so dimensions are discovered by their presence in the group by and are actually optional in the projection. In that particular example I include them in the projection but if I hadn't they wouldn't appear in the final data. I felt this would all be the most expected behavior to end users.

agorajek:
@CircArgs this is real cool feature. I am still unsure where it can lead, but I do agree that a way to explore the DJ dag and query its data through a single DJQL prompt would be Amazing for SQL savvy users! Just not sure how many business users would actually ever see this feature. But scratching your own each is the best way to build good stuff!

@samredai I think this GROUP BY is useful because it is an indication of the intent. Of course DJ could figure out which are dimensions and which are metrics, but if you don't make this "SQL" similar to the other "SQLs" then it will be harder to use. I feel like we should require it to be even more proper SQL so people know better what to expect from it. Example:

```sql
SELECT 
  AGG(default.avg_repair_price) avg_repair_price, 
  default.hard_hat.country, 
  default.hard_hat.city 
FROM 
  metrics 
GROUP BY 
  default.hard_hat.country, 
  default.hard_hat.city 
ORDER BY       
  default.hard_hat.country, 
  default.hard_hat.city 
```

Side comment: why does Github not allow comments on other people comments. This copying down the whole reference is insane... :(

samredai:
@CircArgs @agorajek makes sense! I agree it's a good idea to think about how we can make sure that some valid SQL syntax exists for each functionality since some connector SDKs do client side validation and will fail the query before it even gets to the DJ server. That being said it feels right to keep it very permissive (for now) the way it is in this PR and we can incrementally move towards a single optimal syntax.

datajunction-server/datajunction_server/api/djql.py
```diff
@@ -0,0 +1,91 @@
...
+from fastapi import APIRouter, Depends, Request
+from sqlmodel import Session
+from sse_starlette.sse import EventSourceResponse
+
+from datajunction_server.api.helpers import build_sql_for_dj_query, query_event_stream
+from datajunction_server.models.query import QueryCreate, QueryWithResults
+from datajunction_server.service_clients import QueryServiceClient
+from datajunction_server.utils import get_query_service_client, get_session
+
+router = APIRouter()
```
samredai, line 16: We're starting to use tags to organize things nicely in the swagger UI so it'd be good to add a tag here. Something like: `router = APIRouter(tags=["djql"])`

datajunction-server/datajunction_server/api/djql.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@router.get("/djql/data", response_model=QueryWithResults)
```
samredai, line 19: I like how you referred to it in the docstring as "DJ SQL" since DJQL implies it's some query language other than SQL. So maybe the endpoints should also be `/djsql/...`

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -871,6 +856,49 @@ async def query_event_stream(  # pylint: disable=too-many-arguments
...
 
 
+def build_sql_for_dj_query(  # pylint: disable=too-many-arguments,too-many-locals
+    session: Session,
+    query: str,
+    engine_name: Optional[str] = None,
+    engine_version: Optional[str] = None,
+) -> Tuple[TranslatedSQL, Engine, Catalog]:
+    """
+    Build SQL for multiple metrics. Used by both /sql and /data endpoints
```
samredai, line 866: I think this docstring needs to be updated

datajunction-server/tests/api/djql_test.py
```diff
@@ -0,0 +1,380 @@
...
+    """
+    Test streaming djql
+    """
+    query = """
+    SELECT default.hard_hat.country,
+    default.hard_hat.city
+    FROM default.hard_hat
+        """
+
+    response = client_with_query_service.get(
```
samredai, line 375: I believe this test is failing in the CI because it's missing a few arguments. This should work:
```py
response = client_with_query_service.get(
    "/djql/stream/",
    params={"query": query},
    headers={
        "Accept": "text/event-stream",
    },
    stream=True,
)
```

samredai (approved):
Thanks @CircArgs! I left a few comments including what I think would fix the failed stream test. I'm itching to add a live sql editor to the UI now...👀

datajunction-server/datajunction_server/api/helpers.py
CircArgs: I think this is from copy pasta so we need to change it elsewhere as well

datajunction-server/datajunction_server/construction/dj_query.py
CircArgs: You can query any nodes actually. To my above example:
```sql
SELECT default.hard_hat.country, default.hard_hat.city  FROM default.hard_hat
```

agorajek (approved):
Thanks @CircArgs. This is great. As long as we flush it out before the big opening day we'll be ok :)

shangyian (approved):
Thanks for adding this @CircArgs! Changes look great to me

PR #685, opened 2023-08-02
WIP: Add deactivate and restore namespace endpoints

shangyian:
### Summary

For namespaces, this adds functionality that allows for:
* deactivate (aka "deletion" from the user perspective) - We deactivate the namespace if there are no nodes in it. If there are nodes in it and cascade=false, we raise an error. If there are nodes in it but cascade=true, we deactivate all the nodes inside the namespace and then the namespace.
* restore - This restores the namespace only, unless cascade=true, in which case it additionally restores all nodes in the namespace.

### Test Plan

Still need to add unit tests, but tested locally

- [ ] `make check` passes


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -897,3 +899,101 @@ async def query_event_stream(  # pylint: disable=too-many-arguments
 
             query = query_next
         await asyncio.sleep(stream_delay)  # pragma: no cover
+
+
+def deactivate_node(session: Session, name: str, message: str = None):
```
agorajek, line 904: Should we try and stay consistent with delete/restore? Or did you mean to use delete/restore terminology on the client side only?

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -897,3 +899,101 @@ async def query_event_stream(  # pylint: disable=too-many-arguments
...
+                    and element.node_revisions[-1].status == NodeStatus.INVALID
+                ):  # pragma: no cover
+                    downstream.current.status = NodeStatus.INVALID
+        else:
+            (
+                _,
+                _,
+                missing_parents_map,
+                type_inference_failed_columns,
+            ) = validate_node_data(downstream.current, session)
```
agorajek, line 975: This will change after my PR. https://github.com/DataJunction/dj/pull/680

datajunction-server/datajunction_server/api/namespaces/methods.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+        if node_type
+        else Node.namespace.like(  # type: ignore  # pylint: disable=no-member
+            f"{namespace}%",
+        )
+    )
+
+    list_nodes_query = select(Node.name).where(
+        where_clause,
+    )  # .where(is_(Node.deactivated_at, None))
```
agorajek, line 41: This doesn't seem finished. I think you need to fix it for `deactivated = True`.

datajunction-server/datajunction_server/api/namespaces/methods.py
```diff
@@ -0,0 +1,111 @@
...
+            entity_name=namespace.namespace,
+            node=None,
+            activity_type=ActivityType.RESTORE,
+            details={"message": message or ""},
+        ),
+    )
+    session.commit()
+
+
+def create_namespace(session: Session, namespace: str):
```
agorajek, line 97: What if one already exists?

datajunction-server/datajunction_server/api/namespaces/routes.py
```diff
@@ -0,0 +1,212 @@
...
+    type_: Optional[NodeType] = Query(
+        default=None,
+        description="Filter the list of nodes to this type",
+    ),
+    session: Session = Depends(get_session),
+) -> NodeNameList:
+    """
+    List node names in namespace, filterable to a given type if desired.
+    """
+    node_names = get_nodes_in_namespace(session, namespace, type_)
```
agorajek, line 91: Why two lines?

agorajek (approved):
@shangyian you are fast. I left few comments but most of them can be addressed later.

datajunction-server/datajunction_server/api/helpers.py
shangyian: Yeah, so I figured internally we can be a bit more clear that it's just deactivating and not deleting, but externally (i.e., anything that's potentially user-facing) we should call it "delete" and "restore".

datajunction-server/datajunction_server/api/namespaces/methods.py
shangyian: Makes sense, I changed the name to `include_deactivated`

datajunction-server/datajunction_server/api/namespaces/methods.py
shangyian: So this method was only supposed to be used to create a namespace object + register history, but doesn't include the check for whether a namespace already exists. That's part of the actual API endpoint (since not every use of the method needs that check or would respond the same way).

datajunction-server/datajunction_server/api/namespaces/routes.py
shangyian: good point, removed!

PR #686, opened 2023-08-02
Hard delete node

samredai:
### Summary

This adds a `DELETE /nodes/{name}/hard/` endpoint for hard deleting a node. Hard deleting a node attempts to actually delete all metadata related to this node in the metadata database. In a graph system, this is very destructive and has far reaching effects that aren't easily undone. For that reason, this is an endpoint that we probably shouldn't expose in any clients any time soon (possibly never?) and the API endpoint is probably best kept as an admin functionality. Users should be directed to use the standard `DELETE /nodes/{name}/` which performs a soft delete aka just deactivates the node.

Deleting a node involves two main steps. All downstream nodes are retrieved before deleting the node. The downstream nodes are then all invalidated and a history event is stored for them identifying that the upstream node has been deleted. The second step is specifically for dimension nodes for which we'd retrieve all linked nodes and store a history event that a link has been deleted (as well as invalidate cube nodes if the link was critical to that set of metrics and dimensions being composable).

Here are some examples of using this endpoint.

_All nodes valid_
Screenshot 2023-08-02 at 12 03 07 AM

_hard delete default.repair_order_
```sh
curl -X 'DELETE' \
  'http://localhost:8000/nodes/default.repair_order/hard/' \
  -H 'accept: application/json'
```
_response_
```json
{
  "message": "The node `default.repair_order` has been completely removed.",
  "impact": [
    {
      "name": "default.repair_orders",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.repair_order_details",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.repair_order_transform",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.num_repair_orders",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.avg_repair_price",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.total_repair_cost",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.total_repair_order_discounts",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.avg_repair_order_discounts",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.avg_time_to_dispatch",
      "status": "valid",
      "effect": "broken link"
    },
    {
      "name": "default.repair_orders_cube",
      "status": "invalid",
      "effect": "broken link"
    }
  ]
}
```

_default.repair_orders_cube now shows as invalid_
Screenshot 2023-08-02 at 12 04 14 AM

_the cube's history shows that a link was deleted and the cube was changed to invalid_
Screenshot 2023-08-02 at 12 04 52 AM

hard delete repair_order_details
```sh
curl -X 'DELETE' \
  'http://localhost:8000/nodes/default.repair_order_details/hard/' \
  -H 'accept: application/json'
```
_response_
```json
{
  "message": "The node `default.repair_order_details` has been completely removed.",
  "impact": [
    {
      "name": "default.avg_repair_price",
      "status": "invalid",
      "effect": "downstream node is now invalid"
    },
    {
      "name": "default.total_repair_cost",
      "status": "invalid",
      "effect": "downstream node is now invalid"
    },
    {
      "name": "default.total_repair_order_discounts",
      "status": "invalid",
      "effect": "downstream node is now invalid"
    },
    {
      "name": "default.avg_repair_order_discounts",
      "status": "invalid",
      "effect": "downstream node is now invalid"
    },
    {
      "name": "default.repair_orders_cube",
      "status": "invalid",
      "effect": "downstream node is now invalid"
    }
  ]
}
```

_all metrics defined on default.repair_order_details now show as invalid_
Screenshot 2023-08-02 at 12 05 46 AM

_metric history shows that a hard delete broke a link but the node remained valid, then a dependency was hard deleted and the node's status changed to invalid_
Screenshot 2023-08-02 at 12 11 33 AM

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -472,6 +474,90 @@ def activate_a_node(name: str, *, session: Session = Depends(get_session)):
     )
 
 
+@router.delete(
+    "/nodes/{node_name}/hard/",
+    response_model=List[MaterializationConfigInfoUnified],
+)
+def hard_delete_a_node(
+    node_name: str,
+    *,
```
agorajek, line 483: What is this wildcard for?

agorajek (approved):
Wow, this code looks scary simple! Thanks for doing this @samredai. It will help us keep things clean.

datajunction-server/datajunction_server/api/nodes.py
samredai: It just splits the positional and named arguments. So `hard_delete_a_node("foo", session)` won't work, you'd have to explicitly use `hard_delete_a_node("foo", session=session)`. I use it as more of a visual split to put all of the automatic dependency injected stuff after it, but I can remove it.

agorajek: Ah... visual split. Didn't realize. I don't mind seeing it there. Up to you.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -459,6 +460,89 @@ def restore_node(name: str, *, session: Session = Depends(get_session)):
...
+    """
+    node = get_node_by_name(session, node_name, with_current=True)
+    downstream_nodes = get_downstream_nodes(session=session, node_name=node_name)
+
+    linked_nodes = []
+    if node.type == NodeType.DIMENSION:
+        linked_nodes = get_nodes_with_dimension(session=session, dimension_node=node)
+
+    session.delete(node)
+    session.commit()
```
shangyian, line 483: Should we do everything here in a single commit so that it can be rolled back together if one of these pieces doesn't succeed?

shangyian (approved):
LGTM! Just one question around commit timing

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -459,6 +460,89 @@ def restore_node(name: str, *, session: Session = Depends(get_session)):
...
+        )
+
+    # Revalidate all linked nodes
+    for node in linked_nodes:
+        session.add(  # Capture this in the downstream node's history
+            History(
+                entity_type=EntityType.LINK,
+                entity_name=node_name,
+                node=node.name,
+                activity_type=ActivityType.DELETE,
```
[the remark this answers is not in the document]
samredai: I initially had this but the information didn't seem contextually relevant when I saw it in a UI. The linked node is still valid, just with a link getting dropped which this history event describes, even showing the node that caused the link to break. But the array of impact is not really about this node, it's about the other node that got deleted. It makes more sense then to have the impact stored in one event (for the node that got deleted) and then this event points you to that node.

Which raises another point, once a node is deleted, you can't navigate to it's page in the UI to see the history, even though it exists. @agorajek and I chatted a bit and thought a top level global "Activity" page might be nice that just displays all history with some limit of course.

datajunction-ui/webpack.config.js
```diff
@@ -99,6 +99,7 @@ module.exports = {
   plugins: [
     new HtmlWebpackPlugin({
       template: path.resolve(__dirname, 'public', 'index.html'),
+      favicon: path.resolve(__dirname, 'public', 'favicon.ico'),
```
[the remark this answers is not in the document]
samredai: Screenshot 2023-08-02 at 3 31 34 PM

🍾

datajunction-server/datajunction_server/api/nodes.py
samredai: So does that mean there isn't a way to do this all in one commit?

agorajek: I don't think there is a way, unless we can keep the node (being deleted) as existing but invalid and then re-validate the dag... and here comes the need to manual setting of a node to invalid. Something I never thought was needed.

samredai: That's an interesting idea, we could deactivate the node, then revalidate all of the downstream and linked nodes to get the impact, and then delete the node. We can then catch any issues and back out by just reactivating the node. It all sounds like it would work theoretically but it feels like a lot of steps with gotchas that aren't obvious right now. 🤔

agorajek: Yes @samredai , I don't think we need to improve it right now, but what you said makes perfect sense as a final solution.

shangyian: +1, no need to change it now but we can get there eventually!

PR #689, opened 2023-08-03
Add deactivating materialization to the Python client

shangyian:
### Summary

This adds the ability to deactivate node materializations from the Python client. @agorajek let me know if you think this should be called `delete` rather than `deactivate`, but I figure the latter actually makes more sense in the materialization context. 

### Test Plan

- [x] PR has an associated issue: #528 
- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/_internal.py
+++ b/datajunction-clients/python/datajunction/_internal.py
@@ -377,6 +377,22 @@ class DJClient:
         )
         return response.json()
 
+    def _deactivate_materialization(
+        self,
+        node_name: str,
+        materialization_name: str,
+    ):
+        """
+        Upserts a materialization config for the node.
+        """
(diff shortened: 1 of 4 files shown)


agorajek:
@shangyian great question, and I agree that `delete` seems incorrect here and `deactivate` could be an option, but what do you think about `invalidate` ?

shangyian:
@agorajek Hmm, I think I prefer `deactivate` to `invalidate`. Because invalidate sounds like something that a system should be doing, not something that a user is doing with intention. If someone comes and deactivates a materialization, they are communicating some level of active intention around that choice, whereas invalidation sounds more passive.

agorajek:
Sounds good @shangyian . I think your intuition is correct.

agorajek (approved):

Issue #692, opened 2023-08-04
proposal: landing page image


shangyian:
![image](https://github.com/DataJunction/dj/assets/9524628/65c913a6-9586-449b-8741-0ba888b8724b)

I just spent some time in Figma and came out with the above. What do we think about having this be the landing page image under "introduction" in the docs? @samredai @agorajek @CircArgs @betodealmeida 

Let me know if you want something different on there! (and @CircArgs if you have a vectorized logo for Headjack I can put it on)

samredai:
Yes!

PR #694, opened 2023-08-04
Move to Doks theme and change landing image

shangyian:
### Summary

Screenshot 2023-08-04 at 11 14 15 AM

### Test Plan

- [x] PR has an associated issue: #692
- [x] `make check` passes


docs/i18n/en.yaml
```diff
@@ -0,0 +1,17 @@
+- id: get-started
```
samredai, line 1: This file isn't changed from the theme file so if you remove it here it should get picked up from there.

docs/hugo.toml
```diff
@@ -0,0 +1 @@
+theme = ['doks']
```
samredai, line 1: This can get added to the main config/_default/config.toml file and this file can be removed.

docs/assets/scss/common/_shortcodes.scss
```diff
@@ -0,0 +1,83 @@
+$padding-16: 1rem !default;
```
samredai, line 1: Since this is file is all custom CSS that has to be manually injected into the theme, maybe it would be cleaner if we add the small edits from _code.scss and _app.scss here and then have it load at the end. That way all of our custom css that we're adding to theme is in one place (and maybe rename the file _custom.scss).

docs/config/_default/markup.toml
```diff
@@ -0,0 +1,29 @@
+defaultMarkdownHandler = "goldmark"
```
samredai, line 1: This is unchanged from the theme file so removing it here should pick it up there.

docs/data/docs-versions.yml
```diff
@@ -0,0 +1,53 @@
+# - group: v1.x
```
samredai, line 1: Very nice that this supports versioning! I'm curious to see how that gets wired up with the netlify deploy.

docs/layouts/partials/header/header.html
```diff
@@ -0,0 +1,223 @@
...
+
+<div class="header-bar"></div>
+
+<header class="navbar navbar-expand-lg navbar-light doks-navbar">
+  <nav class="container-{{ if .Site.Params.options.fullWidth }}fluid{{ else }}xxl{{ end }} flex-wrap flex-lg-nowrap" aria-label="Main navigation">
+
+    <a class="navbar-brand order-0" href="{{ .Site.BaseURL | relLangURL }}" aria-label="{{ .Site.Params.Title }}">
+      {{ .Site.Params.Title }}
+      <svg width="30" height="30" viewBox="0 0 152.735065 152.735065" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+        <title>DJ 2</title>
```
samredai, line 17: Is this a typo?

samredai (commented):
@shangyian this is great. I left a number of inline comments but also the `themes/doks` directory should be switched to a submodule. That way we don't have to source control all of the theme code and can easily pull in any bugfixes or new features that get released for it by just updating the submodule commit.

docs/layouts/partials/header/header.html
shangyian: This is just an SVG definition, so the title doesn't actually matter.

agorajek (approved):
@shangyian the new front page looks awesome. No code comments from me, but @samredai 's notes make sense.

docs/hugo.toml
shangyian: Nice, removed!

docs/config/_default/markup.toml
shangyian: Done!

docs/assets/scss/common/_shortcodes.scss
shangyian: So I merged these into a single file, but we still have to override _app.scss because that's what brings them altogether.

PR #695, opened 2023-08-04
Python client cleanup: namespaces and test separation.

agorajek:
### Summary

What the title says plus some minor adjustment of codes in server api namespaces.py.

This is pretty much it. There are few very minor TODOs left plus the README update and the demo notebook.

Note: While I was reviewing all the methods I though that there is a set of methods (especially around catalogs and engines) which seems to be falling out of a typical use by DJClient or DJ Builder. I have kept it in the Builder side for now, but I wonder what you all think about making a 3rd category for the system-admin tasks, say called DJAdmin.

### Test Plan

- [ ] PR has an associated issue: #630 
- [ ] `make check` passes

### Deployment Plan

n/a


datajunction-clients/python/datajunction/client.py
```diff
@@ -17,7 +16,9 @@ class DJClient(_internal.DJClient):
     """
 
     #
-    # List
+    # List objects:
+    # - basic: namespaces, dimensions, metrics, cubes
+    # - TODO (maybe): sources, transforms, nodes...
     #
     def list_namespaces(self, prefix: Optional[str] = None) -> List[str]:
```
shangyian, line 23: fwiw, I think `list_sources` would be useful. I've found myself wanting that while I use the client

datajunction-clients/python/datajunction/client.py
```diff
@@ -50,58 +51,59 @@ def list_metrics(self, namespace: Optional[str] = None) -> List[str]:
...
-        """
-        Return common metrics (names only) for a set of dimensions.
-        """
-        metrics = self.common_metrics_with_details(dimensions=dimensions)
-        return [metric["name"] for metric in metrics]
+        ).json()
+        if name_only:
+            return [metric["name"] for metric in json_response]
+        return [
+            {
```
shangyian, line 98: Would it be better here to return a list of `Metric` objects rather than a dict? Or is the idea that this would eventually provide `paths` or other DAG information, thus making it incompatible with a `Metric` object?

datajunction-clients/python/datajunction/builder.py
```diff
@@ -0,0 +1,388 @@
...
+        name: str,
+        catalog: str,
+        schema_: str,
+        table: str,
+        description: Optional[str] = None,
+        display_name: Optional[str] = None,
+        columns: Optional[List[models.Column]] = None,
+        primary_key: Optional[List[str]] = None,
+        tags: Optional[List[models.Tag]] = None,
+        mode: Optional[models.NodeMode] = models.NodeMode.PUBLISHED,
```
shangyian, line 125: Nice, fixes the not being able to save as draft issue :)

datajunction-clients/python/datajunction/client.py
```diff
@@ -50,58 +51,59 @@ def list_metrics(self, namespace: Optional[str] = None) -> List[str]:
             )
         return self._session.get("/metrics/").json()
 
+    def list_cubes(self, namespace: Optional[str] = None) -> List[str]:
```
shangyian, line 54: Unrelated to this PR, but I wonder if something like `find_cube_by_metrics_and_dimensions` would be useful...

datajunction-clients/python/datajunction/client.py
agorajek: Sounds good. Will do.

datajunction-clients/python/datajunction/client.py
agorajek: Yes, exactly, I think this return type should be minimal, similarly to the dimension side. I think adding paths to the dimensions would also be interesting here.

datajunction-clients/python/datajunction/client.py
agorajek: I am not sure on this. Feels like unless you know of a cube with some content of interest, trying to find cube by providing metrics and dimensions could be a little like shooting in the dark... unless we could provide a list of cubes ordered by closest coverage. Still, I think this is something we may need to see needed first.

samredai: Instead of by metrics and dimensions, I think by a single metric _or_ by a single dimension would definitely be useful. In other words, "show me what cubes use this single metric" would be great for exploring.

samredai (approved):
You raised a good point about the need to separate what's really admin functionality. I wonder if admins really even need a client at all or is it reasonable that they would just call the API directly. It will eventually be controlled through authorization so only authenticated admins can call it and would setup nicely for an "admin page" in the UI.

datajunction-clients/python/datajunction/client.py
agorajek: @shangyian let me close this PR and add this in a follow up. I'd like to refresh the Readme and Notebook first.

shangyian: Sounds great! Thanks @agorajek

shangyian (approved):

Issue #697, opened 2023-08-05
Add user authentication via OAuth


samredai:
This is a set of tasks related to adding support for various OAuth flows.

- [x] Create a router for basic auth oauth flow
- [x]  Create a router for GitHub OAuth
- [x]  Create a router for Google OAuth
- [x]  Allow oauth provider configuration via server settings
- [x]  Add middleware for different oauth flows that inject the user into the request
- [ ]  ~Determine oauth provider rankings when multiple are enabled. In other words, if both a GitHub and Google access token are detected, which should be used?~
- [x]  Design a pattern for injecting custom oauth and document it
- [x]  Block endpoints behind auth (allowing for a list of exceptions, i.e. /docs)

Issue #696, opened 2023-08-05
Add concept of a user


samredai:
This is an issue to track the various items required for adding the concept of a user to DJ.

- [ ] Add a user metadata model
- [ ] Add a POST endpoint for creating a basic auth user
- [ ] Preload a dj:dj user in the demo database
- [ ] Simple oauth username and password flow
- [ ] Decide on whether we add an “ownership” concept and how it would behave

- Add “created_by” field for all entities and set it on creation:
  - [ ] Node
  - [ ] Namespace
  - [ ] Materialization config
  - [ ] Tag

- [ ] “Create User” page for the UI
- [ ] Login page for the UI
- [ ] Display current user in UI
- [ ] Documentation describing how users work

vcknorket:
This is a good one. We also noticed that created_by is super critical when managing. Will help immensely in managing it in enterprise deployment

samredai:
@vcknorket woah I was poking around our issues and can't believe I never saw your messages! GitHub notifications are the ones I always seem to miss....anyway this is in :)

[Here](https://datajunction.io/docs/0.1.0/developers/authentication/) are some docs--for enterprise use cases all you would need to do is override the `get_current_user` dependency. User objects will then be upserted into DJ's metadata db as they are authenticated.

Issue #64, opened 2022-04-06
Authorization


betodealmeida:
Currently there's no authentication/authorization, and the database passwords are exposed in the `/databases/` endpoint. it would be nice to have an auth mechanism, preventing users from seeing database passwords, and also restricting access to databases, catalogs, schemas, tables, columns, etc.

betodealmeida:
One of the ideas is to implement a token based mechanism where authorization is defined by SQL queries.

For example, a power user could have access to a full database "postgres", and a specific schema in "trino":

```sql
-- this is pseudo SQL, and tables are qualifed as database.catalog.schema.table
SELECT * FROM postgres;
SELECT * FROM trino.hive.public;
```

An analyst could have access to only a few columns in a given table:

```sql
SELECT event_time, price FROM postgres.default.public.sales;
```

These queries would be encoded in a token (similar to a JWT), and for every data request for permission.

A cool thing is that we can allow queries to be "stacked" using [macaroons](https://github.com/rescrv/libmacaroons). For example, if a user has a token giving them access to a given table:

```sql
SELECT * FROM trino.hive.public.sales;
```

They can generate a new token with an additional constraint:

```sql
SELECT event_time, price FROM trino.hive.public.sales;
```

This new token can be given to a different user, eg, if they want to view a chart that was built on top of the `sales` table.

PR #698, opened 2023-08-06
Support for Basic and GitHub OAuth

samredai:
### Summary

This adds support for a basic as well as a github oauth flow that's enabled when the server is configured. I'll explain each independently and then describe how these work when multiple are enabled which also outlines the behavior when/if support for more oauth providers are added.

## Structure

Each type of OAuth has a separate router defined in the `/api/authentication/` submodule and any helper functions are stored in a matching named file in the `/internal/authentication/` submodule. For example, `/api/authentication/github.py` and `/internal/authentication/github.py`. The routers are conditionally mounted if you have the appropriate configured values in the `.env` server configuration.

Basic auth is only enabled if `SECRET` is set. (used for password hashing)
GitHub auth is only enabled if additionally `GITHUB_OAUTH_CLIENT_ID` and `GITHUB_OAUTH_CLIENT_SECRET` are set.

## Basic OAuth flow

The flow for basic auth includes creating users as well as logging users in. For creating users (the orange path), there's a simple `/basic/user/` POST endpoint where a username and password can be sent. The info is saved to the database with the password being hashed using the secret set in the env. There's a lot more that can be done with this path such as storing an `is_verified` flag and sending verification emails, but this simple setup is sufficient for the demo environment. A more robust user creation flow can be added through a custom auth router.

For logging users in (the green path), the username and password is provided, the password is compared against the hashed version of the password stored in the db, and if they match, an encrypted JWT is sent back and stored as an HTTP-only cookie.

image

## GitHub OAuth flow

For github oauth, there is no separate user creation flow. The login flow automatically creates a user once they authorize the DJ server to use their github account. The login flow (green path) shows a login request first goes to `/github/login/` which forwards them to an authorization screen for the application registered with GitHub, identified by the `GITHUB_OAUTH_CLIENT_ID` configured. Once the user clicks the big green authorize button, it forwards a code to the `/github/token/` endpoint on the DJ server. Then the token exchange happens and the authorization code is sent to github and exchanged for an access token. A "successfully logged in" message is returned and that token is then stored as an HTTP-only cookie in the user's browser. During this process, if a user doesn't exist in the DJ metadata db, one will automatically be created with the value for `oauth_provider` as `"github"`.

image

## What happens after a user is logged in?

When a user is logged in, that means they have a cookie stored that will either authenticate them via basic auth or via github (both may be enabled). The diagram below shows the order of operations for the middleware that checks for this cookie on every request. In this example, the user is making a request to `/foo/` and first the basic auth middleware checks for a basic auth JWT, if one is found, the user is set on the request state which can be accessed in the endpoint logic if needed. If a basic auth cookie is not found, then it checks for a cookie with a github access token and if found, the user is set on the request state. If neither are found, a 401 unauthorized is returned.

image

## Adding a custom authentication flow

The only requirement for any custom authentication flow is that they store users in the user metadata table and they include middleware that sets the `user` attribute on the request state. That would expose the user to all routes requiring authorization in the same way. As we add authorization logic, that consistency will be important.

## Remaining work and discussion items

1. We'll need to add a create user and a login page for the UI that uses these endpoints. We'd also need some way for it to detect which oauth providers are enabled by the DJ server (maybe an `/oauth/providers/` endpoint that just returns a list).
2. The order of the middleware implies a ranking. A user could technically have both a basic auth and a github auth cookie (could be made rare by a strict UI flow) and whichever we check first would be the "logged in" account that's detected for all requests. There are 3 ways I can think of that we can handle this:
    - Keep it as simple as a ranking and we check the highest rank first then stop once we've found a valid token.
    - Use the same cookie key so that logging in with a different oauth provider automatically logs out of any other logged in accounts. This would involve looking up the external user info during login (i.e. from GitHub) and exchanging it for a DJ session token encoded as a JWT.
    - Support being logged in on multiple accounts, one account per oauth provider. In the UI this would look something like being able to switch between "profiles" where you may see a list with your basic auth username/password account, your github account, and your google account.
3. What do we want our general strategy to be in terms of adding support for oauth providers? I've already confirmed that this works with a google oauth flow--should we continue adding support for various oauth providers or is it ok to just have documentation on how to add a custom oauth provider and leave it to people to implement it themselves? There are many [oauth providers](https://en.wikipedia.org/wiki/List_of_OAuth_providers) out there and I definitely don't expect us to add support for all of them...
4. Although storing the access tokens in a cookie works, there's a few steps we can take to secure it a bit further besides just encrypting it (which this PR already does). For example, we can maintain the concept of a session in our system and exchange the oauth access token for a separate encrypted token that we mint. We'd then set this new DJ session token as a cookie and maintain session state on the server. Much of this goes back to point 3, there are many ways to do auth and I don't know how opinionated we want the OSS project to be on that.

### Test Plan

(Working on adding tests)

- [x] PR has an associated issue: #64 #696 #697 
- [x] `make check` passes


datajunction-query/djqs/config.py
```diff
@@ -1,5 +1,5 @@
 """
-Configuration for the metric repository.
+Configuration for the query service
```
agorajek, line 2: Thanks for cleanup.

datajunction-server/datajunction_server/api/authentication/basic.py
```diff
@@ -0,0 +1,73 @@
...
+        content={"message": "User successfully created"},
+        status_code=HTTPStatus.CREATED,
+    )
+
+
+@router.post("/basic/login/")
+async def login(
+    form_data: OAuth2PasswordRequestForm = Depends(),
+    session: Session = Depends(get_session),
+):
```
agorajek, line 50: Missed the output type?

datajunction-server/datajunction_server/models/user.py
```diff
@@ -0,0 +1,40 @@
...
+    id: Optional[int] = Field(default=None, primary_key=True)
+    username: str
+    password: Optional[str]
+    email: Optional[str]
+    name: Optional[str]
+    oauth_provider: OAuthProvider
+    is_admin: bool = False
+
+
+class UserOutput(BaseModel):
```
agorajek, line 32: Not related to this PR, but I was wondering if adding an `Output` suffix is a common way to create a simplified class for returning it in the HTTP requests?

agorajek (commented):
Thanks @samredai . This looks ok to me but this is a bit of a dog-science to me. Could you at least add a short explanation how you tested this?

samredai:
@agorajek thanks for the early review. I kept this as a draft since I figured it's worth a discussion in the OSS sync first. I'm also working on some diagrams that explain the flow so I'll add those soon.

datajunction-server/datajunction_server/api/main.py
```diff
@@ -37,84 +39,119 @@
...
-            content=exc.to_dict(),
-            headers={"X-DJ-Error": "true", "X-DBAPI-Exception": exc.dbapi_exception},
-        )
-
-    FastAPIInstrumentor.instrument_app(application, tracer_provider=tracer_provider)
-
-    return application
-
-
-app = get_dj_app()
```
shangyian, line 120: Any reason why this is changed to creating the app directly rather than via this function call?

shangyian (approved):
Overall looks great! Thanks for adding @samredai

agorajek (approved):
This is all a mystery to me but thank you for your PR @samredai !

datajunction-server/datajunction_server/api/main.py
samredai: TLDR; the value of clarity in this file is starting to feel worth more than the value of having tracing in the demo environment.

I initially added the function call to allow injecting a custom open telemetry tracer configuration, but this `main.py` is only really used for the demo. So having the otel stuff here didn't really seem informative or useful since our recommendation for someone wanting to instrument otel would probably be to mount all the routers to their own app instance and follow any standard instructions out there for otel instrumentation in fastapi.

datajunction-server/datajunction_server/api/authentication/basic.py
samredai: Updated with all the output types set 👍

datajunction-server/datajunction_server/api/main.py
```diff
@@ -37,84 +39,119 @@
...
+from datajunction_server.models.user import User
 from datajunction_server.utils import get_settings
 
 if TYPE_CHECKING:  # pragma: no cover
     from opentelemetry import trace
 
 _logger = logging.getLogger(__name__)
+settings = get_settings()
 
+UNAUTHENTICATED_ENDPOINTS = [
```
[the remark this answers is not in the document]
samredai: I considered that but started to realize that might need some more thoughtful logic. Particularly when we start to expect and use `request.state.user` in the various endpoints, it wouldn't be possible then for those endpoints to ever be unauthenticated. That is unless we do something like automatically plug in a user `unknown` when DJ is used with no auth configured. That may be all we need but if it's ok, let me try and tackle that in an follow-up PR.

(I do get that a custom DJ setup could have additional custom endpoints appended on and those may not need auth)

datajunction-server/datajunction_server/api/authentication/github.py
```diff
@@ -0,0 +1,87 @@
...
+
+@router.get("/github/login/", status_code=HTTPStatus.FOUND)
+async def login():  # pragma: no cover
+    """
+    Login
+    """
+    settings = get_settings()
+    oauth_client_id = settings.github_oauth_client_id
+    return RedirectResponse(
+        url=github.get_authorize_url(oauth_client_id=oauth_client_id),
```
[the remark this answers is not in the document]
samredai: Done!

datajunction-server/datajunction_server/api/main.py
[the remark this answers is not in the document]
samredai: Created an issue so I remember to follow-up on this: https://github.com/DataJunction/dj/issues/708

PR #700, opened 2023-08-08
Adjust Python client README.

agorajek:
### Summary

Rework the Python READMe file and fix few things I ran into.

### Test Plan

Tried all the commands manually in our Jupyter notebook.

- [x] PR has an associated issue: #630 
- [x] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/datajunction-clients/python/README.md
+++ b/datajunction-clients/python/README.md
@@ -1,12 +1,20 @@
 # DataJunction Python Client
 
+This is a short introduction into the Python version of the Data Junction (DJ) client.
+For a full comprehensive intro into the DJ functionality please check out [datajunction.io](https://datajunction.io/).
+
 ## Installation
+
 To install:
 ```
 pip install datajunction
 ```
 
-## Examples
+## DJ Client : Basic Access
+
+Here you can see how to access and use the most common Data Junction features.
+
+### Examples
 
 To initialize the client:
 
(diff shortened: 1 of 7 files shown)


datajunction-clients/python/README.md
```diff
@@ -16,130 +24,346 @@ from datajunction import DJClient
...
     description="Repair orders",
     catalog="dj",
     schema_="roads",
     table="repair_orders",
 )
-repair_orders.save(mode=NodeMode.PUBLISHED)
 ```
 
-Nodes can also be created as drafts with:
+Nodes can also be created in draft mode:
```
shangyian, line 212: Would it make sense to show this creation path instead?
```
repair_orders = djb.create_source(
    name="repair_orders",
    display_name="Repair Orders",
    description="Repair orders",
    catalog="dj",
    schema_="roads",
    table="repair_orders",
    mode=NodeMode.DRAFT,
)
```

I ask because otherwise the creator will first create the node in published mode (which won't work for nodes with non-existent parents) and then switch it to draft, whereas the former just immediately creates it as a draft. Alternatively we could mention that there are two ways to create a node, one using `create_source` and one with creating the Node object locally. Or we could just leave this to the datajunction.io docs...

shangyian (approved):
Overall looks great -- thanks for updating and fixing these issues! Just left a minor comment but it can be left for another round.

datajunction-clients/python/README.md
```diff
@@ -1,12 +1,20 @@
...
+
 To install:
 ```
 pip install datajunction
 ```
 
-## Examples
+## DJ Client : Basic Access
+
+Here you can see how to access and use the most common Data Junction features.
```
samredai, line 15: Just a small nit: we should eventually settle on a single way to write the full project name across documentation. I see a lot of variations like datajunction, DataJunction, Datajunction, etc.

datajunction-clients/python/README.md
```diff
@@ -16,130 +24,346 @@ from datajunction import DJClient
 dj = DJClient("http://localhost:8000")
 ```
 
-### Catalogs and Engines
+**NOTE**
+If you are running in our demo docker container please change the above URL to "http://dj:8000".
```
samredai, line 28: s/docker container/docker environment/

datajunction-clients/python/README.md
```diff
@@ -16,130 +24,346 @@ from datajunction import DJClient
...
     SELECT
       count(repair_order_id)
     FROM repair_orders
     """,
     description="Number of repair orders",
 )
-num_repair_orders.save()
 ```
+
+### Reference
```
samredai, line 268: Maybe we can eventually use something like pydoc to automatically generate this kind of reference documentation for the entire client library.

samredai (approved):
This is great, thanks @agorajek!

datajunction-clients/python/README.md
agorajek: We should... `DataJunction` and `DJ` ?

datajunction-clients/python/README.md
agorajek: Good point @shangyian . I'll change it to:
```
```python
from datajunction import NodeMode

repair_orders = djb.create_source(
    name="repair_orders",
    display_name="Repair Orders",
    description="Repair orders",
    catalog="dj",
    schema_="roads",
    table="repair_orders",
    mode=NodeMode.DRAFT
)
```

datajunction-clients/python/README.md
agorajek: That would be fantastic.

Issue #702, opened 2023-08-08
Request failed with status code 422 on source node save().


agorajek:
Making the issue JIC I forget to address it:

```python
from datajunction import DJBuilder
djb = DJBuilder("http://dj:8000")

repair_orders = djb.create_source(
    name="default.repair_orders",
    display_name="Repair Orders",
    description="Repair orders",
    catalog="dj",
    schema_="roads",
    table="repair_orders",
)

from datajunction import NodeMode
repair_orders.save(mode=NodeMode.DRAFT)
```

PR #705, opened 2023-08-09
Fix a bug with updating nodes.

agorajek:
### Summary

The bug was reproducible with the unit test (see code), but it was temporarily commented out. Fixed now. This also fixed an issue with a cube creation test as well.

### Test Plan

Unit tests.

- [x] PR has an associated issue: #702 
- [ ] `make check` passes

### Deployment Plan

n/a


shangyian (approved):
Nice!

PR #709, opened 2023-08-10
Fix our Python integration test.

agorajek:
### Summary

What the title says plus adjustments for the new new auth. Also two small doc fixes.

### Test Plan
Ran tests ok.

- [x] PR has an associated issue: #701
- [x] `make check` passes


datajunction-clients/python/tests/test_integration.py
```diff
@@ -603,15 +602,15 @@ def test_integration():  # pylint: disable=too-many-statements,too-many-locals,l
...
-                0: 31921452.0,
-                1: 19100529.0,
-                2: 47819241.0,
-                3: 55808838.0,
-                4: 77991579.0,
-                5: 53141913.0,
-                6: 12969882.0,
-                7: 17111574.0,
-                8: 52569162.0,
+                0: 131364.0,
```
shangyian, line 605: Did you figure out what caused these to change? This makes it seem like we were generating and running the wrong query at some point...

shangyian (approved):
Yay!

datajunction-clients/python/tests/test_integration.py
agorajek: No I did not. I think that the same query at some point must have started to produce different results. Otherwise the test would be failing on the query string comparison before it gets to the results comp. Or maybe we overloaded some tables in the DB when we were building these examples.

PR #710, opened 2023-08-10
Add get_data_by_query_id to server.

agorajek:
### Summary

Because it was missing and we couldn't ask for data based on existing query ID.
There were some inconsistencies wrt handling query ID so I converted everything to `uuid.UUID`. But I am not sure it was the right call. Please lmk.

### Test Plan

- [ ] `make check` passes


samredai:
> There were some inconsistencies wrt handling query ID so I converted everything to `uuid.UUID`

Can you elaborate a bit on this? I think this may be too rigid for any custom query services that don't use UUIDs as query IDs.

agorajek:
> > There were some inconsistencies wrt handling query ID so I converted everything to `uuid.UUID`
> 
> Can you elaborate a bit on this? I think this may be too rigid for any custom query services that don't use UUIDs as query IDs.

Yes, I think some related functionality already expected UUIDs. But I agree with you so let me rewrite it.

datajunction-server/datajunction_server/service_clients.py
```diff
@@ -123,7 +123,7 @@ def submit_query(  # pylint: disable=too-many-arguments
 
     def get_query(
         self,
-        query_id: str,
+        query_id: UUID,
```
shangyian, line 126: Yeah, agreed that we probably shouldn't use UUID here. It was set up this way previously but for our internal query service to work with this, we had to do some additional string parsing to get a uuid extracted for this.

datajunction-server/datajunction_server/api/data.py
```diff
@@ -174,6 +174,22 @@ def get_data(  # pylint: disable=too-many-locals
     return result
 
 
+@router.get(
+    "/data/query/{query_id}",
+    response_model=QueryWithResults,
```
shangyian, line 179: What if data is no longer available for that query ID? Will it just return an empty list or do we need a new query results status?

agorajek: Let me test that... and maybe add a wrapper.

samredai: Instead of a new results status I would just raise a `DJException` that has `http_status_code=HTTPStatus.NOT_FOUND` so it returns a 404.

datajunction-server/datajunction_server/service_clients.py
agorajek: Yes, taking it all out.

agorajek:
This is ready to go.

samredai (approved):

shangyian (approved):
LGTM!

PR #712, opened 2023-08-11
Add column-level lineage endpoint for a node + bugfix

shangyian:
#### Column-level lineage

This PR adds an endpoint to provide column-level lineage for a given node. For each column on the node, we build out a column-level dependency graph, tracing all the way to the columns on the final layer of source node dependencies. 

This search is done by expanding each layer with all columns referenced by the current column's expression. This expansion happens both within a node's query (see the `processed` queue) and across nodes (see the recursive call to `column_level_lineage`). Every time we reach a table with a DJ node attached, we save it in the lineage record and continue to expand the search to upstream nodes. We stop the search whenever we reach columns on a source node.

#### Bugfix

During the compile process, we go through all columns in a DJ node's query and attempt to assign table sources to each column. However, this fails for some cases where we have a highly nested set of subqueries. 

To be more specific, after we've found a potential table candidate for a column, we call `add_ref_column` to determine if we can use it as a table for the column. At some point in this call, it checks to see if the column name matches a column on the table, which fails because it's using the fully qualified column name (i.e., with a namespace) rather than just the column name (which would match the column names on the table).

We fix this by changing `col.alias_or_name(identifier=True)` to `col.alias_or_name.name`. 

### Test Plan

- [ ] `make check` passes


agorajek (approved):
LGTM. It sounds like we did not have that bug recorded in any unit tests.

shangyian:
@agorajek Yeah, unfortunately not... I think to catch this we'll need a suite of nodes that have more complex queries than the roads examples. I'll start adding some more test cases in a separate PR.

PR #713, opened 2023-08-11
Column Lineage UI + more complex transforms and metrics in tests

shangyian:
#### Column Lineage UI

This adds a column lineage view for metric nodes only. It's limited to metrics because column-level lineage isn't very useful for other node types -- i.e., a transform node often refers to numerous upstream columns on its parents, so the resultant graph is too busy to be informative. However, metrics often only use a few columns from its parent nodes, so column-level lineage is actually useful to visualize. 

Also did some refactoring so that both ReactFlow graphs now share `LayoutFlow` and only the data pull is separate for each use case.

#### Test Cases

I added some more complex transform nodes to the Roads example database: `default.regional_level_agg` and `default.national_level_agg`. These build aggregations at the regional and national level. From these transforms, we can then build a more complex metric `default.regional_repair_efficiency`:

Note that this metric actually queries from *two* nodes, which breaks one of the rules we'd set on the metric node's definition. I suspect we should relax this restriction, as the above metric isn't actually possible in DJ if we don't. (We can't cleanly create an upstream transform that aggregates at multiple levels, so metrics have to be able to source from multiple upstream nodes). 

This is what the the `Regional Repair Efficiency` metric looks like in the UI's column lineage view:

Screenshot 2023-08-13 at 8 24 34 AM

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/sql/functions.py
```diff
@@ -1195,29 +1215,37 @@ def infer_type(
     return ct.DateType()
 
 
-class DateDiff(Function):
+class Datediff(Function):
```
shangyian, line 1218: Having it as `DateDiff` meant that we expect the spark function to be `date_diff`, but the actual function is `datediff` without the underscore.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -764,15 +764,15 @@ def check_metric(self):
             )
 
         # must query from a single table
```
shangyian, line 766: @samredai and @agorajek - we should probably discuss this change. I don't remember the original rationale for keeping metrics limited to querying a single node, because there are times that upstream transforms don't cleanly solve for this case. See the description in the PR for more details. Alternatively we could implement derived metrics to support something like this, but I think it's easier to just relax this restriction until we get around to that.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -764,15 +764,15 @@ def check_metric(self):
...
-        if (
-            len(tree.select.from_.relations) != 1
-            or tree.select.from_.relations[0].extensions
-        ):
-            raise DJInvalidInputException(
-                http_status_code=HTTPStatus.BAD_REQUEST,
-                message=f"Metric {self.name} selects from more than one node, "
-                "should only select from a single node",
-            )
+        # if (
```
agorajek, line 767: missed cleanup ?

datajunction-server/tests/api/metrics_test.py
agorajek, line 1: forgot to cleanup?

agorajek (approved):
So awesome. I am going to use this page for demo right away. Reviewed the backed code, but JS is still magic to me.

datajunction-server/datajunction_server/models/node.py
shangyian: @agorajek So technically it is possible to create a single transform, but I wanted to be able to label the primary key on the transform, which would then be different for `default.regional_level_agg` and `default.national_level_agg`. That's what I meant by "unclean".

PR #715, opened 2023-08-11
Add missing client list_ methods.

agorajek:
### Summary

1. Added all remaining list_* methods to the Python client.
2. Added a special error message when creating cube node. Because the current error may be not enough / confusing.

### Test Plan

unit tests

- [x] PR has an associated issue: #630 
- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/_internal.py
+++ b/datajunction-clients/python/datajunction/_internal.py
@@ -203,13 +203,25 @@ class DJClient:
     def _get_nodes_in_namespace(
         self,
         namespace: str,
-        type_: Optional[str] = None,
+        type_: Optional[models.NodeType] = None,
     ):
         """
-        Retrieves all nodes in the namespace
+        Retrieves all nodes in given namespace.
         """
         response = self._session.get(
-            f"/namespaces/{namespace}/" + (f"?type_={type_}" if type_ else ""),
+            f"/namespaces/{namespace}/" + (f"?type_={type_.value}" if type_ else ""),
+        )
+        return response.json()
+
+    def _get_all_nodes(
+        self,
+        type_: Optional[models.NodeType] = None,
+    ):
+        """
+        Retrieve all nodes of a given type.
+        """
+        response = self._session.get(
+            "/nodes/" + (f"?node_type={type_.value}" if type_ else ""),
         )
         return response.json()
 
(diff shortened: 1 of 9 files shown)


datajunction-clients/python/datajunction/_internal.py
```diff
@@ -203,13 +203,25 @@ def process_results(results) -> "pd.DataFrame":
...
+
+    def _get_all_nodes(
+        self,
+        type_: Optional[models.NodeType] = None,
+    ):
+        """
+        Retrieve all nodes of a given type.
+        """
+        response = self._session.get(
+            "/nodes/" + (f"?node_type={type_.value}" if type_ else ""),
```
samredai, line 224: This is one of two ways to get a list of nodes of a specific type--there's also the /metrics and /dimensions endpoints. I wonder if we should get rid of those so there's just a single way to do this. On a similar note, I wonder if this /nodes endpoint should also have an optional `?namespace=...` argument and then we can get rid of the /namespaces/{namespace}/ endpoint. Not saying this all needs to be done in this PR but curious to hear your thoughts on it.

samredai (approved):
LGTM, just had a comment that's more about how the server API is laid out.

datajunction-clients/python/datajunction/_internal.py
agorajek: Totally agree with simplifying the server side. We don't need separate listing methods for each node type on the server side. I was actually going to do it in the same PR, just ran out of time. Will do in a follow-up.

agorajek: https://github.com/DataJunction/dj/issues/716

datajunction-clients/python/datajunction/client.py
```diff
@@ -31,31 +29,76 @@ def list_namespaces(self, prefix: Optional[str] = None) -> List[str]:
...
+            return self._get_nodes_in_namespace(
+                namespace=namespace,
+                type_=models.NodeType.CUBE,
+            )
+        return self._get_all_nodes(type_=models.NodeType.CUBE)
+
+    #
+    # List other nodes: sources, transforms, all.
+    #
+    def list_sources(self, namespace: Optional[str] = None) -> List[str]:
```
shangyian, line 66: Should there also be an endpoint to list source nodes that were created via the register-a-table API endpoint? I know that users can also do that just by calling `list_sources(namespace="source")` (or whatever the default namespace was configured to be server-side), but they might not know what the default registered table sources namespace is. 

No need to do it in this PR, of course!

shangyian (approved):
LGTM!

PR #720, opened 2023-08-15
UI Login/Logout functionality, switch backend to support cookies and bearer tokens

samredai:
### Summary

This is a follow up to #698 and brings the new auth logic all the way to the UI by adding a login flow that supports basic auth and GitHub auth. I also simplified some of the auth logic on the server. Now the token is created the same way regardless of oauth provider, allowing for a single middleware instance. The backend also now expects tokens as part of authorization headers (using the `HTTPBearer` dependency that comes with fastapi).

https://github.com/DataJunction/dj/assets/43911210/6ca2a2a1-64fa-4c6e-b34d-89167b0ab3b7

Fastapi also detects the `HTTPBearer` dependency and the swagger documentation shows which endpoints are behind auth while also letting you test the entire login flow.

https://github.com/DataJunction/dj/assets/43911210/207bc104-cf05-4c28-b55a-7a1a0294cae4

Here is what the full GitHub OAuth flow looks like when you include the frontend.

image

Some remaining items:
- Add a signup button to the login page that sends you to a simple create a user page
- ~Design a "service api key" flow where logged in users can generate API keys to use in clients/dashboards/apps/etc~...The /token endpoint that's been added to this PR provides short-lived tokens which might be enough
- The expiration date embedded in the JWT currently isn't being checked. A simple check needs to be added server side to the auth middleware.
- ~The UI needs to respond to expired or malformed token responses from the server and forward the user to the login page (probably also delete the cookie).~...The server now controls the logged in state of the UI through a cookie flag `__djlif`
- Extend the auth docs to include the UI part of it and explain the `REACT_DISABLE_AUTH` and `REACT_ENABLE_GITHUB_OAUTH` config parameters. The `REACT_DISABLE_AUTH` flag is meant to be there temporarily as an escape hatch while prototyping out the auth features. It has to be removed whenever we're ready for auth to be run in any serious environment since this is modifiable by the client (even though changing it will still hit 403s with a backend server that requires auth)
- Show the little user avatar in the top navbar that calls `/whoami/`. The logout button could be folded under there in some sort of dropdown as is typical.
- When auth fails in the UI, display something to the user such as a toast notification.

### Test Plan

- [x] PR has an associated issue: #64 #696 #697
- [x] `make check` passes


datajunction-ui/src/app/pages/LoginPage/assets/sign-in-with-google.png
agorajek, line 1: What is the code/file that handles the google auth path?

agorajek (approved):
This looks awesome. Thanks for the videos in the description. Perhaps make the remaining TODOs into an issue.

datajunction-server/datajunction_server/api/authentication/github.py
```diff
@@ -81,19 +80,19 @@ async def get_access_token(
...
+        raise DJException(
+            http_status_code=HTTPStatus.UNAUTHORIZED,
+            errors=DJError(
+                code=ErrorCode.OAUTH_ERROR,
+                message=(
+                    "Could not retrieve user using the GitHub provided access token"
+                ),
+            ),
+        )
+    response.set_cookie(key="__dj", value=encrypt(user.username))
```
betodealmeida, line 97: Nit, it would be nice if `__dj` was a contant.

datajunction-server/datajunction_server/api/authentication/whoami.py
```diff
@@ -0,0 +1,22 @@
...
+from fastapi.security import HTTPBearer
+
+from datajunction_server.models.user import UserOutput
+from datajunction_server.utils import get_settings
+
+settings = get_settings()
+router = APIRouter(
+    tags=["Who am I?"],
+    dependencies=[Depends(HTTPBearer())] if settings.secret else [],
+)
```
betodealmeida, line 14: Maybe we should create a `SecureAPIRouter` to remove the need for declaring dependencies everywhere, something like:

```python
class SecureAPIRouter(APIRouter):

    def __init__(self, *args: Any, **kwargs: Any):
        settings = get_settings()
        return super().__init__(
            *args,
            dependencies=[Depends(HTTPBearer())] if settings.secret else [],
            **kwargs,
        )
```

datajunction-server/datajunction_server/api/dimensions.py
```diff
@@ -18,7 +19,7 @@
 from datajunction_server.utils import get_session
 
 _logger = logging.getLogger(__name__)
-router = APIRouter(tags=["dimensions"])
+router = APIRouter(tags=["dimensions"], dependencies=[Depends(HTTPBearer())])
```
betodealmeida, line 22: Should this be `dependencies=[Depends(HTTPBearer())] if settings.secret else [],`?

datajunction-server/datajunction_server/internal/authentication/basic.py
```diff
@@ -71,18 +71,21 @@ async def parse_basic_auth_cookie(
...
     try:
-        jwt = get_jwt(request=request)
-    except (JWTError, AttributeError):
-        pass
-    encrypted_username = jwt.get("sub") if jwt else None
-    username = decrypt(encrypted_username) if encrypted_username else None
+        if any(request.url.path.startswith(endpoint) for endpoint in SSE_ENDPOINTS):
+            auth_token = request.query_params.get("token")
+        else:
+            auth_token = await get_token(request.headers.get("Authorization"))
```
betodealmeida, line 79: Maybe we should just always allow passing a token via query params?

datajunction-server/datajunction_server/internal/authentication/basic.py
```diff
@@ -71,18 +71,21 @@ async def parse_basic_auth_cookie(
...
+    except (JWTError, AttributeError) as exc:
+        _logger.error(str(exc))
+    username = decrypt(auth_token) if auth_token else None
+    _logger.info("User detected as %s", username)
     user = None
     try:
         user = session.exec(select(User).where(User.username == username)).one()
     except NoResultFound:
-        pass
+        _logger.error("Cannot find user %s", username)
```
betodealmeida, line 88: I think you need to do more cleaning here, otherwise `user` might be undefined in line 90 and will raise another exception. We probably need to set `user = None` here too.

Alternatively you can do `one_or_none()` in line 86, and remove the `try/except` block, I think.

datajunction-server/datajunction_server/internal/authentication/github.py
```diff
@@ -58,7 +51,7 @@ def get_github_user(encrypted_access_token: str) -> Optional[User]:  # pragma: n
         pass
     if existing_user:
         _logger.info("OAuth user found")
-        user = existing_user
+        user = existing_user  # type: ignore
```
betodealmeida, line 54: Why do we need `type: ignore` here? Maybe we just need to `cast` `existing_user` to a `User`?

datajunction-server/datajunction_server/internal/authentication/jwt.py
```diff
@@ -68,3 +79,13 @@ def decrypt(value: str) -> str:
...
     return jwe.decrypt(value, settings.secret).decode("utf-8")
+
+
+async def get_token(
+    authorization: str = Header(default="Bearer "),
+) -> str:
+    """
+    Get an authorization bearer token from an authorization header
+    """
+    _, token = authorization.split(" ")
```
betodealmeida, line 90: I couldn't find anything, but I'd think there'd be a helper function in FastAPI that returns the token so we don't have to do this parsing ourselves?

datajunction-ui/src/app/index.tsx
```diff
@@ -11,48 +11,61 @@ import { NamespacePage } from './pages/NamespacePage/Loadable';
...
 import { SQLBuilderPage } from './pages/SQLBuilderPage/Loadable';
 import { NotFoundPage } from './pages/NotFoundPage/Loadable';
+import { LoginPage } from './pages/LoginPage';
 import { Root } from './pages/Root/Loadable';
 import DJClientContext from './providers/djclient';
 import { DataJunctionAPI } from './services/DJService';
+import { CookiesProvider, useCookies } from 'react-cookie';
 
 export function App() {
+  const [cookies, setCookie] = useCookies(['__dj']);
```
betodealmeida, line 21: Same here, `__dj` should be a constant.

datajunction-ui/src/app/services/DJService.js
```diff
@@ -1,12 +1,24 @@
...
+};
 const DJ_URL = process.env.REACT_APP_DJ_URL
   ? process.env.REACT_APP_DJ_URL
   : 'http://localhost:8000';
 
 export const DataJunctionAPI = {
   node: async function (name) {
-    const data = await (await fetch(DJ_URL + '/nodes/' + name + '/')).json();
+    const data = await (
+      await fetch(DJ_URL + '/nodes/' + name + '/', {
```
betodealmeida, line 18: Nit, it's cleaner to do:
```suggestion
      await fetch(`${DJ_URL}/nodes/${name}/`, {
```

Same below in the other `fetch`es.

betodealmeida (approved):
This looks great! I left a few comments and nits.

datajunction-server/datajunction_server/api/health.py
```diff
@@ -6,12 +6,17 @@
 from typing import List
 
 from fastapi import APIRouter, Depends
+from fastapi.security import HTTPBearer
 from sqlalchemy import select
 from sqlmodel import Session, SQLModel
 
-from datajunction_server.utils import get_session
+from datajunction_server.utils import get_session, get_settings
```
shangyian, line 13: Does the health check need auth?

datajunction-server/datajunction_server/internal/authentication/jwt.py
shangyian: It should be possible to do something like this right?

```
from fastapi.security.http import HTTPAuthorizationCredentials, HTTPBearer

async def get_token(
    authorization: Optional[HTTPAuthorizationCredentials] = Depends(HTTPBearer(auto_error=False)),
) -> str:
    if auth is None:
        raise DJException(
            http_status_code=HTTPStatus.UNAUTHORIZED,
            errors=DJError(
                code=ErrorCode.OAUTH_ERROR,
                message=" ... ",
            ),
        )
    return auth.credentials
```

datajunction-ui/.env
```diff
@@ -1,2 +1,2 @@
 REACT_APP_DJ_URL=http://localhost:8000
-REACT_USE_SSE=true
```
shangyian, line 2: Why has this been removed?

shangyian (approved):
Nice!! 🥳

datajunction-server/datajunction_server/api/authentication/github.py
samredai: Good idea, done!

datajunction-ui/src/app/pages/LoginPage/assets/sign-in-with-google.png
samredai: I actually have that locally but didn't add it as part of this PR. After the dust settles a bit and we're good with this design it'll be really easy to add google. (This PR just has the image for the sign in with google button)

datajunction-server/datajunction_server/api/authentication/whoami.py
samredai: Love this. A friend gave me the idea to also wrap HTTPBearer so now we have a `DJHTTPBearer` that's set as part of this `SecureAPIRouter`...very clean!

datajunction-server/datajunction_server/api/dimensions.py
samredai: Fixed by your other suggestion to move to a central `SecureAPIRouter`

datajunction-server/datajunction_server/internal/authentication/basic.py
samredai: The `DJHTTPBearer` that I added is now more flexible and can accept either a jwt cookie or a bearer token. I found out how to make the client side event source aware of the http-only cookie so now this actually works identical across the board (no need to treat `SSE_ENDPOINTS` specially). I'll follow-up with a PR that allows for the token as a query param...any suggestions on the best way to do that welcome. :)

datajunction-server/datajunction_server/internal/authentication/basic.py
samredai: I've broken this out to a `get_user` function that uses `one_or_none()` and raises a DJException when a user can't be found. This feels cleaner since we can now easily differentiate between the user not existing and the password being incorrect. (My understanding though is there's a benefit to obscuring that to clients, something about a timing attack)

datajunction-server/datajunction_server/internal/authentication/jwt.py
samredai: Nice! I used `fastapi.security.utils.get_authorization_scheme_param` exactly how the `HTTPBearer` class does in the `DJHTTPBearer` wrapper, so it's hitting the same logic as this example. 👍

datajunction-ui/src/app/index.tsx
samredai: Done!

datajunction-server/datajunction_server/api/health.py
samredai: Good point, I removed auth there. 😅

datajunction-ui/.env
samredai: Whoops, I put it back!

datajunction-server/datajunction_server/internal/authentication/github.py
samredai: That fixed it!

samredai:
@agorajek @betodealmeida @shangyian thanks for all the feedback. Lots of changes and now it's much much cleaner. 🍾

Issue #718, opened 2023-08-15
SQL that tries to access struct types is buggy


shangyian:
* Type inference on struct types works
* The built SQL for metrics is buggy and removes the struct field name

PR #721, opened 2023-08-15
Fix bug with building queries with struct-type columns

shangyian:
### Summary

When we build SQL that accesses struct types, the built SQL (a) doesn't have the right aliased table name for the struct-type column and (b) doesn't include both the name of the column and the subscript. There were a couple of causes:
* We don't add a reference to the struct-type column onto the table
* We don't indicate that the `add_ref_column` call was successful
* We need a way to track the name for columns that reference a subscript on a `StructType` (see `_struct_subscript` variable on `ast.Column`)

### Test Plan

Added a test for the particular struct reference within a metric that was failing.

- [ ] PR has an associated issue: #718 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -701,6 +701,7 @@ class Column(Aliasable, Named, Expression):
         repr=False,
         default=None,
     )
+    _struct_subscript: Optional["Name"] = field(repr=False, default=None)
     _type: Optional["ColumnType"] = field(repr=False, default=None)
     _expression: Optional[Expression] = field(repr=False, default=None)
     _is_compiled: bool = False
@@ -747,6 +748,9 @@ class Column(Aliasable, Named, Expression):
         """
         self._expression = expression
         return self
+
+    def add_struct_subscript(self, struct_subscript: "Name"):
+        self._struct_subscript = struct_subscript
 
     def add_table(self, table: "TableExpression"):
         self._table = table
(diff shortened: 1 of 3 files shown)


agorajek (approved):
lgtm but it would be cool if you could have a look @CircArgs .

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -1043,7 +1052,7 @@ def add_ref_column(
                         column.add_type(col.type)
                         return True
 
-        for col in self.columns:
+        for col in self._columns:
```
CircArgs, line 1055: Why the switch to the obfuscated attribute? I think the property does some filtering in a case like:

```sql
SELECT column1, column2
FROM tableName(column1, column2)
```
where I think spark would only read `column1, column2` from the table and so no other columns would be legally referenced in the query despite maybe actually being in the table or node columns and seen during compiling the table with the node metadata.

CircArgs: so if `tableName` has columns `column1, ...column3` then `.columns` will only show `column1, column2` but `._columns` would show them all.

CircArgs:
@shangyian I vaguely recall some struct stuff earlier on after redoing the ast and parser. I guess we never handled it or handled it fully?

datajunction-server/datajunction_server/sql/parsing/ast.py
shangyian: Good point - I'm not sure why I changed it, but it was likely just while I was testing something. I switched it back.

shangyian:
> @shangyian I vaguely recall some struct stuff earlier on after redoing the ast and parser. I guess we never handled it or handled it fully?

@CircArgs Yeah, I don't think we had any test cases for this. The problem is that struct references still use the same namespace / names setup that we have, but we always assume that the name refers to the column name, whereas for struct references, it could be the struct subscript instead. I actually did a second round on this PR where I added a `is_struct_ref` boolean, and if it's set to true, we can use the `struct_column_name` and `struct_subscript` properties depending on which we need.

CircArgs (approved):

PR #722, opened 2023-08-15
Allow for dimensions and transforms to be created without a catalog

shangyian:
### Summary

We should allow users to create dimensions and transforms that don't have a catalog. This makes it easier to create a dimension on-the-fly, without having to create source table that underlies the dimension. This is often useful if the dimension is just a simple enum-like mapping that can be maintained as a query in DJ rather than as a fully materialized table. An example such dimension might be:
```
name: default.title_dimension
primary_key: title_code
description: Title to title code mapping
query: 
  SELECT 0 AS title_code, 'Agha' AS title 
  UNION ALL SELECT 1, 'Abbot' 
  UNION ALL SELECT 2, 'Akhoond' 
  UNION ALL SELECT 3, 'Apostle'
```

### Test Plan

Added a unit test with the above example. Also tested with some cases locally that needed this feature.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -505,7 +505,10 @@ def link_dimension(
         name=dimension,
         node_type=NodeType.DIMENSION,
     )
-    if node.current.catalog.name != dimension_node.current.catalog.name:
+    if (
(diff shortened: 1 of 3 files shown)


agorajek (approved):
This is a no brainer. Thanks @shangyian for the fix!

PR #723, opened 2023-08-15
graphql start

CircArgs:
### Summary

Starting a graphql api with strawberry pydantic support.

### Test Plan

unit tests

- [ ] `make check` passes


shangyian (approved):
Looks great, thanks for starting this up again!

datajunction-server/datajunction_server/api/graphql/engines.py
```diff
@@ -0,0 +1,35 @@
...
+    *,
+    info: Info = None,
+) -> List[EngineInfo]:
+    """
+    List all available engines
+    """
+    session = info.context["session"]  # type: ignore
+    return [
+        EngineInfo.from_pydantic(engine)  # type: ignore #pylint: disable=E1101
+        for engine in session.exec(select(Engine))
```
shangyian, line 34: I wonder if this select query should actually be very specific about which fields it requests on the model? I guess it doesn't matter as much for Engine since most setups won't have a ton of Engines, but when we get to implementing the Node endpoints, it'll be nice to have the GraphQL APIs only query for the specific field requested fields.

PR #726, opened 2023-08-16
Make "Compile SQL" button only call backend when requested

shangyian:
### Summary

The "Show Compiled SQL" toggle in the UI should only call the backend when a user actually requests it. Alternatively we should just make the compilation faster and/or cache the compiled node.

### Test Plan

- [x] `make check` passes


agorajek (approved):
LGTM

samredai (approved):

PR #727, opened 2023-08-18
Ast serde

CircArgs:
### Summary

Serializing and deserializing ASTs maintaining all information even after compilation into a flat form that is json serializable

### Test Plan

unit tests

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -211,11 +211,41 @@ class Node(ABC):
                 return curr_parent
             curr_parent = curr_parent.parent
 
-    def flatten(self) -> Iterator["Node"]:
+    def flatten(
+        self,
+        obfuscated: bool = False,
+        named: bool = False,
+    ) -> Iterator["Node"]:
         """
         Flatten the sub-ast of the node as an iterator
         """
-        return self.filter(lambda _: True)
+        seen = set()
+
+        def _flatten(args):
+            parent_key = None
+            if named:
+                parent_key, node = args
+            else:
+                node = args
+            if id(node) not in seen:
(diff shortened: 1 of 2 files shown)


shangyian (commented):
Thanks for adding this! Have you done any profiling to see how long the deserialization takes? I can get you some examples of huge queries if needed.

CircArgs:
@shangyian 
> Thanks for adding this! Have you done any profiling to see how long the deserialization takes? I can get you some examples of huge queries if needed.

Just ran some to get an idea:

parsing all the spark tpcds spark queries:
`3.53 s ± 23.9 ms per loop`

serializing all tpcds spark queries:
`83.1 ms ± 1.76 ms per loop`

deserializing them:
`31.8 µs ± 950 ns per loop`

shangyian (approved):
@CircArgs that's great!

shangyian:
@CircArgs Hmm, actually I tried this with one of our use cases -- specifically some metrics that depend on several layers of transforms, each of which have fairly nested subqueries. For those metrics, I serialized the compiled query AST and then deserialized it, but it takes a long time to deserialize (more than five minutes per metric). Maybe there's something else going on, but parsing and recompiling is faster.

CircArgs:
@shangyian I can hold off on this for now then. I wonder if there's a big distinction with just serializing/de non-compiled queries vs compiled queries. My timings were all non-compiled

shangyian:
@CircArgs Maybe it's because some compiled queries, if they're pulling together many layers of transforms, can be huge. But I would still expect this to be faster than actually having to compile the queries. 🤔 

On that thought, we also need to make the queries built more efficient / readable by removing all the columns that aren't used.

agorajek (approved):
Thanks for this work @CircArgs . Can't wait to see it plugged in.

Issue #729, opened 2023-08-18
Add a few extra attributes in the `GET /namespace/{namespace}/` endpoint to facilitate efficient table display


samredai:
The logical progression of navigating through the DJ dag is starting to become clear.
1. List all the namespaces (returns `List[str]`)
2. Select a namespace and list all the nodes in a namespace (returns `List[str]`)
3. Select a node and view all the node details (returns `NodeOutput`)

For 2., we typically want to show a bit more detail about the nodes in the namespace. Not every single detail about the node but a few high level ones, enough to display in a table view. Since it just returns a list, in the UI we end up calling the `/node/{name}/` endpoint to get the full details on each node and pluck out the handful of attributes we want to show in the table. I'd like to propose that we instead update that endpoint to return `List[NodeMinimumDetail]` that includes the handful of high level attributes for each node, That would allow us to make one call whenever we're displaying a high level view of nodes in a namespace instead of one call per node.

Keeping in mind that once the graphql API is ready, that would effectively provide this solution everywhere.

PR #730, opened 2023-08-18
Some minimal details in the response for `POST /namespace/{namespace}/`

samredai:
### Summary

This adds a minimal amount of detail to the response of the `GET /namespace/{namespace}/` endpoint (name, display_name, type, status, mode, updated_at). See issue #729 for some of the reasoning around this.

Now when going to the page for a namespace in the UI, it loads significantly faster since it only needs to make a single request that runs a single query to get just enough information to display the table (as opposed to a single request per node).

https://github.com/DataJunction/dj/assets/43911210/0dce33ec-9416-4e52-8403-5e78b007d89c

### Test Plan

- [x] PR has an associated issue: #729 
- [x] `make check` passes


shangyian:
Awesome! We can eventually switch to GraphQL, but this will keep things performant for now

datajunction-server/datajunction_server/api/namespaces.py
```diff
@@ -61,23 +66,25 @@ def create_node_namespace(
...
 @router.get(
     "/namespaces/",
-    response_model=List[str],
+    response_model=List[NamespaceOutput],
     status_code=200,
 )
 def list_namespaces(
     session: Session = Depends(get_session),
-) -> List[str]:
+) -> List[NamespaceOutput]:
```
shangyian, line 74: Will this cause the Python client to fail? I think at the moment it assumes that this endpoint returns a list of strings

samredai: Yep, fixed! I just parse out the names right after the call.

samredai: In the future we might want to use the full payload to generate an array of objects or maybe even just a single NodeList object that prints as a nice looking table. 🤔

PR #731, opened 2023-08-18
various client enhancements: linked nodes, more tests, bugfix

shangyian:
### Summary

Various client enhancements:
* Added `dimension.linked_nodes()`, which allows users to discover all nodes linked to a particular dimension. This is equivalent to what's surfaced in the UI under the "Linked Nodes" tab. 
* Fixed a bug where we don't refresh the Node client object after creating a node. This results in the node object not having any columns, for example, which is confusing if someone has just created a transform.
* Bumped the client version to match what's been released in PyPI.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -101,21 +101,6 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
     #
     # Nodes: SOURCE
     #
-    def source(self, node_name: str) -> "Source":
-        """
-        Retrieves a source node with that name if one exists.
-        """
-        node_dict = self._verify_node_exists(
-            node_name,
-            type_=models.NodeType.SOURCE.value,
-        )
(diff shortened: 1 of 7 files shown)


datajunction-clients/python/datajunction/builder.py
```diff
@@ -101,21 +101,6 @@ def restore_node(self, node_name: str) -> None:
     #
     # Nodes: SOURCE
     #
-    def source(self, node_name: str) -> "Source":
```
shangyian, line 104: @agorajek fyi, I moved these to the regular non-builder client because I figured they're meant to be read-only client objects. This way users can do something like:
```
repair_order_dim = dj.dimension("default.repair_order")
repair_order_dim.linked_nodes()
> [ ... ]
```

agorajek: Sounds good!

agorajek (approved):
lg

PR #735, opened 2023-08-19
Store column lineage for reuse

shangyian:
### Summary

Store column-level lineage on the `noderevision` table when the node is created and when the node's query is updated. This makes it so that we only calculate lineage once, and further calls for lineage will just fetch the stored lineage object. For nodes with very large lineage trees, this will significantly reduce the amount of time spent calculating lineage.

### Test Plan

- [x] `make check` passes


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -380,6 +376,12 @@ def save_node(
     node.current_version = node_revision.version
     node_revision.extra_validation()
 
+    if node_revision.type == NodeType.METRIC:
```
agorajek, line 379: Your PR summary does not mention this being specific for METRIC nodes. I understand this may be expensive to maintain for all node columns, but now when I think of measures coming (possibly to the picture) the doing this for Metrics only may stop making sense.

shangyian: @agorajek Good point, I can add lineage for all other nodes as well. It's not actually that expensive assuming that we only do it once at node creation/update time. If it is too time consuming, I was also thinking about turning it into something that's a background task that gets kicked off async when the node is created or updated so that it's non-blocking.

shangyian: @agorajek Good point, I can add lineage for all other nodes as well. It's not actually that expensive assuming that we only do it once at node creation/update time. If it is too time consuming, I was also thinking about turning it into something that's a background task that gets kicked off async when the node is created or updated so that it's non-blocking.

agorajek: @shangyian I think having a lineage on every column of every node is the ultimate goal of this. So if you think this is doable I'd add it from the start and worry about the performance later.

shangyian: @agorajek Done! I excluded source and cube nodes though, as those aren't quite the same pattern as the other node types that have queries.

agorajek (approved):
Sorry I missed your commit. Nice! Thanks for the lineage store.

Issue #725, opened 2023-08-16
Hard delete doesn't propagate to downstreams beyond first layer


shangyian:
Hard deleting a node invalidates only the first layer of downstream nodes, but it doesn't propagate down to all subsequent layers of downstreams. For example, when a source node is hard deleted, it will invalidate a transform that depends on it, but the metrics that depend on the transform remain valid. They should all be invalid.

samredai:
Thanks @shangyian, good catch! It looks like it's just relying on [get_downstream_nodes](https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/api/helpers.py#L245). I wonder if that function isn't including the metrics in your example.

PR #739, opened 2023-08-22
When hard deleting a node, completely remove its revisions and relations

shangyian:
### Summary

When we hard delete a node, we should also completely remove its associated node revisions, and for each node revision, all of the associated node relations entries. The current logic doesn't do this, leaving these orphaned node revisions to cause issues later, especially if a node by the same name is later recreated.

### Test Plan

- [ ] PR has an associated issue: #725 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/models/node.py
+++ b/datajunction-server/datajunction_server/models/node.py
@@ -636,7 +636,12 @@ class NodeRevision(NodeRevisionBase, table=True):  # type: ignore
         },
     )
 
-    parent_links: List[NodeRelationship] = Relationship()
+    parent_links: List[NodeRelationship] = Relationship(
+        sa_relationship_kwargs={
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/models/node.py
```diff
@@ -636,7 +636,12 @@ class NodeRevision(NodeRevisionBase, table=True):  # type: ignore
         },
     )
 
-    parent_links: List[NodeRelationship] = Relationship()
+    parent_links: List[NodeRelationship] = Relationship(
+        sa_relationship_kwargs={
+            "primaryjoin": "NodeRevision.id==NodeRelationship.child_id",
+            "cascade": "all, delete",
```
shangyian, line 642: This lets us cascade the `NodeRevision`'s delete to the `NodeRelationship` table

agorajek (approved):
Super nice catch. Thank you!

PR #743, opened 2023-08-24
Fix Python integration test.

agorajek:
### Summary

Because it was dependent on our demo database. Now it is fully isolated and its namespace gets deleted on a successful run.

### Test Plan

`pdm run pytest --integration -- tests/test_integration.py`

- [x] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/datajunction-clients/python/tests/test_integration.py
+++ b/datajunction-clients/python/tests/test_integration.py
@@ -29,7 +29,7 @@ def test_integration():  # pylint: disable=too-many-statements,too-many-locals,l
             matching_namespace = existing_namespace
     assert matching_namespace
 
-    # Create a source
+    # Create a sources
     dj.create_source(
         name=f"{namespace}.repair_orders",
         description="Repair orders",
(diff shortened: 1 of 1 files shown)


datajunction-clients/python/tests/test_integration.py
```diff
@@ -737,3 +505,6 @@ def test_integration():  # pylint: disable=too-many-statements,too-many-locals,l
 
     # Check that publishing transform 3 works
     transform_3.publish()
+
+    # Delete the test namespace
+    dj.delete_namespace(namespace=namespace, cascade=True)
```
shangyian, line 510: Not related to this PR, but I wonder if we need a hard-delete for namespaces as well? It's not ideal that running the integration test locally (or even worse, someone could run this against an actual deployed environment too) would create a bunch of deactivated namespaces at the end. Ideally these would be fully deleted

shangyian (approved):
Awesome, thanks for fixing!

datajunction-clients/python/tests/test_integration.py
agorajek: Oh yes, that would be useful to have and this integration test would be a good use case. 
Made an issue: https://github.com/DataJunction/dj/issues/744

Issue #746, opened 2023-08-25
Add dimension listing for nodes


shangyian:
We currently only enable this for metrics, but it should be available for all nodes

PR #747, opened 2023-08-25
Add listing of dimension attributes for any node

shangyian:
### Summary

Adds an endpoint to list dimension attributes for any node. This opens up exploring the dimensions graph from any node, not just from metric nodes.

### Test Plan

- [x] PR has an associated issue: #746 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -54,6 +54,7 @@ from datajunction_server.models.node import (
     CreateCubeNode,
     CreateNode,
     CreateSourceNode,
+    DimensionAttributeOutput,
     LineageColumn,
     Node,
     NodeMode,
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -833,6 +834,21 @@ def list_node_dag(
     return list(set(cast(List[Node], dimension_nodes) + downstreams + upstreams))  # type: ignore
 
 
+@router.get(
+    "/nodes/{name}/dimensions/",
```
agorajek, line 838: Q unrelated to this PR... you mentioned that somewhere we have a problem with the slashes at the end of the request. Isn't that b/c we add them here?

agorajek (approved):
+1

datajunction-server/datajunction_server/api/nodes.py
shangyian: It is unfortunately far more complex 😢 https://github.com/tiangolo/fastapi/discussions/7298

I'll look into adding one of the options there to DJ

Issue #744, opened 2023-08-24
Add a hard_delete_namespace() method.


agorajek:
Similar to `hard_delete_node()` that would be useful for cleanup purposes.

PR #750, opened 2023-08-29
Add hard deletion for namespaces + bugfixes

shangyian:
### Summary

This adds "hard" deletion for namespaces, including an option to cascade the hard delete to nodes within the namespace. This opens up an option to completely remove namespaces (and the nodes within them) from the DJ database, which is useful in certain contexts, where namespace deactivation will just litter the database with unnecessary state. We should eventually put this behind an admin gate. 

This PR also fixes two bugs around namespace functions:
* When we try to list nodes under a namespace, we should first check to make sure the namespace exists and raise an error if it doesn't.
* When listing nodes under a namespace, we shouldn't match on any namespace that contains the search string, as trying to search for namespace `foo.bar` will also (erroneously) match on `default.foo.bar`. Instead, we need to match on the search namespace being the prefix.

### Test Plan

Locally + added unit test

- [x] PR has an associated issue: #744
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/internal/namespaces.py
+++ b/datajunction-server/datajunction_server/internal/namespaces.py
@@ -4,9 +4,12 @@ Helper methods for namespaces endpoints.
 from datetime import datetime
 from typing import Dict, List
 
+from sqlalchemy import or_
 from sqlalchemy.sql.operators import is_
 from sqlmodel import Session, col, select
 
+from datajunction_server.api.helpers import get_node_namespace, hard_delete_node
+from datajunction_server.errors import DJActionNotAllowedException
 from datajunction_server.models import History
 from datajunction_server.models.history import ActivityType, EntityType
 from datajunction_server.models.node import Node, NodeNamespace, NodeRevision, NodeType
@@ -22,6 +25,7 @@ def get_nodes_in_namespace(
     """
     Gets a list of node names in the namespace
     """
+    get_node_namespace(session, namespace)
     list_nodes_query = select(
         Node.name,
         Node.display_name,
@@ -33,7 +37,10 @@ def get_nodes_in_namespace(
         NodeRevision.mode,
         NodeRevision.updated_at,
     ).where(
-        col(Node.namespace).contains(namespace),  # pylint: disable=no-member
+        or_(
+            col(Node.namespace).like(f"{namespace}.%"),  # pylint: disable=no-member
+            Node.namespace == namespace,
+        ),
         Node.current_version == NodeRevision.version,
         Node.name == NodeRevision.name,
         Node.type == node_type if node_type else True,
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -33,7 +37,10 @@ def get_nodes_in_namespace(
         NodeRevision.mode,
         NodeRevision.updated_at,
     ).where(
-        col(Node.namespace).contains(namespace),  # pylint: disable=no-member
+        or_(
+            col(Node.namespace).like(f"{namespace}.%"),  # pylint: disable=no-member
```
agorajek, line 41: Are you sure we need the dot between `{namespace}` and `%` ? This could cause some confusion imo.

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -108,3 +115,35 @@ def create_namespace(session: Session, namespace: str):
...
+
+
+def hard_delete_namespace(session: Session, namespace: str, cascade: bool = False):
+    """
+    Hard delete a node namespace.
+    """
+    node_names = session.exec(
+        select(Node.name).where(
+            or_(
+                col(Node.namespace).like(f"{namespace}.%"),  # pylint: disable=no-member
```
agorajek, line 127: ditto

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -108,3 +115,35 @@ def create_namespace(session: Session, namespace: str):
         ),
     )
     session.commit()
+
+
+def hard_delete_namespace(session: Session, namespace: str, cascade: bool = False):
+    """
+    Hard delete a node namespace.
+    """
+    node_names = session.exec(
```
agorajek, line 124: Where do we handle deleting other namespaces under the one being deleted?

agorajek (commented):
Thanks for starting this. Two questions in line.

datajunction-server/datajunction_server/internal/namespaces.py
shangyian: @agorajek I think the problem here is that if we don't have this `.`, we'll end up matching more than we should. For example, let's say we want to search for `namespace=foo`, without `.` it'll match to `foobar.group1` and `foo.bar.group1`, but `foobar.group1` shouldn't be included in there because it doesn't actually match the full namespace chunk.

datajunction-server/datajunction_server/internal/namespaces.py
shangyian: Good point! I'll add a check for that

datajunction-server/datajunction_server/internal/namespaces.py
agorajek: I see your point. I think I used the namespace prefix somewhere in the code that was using the logic w/o an implicit dot, but that might have been for listing only. On the 2nd thought I think your approach is safer here, because this is deleting.

PR #753, opened 2023-08-29
Streamline unit tests

shangyian:
### Summary

This PR kicks off some changes to the unit test examples setup so that we can start cleaning up our unit tests to only load the examples that are relevant for that test. In the existing setup, we load all examples, which has become quite large over time. This repeated loading of all examples has caused unit tests to take >1h to run (especially after #735).

* Splits up the `EXAMPLES` list into different groups: `ROADS`, `NAMESPACED_ROADS`, `EVENT`, `BASIC`, `DBT` etc. This list is now a dictionary that maps the group name to a list of requests. Most tests will just want to load one of these groups, but some will load more than one.
* Adds a `client_example_loader` fixture that provides a callable for loading examples into a `TestClient` session
* The existing `client_with_examples` fixture stays the same so that we can remain backwards compatible. As we clean up the unit tests, I would expect only a few tests to use this fixture, as most only operate in one of the groups above.
* Added several convenience fixtures for the different groups like `client_with_roads` and `client_with_service_setup`. These might be able to be generated automatically based on the keys to `EXAMPLES`, but I haven't investigated.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-server/tests/api/client_test.py
+++ b/datajunction-server/tests/api/client_test.py
@@ -119,12 +119,12 @@ repairs_cube = dj.create_cube(
 
 
 def test_generated_python_client_code_adding_materialization(
-    _client_with_query_service: TestClient,
+    client_with_query_service_example_loader,
 ):
     """
     Test that generating python client code for adding materialization works
     """
-    custom_client = _client_with_query_service(["BASIC"])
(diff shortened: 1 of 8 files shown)


agorajek:
This is great. Let's see how much we can speed up these server tests.

PR #754, opened 2023-08-30
Show revision history when auth enabled

youngman-droid:
### Summary

Trivial fix to allow revision history to show with auth turned on

### Test Plan

No tests changed

- [x] `make check` passes


shangyian (commented):
Thanks for the fix! Can you also run `yarn prettier --write .` and `yarn lint --fix` in the `datajunction-ui` directory? (I'm not sure why it's not enabled in the GH action, but we can look into fixing that)

PR #756, opened 2023-08-30
Make Python test execution dependent on code change.

agorajek:
### Summary

What the title says. Using the https://github.com/dorny/paths-filter features. I combined some steps together to save on redundant code. If this works I'd like to clean up the module names for query and reflection dirs.

### Test Plan

In the wash... but I'll try to add bogus changes to make sure that the full test matrix is healthy:
- 1st commit run in seconds (as expected)
- 2nd commit runs in correct dir of the client. 

### Deployment Plan

n/a

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,19 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            client:
+              - 'datajunction-clients/python/**'
+            server:
+              - 'datajunction-server/**'
+            djqs:
+              - 'datajunction-query/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  build-main:
```
shangyian, line 14: I wonder if this change caused it to wait for these to run?

Screenshot 2023-08-30 at 5 01 46 PM

shangyian (approved):
Thanks for adding this @agorajek!

.github/workflows/test.yml
agorajek: thank you, I will put that back

PR #757, opened 2023-09-01
update tests with smaller fixtures to increase efficiency

shangyian:
### Summary

Continues the work started in #753 

Updates a few more unit tests to use the less expensive fixtures.

Tests are now back down to ~30min!

### Test Plan

- [ ] PR has an associated issue: #753 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/tests/api/data_test.py
+++ b/datajunction-server/tests/api/data_test.py
@@ -20,12 +20,12 @@ class TestDataForNode:
 
     def test_get_dimension_data_failed(
         self,
-        client_with_examples: TestClient,
+        client_with_account_revenue: TestClient,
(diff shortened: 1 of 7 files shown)


agorajek (approved):
Thanks!

shangyian:
@agorajek it would be even better if we could reuse the TestClient fixture across functions. At the moment we're just saving on loading a less expensive TestClient fixture, but in the `/sql` endpoint tests, we should be able to just reuse the same TestClient since they're all read-only calls that never modify the database. Unfortunately it's not as straightforward as just saying `@pytest.fixture(scope="class")` because there are various upstream fixtures that are defined at the function level.

Issue #758, opened 2023-09-01
Add an option to configure a SQL transpilation library


shangyian:
This would allow for transpiling between SQL dialects.

PR #759, opened 2023-09-01
Add SQL transpilation option between dialects

shangyian:
### Summary

This adds an optional plugin to configure a SQL transpilation library for transpiling between SQL dialects. 

We'll use sqlglot by default, but additional transpilation libraries can be configured by implementing the `SQLTranspilationPlugin` base class with a custom `transpile_sql` method that is tailored to the library. Note that the base plugin implementation will handle checking that the package exists before loading and using it, or skipping transpilation if the package cannot be loaded. Thus, we'll always default to providing a generated query where we can, but will transpile if the transpilation library is available.

### Test Plan

Added tests

- [x] PR has an associated issue: #758 
- [x] `make check` passes


datajunction-server/datajunction_server/transpilation.py
```diff
@@ -0,0 +1,92 @@
...
+                raise DJException(
+                    message=f"Not installed: {self.package_name}",
+                ) from import_err
+
+    @abstractmethod
+    def transpile_sql(
+        self,
+        query: str,
+        *,
+        input_dialect: Optional[Dialect] = Dialect.SPARK,
```
agorajek, line 43: So this default value shuld be `None`, correct?

datajunction-server/datajunction_server/transpilation.py
```diff
@@ -0,0 +1,92 @@
...
+    Implement sqlglot as a transpilation option
+    """
+
+    package_name: str = "sqlglot"
+
+    def transpile_sql(
+        self,
+        query: str,
+        *,
+        input_dialect: Optional[Dialect] = Dialect.SPARK,
```
agorajek, line 60: And here?

agorajek (commented):
Thanks for the PR @shangyian . Let's hold with pushing this to next week, but I think this is a good approach.

cc @betodealmeida @samredai and @CircArgs , you may want to check it out as well.

datajunction-server/datajunction_server/transpilation.py
shangyian: Good point, changed that!

samredai:
This looks great so far and the approach makes sense to me. How are you planning to integrate this into the /sql/ and /data/ endpoint? Since we don't intend to store the transpiled SQL and it's just something that happens at runtime right before the response goes out, I wonder if making it an injected dependency would be the best approach. That might allow for an alternative mechanism to add a completely custom transpiler without using the `SQLTranspilationPlugin` base class.

That's technically possible today but just needs to be documented and if DJ's internal transpiler logic uses the same approach (dependency injection), it'll serve as a good example for someone wanting to add an entirely custom one.

shangyian:
> How are you planning to integrate this into the /sql/ and /data/ endpoint?

It's already integrated via the TranslatedSql model. The idea is that the DJ server admin could choose a SQL transpilation library (or none), and this lib would get loaded at server start and get used to transpile sql as a part of the model. If the server admin thinks that DJ's own transpiler is sufficient, they won't configure any external libraries and DJ will return queries exactly as before.

agorajek (approved):
Looks good!

Issue #761, opened 2023-09-05
Modifying primary key of a node should remove existing primary key columns


shangyian:
When a node has existing columns marked as part of its primary key, updating its PK to new columns should remove the primary key attribute from the existing columns. Otherwise it will look as if we're never updating the primary key.

An example: `default.repair_orders` has PK `repair_order_id`. Let's say we want to change this to `hard_hat_id`. We would call:

```
curl -X 'PATCH' \
  'http://localhost:8000/nodes/default.repair_orders' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "primary_key": [
    "hard_hat_id"
  ]
}'
```

However, now the new PK is `repair_order_id`, `hard_hat_id` because we don't remove the old PK attribute on `repair_order_id`.

PR #764, opened 2023-09-06
Fix two node update bugs

shangyian:
### Summary

(1) When an update touches a node's primary key, removing a primary key column should get reflected in the update. Otherwise it will only be possible to add primary key columns and not remove them.
(2) Minor node changes don't result in parent links being saved. Since we're always creating a new node revision, we always have to regenerate and save the parents.

### Test Plan

Added tests

- [x] PR has an associated issue: #761 #762 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -620,8 +620,8 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
         status=old_revision.status,
     )
 
-    # Link the new revision to its parents if the query has changed and update its status
-    if new_revision.type != NodeType.SOURCE and (query_changes or pk_changes):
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -620,8 +620,8 @@ def create_new_revision_from_existing(  # pylint: disable=too-many-locals,too-ma
         status=old_revision.status,
     )
 
-    # Link the new revision to its parents if the query has changed and update its status
-    if new_revision.type != NodeType.SOURCE and (query_changes or pk_changes):
```
agorajek, line 624: Wondering why we had this condition as so specific...

agorajek (approved):
Nice catch @shangyian !

Issue #749, opened 2023-08-29
Node creation UI(s)


shangyian:
Nodes
- [x] UI to register tables as source nodes
- [x] Creation UI for transform/dimension/metric nodes (any node with a query)
- [x] Edit UI for transform/dimension/metric nodes
- [x] Node deactivation via UI
- [x] Creation UI for cubes (selectors for metrics and dimension attributes)

Namespaces
- [x] Create namespace via UI: done, but we should change the interface to be more intuitive
- [ ] Delete namespace via UI

Columns
- [x] A column can be linked to a dimension via the node columns page UI
- [x] The link between a column and a dimension node can be removed via UI
- [x] Column attributes can be set via the node columns page UI (added and removed)

Materialization
- [x] UI to create materialization

PR #770, opened 2023-09-13
Setting Column Attributes + Dimension Links

shangyian:
### Summary

This PR has three changes related to setting column attributes and adding dimension links:
1. The backend API has been modified to set attributes on a single column per API call. The primary advantage of this is that we can now support adding, modifying, or removing attributes, whereas before we only supported adding attributes to columns. This is done by assuming that each call to `POST /nodes/{node}/columns/{column}/attributes` will just set all attributes to the list of attributes provided in that call. Setting the list of attributes to `[]` will remove all column attributes.
2. Added a UI element to edit column attributes on the node page's "Columns" tab.
3. Modified the Python client to work with the single column attribute-setting approach.
4. Added the ability to link dimensions to a column in the UI.

https://github.com/DataJunction/dj/assets/9524628/fa2c491d-982d-4c29-8c84-a48c33b0772f

### Test Plan

- [x] PR has an associated issue: #749 
- [x] `make check` passes


datajunction-clients/python/datajunction/models.py
```diff
@@ -59,9 +59,8 @@ class ColumnAttribute(BaseModel):
     Represents a column attribute
     """
 
-    attribute_type_namespace: Optional[str] = "system"
-    attribute_type_name: str
-    column_name: str
+    namespace: Optional[str] = "system"
```
agorajek, line 62: I think this simplification makes sense.

Not relate to this code change, but I am a bit confused how do we set a value for a linked dimension or some other user input attribute?

datajunction-ui/src/app/pages/NodePage/EditColumnPopover.jsx
```diff
@@ -0,0 +1,86 @@
...
+    const response = await djClient.setAttributes(node, column, attributes);
+    if (response.status === 200 || response.status === 201) {
+      setStatus({ success: 'Saved!' });
+    } else {
+      setStatus({
+        failure: `${response.json.message}`,
+      });
+    }
+    onSubmit();
+    // window.location.reload();
```
agorajek, line 27: Needed or not?

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -420,8 +419,7 @@ def validate_node_data(  # pylint: disable=too-many-locals
     errors = []
     if missing_parents_map or type_inference_failures or invalid_required_dimensions:
         # update status (if needed)
-        if validated_node.mode == NodeMode.DRAFT:
-            validated_node.status = NodeStatus.INVALID
+        validated_node.status = NodeStatus.INVALID
```
agorajek, line 422: The comment above does not need `(if needed)` now.

agorajek (approved):
Looks good. Few minor comments in-line. Plus unit tests are failing.

datajunction-clients/python/datajunction/models.py
shangyian: Yeah, this brings it to consistency with that setup, because linked dimensions are set via [`POST /nodes/{name}/columns/{column}`](http://localhost:8000/docs#/nodes/link_dimension_nodes__name__columns__column__post) with the dimension passed.

PR #772, opened 2023-09-14
Add measures to DJ data model.

agorajek:
### Summary

Add Measure to data model so we can start moving Metric nodes onto these.

I added the new table plus the new foreign key (on column) to the initial migration because SQLlite does not have an ability to perform that action after table is created.

### Test Plan

make check + test

- [ ] `make check` passes


datajunction-server/datajunction_server/models/measure.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Type of allowed aggregation for a given measure.
+    """
+
+    ADDITIVE = "additive"
+    NON_ADDITIVE = "non-additive"
+    SEMI_ADDITIVE = "semi-additive"
+
+
+class MeasureColumn(BaseSQLModel, table=True):  # type: ignore
```
shangyian, line 27: Sorry I know I added this, but in thinking through it more, it might make more sense just to remove the link table `MeasureColumn` and keep `columns: List["Column"]` on `Measure` and `measure: Measure` on `Column`. Since it will be a one(-measure)-to-many(-columns) relationship, we can just store the measure_id foreign key on `Column`, and have a `measure` field on `Column` that's backpopulated.

agorajek: Hmm...  I see. Yes that should work and it should be just as fast for both searching both ways.

agorajek: @shangyian I made some changes but I am not sure this is exactly what we need. Let's go over this tomorrow.

shangyian: @agorajek nice, these changes look good to me! I just pulled it down and played around with it by adding a test -- seems to work as expected. But we can discuss your concerns tomorrow nonetheless

shangyian (approved):

datajunction-server/datajunction_server/models/measure.py
agorajek: Thanks for the unit test @shangyian !

datajunction-server/alembic/versions/2023_07_07_2006-4e1ff36c27c6_initial_migration.py
```diff
@@ -176,6 +176,23 @@ def upgrade():
             name=op.f("pk_catalogengines"),
         ),
     )
+    op.create_table(
```
shangyian, line 179: @agorajek This actually breaks the existing db migration setup, unless we squeeze all of the migrations into this initial commit

Issue #769, opened 2023-09-12
Node Update Bugs: Child dependencies aren't updated to invalid based on column references


shangyian:
When a node's query is updated such that it has different output columns, its downstream children aren't checked to see if they are still referencing existing columns on their parents. Furthermore, the actual affected children may not just be limited to the node's immediate children, but may fan out across the entire graph of dependencies. 

```mermaid
graph TD;
    X-->A;
    Y-->A;
    A-->B;
    A-->C;
    B-->D;
    B-->E;
    C-->F;
    style A stroke:#333,stroke-width:4px
```
When node A's query is updated:
* Node status changes should only affect either A or A's children, but should not affect A's parents.
* The links between A and its parents may be affected (i.e., the shape of the graph may change).

The status of any given node can be affected by:
  * The existence of parent references
  * The statuses of parent references
  * The columns of parent references

| Initial  |  Query Changes      |   Effects |
| ------------- | ----------- | ---------- |
| Invalid            | Node A's query was updated to reference parents that exist | <li>A's status will change to valid.</li><li>All of A's children should be revalidated</li> | 
| Invalid | Node A's query was updated to reference the same parents or parents that don't exist | <li>A's status will stay the same (invalid)</li><li>None of A's children need to be revalidated, since A's status will stay the same</li> | 
| Valid | Node A's query was updated to reference one or more parents that don't exist | <li>A's status will change to invalid</li><li>All of A's children need to be revalidated and will likely all be invalid</li>
| Valid | Node A's query was updated to reference the same parents or parents that do exist | <li>A's status will stay the same (valid)</li><li>All of A's children need to be revalidated, since A may have new columns.</li> |

PR #773, opened 2023-09-17
Updating node should update child dependencies based on column changes

shangyian:
### Summary

There are two main parts to this PR:
* Refactor node validation function `validate_node_data` to have no side effects by returning a `NodeValidator` object instead of directly changing the original node revision.
* After a node gets updated, we should revalidate all downstream nodes and take into account column changes (name changes and type changes) as well as status changes

### Test Plan

- [x] PR has an associated issue: #769 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -9,6 +9,7 @@ import json
 import logging
 import time
 import uuid
+from dataclasses import dataclass, field
 from datetime import datetime
 from http import HTTPStatus
 from typing import Dict, List, Optional, Set, Tuple, Union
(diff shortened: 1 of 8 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -345,29 +351,69 @@ def find_bound_dimensions(
...
     data: Union[NodeRevisionBase, NodeRevision],
     session: Session,
-) -> Tuple[
-    NodeRevision,
-    Dict[NodeRevision, List[ast.Table]],
-    Dict[str, List[ast.Table]],
-    List[str],
-    List[DJError],
-]:
+) -> NodeValidator:
```
agorajek, line 403: Awesome!

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -424,25 +411,34 @@ def save_node(
     session.refresh(node.current)
 
 
+def add_lineage_to_node(session: Session, node_revision: NodeRevision):
```
agorajek, line 414: I think a better way is to "never" mutate the arguments of the function. I would write it this way:
```
def get_node_lineage(session: Session, node_revision: NodeRevision):
    ...
    return lineage or []
```
And then use it:
```
some_node_revision.lineage = get_node_lineage(session, some_node_revision)
```
I know we violated this all over the place here, but since you are doing this cleanup already... maybe we can start making small steps?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -480,24 +466,151 @@ def _update_node(
...
+                }
+                event = History(
+                    entity_type=EntityType.NODE,
+                    entity_name=child.name,
+                    node=child.name,
+                    activity_type=ActivityType.STATUS_CHANGE,
+                    details={
+                        "upstreams": [
+                            history_events[entry.name]
+                            for entry in changelog
```
agorajek, line 576: I have a feeling this will cause the same nodes to be added to event history multiple times.

agorajek (commented):
@shangyian thank you for this PR. I have some questions in line but overall looks good.

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -376,50 +422,63 @@ def validate_node_data(  # pylint: disable=too-many-locals
...
+        return node_validator
 
     # Add aliases for any unnamed columns and confirm that all column types can be inferred
     query_ast.select.add_aliases_to_unnamed_columns()
 
+    # Invalid parents will invalidate this node
+    invalid_parents = {
+        parent.name
+        for parent in node_validator.dependencies_map
+        if parent.type != NodeType.SOURCE and parent.status == NodeStatus.INVALID
```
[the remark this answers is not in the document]
shangyian: So I noticed that sometimes our source nodes are `invalid`. This is obviously a bug, but I'm not sure where in the code this is happening and so decided to add this check for now, until we fully fix it.

datajunction-server/datajunction_server/internal/nodes.py
shangyian: Good point! Also I think this change could just be made on the upstream function `get_column_level_lineage` rather than in a separate function.

datajunction-server/datajunction_server/internal/nodes.py
shangyian: That is the intention -- in the sense that we want all of the upstream actions up until this node to be recorded in the event history of that node, so that we know what happened up to the point this node got changed.

datajunction-server/datajunction_server/api/helpers.py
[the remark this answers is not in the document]
agorajek: Oh, sg, but please add a comment about it in the code.

datajunction-server/datajunction_server/internal/nodes.py
agorajek: I see! That makes sense...

datajunction-server/datajunction_server/internal/nodes.py
shangyian: @agorajek Done! I removed `add_lineage_to_node` and just pushed the conditional checks into the upstream function.

agorajek (approved):
ship_it

PR #776, opened 2023-09-19
UI for registering tables, deleting nodes, adding namespaces

shangyian:
#### Added page to register tables:

Screenshot 2023-09-20 at 12 03 42 AM

Screenshot 2023-09-20 at 12 10 42 AM

#### Node deletion

The delete node icon was added previously to the list nodes page, but clicking it wasn't wired up to actually deactivate nodes. This change wires it up with the right API call. Note that the current setup is prone to fat-fingering node deletions, but this can be fixed later.

#### Add namespace popover

Provides a small popover in the hierarchical namespaces sidebar to add a new namespace

https://github.com/DataJunction/dj/assets/9524628/228e7171-82f3-471d-aa6a-1986d30d317c

### Test Plan

- [x] PR has an associated issue: #749 
- [x] `make check` passes

### DIFF
--- a/datajunction-ui/package.json
+++ b/datajunction-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "datajunction-ui",
-  "version": "0.0.1-rc.22",
+  "version": "0.0.1-rc.21",
   "description": "DataJunction Metrics Platform UI",
   "module": "src/index.tsx",
   "repository": {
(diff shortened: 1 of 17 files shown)


agorajek (approved):
Neat! The only thing I would ask for is to add a typeahead for schema and tables at some point. But for now this is awesome.

Issue #777, opened 2023-09-19
Add Display Name to the node columns.


agorajek:
By default we can create it from the column name by:
- replacing `_` with space
- capitalizing first letters

With the ability to override any specific column display name.

PR #783, opened 2023-09-20
Add display name to columns

shangyian:
### Summary

Adds display name to node columns. For the time being, we just set column display names to be a cleaned up version of the column name itself (i.e., we replace underscores with spaces and capitalize the initial word). We can later add an endpoint that exposes the column display names to be editable via the API

### Test Plan

- [x] PR has an associated issue: #777 
- [x] `make check` passes

### Deployment Plan

I tested the database migration locally (and updated dj.demo.db). Note that the migration SQL is compatible with sqlite and cockroachdb, but can't guarantee compatibility with other databases.


agorajek (approved):
Wow, that was fast!

Issue #733, opened 2023-08-18
Freeing Metrics through Measures and Dimensions.


agorajek:
**Main Problem:**
Metric nodes are currently defined on a specific node and it becomes hard to reuse the same Metric definition across DJ dag. So, even if we had an exact copy of a DJ node, that is a parent to a given metric, that other node would not be able to serve as a source of data for the same metric. Let's remember that Cube nodes serve as a way to reuse the concept of Metrics and Dimensions, but Cube nodes don't work well for any arbitrary table that we would want to plug into DJ dag.

**Proposed Solution:**
It is important to preserve the current structure of DJ client APIs while introducing any new model. This would guarantee that any DJ read-only clients are unaffected by these changes.

**The gist**: instead of defining DJ Metrics on top of a single node, we can introduce the concept of Measures and define Metrics on top of these Measures. 

**The details:**

First let's review Measures:
1. Measures would be separate entities of the following structure: {name: str, columns: [<node.column>, ...]}, 
    example: `{name: revenue, columns: [online_sales.revenue, store_sales.revenue, membership.revenue]}`
2. Measures (since they are separate entities) would have a separate namespace (from nodes).
3. Measure validation would ensure (for simplicity):
    - that all the columns each Measure points to are of the same type.
    - that only one column on any node can point to the same measure
4. So in order for node columns to be described as a Measure we would need a new column attribute: `{"measure": <measure_id>}`

Now back to Metrics:
1. As before, Metrics would be aggregate functions.
2. As before, Metrics would be allowed to have required dimensions.
3. But, Metrics would not be full queries anymore, and instead they should have:
    - the Metric expression (valid SQL expression based on Measures and Dimension Attributes[*])
    - the list of Required Dimensions (aka PK of the Metric)
4. Each Measure would effectively "be connected" to the following entities:
    - Measures used in the Metric expression
    - Dimensions which attributes are used in the Metric expression
    - Dimensions listed as Required Dimensions

[*] For the Dimension Attributes to be discoverable and accesible when applying Metric expresssions we will need additional column property: `{"dim_attribute": "<dim_node>.<dim_column>"}`.

The main fallout of the above solution is that Metrics become loosely coupled and they become floating definitions applicable (or computable) on any node where the required set of Measures and Dimensions is available. This surely makes the metric validation and discovery harder if we decide to keep the validation of Metrics tied to any DJ dag changes.

**Some Open Questions:**
1. Do we want Metrics to remain 1st class Nodes?
[...]

PR #785, opened 2023-09-21
Measures CRUD API

shangyian:
### Summary

This PR adds API endpoints for:
* `POST /measures`: creating measures
* `PATCH /measures/{name}`: editing measures
* `GET /measures`: listing measures (only shows the measure name)
* `GET /measures/{name}`: shows detailed info for the measure, including the column(s)

I'll punt the delete measures endpoint to a different PR, after we implement the functionality to reference them in metrics.

### Test Plan

Added a series of unit tests. Will add some example measures to the demo db as well.

- [x] PR has an associated issue: #733 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/measures.py
+++ b/datajunction-server/datajunction_server/api/measures.py
@@ -3,7 +3,7 @@ Measures related APIs.
 """
 
 import logging
-from typing import List
+from typing import List, Optional
 
 from fastapi import Depends
 from sqlmodel import Session, select
@@ -54,7 +54,7 @@ def get_node_columns(session: Session, node_columns: List[NodeColumn]) -> List[C
         available = [
             col for col in node.current.columns if col.name == node_column.column
         ]
-        if len(available) != 1:
+        if len(available) == 0:
             raise DJDoesNotExistException(
                 message=f"Column `{node_column.column}` does not exist on "
                 f"node `{node_column.node}`",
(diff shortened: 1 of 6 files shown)


datajunction-server/datajunction_server/api/measures.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Finds all the specified node columns or raises if they don't exist
+    """
+    measure_columns = []
+    for node_column in node_columns:
+        node = get_node_by_name(session, node_column.node)
+        available = [
+            col for col in node.current.columns if col.name == node_column.column
+        ]
+        if len(available) != 1:
```
agorajek, line 57: nit: the error message fits only the ` == 0` case so I would either adjust it or add the additional case.

datajunction-server/datajunction_server/api/measures.py
```diff
@@ -0,0 +1,145 @@
...
+            raise DJDoesNotExistException(
+                message=f"Column `{node_column.column}` does not exist on "
+                f"node `{node_column.node}`",
+            )
+        measure_columns.extend(available)
+    return measure_columns
+
+
+@router.get("/measures/", response_model=List[str])
+def list_measures(*, session: Session = Depends(get_session)) -> List[str]:
```
agorajek, line 67: We could do add a `prefix` argument here and have this case be server when that arg is empty?

agorajek (approved):
Here come the measures :). Two nits inline.

Issue #787, opened 2023-09-23
/sql endpoint for non-metrics should not include dims in group by


shangyian:
When requesting SQL for a non-metric node (i.e., transform, dimension, or source nodes), the generated SQL should not include requested dimension attributes in the `GROUP BY` clause. Instead, it should just join the appropriate nodes to bring in the dimension attribute and include it in the `SELECT` clause.

PR #789, opened 2023-09-23
SQL for non-metric nodes should not include dimension attributes in group by

shangyian:
### Summary

When requesting SQL for a non-metric node (i.e., transform, dimension, or source nodes), the generated SQL should not include requested dimension attributes in the `GROUP BY` clause. Instead, it should just join the appropriate nodes to bring in the dimension attribute and include it in the `SELECT` clause. This PR adds a small check based on node type to determine whether or not to include the dimension attribute in the `GROUP BY`.

### Test Plan

Locally. Added a unit test for requesting dimension attributes on a source (`default.repair_orders`, which can access `default.hard_hat` after two hops away on the dimensions graph).

- [x] PR has an associated issue: #787 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -368,6 +368,7 @@ def add_filters_dimensions_orderby_limit_to_query_ast(
     dimensions: Optional[List[str]] = None,
     orderby: Optional[List[str]] = None,
     limit: Optional[int] = None,
+    include_dimensions_in_groupby: bool = True,
 ):
     """
     Add filters and dimensions to a query ast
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/construction/build.py
```diff
@@ -530,6 +532,7 @@ def build_node(  # pylint: disable=too-many-arguments
         dimensions,
         orderby,
         limit,
+        include_dimensions_in_groupby=(node.type == NodeType.METRIC),
```
agorajek, line 535: I think this is good (for now), but I have a feeling that `get_query(...)` function becomes a little too overloaded. I think that it would be easier if we had a simple query generation method that behaves the same way for any node in DJ dag. And in addition to that we could have a special methods for generating SQL from multiple metrics and dimensions. And we have the latter one already. I am just not sure if we should make the `get_query` function do different things for different node types...

agorajek (approved):
Feel free to merge if you think this is a good approach, but please check out my inline comment.

datajunction-server/datajunction_server/construction/build.py
shangyian: Will merge this for now and we can decide on whether a different interface is needed afterwards

Issue #790, opened 2023-09-25
History events should be updated with a user


shangyian:
* The `History` model already has a `user` field, which we can use to reference the `username` field in the `User` model.
* `username` in the `User` model may require a unique constraint
* We can introduce a foreign key relationship in SQLAlchemy, but keep the join as a lazy join

PR #791, opened 2023-09-25
Include the current user when storing history events

shangyian:
### Summary

This PR adds functionality to store the current user every time we create a history event, allowing us to track which user is associated with a particular action. We use the `user` field on the `History` model, and we store the user's username.

### Test Plan

Locally + unit tests

- [x] PR has an associated issue: #790 
- [x] `make check` passes

### Deployment Plan

N/A since we can't find users for previously saved history events


agorajek (approved):
Looks good.

Issue #792, opened 2023-09-25
Support for updating cubes via API


shangyian:
At the moment, we don't actually support updating cubes. While we claim to have an `update node` endpoint, this endpoint only accepts input arguments that work for source/transform/dimension/metric nodes and not cubes.

PR #795, opened 2023-09-26
Cube Node Fixes

shangyian:
### Summary

This PR has a number of cube-related fixes:

#### Cube Updates

Cube updates weren't working before. If someone tried to update a cube using the existing update-node endpoint, they would fail out because we didn't allow for specifying metrics and dimensions in the input. Furthermore, the generic node update sequence doesn't actually update the metrics and dimensions in a cube. 

This change separates out the cube node update into a separate path to handle updates correctly, including:
* Minor updates like changes to `display_name` and `description`
* Major updates like changes to the metrics and dimensions in the cube
* Updates and reschedules all materializations if there are major updates
* Propagates upstream node updates to cubes

#### Druid Materialization Fix

Druid materialization for cube nodes were generating measures columns (ingested to Druid) with column names that were too long to be queried via Druid SQL. This means that we create Druid datasources that are unqueryable later. 

This PR limits the measure column names to be 128 characters max. We use a readable name, generated via the same method as before, but truncate it so that we have enough space to append a crc32 hash (8 characters). This yields measure column names like `orders3845127662_sum`.

#### Druid Materialization Columns

In order to materialize cubes to a Druid datasource, we first materialize it to a table in the data warehouse. This change will save the column names and types on that table to the default cube materialization config, which will be returned every time we ask for the cube.

### Test Plan

- [x] PR has an associated issue: #792
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -477,19 +477,7 @@ def update_node_with_query(
 
     session.add(new_revision)
     session.add(node)
-
-    session.add(
-        History(
-            entity_type=EntityType.NODE,
-            entity_name=node.name,
-            node=node.name,
-            activity_type=ActivityType.UPDATE,
-            details={
-                "version": new_revision.version,  # type: ignore
-            },
-            user=current_user.username if current_user else None,
-        ),
-    )
+    session.add(node_update_history_event(new_revision, current_user))
 
     if new_revision.status != old_revision.status:  # type: ignore
         session.add(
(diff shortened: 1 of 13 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -413,13 +415,42 @@ def save_node(
...
+    if node.type == NodeType.CUBE:
+        node_revision = update_cube_node(
+            session,
+            node.current,
+            data,
+            query_service_client,
+            current_user,
+        )
+        return node_revision.node if node_revision else node
+    return update_node_with_query(
```
agorajek, line 438: Is SOURCE node part of this code path? If so `update_node_with_query` seems to be doing more than its name says. And of not should we catch the SOURCE node case here somehow?

agorajek (approved):
This is great! There is one question in line where I wasn't sure what's going on.

PR #796, opened 2023-09-27
During cube materialization, pass columns of the intermediate table to query service

shangyian:
### Summary

During cube materialization, pass columns of the intermediate table over to the call to the query service. This enables us to create the table with the columns first.

### Test Plan

- [x] `make check` passes


agorajek (approved):
`intermediate_columns` is a little mysterious, but I can't find a better name so let's go with that. Maybe a comment above this line:
```
# cube dataset column list may include metric components
...
```

PR #802, opened 2023-09-29
pass columns to query service for druid materialization

shangyian:
### Summary

Instead of calling the query service with `intermediate_columns` for Druid materialization, use `columns`, which is also used by generic materialization.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
I am not sure actually about this, but lets give it a try.

Issue #779, opened 2023-09-19
Adjust the DJ query to select only columns that are used upstream.


agorajek:

shangyian:
Hmm, I wonder what changed after the linked PR on this issue https://github.com/DataJunction/dj/issues/458

PR #803, opened 2023-09-29
Optimize SQL generation for multiple metrics

shangyian:
### Summary

When we build SQL for multiple metrics, we should build a more optimized query that doesn't produce separate scans of each metric's parent, which scales the redundant scans based on the number of metrics in the query. This PR changes the multiple metrics building so that we:
1. Group metrics by their parent nodes.
2. Build a query for each parent node with the dimensional attributes referenced joined in
3. For each metric that references the parent node, insert the metric expression into the parent node's select and build the parent query.
4. Set the rest of the parent query's attributes (filters, orderby, limit etc)
5. Join together the transforms on the shared dimensions
6. Select all the requested metrics and dimensions in the final SELECT

### Test Plan

Lots of local test + unit tests.

- [ ] PR has an associated issue: Somewhat related to #458 #779, but this is somewhat different in that it's about not reexecuting the same metric query over and over
- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -62,6 +62,7 @@ from datajunction_server.sql.parsing import ast
 from datajunction_server.sql.parsing.backends.antlr4 import SqlSyntaxError, parse
 from datajunction_server.sql.parsing.backends.exceptions import DJParseException
 from datajunction_server.typing import END_JOB_STATES, UTCDatetime
+from datajunction_server.utils import LOOKUP_CHARS, SEPARATOR
 
 _logger = logging.getLogger(__name__)
 
@@ -746,7 +747,7 @@ def validate_orderby(
     """
     invalid_orderbys = []
     for orderby_element in orderby:
-        if orderby_element not in metrics + dimension_attributes:
+        if orderby_element.split(" ")[0] not in metrics + dimension_attributes:
             invalid_orderbys.append(orderby_element)
     if invalid_orderbys:
         raise DJException(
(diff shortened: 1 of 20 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -849,7 +853,7 @@ def build_sql_for_multiple_metrics(  # pylint: disable=too-many-arguments,too-ma
                     ColumnMetadata(name=col.name, type=str(col.type))
                     for col in (metric_columns + dimension_columns)
                 ],
-                dialect=engine.dialect,
+                dialect=materialized_cube_catalog.engines[0].dialect,
```
agorajek, line 856: With Druid as the only engine for cubes this is ok. But we should probably make a comment here that this is a temporary shortcut.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -295,7 +295,7 @@ def create_cube_node_revision(  # pylint: disable=too-many-locals
     dimension_attribute = session.exec(
         select(AttributeType).where(AttributeType.name == "dimension"),
     ).one()
-    dimensions_set = {dim.rsplit(SEPARATOR, 1)[1] for dim in data.dimensions}
+    dimensions_set = {dim.replace(SEPARATOR, "_DOT_") for dim in data.dimensions}
```
agorajek, line 298: If its not too hard lets define this in constants. It will be easier to find later.

datajunction-server/datajunction_server/construction/build.py
agorajek, line 1: Wow, this is so much goodies.

agorajek (approved):
@shangyian this is fantastic! Thank you.

PR #805, opened 2023-10-04
Display Name to Node Columns

youngman-droid:
### Summary

Adds Display Name to Node Columns, also adds api to change display name

### Test Plan

- [  #777] PR has an associated issue: 
- [ ] `make check` passes


datajunction-server/datajunction_server/models/history.py
```diff
@@ -47,6 +47,7 @@ class EntityType(str, Enum):
     NODE = "node"
     QUERY = "query"
     TAG = "tag"
+    DISPLAY_NAME = "display_name"
```
shangyian, line 50: Hmm, I think we don't need a specific 'display_name' entity in history. The entity should be `column` and the action `update`

datajunction-ui/src/mocks/mockNodes.jsx
```diff
@@ -1188,89 +1207,94 @@ export const mocks = {
...
             name: 'default.date_dim',
           },
         },
         {
-          name: 'hire_date',
-          type: 'date',
-          attributes: [],
-          dimension: {
-            name: 'default.date_dim',
+          "name": "hire_date",
```
shangyian, line 1238: Not sure why the formatting changed here, but that's fine

shangyian (approved):
Changes look good, thanks for the updates!

PR #806, opened 2023-10-04
Add Tags + Modify Tags on Node

shangyian:
### Summary

This PR makes it so that we can (a) create tags and (b) tag nodes.

The main changes include:
* Changed the "tag a node" endpoint to be an "update tags on a node" endpoint. This means that calling `POST /nodes/{name}/tags` will update the node with the list of provided tags, which allows for both adding tags to a node and removing tags from the node. The previous setup only allowed us to add tags to a node.
* Added a select field of available tags on the node creation and edit pages.
* Added a link from each tag on the node page to a tag page that displays a list of nodes for that tag
* Display names of tags should be editable
* Added a page for creating new tags

#### Tagging nodes

https://github.com/DataJunction/dj/assets/9524628/177efeeb-35c7-4987-a4b8-5965e44b0e77

#### Tags page

Screenshot 2023-10-04 at 8 43 59 AM

#### Create Tag

Screenshot 2023-10-04 at 3 52 39 PM

### Test Plan

- [ ] PR has an associated issue
- [ ] `make check` passes


datajunction-server/datajunction_server/api/tags.py
```diff
@@ -144,5 +163,8 @@ def list_nodes_for_a_tag(
             http_status_code=404,
         )
     if not node_type:
-        return sorted([node.name for node in tag.nodes])
-    return sorted([node.name for node in tag.nodes if node.type == node_type])
+        return sorted([node.current for node in tag.nodes], key=lambda x: x.name)
+    return sorted(
```
samredai, line 167: Do you think it's better to leave sorting to the frontend (UI, clients, etc)? Tags may not grow quickly enough for there to be a large number of them but when that happens, I'm wondering about centralizing the sort to the backend server vs. distributing it among the clients. Either most likely will work for a looong time so maybe it's not worth worrying about right now.

samredai (approved):
LGTM! Just had one small comment, nothing blocking.

CircArgs (approved):

datajunction-server/datajunction_server/api/tags.py
agorajek: I agree on this. Seems nice but unnecessary.

agorajek (approved):
Fantastic, as always.

datajunction-server/datajunction_server/api/tags.py
shangyian: Good point, might be easier to just stick to client-side sorting. Although I was thinking that if we have a defined order, it might be simpler for the server to always maintain that order (i.e., column order on a node). In the case of tags though, alphabetical order is somewhat of an arbitrary choice.

PR #807, opened 2023-10-04
Synchronized release management

agorajek:
### Summary 

This PR is to make the release management easier. All the details are explained in the docs included in this PR. 

At the high level:
- added a PR action to test for matching version numbers, so we don't make a mistake easily (version-test.yml)
- version-bump.yml will bump your versions across the board, staying withing the current cycle (alpha, beta, etc) until we are ready to switch
- version-publish.yml will publish everything to pypi/npm/etc

Also... I pushed all the versions to 0.0.1a30 to have a common start.

Will need some help to add Java client when time comes.

### Test Plan
1. tested `version-test.yml` in this PR back and forth... it is pretty hard to get the multiple var comparisons in the if statement right.
2. `version-bump.yml` and `version-publish.yml` will test after merging this PR. cause I don't have a good idea how to do it now.

- [ ] PR has an associated issue: #811
- [ ] `make check` passes

### Deployment Plan

n/a


.github/workflows/weekly-version-bump.yml
```diff
@@ -0,0 +1,73 @@
...
+          echo "Bump DJ Server version ..."
+          cd datajunction-server
+          hatch version ${{ github.event.inputs.bump }}
+          export NEW_VERSION=v$(hatch version)
+
+          echo "Copy version to DJ Query Service ..."
+          cp ./datajunction_server/__about__.py ../datajunction-query/djqs/
+
+          echo "Apply to DJ Python Client ..."
+          cp ./datajunction_server/__about__.py ../datajunction-clients/python/datajunction/
```
shangyian, line 58: Hmm, wait, these might not have the same versions right? The client and the server are already not in sync in terms of released versions, and plus in the future if someone does a manual release of either the client or the server alone, these will get out of sync.

.github/workflows/weekly-version-bump.yml
```diff
@@ -0,0 +1,73 @@
+name: Bump versions of all DJ Services and Clients.
```
shangyian, line 1: Do we actually want a weekly auto-release? Or is it better off as a manual thing? Then we can be more intentional about when we release

.github/workflows/weekly-version-bump.yml
agorajek: I'd like to put them in sync. But you are right that someone could always run one of these locally and ship to pypi... hmmm. I can switch to making these versions independent.

.github/workflows/weekly-version-bump.yml
agorajek: For the minor/major version we would still run this workflow manually, but the weekly automation would only bump the patch version. I think this should be automated.

shangyian: Ok sounds good 👍

samredai: Instead of the patch version, why don't we publish [release candidates](https://peps.python.org/pep-0440/#pre-releases) with the `rc` tag in the version? That way these weekly automated pushes don't get picked up automatically everywhere the client isn't pinned. (pip will ignore `rc` versions unless specifically requested in the install command)

agorajek: Will pip also ignore the `alpha` and `beta` version tags? Because we are now using `alpha` everywhere. I don't think we should make this too complex at this stage of the project.

agorajek: Thinking more about it... once we get these versions in sync manually. What's wrong with trying to keep them in-sync... and yes, the weekly-auto process would only do alpha/beta/rc. And when we want to bump a big version we would run this workflow with a different argument (minor, major. etc). I still think that keeping them all in sync would be easy and great way to lower our brain cycles. What am I missing?

shangyian: Yeah, that's a good point. If we're going to do auto-releases, then it will be easier if we just keep all the versions aligned, and if we do a few extra releases that have nothing in them, that's fine

samredai: Yeah I'm ok with that too. And I heard a great point the other day that we can always just include a simple argument to skip releases of certain components, for example if we're doing a release of everything from 0.0.1 to 0.0.2 but nothing has changed in the client, we could always add a flag to skip the client. Doesn't seem like an issue that the client version would go from 0.0.1 to 0.0.3 and it would allow everything to stay in sync.

datajunction-clients/python/Makefile
```diff
@@ -7,14 +7,7 @@ lint:
...
-	@git add pyproject.toml
-	@git commit -m "v$$(poetry version -s)"
-	@git tag v$$(poetry version -s)
-	@git push
-	@git push --tags
-	@poetry version
-
-release:
-	@poetry publish --build
+dev-release:
```
shangyian, line 10: Should this also include an alpha-release option, since we've been mostly doing alpha releases?

docs/content/0.1.0/docs/developers/releasing-new-version.md
```diff
@@ -0,0 +1,69 @@
...
+weight: 100
+title: "Releasing New Version"
+---
+
+DataJunction project publishes all of its backend services and the Python client library to [PyPI](https://pypi.org/):
+
+- datajunction-clients/python
+- datajunction-query
+- datajunction-server
+- datajunction-reflection (NOT YET IMPLEMENTED)
```
shangyian, line 11: We actually do publish the reflection service here: https://pypi.org/project/datajunction-reflection/

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,108 @@
...
+          git commit -m "Bumping DJ to version $NEW_VERSION ."
+          git checkout -b releases/version-$NEW_VERSION
+          git push --set-upstream origin releases/version-$NEW_VERSION -f
+
+          echo "Make a tag ..."
+          git tag -a version-$NEW_VERSION -m version-$NEW_VERSION
+          git push origin version-$NEW_VERSION
+
+          echo "Create a PR ..."
+          gh pr create -B main -H "releases/version-$NEW_VERSION" --title "Make DataJunctino version $NEW_VERSION" --body "This is an automated PR triggered by the Weekly Version Bump action. Merging this PR will publish all the component for version $NEW_VERSION ."
```
shangyian, line 106: typo here fyi

shangyian (approved):
Overall this looks fine, let's try it out! Thanks for adding

docs/content/0.1.0/docs/developers/releasing-new-version.md
agorajek: Oh, indeed.

datajunction-clients/python/Makefile
agorajek: I don't think so, because you only want to do the Alpha (which is our current cycle) in a synchronized fashion and not individually.

docs/content/0.1.0/docs/developers/releasing-new-version.md
agorajek: Fixed.

samredai (approved):
Thanks @agorajek!

PR #808, opened 2023-10-05
Access control

CircArgs:
### Summary

Adds authorization models and implements authorization on some routes.
Future routes to come in successive PRs

### Test Plan

Tests WIP. Manual tests all passed

- [x] `make check` passes


datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,436 @@
...
+if TYPE_CHECKING:
+    from datajunction_server.sql.parsing.ast import Column
+
+
+class ResourceRequestVerb(Enum):
+    """
+    Types of actions for a request
+    """
+
+    VIEW = "view"
```
agorajek, line 35: What do you mean by VIEW? To see whether the resource exists? In other-words see its name, w/o seeing its content? If do, didn't think we would need this kind of permission. I'd say we should be able to see (or LIST) all the objects and READ only those which we are allowed to READ.

CircArgs: >  In other-words see its name, w/o seeing its content?

Yeah.

Nothing in this design forces you to restrict viewing. It just provides the option. I'm fairly confident a place like Travelers would even want to restrict viewing

samredai: +1 To differentiating view and read. Being stingy with read access is really about preventing people from authoring downstream nodes on top of your node. However you still may want people to have view access, particularly from the UI. Giving someone view but not giving them read is a way to say "You can see how I defined my node, but I don't want you to be allowed to build on top of it."

Another case I can think about is people who want to "work in the darkness" until they're ready to share their nodes more broadly. We sort of do this with draft but with view access that may actually be a better way to do this. They can model their entire DAG without giving anyone view access, and then they can grant view to everyone when they're ready.

samredai: This is similar to the `REFERENCE` permission in [snowflake](https://docs.snowflake.com/en/user-guide/security-access-control-privileges).

> Grants the ability to view the structure of an object (but not the data).

agorajek: Ok... then i think the verbs should be like this:
- LIST : you can see the node/resource exist at all, aka its name
- ACCESS  : you can see the node in RO mode and you can access its data, i don't think these should be separate
- EDIT : you can modify the node 
- ADMIN : you can grant provs/verbs on this node to others

I would not use READ or WRITE here because these terms can be easily confused between node attribute access and the data access.

shangyian: I'm still unsure about use cases around column-level access. Because if you have node-level access, but not column-level access, that doesn't make sense since the former would give you access to the query, which gives you access to the latter. But if you have column-level access and not node-level, then you would have a very difficult time understanding how this column fits into the larger picture of the node, since you wouldn't have access to the query.

datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+class ResourceObjectBase(BaseModel):
+    """
+    Base class for resource objects
+    that are passed to injected validation logic
+    """
+
+    name: str  # name of the node
+    created_by: str
```
shangyian, line 47: Does this need an `updated_by` as well? (Can be introduced later, of course)

datajunction-server/datajunction_server/models/access.py
agorajek: Yeah... I think we can push the problem of specific data access permissions to a later time. The reason I am saying that is because you can already create a transform node B (with a smaller scope) of the transform node A, and give access to node B instead of A.

samredai: @olek Now that you say it, I agree `READ` and `WRITE` could be seen as vague. Here's my thoughts on your suggestion:
- `LIST` implies that you can only see it in a list, this is not what's meant by `VIEW` above which means you can fully view the node in lists as well as all of it's metadata in the UI (the query, the columns, materializations, etc). I'm ok with having `LIST` as well but I don't think it replaces the need for `VIEW` behavior (or whatever we call it).
- `ACCESS` feels like a very vague name. I think the verb should best describe what it is that you can access. As for the description, you don't think there's a case for allowing others to discover a node they don't have read access to? I can imagine that would be a common precursor to requesting access to read the data (after reading the description and seeing the column names/types).
- I'm ok with `EDIT` and `ADMIN` but they're still broad brushes. For example, if I have `EDIT` can I delete the node? (having separate `WRITE` and `DELETE` permissions is common in other systems). If I want to give someone the ability to grant others access to the node without allowing them to delete the node, how can I do that without making them an `ADMIN`?

In summary, I don't think this is an either/or in terms of granularity, meaning we could start with granular permissions and then also add more powerful permissions that encompass others. `LIST` and `VIEW` can be individual permissions but `READ` can be a broader brush that lets you do everything allowed by `LIST` and `VIEW` as well as allows you to read the data.

datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,436 @@
...
+        if node.id
+        in {
+            request.access_object.revision_id
+            for request in validation_results
+            if request.approved
+        }
+    ]
+
+
+def validate_access_namespaces(
```
samredai, line 360: nit: I think this models directory should only contain the model classes since I can imagine these authorization related functions will grow in number over time. These functions would probably be better located somewhere under the internal directory, maybe in a new submodule called `authorization`?

datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,436 @@
...
+            for request in validation_results
+            if request.approved
+        }
+    ]
+
+
+ValidateAccessFn = Callable[[AccessControl], None]
+
+
+# Dummy default if not dependency injected
```
samredai, line 406: This comment can just go in the docstring

datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,436 @@
...
+    validation_results = access_control.validate()
+    if raise_:
+        access_control.raise_if_invalid_requests()#pragma: no cover
+
+    return [
+        namespace
+        for namespace in namespaces
+        if namespace
+        in {
+            request.access_object.name
```
samredai, line 396: Could this logic be added as a helper method to this class? I imagine there will be many areas of the source code where you want to inspect one of these access control instances to ask for a list of all approved or denied requests with an argument to filter by type (namespace/node/tag, etc)

datajunction-server/datajunction_server/models/access.py
CircArgs: I was afraid it might carry onto someone implementing their own and confuse them

datajunction-server/datajunction_server/models/access.py
CircArgs: @shangyian this class exists as a layer between DJ internals and the injected code. The goal is to just provide some features the injected authorization code would want to look at to approve or deny access. You think they might want to know who updated for that?

datajunction-server/datajunction_server/models/access.py
samredai: Ah I see, that's a good point. Maybe it can be reworded to get the same message across but also provide clarity? Such as describing it as a placeholder/example access validation dependency that approves all requests.

datajunction-server/datajunction_server/models/access.py
agorajek: @samredai many good points. But instead of responding to all your comments let me try another take on this initial list of verbs for nodes:
- VIEW : you can see the definition of the resource/node, its name, attributes, etc. (I don't think we need a special LIST permission, #famouslastwords)
- QUERY :  you can query the data from a given node
- EDIT : you can modify a node (includes deletion, because modifying its query to SELECT 1 is equivalent to soft deletion)
- GRANT : you can grant/revoke permissions to other on a given node (should be better than ADMIN).

Let's try it for tags:
- VIEW : you can see the tag and its nodes, description etc
- EDIT : you can modify a tag and add/remove nodes from it 
- GRANT : you can grant/revoke permission on a tag

CircArgs: @agorajek It doesn't sound like you are thinking of the verbs as verbs in your descriptions. 
> you can...
 The verbs are not permissions themselves which **'you can'** would suggest. They are the actions that are being attempted. 

I would rephrase as 
> VIEW : You ~~can~~ _are attempting to_ see the definition of the resource/node, its name, attributes, etc. 
> ... so on

`GRANT` definitely isn't a verb it sounds like it belongs to the logic behind the injected validation code which could be.

samredai: Got some advice from a reliable source who's worked a lot in the security space. The piece of advice was to try and stick to a set (or subset) of permissions from any popular and well-used system out there.

What do we all think about just starting with a handful of permissions defined by postgres? https://www.postgresql.org/docs/current/ddl-priv.html

agorajek: @CircArgs i was definitely thinking about verbs, but perhaps I made few thinking short-cuts that were confusing. In the example of GRANT, I just though that it could be easier to understand as ADMIN. All of these terms I mentioned could be seen as nouns or verbs, but let's not go there...

@samredai I am OK with just taking some existing set of permissions and applying them to nodes, tags and namespaces. My concern here is that some database related concepts will not match well with DJ concepts. I thought we could come up with a well-fit pattern here, but you prefer to re-use something I would suggest looking at existing semantic layer systems... here is an example from [Microstrategy](https://www2.microstrategy.com/producthelp/Current/SystemAdmin/WebHelp/Lang_1033/Content/Controlling_access_to_objects__Permissions.htm) . Notice that Browse/Read/Write/Delete/Control/Use sound much more meaningful for DJ than the database examples (like Postgres or Snowflake).

datajunction-server/datajunction_server/models/access.py
shangyian: I see, in that case agreed that they shouldn't need updated_by. But then would it be clearer to just call this `owner` rather than `created_by`, since I assume that the owner of the node better represents access and may have changed from the user who created the node?

datajunction-server/datajunction_server/models/access.py
shangyian: @agorajek and @CircArgs, plus one to borrowing from Microstrategy. I like the following:
- Browse: Can see the entity's name in any list view
- Read: Can view all detailed info on the entity
- Write: Can modify the entity
- Delete: Can delete the entity (we can decide if there's an additional layer of permission needed for hard-deletes)
- Use: Can reference the entity in a different entity, i.e., a downstream node
- Execute: Can retrieve data for entities that have data (i.e., nodes)
- Full: All of the above, plus the ability to change ACL on the entity

CircArgs: @shangyian Is USE assuming people can create nodes based on others they may not be able to READ?

shangyian: I was thinking that `USE` is more like a case where you can `READ` the node, but you might not be able to use it in any downstream nodes, and `USE` is explicitly granting that permission.

samredai: +1, love where we landed on this! I agree that `USE` would be useful. It's a verb that signals an attempt to "use" the node in a creation/update of another node. The injected code can decide how it wants to treat that.

datajunction-server/datajunction_server/models/access.py
```diff
@@ -0,0 +1,437 @@
...
+    verb: ResourceRequestVerb,
+    user: Optional[User],
+    nodes: Iterable[Union[NodeRevision, Node]],
+    raise_: bool = False,
+) -> List[Union[NodeRevision, Node]]:
+    """
+    Validate the access of the user to a set of nodes
+    """
+    if user is None:
+        return list(nodes)  # pragma: no cover
```
samredai, line 346: More reasons to make auth mandatory 😄 . We'll have to remember to come back and change this back to raising an exception, or remove the `if user is None` check entirely.

CircArgs: I intend to remove all these optional checks including any for access_control once the user is required and access control is implemented for all routes

samredai:
@CircArgs I pulled this down and ran it locally. It's working great! The only issue I noticed is when access validation fails and raises a DJException, it needs to use `HTTPStatus.FORBIDDEN` instead of `HTTPStatus.UNAUTHORIZED` because otherwise this snippet below in the dj exception handler will get triggered and will clear the jwt and basically force a logout. So when I tweaked it to test how the UI responds, it would just log you out on the next click.
```py
    # If unauthorized, clear out any DJ cookies
    if exc.http_status_code == HTTPStatus.UNAUTHORIZED:
        response.delete_cookie(AUTH_COOKIE, httponly=True)
        response.delete_cookie(LOGGED_IN_FLAG_COOKIE)
```
Also about the UI, we'll need to make some changes there to respond to these access errors and provide some sort of "access denied" visual. Definitely doesn't need to be part of this PR but thought I'd mention it here.

datajunction-server/datajunction_server/models/access.py
CircArgs: Okay here are the verbs now:

```python
class ResourceRequestVerb(Enum):
    """
    Types of actions for a request
    """

    BROWSE = "browse"
    READ = "read"
    WRITE = "write"
    USE = "use"
    EXECUTE = "execute"
    DELETE = "delete"
```

samredai (approved):

PR #810, opened 2023-10-05
Sign Up page to create a basic auth user

samredai:
### Summary

This adds a signup form to the login screen that can be used to create a basic auth user (not using google oauth or github oauth). This gets us one step closer to always requiring auth and is worth serious consideration on if we want to do that in the OSS project as an alternative to what we recently agreed on in a recent sync--setting the user as "unknown" when auth is globally disabled.

In other words, creating a basic auth user through the signup page is such a low bar that maybe we can make it the lowest bar :)

https://github.com/DataJunction/dj/assets/43911210/018be113-b227-4582-8663-df0cda207e52

### Test Plan

Added UI tests for the signup page and tested signing up in the local docker demo environment.

- [x] PR has an associated issue: #697 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/authentication/basic.py
+++ b/datajunction-server/datajunction_server/api/authentication/basic.py
@@ -24,6 +24,7 @@ router = APIRouter(tags=["Basic OAuth2"])
 
 @router.post("/basic/user/")
 async def create_a_user(
+    email: str = Form(),
     username: str = Form(),
     password: str = Form(),
     session: Session = Depends(get_session),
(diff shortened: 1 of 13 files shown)


CircArgs (approved):

CircArgs:
Cool! Let's get rid of users being optional in routes and require basic auth?

CircArgs:
Also the signing page reminds me of early 2000s sites. it looks like it belongs to 2000 Ebay:

agorajek:
I think I like this idea! @betodealmeida do you have any thoughts from your OSS experience?

shangyian (approved):
Nice!

samredai:
> Cool! Let's get rid of users being optional in routes and require basic auth?

Agreed. I was initially thinking to do this by injecting "unknown" when auth is disabled but requiring auth feels like the better path.

PR #813, opened 2023-10-10
Config file based configuration for Query Service

samredai:
### Summary

This makes a few changes to the query service to make it more easily configurable and more intuitive to extend support for other kinds of query engines.

- Catalogs and engines can now be configured via a YAML/TOML file (an .env variable points to the config file, i.e. `CONFIGURATION_FILE=./config.djqs.yml`)
- The `Engine` model has been expanded to include `type` (which so far can be `duckdb` or `sqlalchemy`) and `extra_params` where engine specific parameters can be set.
- Catalogs and engines can still be created through API requests, but since it's reasonable that some may want to disable that and only manage server configuration through a config file, `ENABLE_DYNAMIC_CONFIG=False` can be set in the `.env` and the router for creating catalogs and engines won't be mounted.

cc: @Geo881 

### Test Plan

Tested the following config file:
```yaml
engines:
  - name: duckdb
    version: 0.7.1
    type: duckdb
    uri: duckdb:////code/docker/default.duckdb
    extra_params:
catalogs:
  - name: warehouse
    engines:
      - duckdb
```
Ran `docker compose --profile demo up` and tested the SQL page in the UI as well as the catalog and engine list endpoints for the query service API.

Also tested the TOML equivalent:
```toml
[[engines]]
name = "duckdb"
version = "0.7.1"
type = "duckdb"
uri = "duckdb:////code/docker/default.duckdb"

[engines.extra_params]
location = "/code/docker/default.duckdb"

[[catalogs]]
name = "warehouse"
engines = [ "duckdb" ]
```
and the JSON equivalent:
```json
{
  "engines": [
    {
      "name": "duckdb",
      "version": "0.7.1",
      "type": "duckdb",
      "uri": "duckdb:////code/docker/default.duckdb",
      "extra_params": {
        "location": "/code/docker/default.duckdb"
      }
    }
  ],
  "catalogs": [
    {
      "name": "warehouse",
      "engines": [
        "duckdb"
      ]
    }
  ]
}
```
- [x] `make check` passes


shangyian (approved):
Looks good, thanks for adding! I'm guessing we could enable something similar for the server

datajunction-query/config.djqs.yml
agorajek, line 1: Maybe put the config.* files into a config dir? It looks like there may be more than one?

datajunction-query/djqs/api/main.py
```diff
@@ -30,10 +34,13 @@
         "url": "https://mit-license.org/",
     },
 )
-app.include_router(catalogs.router)
-app.include_router(engines.router)
+app.include_router(catalogs.get_router)
+app.include_router(engines.get_router)
 app.include_router(queries.router)
 app.include_router(tables.router)
+if settings.enable_dynamic_config:  # pragma: no cover
```
agorajek, line 41: Does it make sense to load the config above, if here you find out that `enable_dynamic_config=True` ?

datajunction-query/djqs/api/main.py
```diff
@@ -14,13 +14,17 @@
...
+from djqs.config import load_djqs_config
 from djqs.exceptions import DJException
-from djqs.utils import get_settings
+from djqs.utils import get_session, get_settings
 
 _logger = logging.getLogger(__name__)
 
-
 settings = get_settings()
+if settings.configuration_file:  # pragma: no cover
```
agorajek, line 24: Too hard to test? This seems like a big part of this PR...

datajunction-query/djqs/api/main.py
agorajek, line 41: ditto: wrt to unit test exclusion

agorajek (commented):
Nice addition. But I have couple of questions and comments in line.

datajunction-query/alembic/versions/2023_10_09_1858-f3407a1ec625_add_type_and_extra_para_ms_field_for_.py
```diff
@@ -0,0 +1,37 @@
+"""Add type and extra_para
+ms field for engines
```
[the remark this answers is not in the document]
samredai: *dogs, and fixed 😄

datajunction-query/config.djqs.yml
samredai: Can you give an example of why there would be more than one? I was thinking all catalogs and engines would be defined in a single config file.

datajunction-query/djqs/api/main.py
samredai: I didn't see any reason to prevent a setup where both a configuration file is used for priming the system with a set of catalogs and engines, but then more catalogs and engines can be added dynamically through API calls.

datajunction-query/config.djqs.yml
agorajek: Oh, never mind I think. The `djqs` infix confused me.

agorajek (approved):
awesome_sauce

PR #814, opened 2023-10-10
Fix node access control check 

shangyian:
### Summary

This fixes a bug in `validate_access_nodes` so that during the resource access validation check, we distinguish between node revision ids and node ids. It was previously just using node revision ids, which then fails validation if we get passed a list of nodes.

### Test Plan

Locally and beyond. Discovered upon deployment

- [x] `make check` passes

### Deployment Plan

N/A


samredai (approved):
Great catch!

Issue #818, opened 2023-10-12
Remove automated alias generation when saving metric nodes


shangyian:
When we save metric nodes, we automatically add an alias to the metric, and we raise an error if the user defines an alias that doesn't match this generated alias. We should just remove this functionality entirely and generate the automated alias at metric build time. The saved metric SQL should reflect exactly what the user types.

PR #819, opened 2023-10-13
Remove automated metric alias generation

shangyian:
### Summary

This change removes the automated metric alias generation upon metric creation and instead pushes it to happen in the metric query building phase. The columns attached to the node will still reference the right aliases, but the node's saved query will remain exactly as the user typed it. When we build metric queries, we generate an appropriate alias for the metric expression and put it into the query then. 

This should make creating metrics easier for the user by removing the cognitive load of understanding why an alias was suddenly appended to their metric expression, or why their chosen alias failed.

### Test Plan

Locally

- [x] PR has an associated issue: #818 
- [x] `make check` passes

### Deployment Plan

N/A


datajunction-server/tests/construction/fixtures.py
```diff
@@ -145,14 +145,14 @@ def build_expectation() -> Dict[str, Dict[Optional[int], Tuple[bool, str]]]:
         """basic.num_comments""": {
             None: (
                 True,
-                """SELECT  COUNT(1) AS cnt
+                """SELECT  COUNT(1) AS basic_DOT_num_comments
```
CircArgs, line 148: I like how `_DOT_` is still surviving. When DJ is a trillion dollar project, it will be one of those quirks discussed at dinner tables worldwide.

CircArgs (approved):

datajunction-server/tests/construction/fixtures.py
agorajek: And there will be a fraction who will say that this is because of the project roots in the department of transportation.

datajunction-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "datajunction-ui",
-  "version": "0.0.1-rc.26",
+  "version": "0.0.1-rc.27",
```
agorajek, line 3: Did you mean to change this line?

agorajek:
Does it mean that whatever the user sets as alias (if at all) will just stay in the SQL but will be never used?

shangyian:
@agorajek Yeah, if the user sets an alias for the metric, it'll get saved as a part of the metric's query, but the alias will get overwritten when we build SQL for that metric. At the moment we're raising an error if someone sets an alias that's not the allowed alias for a metric expression, which is confusing/frustrating for users

agorajek (approved):

Issue #152, opened 2022-04-30
Add support for CTEs


betodealmeida:

CircArgs:
Long solved since new parser & ast

PR #824, opened 2023-10-16
Add support for CTEs and fix bug with struct types + aliased tables

shangyian:
### Summary

This change adds support for CTEs in node queries, so users can author transform nodes like:
```sql
WITH ro as (
  SELECT
    repair_order_id,
    STRUCT(
      required_date as required_dt,
      order_date as order_dt,
      dispatched_date as dispatched_dt
    ) relevant_dates
  FROM default.repair_orders
)
SELECT
  SUM(DATEDIFF(ro.relevant_dates.dispatched_dt, ro.relevant_dates.order_dt)) AS dispatch_delay_sum,
  COUNT(ro.repair_order_id) AS repair_orders_cnt
FROM ro
```

It also fixes a bug where struct types that source from aliased tables (i.e., in the above query, `relevant_dates`, a struct, comes from a table reference aliased `ro`, which would not work at the moment since it fails to handle the case where a column has multiple names in its namespace).

### Test Plan

- [x] PR has an associated issue: #152 #495 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -435,7 +435,10 @@ def validate_node_data(  # pylint: disable=too-many-locals,too-many-statements
             else validated_node.query
         )
         query_ast = parse(formatted_query)  # type: ignore
-        dependencies_map, missing_parents_map = query_ast.extract_dependencies(ctx)
+        (
+            dependencies_map,
+            missing_parents_map,
+        ) = query_ast.bake_ctes().extract_dependencies(ctx)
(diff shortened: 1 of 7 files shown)


CircArgs:
Didn't we already support ctes? The AST has had support for a long time

agorajek (approved):
Wow, awesome fix! Thanks @shangyian

shangyian:
@CircArgs This change is not for the AST but rather allowing them in the node queries. In the current setup we get an error if we use a CTE (column type inference failures for anything that references the CTE columns) despite it being parseable by the AST.

PR #829, opened 2023-10-18
Use background task for tasks that can be done async

shangyian:
### Summary

This PR modifies some API endpoints to use background tasks for any tasks that do not need to block the API response output. Changes include:
* Column lineage saving on node creates/updates
* Scheduling materialization jobs

This reduces the amount of time we take to return a response to certain API calls.

It also fixes a bug when recreating deactivated nodes, where the deactivated node should be treated differently depending on whether it's a source, a {transform|dimension|metric} or a cube.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/materializations.py
+++ b/datajunction-server/datajunction_server/internal/materializations.py
@@ -218,8 +218,8 @@ def create_new_materialization(
         if not temporal_partition:
             raise DJInvalidInputException(
                 "The cube materialization cannot be configured if there is no "
-                "temporal partition specified on the cube. Please set at least one cube"
-                "element with a temporal partition.",
+                "temporal partition specified on the cube. Please make sure at "
+                "least one cube element has a temporal partition defined",
             )
         default_job = [
             conf
--- a/datajunction-server/datajunction_server/internal/materializations.py
+++ b/datajunction-server/datajunction_server/internal/materializations.py
@@ -187,6 +187,7 @@ def create_new_materialization(
     """
     generic_config = None
     engine = get_engine(session, upsert.engine.name, upsert.engine.version)
+    temporal_partition = current_revision.temporal_partition_columns()
     if current_revision.type in (
         NodeType.DIMENSION,
         NodeType.TRANSFORM,
@@ -214,6 +215,12 @@ def create_new_materialization(
     if current_revision.type == NodeType.CUBE:
         # Check to see if a default materialization was already configured, so that we
         # can copy over the default cube setup and layer on specific config as needed
+        if not temporal_partition:
+            raise DJInvalidInputException(
+                "The cube materialization cannot be configured if there is no "
+                "temporal partition specified on the cube. Please set at least one cube"
+                "element with a temporal partition.",
+            )
         default_job = [
             conf
             for conf in current_revision.materializations
@@ -239,7 +246,6 @@ def create_new_materialization(
                     f"engine `{engine.name}`."
                 ),
             ) from exc
-    temporal_partition = current_revision.temporal_partition_columns()
     materialization_name = (
         f"{temporal_partition[0].name}_{engine.name}"
         if temporal_partition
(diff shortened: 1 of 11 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -839,33 +857,42 @@ def _create_node_from_inactive(
         if previous_inactive_node.type != new_node_type:
             raise DJException(  # pragma: no cover
                 message=f"A node with name `{data.name}` of a `{previous_inactive_node.type.value}` "  # pylint: disable=line-too-long
-                "type existed before. If you want to re-created with a different type now, "
-                "you need to remove all the traces of the previous node with a <TODO> command.",
+                "type existed before. If you want to re-create it with a different type, "
+                "you need to remove all traces of the previous node with a hard delete call: "
+                "DELETE /nodes/{node_name}/hard",
```
samredai, line 862: Have we changed our position on clients freely using the hard delete endpoint? I don't have any problems if so but if we want to keep that as a server admin utility then we shouldn't mention it in this error response.

samredai (approved):
Overall looks good! Question about materialization: when a node is updated and materializations are scheduled via background tasks, how are we expecting scheduling issues to be debugged. Is the idea that it's something that should be part of the operational monitoring infra that picks up the errors in the log?

datajunction-server/datajunction_server/internal/nodes.py
shangyian: @samredai re: materialization, I was thinking that as a next step I could add notifications, which would then make it so that in the scheduling background task, we can catch exceptions and add it to the notifications queue. 

re: hard delete: good point, I guess we can keep this message and just block if someone tries to call that endpoint and doesn't have permissions. Although I think users could judiciously use this for nodes that they own, so as long as we check for ownership, maybe we can just allow for hard deletes?

datajunction-server/datajunction_server/internal/materializations.py
```diff
@@ -214,6 +215,12 @@ def create_new_materialization(
     if current_revision.type == NodeType.CUBE:
         # Check to see if a default materialization was already configured, so that we
         # can copy over the default cube setup and layer on specific config as needed
+        if not temporal_partition:
+            raise DJInvalidInputException(
+                "The cube materialization cannot be configured if there is no "
+                "temporal partition specified on the cube. Please set at least one cube"
```
agorajek, line 221: Maybe make it a little clearer by `... Please make sure at least one cube element has a temporal partition defined.`

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -438,12 +435,13 @@ def save_node(
     session.refresh(node.current)
 
 
-def update_any_node(
+def update_any_node(  # pylint: disable=too-many-arguments
```
agorajek, line 438: I think I will just turn this pylint rule off. It seems a little ridiculous.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -534,13 +533,21 @@ def update_node_with_query(
...
-            query_service_client,  # type: ignore
+        background_tasks.add_task(
+            schedule_materialization_jobs,
+            materializations=node.current.materializations,
+            query_service_client=query_service_client,
         )
         session.add(new_revision)
         session.commit()
 
+    if background_tasks:
```
agorajek, line 544: Hmm. So why you didn't check the existence of background_tasks when adding `schedule_materialization_jobs` to it?

datajunction-server/datajunction_server/internal/nodes.py
agorajek: If we ask for a confirmation on hard deletes, then I don't see why hiding this feature. People should be able to delete their nodes if they so desire, unless EY consultants tell us not to ;)

agorajek (approved):
Looks good. First time seeing the BackgroundTasks in action. Seems like a great feature, especially if you can add some callback to it.

datajunction-server/datajunction_server/internal/nodes.py
shangyian: So that's because we always pass in that argument, but I think I'll include that check anyway just in case this changes in the future

datajunction-server/datajunction_server/internal/materializations.py
shangyian: Good point, changing this

PR #832, opened 2023-10-20
Bubble up some of the access internals

samredai:
### Summary

This makes some small changes to the authorization logic. The main change is that it bubbles up the `Resource` class, creating it directly where access validation is applied. It was originally abstracted away by the validate access function which takes `Node` and `NodeRevision` objects directly and creates the `Resource` instance within the function.

The main reason for this is to support performing access validation when you don't have full `Node` or `NodeRevision` instances available. The case that prompted this change was that I was selecting only a few columns from the node table (name, display_name, description), but I still wanted to filter the results down to those accessible by the user. This change lets me use the information available to create a bunch of Resource instances and then pass them to the validate access function.

### Test Plan

Updated tests, ran `make tests`.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -33,7 +33,7 @@ from datajunction_server.constants import NODE_LIST_MAX
 from datajunction_server.errors import DJException, DJInvalidInputException
 from datajunction_server.internal.access.authentication.http import SecureAPIRouter
 from datajunction_server.internal.access.authorization import (
-    validate_access_placeholder,
+    validate_access,
     validate_access_requests,
 )
 from datajunction_server.internal.nodes import (
@@ -183,7 +183,7 @@ def list_nodes(
     session: Session = Depends(get_session),
     current_user: Optional[User] = Depends(get_current_user),
     validate_access: access.ValidateAccessFn = Depends(  # pylint: disable=W0621
-        validate_access_placeholder,
+        validate_access,
     ),
 ) -> List[str]:
     """
@@ -220,7 +220,7 @@ def list_all_nodes_with_details(
     session: Session = Depends(get_session),
     current_user: Optional[User] = Depends(get_current_user),
     validate_access: access.ValidateAccessFn = Depends(  # pylint: disable=W0621
-        validate_access_placeholder,
+        validate_access,
     ),
 ) -> List[NodeIndexItem]:
     """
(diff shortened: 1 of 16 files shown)


datajunction-server/tests/api/sql_test.py
```diff
@@ -828,7 +830,7 @@ def _validate_access(access_control: access.AccessControl):
         return _validate_access
 
     app = client_with_roads.app
-    app.dependency_overrides[validate_access] = validate_access_override
+    app.dependency_overrides[validate_access_placeholder] = validate_access_override
```
shangyian, line 833: Hmm, can we continue to call this dependency `validate_access`, both so that there are fewer breaking changes downstream and so that when we do implement this, we don't need to rename it from "placeholder"?

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -231,21 +241,34 @@ def list_all_nodes_with_details(
...
+        for approval in validate_access_requests(
+            validate_access,
+            current_user,
+            [
+                access.ResourceRequest(
+                    verb=access.ResourceRequestVerb.BROWSE,
+                    access_object=access.Resource(
+                        name=row.name,
+                        resource_type=access.ResourceType.NODE,
+                        owner="",
```
shangyian, line 264: Is this empty because we're not yet storing owner on a node?

shangyian (approved):
LGTM! Only one minor comment

datajunction-server/datajunction_server/api/nodes.py
samredai: Yes that's exactly right. I'm planning to follow up with a PR that stores owner. Sort of a side discussion but I have some opinions on the usefulness of an owner field. I think we should have "created_by" here instead. To me, "ownership" is something that's implied by who created the node plus an analysis of the activity/history.

datajunction-server/tests/api/sql_test.py
samredai: Oh thanks for catching this. I originally did this because I named the other function `validate_access` but I eventually changed that to `validate_access_requests` to be more explicitly named so I'll change this back.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Yeah, +1 to created_by, which we can derive from the node history

datajunction-server/tests/api/sql_test.py
samredai: Fixed 👍

CircArgs (approved):

PR #833, opened 2023-10-21
Metric kind, direction, and unit

shangyian:
### Summary

This change adds metric kind, direction and unit as additional metadata on a metric node.

This creates a new `MetricMetadata` model, which is a one-to-one mapping to the `NodeRevision` model and is only populated for metric nodes. `MetricMetadata` has `direction` and `unit` on it as available metadata fields.

* <s>Kind: The type of metric, i.e., count, ratio, rate etc. This could use some more eyes to decide if it's what we really want.</s> I removed `kind` since it doesn't seem to add much and the values here could be represented by `unit` instead
* Direction: the direction of the metric that's considered good, i.e., higher is better, lower is better, or neutral
* Unit: The unit of measure for the metric. At the moment this is just a curated list of enums, but later these units could be stored in a database and customized by the user.

The API serves the potential values of each of these fields under the following endpoint: `GET /metrics/metadata`

Also propagated changes to the UI so that we can edit these fields for metric nodes.

Screenshot 2023-10-24 at 7 10 09 AM

Screenshot 2023-10-24 at 7 10 55 AM

### Test Plan

Tested locally

- [x] `make check` passes


datajunction-ui/src/app/pages/AddEditNodePage/index.jsx
```diff
@@ -281,6 +311,54 @@ export function AddEditNodePage() {
...
+                    >
+                      <ErrorMessage name="metric_direction" component="span" />
+                      <label htmlFor="MetricDirection">Metric Direction</label>
+                      <Field
+                        as="select"
+                        name="metric_direction"
+                        id="MetricDirection"
+                      >
+                        {metricDirections.map(direction => (
+                          <option value={direction}>
```
samredai, line 341: Should we explicitly catch the default values of "Neutral", Unknown", and "Unspecified" and then set a placeholder value instead like "Set metric direction..."? I think that might be more inviting for users to go ahead and set a value even though it's optional. On that same note, can we show an asterisk on the required fields so that it's clear these are optional?

samredai (approved):
Thanks @shangyian! LGTM but I left a couple suggestions.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -505,6 +506,127 @@ def is_available(
...
+    """
+    The direction of the metric that's considered good, i.e., higher is better
+    """
+
+    HIGHER_IS_BETTER = "higher_is_better"
+    LOWER_IS_BETTER = "lower_is_better"
+    NEUTRAL = "neutral"
+
+
+class MetricKind(str, enum.Enum):
```
agorajek, line 519: Have you considered `MetricType` ? Kind sounds weird, but maybe we'll get used to it.

Also, I feel a better the default would be `GENERIC` instead of `UNSPECIFIED`, but its your choice.

datajunction-server/datajunction_server/models/node.py
```diff
@@ -505,6 +506,127 @@ def is_available(
...
+
+    # A ratio is one quantity divided by another quantity, with the two sharing the same units
+    RATIO = "ratio"
+
+    # A rate is like ratio, but with different units of measure
+    RATE = "rate"
+
+    # A proportion is like ratio, but it compares a part to a whole.
+    # The possible values range from 0 to 1
+    PROPORTION = "proportion"
```
agorajek, line 538: This is not much different than ratio. In general I think we should start really small here and add new items when it is clear we need them.

agorajek (approved):
Thanks @shangyian. I left few semi-philosophical questions in line. But we shouldn't debate them for too long.

datajunction-ui/src/app/pages/AddEditNodePage/index.jsx
```diff
@@ -242,6 +259,19 @@ export function AddEditNodePage() {
...
+  // Get metric metadata values
+  useEffect(() => {
+    const fetchData = async () => {
+      const directions = await djClient.listMetricMetadata.Direction();
+      setMetricDirections(directions);
+      const units = await djClient.listMetricMetadata.Unit();
+      setMetricUnits(units);
+      const kinds = await djClient.listMetricMetadata.Kind();
+      setMetricKinds(kinds);
+    };
```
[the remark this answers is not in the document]
shangyian: Good point! I moved them all to a single endpoint

datajunction-ui/src/app/pages/AddEditNodePage/index.jsx
shangyian: Makes sense, I added a blank option to each of these, so that it clearly indicates it's optional. Also added asterisks to the required fields

datajunction-server/datajunction_server/models/node.py
shangyian: So I ended up removing `MetricKind` altogether, because it seems like the values captured here could be encapsulated in `MetricUnit` instead (i.e., percentage or proportion). There are also too many values here, making it hard to reason about what this field is meant to be.

Issue #821, opened 2023-10-16
Snowflake Query Service


Geo881:
- [ ] Add snowflake to the engine type enum
- [ ] Update the json schema to validate snowflake specific parameters 
- [ ] Use snowflake client with the user configured parameters to query a snowflake warehouse
- [ ] Convert results to a DJ query results object
- [ ] Mock snowflake python client and write tests

PR #834, opened 2023-10-21
adding Snowflake support for DJQS

Geo881:
### Summary

This adds snowflake support for the OSS query service.
Below is an example of adding a snowflake warehouse with a snowflake engine configured.

```yaml
engines:
  - name: duckdb
    version: 0.7.1
    type: duckdb
    uri: duckdb:////code/docker/default.duckdb
    extra_params:
  - name: snowflake
    version: 1.0
    type: snowflake
    uri: snowflake:////
    extra_params:
      user: <USER HERE>
      account: <ACCOUNT NAME HERE>
      database: <DATABASE NAME HERE>
catalogs:
  - name: warehouse
    engines:
      - duckdb
  - name: snowflake
    engines:
      - snowflake
```
Some follow ups that are needed:
- Selecting the engine for catalogs with multiple engines is not exposed everywhere. ie the SQL page in the UI
-  The engine has to be defined twice, currently once in the query service and once in the main service. It seems like we can just define this once in the query service and then the main service can call the query service to discover the available engines for a given catalog.

### Test Plan

I loaded the Roads example into a snowflake warehouse and configured the query service to connect with it. I started up a docker compose environment and tested it out in the UI.

- [x] PR has an associated issue: #821
- [x] `make check` passes

### DIFF
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -98,10 +98,8 @@ def run_query(
         return run_duckdb_query(query, conn)
     if engine.type == EngineType.SNOWFLAKE:
         conn = snowflake.connector.connect(
-            user=engine.extra_params["user"],
-            account=engine.extra_params["account"],
+            **engine.extra_params,
             password=os.getenv("SNOWSQL_PWD"),
-            database=engine.extra_params["database"],
         )
         cur = conn.cursor()
 
@@ -143,7 +141,7 @@ def run_duckdb_query(
 
 def run_snowflake_query(
     query: Query,
-    cur,
+    cur: snowflake.connector.cursor.SnowflakeCursor,
 ) -> List[Tuple[str, List[ColumnMetadata], Stream]]:
     """
     Run a query against a snowflake warehouse
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -3,10 +3,12 @@ Query related functions.
 """
 
 import logging
+import os
 from datetime import datetime, timezone
 from typing import List, Tuple
 
 import duckdb
+import snowflake.connector
 import sqlparse
 from sqlalchemy import create_engine, text
 from sqlmodel import Session, select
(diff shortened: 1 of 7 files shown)


datajunction-query/djqs/engine.py
```diff
@@ -94,6 +96,17 @@ def run_query(
             )
         )
         return run_duckdb_query(query, conn)
+    if engine.type == EngineType.SNOWFLAKE:
+        conn = snowflake.connector.connect(
+            user=engine.extra_params["user"],
+            account=engine.extra_params["account"],
+            password=os.getenv("SNOWSQL_PWD"),
+            database=engine.extra_params["database"],
+        )
```
samredai, line 105: There are a number of [other parameters](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-example) to this `connect(...)` function. I wonder if this would be better if we just use unpacking here to allow people to define any of them. We should still plug-in the password from this `SNOWSQL_PWD` environment, so I'm thinking something like:

```py
    if engine.type == EngineType.SNOWFLAKE:
        conn = snowflake.connector.connect(
            **extra_params,
            password=os.getenv("SNOWSQL_PWD"),
        )
```

datajunction-query/tests/api/queries_test.py
```diff
@@ -593,3 +593,75 @@ def test_submit_duckdb_query(
...
+
+
+@mock.patch("djqs.engine.snowflake.connector")
+def test_submit_snowflake_query(
+    mock_snowflake_connect,
+    session: Session,
+    client: TestClient,
+) -> None:
+    """
+    Test submitting a Spark query
```
samredai, line 605: This docstring needs to be updated to "Test submitting a Snowflake query"

samredai (approved):
Thanks @Geo881, this is awesome! Overall looks good to me but I left a few comments/suggestions. We also need to add docs describing how to configure this--in fact the DJQS docs in general are outdated. I think it's ok to tackle that in a follow-up PR.

datajunction-query/djqs/engine.py
```diff
@@ -128,6 +141,20 @@ def run_duckdb_query(
     return output
 
 
+def run_snowflake_query(
+    query: Query,
+    cur,
```
shangyian, line 146: minor: can we get the type of this cursor here?

shangyian (approved):
Nice!! Thanks for adding this

datajunction-query/djqs/engine.py
samredai: +1, `cur: snowflake.connector.cursor.SnowflakeCursor`

agorajek:
Nice addition! Thanks @Geo881

datajunction-query/djqs/engine.py
Geo881: @samredai I think this is cleaner and more flexible, thanks for for the suggestion. I will make the changes.

datajunction-query/djqs/engine.py
Geo881: @shangyian definitely, thanks for pointing that out.

Issue #420, opened 2023-03-24
Parse contexts that need to be handled by the DJ AST visitor


samredai:
There are a few remaining parse contexts that still need to be handled by the DJ AST visitor. Adding them here so we don't lose track of these.

- [ ] SubscriptContext
- [ ] BooleanLiteralContext
- [ ] ArithmeticUnaryContext
- [ ] ExtractContext
- [ ] AliasedRelationContext
- [ ] SetConfigurationContext (maybe?)
- [ ] TimestampaddContext
- [ ] TimestampdiffContext
- [ ] SubstringContext
- [ ] PositionContext
- [ ] CurrentLikeContext

The following are probably things we won't allow in node queries so we just need to catch them and raise a clear error:
- [ ] DropTableContext
- [ ] CreateViewContext
- [ ] ShowPartitionsContext
- [ ] DmlStatementContext (insert, delete, update, etc)
- [ ] ManageResourceContext (hive-like ADD statements)

shangyian:
FYI -- just added support for `BooleanLiteralContext` while in the middle of getting the type inference tests to work.

shangyian:
Added `SubscriptContext`. This is a lot cleaner with the new ast!

PR #835, opened 2023-10-23
Add support for inline tables using VALUES + misc bugfixes

shangyian:
### Summary

This change adds support for inline tables using `VALUES (...)` by implementing a visitor for `InlineTableContext`:

```
SELECT
  w.col1 AS one,
  w.col2 AS two
FROM VALUES
  ('1', 1),
  ('11', 1) AS w
```

It also fixes a few bugs:
*  Subscripts should allow for accessing struct types (i.e., `somestruct['abcd']` will access the `abcd` field on `somestruct`)
* `MIN` should work on timestamps and date types
* `TINYINT` and `SMALLINT` should be exposed as DJ types (otherwise `CAST(... AS TINYINT)` won't work)

### Test Plan

- [x] PR has an associated issue: #420 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -245,7 +245,9 @@ def create_node_revision(
     ]
     node_revision.required_dimensions = node_validator.required_dimensions
     new_parents = [node.name for node in node_validator.dependencies_map]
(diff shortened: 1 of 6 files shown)


agorajek (approved):
Wow, so cool to see this SQL feature added.

samredai (approved):

Issue #836, opened 2023-10-23
We should remove the flag that allows disabling authentication


samredai:
We need to remove the flag for disabling authentication. This way we can switch the type of the injected `current_user` object to be `User` instead of `Optional[User]`. Basic auth is now supported in the demo environment including with a signup page.

If for some reason someone absolutely wanted to disable authentication, they can do so by creating and providing their own implementation of the `DJHTTPBearer` injected dependency that does something like hard codes a global user.

PR #838, opened 2023-10-24
Remove disabling auth and set created_by for nodes, namespaces, and tags

samredai:
### Summary

---

**DISCLAIMER**: Once this PR is merged, there could be implications to some custom oauth flows, specifically if you do not save the DataJunction user object for users coming through oauth. As nodes, namespaces, and tags are created, it's expected that the user object passed in through the auth plugin has a user ID that exists in the datajunction metadata database and can be directly linked to the entities.

---

This PR does two main things:
- Removes the ability to disable auth, making `current_user` no longer optional
- Sets a `created_by` parameter when nodes, namespaces, and tags are created that links to a user by their ID

A few extra things in this PR:
- Updated the demo database
- Updated the roads notebook to use a requests session so that the login cookies can be used to authenticate the requests
- Add an optional username and password for the DJClient init for servers that use basic auth (optionally a `DJ_PWD` env var for the password)

### Test Plan

Wrote tests and also ran `docker compose --profile=demo up` to test running the notebook and validating that `created_by` is being set in the metadata database.

- [x] PR has an associated issue: #836 
- [x] `make check` passes

### Deployment Plan

N/A


datajunction-server/alembic/versions/2023_10_23_2051-363991e344b8_add_created_by_for_nodes_namespaces_and_.py
```diff
@@ -0,0 +1,61 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "363991e344b8"
+down_revision = "58e246fba34b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
shangyian, line 22: Should this migration include a stage that will populate the existing nodes' `created_by` fields with the username extracted from history events? That would make the info more complete where possible

datajunction-clients/python/datajunction/models.py
```diff
@@ -121,6 +122,7 @@ class Tag(BaseModel):
     name: str
     display_name: str
     tag_type: str
+    created_by: int = Field(foreign_key="users.id")
```
shangyian, line 125: Should this be a `created_by_id: int`, and then there be an additional Relationship model like:
```
created_by: User = Relationship(
    sa_relationship_kwargs={
        "lazy": "joined",
    },
)
```

This would make it so that the user object can be accessed via `created_by`

shangyian (approved):
Overall LGTM! I left a few suggestions in the comments, but they're more nice-to-haves

datajunction-server/alembic/versions/2023_10_23_2051-363991e344b8_add_created_by_for_nodes_namespaces_and_.py
agorajek: +1. I didn't realize that steps like that could be included in the migration itself. But this makes total sense, because this step could just be a SQL update statement.

agorajek (approved):
This means business from now on :)

datajunction-clients/python/datajunction/models.py
samredai: Oh yeah, that's definitely better, I'll update this

datajunction-server/alembic/versions/2023_10_23_2051-363991e344b8_add_created_by_for_nodes_namespaces_and_.py
samredai: That's a great idea, let me try that out 👀

samredai: @shangyian take a look when you have a chance. I insert a new user `"unknown"` as part of the upgrade and backfill the `created_by` values everywhere with that user ID. Is this what you imagined?

shangyian: @samredai this looks good to me! I guess another option is we could look at each node's history events and extract the user from there, but that might be too smart, since that user won't ever be able to log in.

samredai: Ah right, I'm re-reading your first comment where you mentioned using the history events. Good point about how they wouldn't be able to login. Hopefully this is an easy cleanup of existing nodes where system admins can just adhoc reassign all of these "unknown" users based on the node history and then new creations are properly assigned.

agorajek: Looks good imo. And we should be ok for now, but I think we should make sure our integration test captures some of the actions that are affected here.

samredai:
Closing this since this has been implemented in #1101 and #1113

PR #840, opened 2023-10-25
add in memory cache with configurable interval

samredai:
### Summary

This adds caching to the `GET /nodes/details/` endpoint which is loaded by the UI on every page load to build the client side search index. The cache interval is configurable and set to a default of 60 seconds.

In the local demo environment which just has a local sqlite database for the metadata store, this decreased the time of the end to end API call from about 200ms to <1ms (with the 200ms call happening globally every 60 seconds as the cache is refreshed).

In production environments making calls to a remote metadata database, we see about a 300-500ms end to end network call so reducing that to <1ms will be great. Let me know what you all think about the default cache interval of 60 seconds. This penalty for longer cache windows is that it will take longer for a new node to show up in search after it's created.

### Test Plan

Tested out the UI and monitored network activity using chrome dev tools.

- [ ] `make check` passes

### DIFF
--- a/datajunction-clients/python/pdm.lock
+++ b/datajunction-clients/python/pdm.lock
@@ -6,7 +6,7 @@ groups = ["default", "pandas", "test"]
 cross_platform = true
 static_urls = false
 lock_version = "4.3"
-content_hash = "sha256:9cb6487edffc637dd1b8b893fa7a07c736acacef75539490aeedb8cd8b556f3a"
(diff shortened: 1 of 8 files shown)


shangyian (approved):
Nice!

agorajek (approved):
This is awesome, but just wondering if we cold start with 10sec instead of a minute. I feel like that would be nicer user experience.

PR #841, opened 2023-10-27
Fully qualified names on transform sql + additional column metadata

shangyian:
### Summary

cc @matthewwardrop 

#### Fully Qualified Column Names

When generating SQL for any node (source, transform, metric etc), make sure that the generated column names are fully qualified names (i.e., includes the referenced node name and the column name). This makes it so that we can easily disambiguate where a particular dimension column comes from and is important if there are cases where the starting node has a column name that's the same as a dimension column that's brought in, or where there are multiple dimension attributes with the same column name but sourcing from different dimension nodes.

#### Additional Column Metadata

We can have better column metadata that provides a mapping between the generated SQL column names and the referenced dimension or metric columns. For example:

```
[
    {
        "name": "default_DOT_dispatcher_DOT_company_name",
        "column": "company_name",
        "node": "default.dispatcher",
        "type": "string",
    },
    {
        "name": "default_DOT_hard_hat_DOT_city",
        "column": "city",
        "node": "default.hard_hat",
        "type": "string",
    },
    {
        "name": "default_DOT_municipality_dim_DOT_local_region",
        "column": "local_region",
        "node": "default.municipality_dim",
        "type": "string",
    },
    {
        "name": "default_DOT_num_repair_orders",
        "column": "default_DOT_num_repair_orders",
        "node": "default.num_repair_orders",
        "type": "bigint",
    },
]
```

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/models/query.py
```diff
@@ -51,6 +51,17 @@ def __hash__(self):
         return hash((self.name, self.type))
 
 
+class ColumnMetadataDetails(BaseSQLModel):
```
agorajek, line 54: Why not just replace the above `ColumnMetadata` or why not simply add the `column` and `node` attributes to `ColumnMetada`?

shangyian: I thought about that, but in some cases it's unnecessary. For example, if we're generating materialization and storing config, we only need the column name and the column type, since that's just consumed internally by DJ and not by users.

agorajek: I see, but these fields could be optional right? Maybe it is better to have a separate class but I don't see it yet.

shangyian: @agorajek on second thought, that's a good point! I moved the fields and made them optional on the `ColumnMetadata` class.

agorajek (approved):
Awesome!

Issue #842, opened 2023-10-27
Add ability to show the diff for two source node revisions.


agorajek:
And add it to the Python client.

We also need to expose the `refresh_source_node` API in Python client as well.

PR #844, opened 2023-10-28
Open up source node refresh in Python client.

agorajek:
### Summary

This is a 1st PR in a small series to build a source node monitoring job. This job will refresh the source nodes and will show its owner the impact that the change in their node has on the DJ dag. We can debate if such a change should be applied immediately or not, but for let's assume it will, because that is the reality and we don't want DJ dag to be based on wishful thinking.

### Test Plan

Unit tests.

- [ ] PR has an associated issue: #842 
- [ ] `make check` passes

### Deployment Plan

nah


shangyian (approved):
LGTM!

PR #845, opened 2023-10-28
Filter Pushdowns + Optimized Dimension Joins

shangyian:
### Summary

* **Filter pushdowns:** Pushdown filters to the upstream transforms (based on linked dimensions) where possible. This means that if the downstream transform has a filter set, and the upstream transform has that filter's dimension attribute linked on an FK column, it'll try to filter the upstream transform where possible
* **Optimized dimension joins:** If a user requests a dimension node's primary key column, don't join the dimension node into the generated SQL, since it's unnecessary and we can just use the original node's foreign key column

### Test Plan

Tested locally

- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -278,6 +278,9 @@ def _build_tables_on_select(
     tables: Dict[NodeRevision, List[ast.Table]],
     memoized_queries: Dict[int, ast.Query],
     build_criteria: Optional[BuildCriteria] = None,
+    filters: Optional[List[str]] = None,
+    dimensions: Optional[List[str]] = None,
+    access_control=None,
 ):
     """
     Add all nodes not agg or filter dimensions to the select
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/construction/build.py
```diff
@@ -287,16 +291,49 @@ def _build_tables_on_select(
             Optional[ast.Table],
             _get_node_table(node, build_criteria),
         )  # got a materialization
+        fk_column_mapping = {
+            ",".join([pk.name for pk in col.dimension.current.primary_key()]): col
```
agorajek, line 295: This comma separated list of names... would it be safer to sort it by name or does the `primary_key()` method returns them in order?

datajunction-server/tests/api/sql_test.py
```diff
@@ -756,6 +812,7 @@ def test_sql_with_filters(  # pylint: disable=too-many-arguments
         params={"dimensions": dimensions, "filters": filters},
     )
     data = response.json()
+    print("SQLLL", data["sql"])
```
agorajek, line 815: testing leftover?

datajunction-server/tests/api/sql_test.py
```diff
@@ -270,6 +270,60 @@ def test_sql(
...
+                  default_DOT_long_events.event_id default_DOT_long_events_DOT_event_id,
+                  default_DOT_long_events.event_latency default_DOT_long_events_DOT_event_latency
+                FROM (
+                  SELECT default_DOT_event_source.country,
+                          default_DOT_event_source.device_id,
+                          default_DOT_event_source.event_id,
+                          default_DOT_event_source.event_latency
+                  FROM logs.log_events AS default_DOT_event_source
+                    WHERE default_DOT_event_source.event_latency > 1000000
+                     AND default_DOT_event_source.country = 'ABCD'
```
agorajek, line 296: Nice!

agorajek (approved):
One small question and typo. But the PR looks great!

One small thing to keep in mind is that in a certain scenario, pushing the filter up to the dimension node may actually be breaking a SQL logic, but only if the downstream SQL transforms the values of the filtered dimension attributes, which should be rare, and we can cross that bridge when we get there.

datajunction-server/datajunction_server/construction/build.py
shangyian: Good point, I'll add sorting by name. (Not that we support multi-column primary keys at the moment...)

PR #848, opened 2023-10-31
Measures SQL

shangyian:
### Summary

This PR adds an endpoint `GET /sql/measures` that returns the measures SQL for a set of metrics with dimensions and filters. This SQL can be used to produce an intermediate table where the full, non-aggregated measures and dimensions data are needed. In a follow-up PR, I'll be using this functionality to produce the intermediate table needed for loading into an analytics database (e.g., Druid). 

An example query to this endpoint:
```
GET /sql/measures?metric=default.num_repair_orders&metric=default.avg_length_of_employment&dimension=default.hard_hat.state
```

The endpoint also returns columns with semantic types:
```
[
    {
        "column": "discount",
        "name": "default_DOT_repair_orders_fact_DOT_discount",
        "node": "default.repair_orders_fact",
        "semantic_type": "measure",
        "type": "float",
    },
    {
        "column": "price",
        "name": "default_DOT_repair_orders_fact_DOT_price",
        "node": "default.repair_orders_fact",
        "semantic_type": "measure",
        "type": "float",
    },
    {
        "column": "dispatcher_id",
        "name": "default_DOT_dispatcher_DOT_dispatcher_id",
        "node": "default.dispatcher",
        "semantic_type": "dimension",
        "type": "int",
    },
]
```

cc @matthewwardrop 

#### duckdb 

I changed the various `/sql` and `/data` endpoint tests to actually run the generated SQL on the roads tables in duckdb. This makes the tests more robust, since eyeballing that the generated SQL looks right isn't as confidence-inducing as just running the queries against an actual database.

#### roads changes

I added a new intermediate transform `default.repair_orders_fact` that stitches together the `repair_orders` and `repair_order_details` tables, and switched some of the metrics to source from this fact transform. This made it easier to test metrics that spanned multiple parent transforms. For example, there are a set of metrics that source from `default.repair_orders_fact` (i.e., `default.num_repair_orders`) and a set of metrics that source from `default.hard_hat` (`default.avg_length_of_employment`), but share the same dimensions, and this set up made it easier to requesting measures SQL across parent transforms.

- [ ] `make check` passes


samredai (approved):
🚀

matthewwardrop:
Thanks for putting this together Yian! To close the loop from our discussion, we need to identify the semantic entity associated with the column, and not just the transform node that actually served the data. For example, I could imagine the following payload:

```
[
   {
        "name": "default_DOT_dispatcher_DOT_dispatcher_id",
        "node": "my.dispatchers",
        "column": "id",
        "semantic_type": "dimension",
        "semantic_entity": "default.dispatcher.dispatcher_id",
        "type": "int",
    },
...
]

File: datajunction-server/datajunction_server/api/sql.py [hunk 1]
```diff
@@ -25,6 +26,37 @@
...
+    validate_access: access.ValidateAccessFn = Depends(  # pylint: disable=W0621
+        validate_access,
+    ),
+) -> TranslatedSQL:
+    """
+    Return the measures SQL for a set of metrics with dimensions and filters.
+    This SQL can be used to produce an intermediate table with all the measures
+    and dimensions needed for an analytics database (e.g., Druid).
+    """
+    return get_measures_query(
```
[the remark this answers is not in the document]
shangyian: Agreed that this might be useful! I'm going to hold off on it now and add this in a follow-up PR.

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -702,6 +712,47 @@ def build_node(  # pylint: disable=too-many-arguments
...
+            projection.append(expression)
+        else:
+            expression = expression.copy()
+            projection.append(expression)  # type: ignore
+    built_ast.select.projection = projection
+
+    if built_ast.select.where:
+        for col in built_ast.select.where.find_all(ast.Column):
+            if hasattr(col, "alias"):  # pragma: no cover
+                col.alias = None
```
agorajek, line 744: Why do we need to cleanup the aliases in WHERE and GROUP BY?

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -1094,3 +1156,207 @@ def build_ast(  # pylint: disable=too-many-arguments
...
     _logger.info("Finished building query in %s", end - start)
     return query
+
+
+def metrics_to_measures(
+    session: Session,
+    metric_nodes: List[Node],
+) -> DefaultDict[str, Set[str]]:
+    """
+    Returns a mapping between each metric node and a list of necessary measures for that metric.
```
agorajek, line 1166: If we solidify the concepts of measures and dimension attributes as semantic types then this will likely become `metrics_to_semantic_types()`.

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -98,7 +98,15 @@ def check_convergence(path1: List[str], path2: List[str]) -> bool:
         if partial1 == partial2:
             return True
 
-    return False
+    # TODO: Once we introduce dimension roles, we can remove this.  # pylint: disable=fixme
```
agorajek, line 101: Yeah... let's make a call on this soon.

agorajek (approved):
@shangyian super cool!

datajunction-server/datajunction_server/construction/build.py
CircArgs: @agorajek https://github.com/DataJunction/dj/blob/347abfd961bab9aab3f273aedcbe262cd2e24e16/datajunction-server/datajunction_server/sql/parsing/ast.py#L1005 aliases get appended when SQL is rendered. There's no check on the ancestor nodes to determine if the alias would be valid.

datajunction-server/datajunction_server/sql/dag.py
shangyian: Yeah, this seems like an easy thing to add and it's really more for organizational purposes

PR #857, opened 2023-11-08
[bugfixes] Add default catalog for nodes without them, redundant dimensions should not be included

shangyian:
#### Support catalog-less dimension nodes

This is a bug that shows up when we use a non-SQLite database for metadata: SQLite doesn't enforce that foreign keys constraints must reference an actual value that exists in the referenced table, but other databases (Postgres, CockroachDB etc) do. Therefore, when using other databases for metadata, we end up being unable to link dimension nodes that don't have a catalog (i.e., they are pure SQL) to catalog-ed nodes. This fixes it by creating a default "unknown" catalog for these types of nodes.

#### Redundant dimensions

When asking for the shared dimensions between several metric nodes, we return columns that are tagged with type "dimension" and are linked out to a dimension node. However, this leaves us with redundant dimensions, where we return both the foreign key on a node and the referenced primary key. This change makes it so that we only return the primary key, which is typically how users will think about their dimensions referencing.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/catalogs.py
+++ b/datajunction-server/datajunction_server/api/catalogs.py
@@ -21,6 +21,8 @@ _logger = logging.getLogger(__name__)
 settings = get_settings()
 router = SecureAPIRouter(tags=["catalogs"])
 
+UNKNOWN_CATALOG_ID = 0
+
 
 @router.get("/catalogs/", response_model=List[CatalogInfo])
 def list_catalogs(*, session: Session = Depends(get_session)) -> List[CatalogInfo]:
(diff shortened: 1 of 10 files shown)


datajunction-server/datajunction_server/api/catalogs.py
```diff
@@ -21,6 +21,8 @@
 settings = get_settings()
 router = SecureAPIRouter(tags=["catalogs"])
 
+UNKNOWN_CATALOG_ID = 0
```
agorajek, line 24: Are you sure all the known catalogs get a positive integer id?

agorajek (approved):
Thanks!

datajunction-server/datajunction_server/api/catalogs.py
shangyian: So at the moment the id is an incremental id set by the database and it starts at 1. This should be true for both SQLite and Postgres. So unless someone is manually creating a catalog entry and setting the id to 0, we should be safe.

PR #859, opened 2023-11-10
Modify Druid materialization flow to use the measures query

shangyian:
### Summary

This PR switches the Druid materialization flow to use the measures query from the measures SQL endpoint. This allows us to remove logic around decomposing metrics into measures, as that was likely an over-optimization that complicates the Druid ingestion process.

This also provides a few additional useful outputs:
* Metric to measures mapping
* Metric expressions rewritten based on the intermediate measures output table

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build.py
+++ b/datajunction-server/datajunction_server/construction/build.py
@@ -731,10 +731,21 @@ def rename_columns(built_ast: ast.Query, node: NodeRevision):
             if expression.alias_or_name.name in node_columns:  # type: ignore
                 alias_name = node.name + SEPARATOR + expression.alias_or_name.name  # type: ignore
             expression = expression.copy()
+            expression.set_semantic_entity(alias_name)  # type: ignore
             expression.set_alias(ast.Name(amenable_name(alias_name)))
             projection.append(expression)
         else:
             expression = expression.copy()
+            if isinstance(expression, ast.Aliasable) and not isinstance(
+                expression,
+                ast.Wildcard,
+            ):
+                column_ref = expression.alias_or_name.identifier()
+                if column_ref in node_columns:  # type: ignore
+                    alias_name = node.name + SEPARATOR + column_ref  # type: ignore
+                    expression.set_semantic_entity(alias_name)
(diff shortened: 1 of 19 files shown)


datajunction-server/datajunction_server/construction/build.py
```diff
@@ -1161,21 +1181,31 @@ def build_ast(  # pylint: disable=too-many-arguments
...
             if col.table:  # pragma: no cover
-                metric_to_measures_mapping[col.table.dj_node.name].add(  # type: ignore
+                parents_to_measures[col.table.dj_node.name].add(  # type: ignore
+                    col.alias_or_name.name,
+                )
+                metric_to_measures_mapping[metric_node.name].add(
                     col.alias_or_name.name,
                 )
-    return metric_to_measures_mapping
+    return parents_to_measures, metric_to_measures_mapping
```
agorajek, line 1208: I agree we don't need the `_mapping` suffix. You may want to adjust the 2nd variable as well.

datajunction-server/tests/api/cubes_test.py
```diff
@@ -62,27 +62,27 @@ def test_read_cube(client_with_account_revenue: TestClient) -> None:
...
-          FROM accounting.account_type_table AS default_DOT_account_type_table)
-          AS default_DOT_account_type
-          GROUP BY  default_DOT_account_type.account_type_name
-        )
-        SELECT  default_DOT_account_type.default_DOT_number_of_account_types,
-            default_DOT_account_type.default_DOT_account_type_DOT_account_type_name
-         FROM default_DOT_account_type
-        """,
-    )
+    # assert compare_query_strings(
```
agorajek, line 65: You might have missed this spot to delete or uncomment.

agorajek (approved):
@shangyian the druid field names look so much cleaner now. Thank you!

datajunction-server/datajunction_server/construction/build.py
shangyian: good point, removed!

PR #860, opened 2023-11-14
Cube Dimension Values

shangyian:
### Summary

This PR adds two endpoints that allow for retrieving dimensions values for a cube. This is useful for cases where we need a list of cube-scoped dimensions values, often for building out a dashboard based on the cube.

These endpoints provide the ability to:
* Ask for all unique dimension values for each dimension attribute in the cube, as well as any combination of dimension attributes
* Ask for counts for each of the dimension values
* Ask for dimensions values given certain filters
* Limit the number of dimension values returned

Example:
```
curl -X 'GET' \
  'http://localhost:8000/cubes/default.repair_orders_cube/dimensions/data?dimensions=default.hard_hat.postal_code&dimensions=default.hard_hat.state&include_counts=true&async_=false' \
  -H 'accept: application/json'
```
Returns
```
{
  "dimensions": ["default.hard_hat.postal_code", "default.hard_hat.state"],
  "values": [
    {
      "value": ["33125", "MI"],
      "count": 25
    },
    {
      "value": ["37421", "NJ"],
      "count": 16
    },
    {
      "value": ["71730", "PA"],
      "count": 16
    },
    {
      "value": ["13440", "MA"],
      "count": 9
    },
    {
      "value": ["42001", "GA"],
      "count": 9
    },
    {
      "value": ["85021", "AZ"],
      "count": 4
    },
    {
      "value": ["27292", "CT"],
      "count": 4
    },
    {
      "value": ["74403", "OK"],
      "count": 1
    },
    {
      "value": ["14304", "NY"],
      "count": 1
    }
  ],
  "cardinality": 9
}
```

### Test Plan

Locally, with roads and other test data

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/cubes.py
+++ b/datajunction-server/datajunction_server/api/cubes.py
@@ -118,7 +118,10 @@ def get_cube_dimension_values(  # pylint: disable=too-many-locals
         validate_access,
     )
     if cube.availability:
-        catalog = get_catalog_by_name(session, cube.availability.catalog)  # type: ignore
+        catalog = get_catalog_by_name(  # pragma: no cover
+            session,
+            cube.availability.catalog,  # type: ignore
+        )
     else:
         catalog = cube.catalog
     query_create = QueryCreate(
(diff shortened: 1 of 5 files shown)


agorajek:
This is great @shangyian. 

But in the interim, would you consider adding an endpoint that given the same set of inputs, that is cube and dimension list gives you only the cardinality and perhaps min/max value (and maybe top n values - if you really want to "go crazy"). I think getting the stats w/o the data will prevent the users from getting accidental OOM. Once they know that cardinality is manageable then then can ask for the data. OK to add in later PRs imo.

agorajek (approved):

shangyian:
@agorajek So right now the default limit is set to 50K (and is controllable by setting the `limit` query param), so you can't get more than 50K dimensions if you request without `limit` being set. But if that's a bit too many, we can definitely drop the default to something like 100. The difficulty is that it's not clear who is consuming this - if it's an actual dashboard, I would imagine they want no more than 10, but a sensible default will differ for every use case. 

Another thing I think we should consider is setting a hard limit at X, after which we force the user to paginate rather than serving up all the data in a single request. And this goes for all endpoints, not just this one.

PR #864, opened 2023-11-21
Use an enum for semantic type

shangyian:
### Summary

Mostly just code cleanup -- instead of returning a string for the `semantic_type` column property, this creates a `SemanticType` enum for all the valid values and sets the type of the property to be the enum. 

### Test Plan

Ran unit tests to check that 

- [ ] `make check` passes


shangyian:
@agorajek Yeah, I think we probably won't need to, because when we're talking about dimensions, we seem to only return dimension attributes. Dimension nodes seem to almost function as a hierarchy.

PR #867, opened 2023-11-21
Exporting/downloading a namespace to a local YAML project

samredai:
### Test Plan

- [ ] `make check` passes


agorajek (approved):
This is great! Wrt to:
> Another follow-up feature is to identify these YAML managed definitions via server side metadata

Do you think we could plug the yaml project-root namespace metadata into the server side namespace? And this way the UI could simply check the namespace to decide if there is any yaml-project-root in its path.

samredai:
@agorajek that sounds right to me and for that reason we should probably allow optionally setting a git url in the project definition so we can detect if a deployment is just the same project being redeployed or if someone is starting a new YAML based project and just happens to use the same namespace. It'd also be useful to show this git url in the UI.

PR #868, opened 2023-11-22
Fix various bugs with SQL generation + data retrieval

shangyian:
### Summary

This PR fixes various bugs:
* Remove the dual pathway of generating SQL from the Python client (i.e., calling `/sql/{node_name}` vs `/sql?metrics=...`), The latter pathway is the most up-to-date, so the Python client has been updated to just use that endpoint for both SQL for a single metric and multiple metrics.
* Fix an issue where aliases show up in the `WHERE` clause of a generated query. There's probably a more generic way to fix this when assembling the final query string (i.e., in `__str__`), but I'm adding this fix in for now to unblock users.
* Fix an issue where `/data` doesn't use the same SQL as generated `/sql`, caused by not passing in the access control validator.

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -817,6 +818,7 @@ def build_sql_for_multiple_metrics(  # pylint: disable=too-many-arguments,too-ma
         session,
         metrics,
         dimensions,
+        check_dimensions=False,
```
agorajek, line 821: Why we don't need to "check dimensions" in a cube?

agorajek (approved):
Looks great! Just one question in-line.

datajunction-server/datajunction_server/api/helpers.py
samredai: @shangyian can correct me but it looks like this `check_dimensions` flag just checks that dimensions have been provided which is something we already check at the front of the API call logic. The logic here still validates the provided dimension attributes exist and are compatible with the metrics, regardless of the value for the `check_dimensions` arg. Maybe that arg could have a better name? Something like `dimensions_required`? Or maybe we should leave this check out completely from the `validate_cube` inner logic and leave it to the calling code to do the `if not dimensions:` check and raise if it needs to.

shangyian: Yeah, I agree that it should be renamed -- I changed it to `require_dimensions`. I opted not to remove it because it's called in five other places, and it's nicer not to repeat the logic that many times.

PR #870, opened 2023-11-30
Cube Builder UI

shangyian:
### Summary

Add a cube builder to the UI

https://github.com/DataJunction/dj/assets/9524628/13d22347-067b-45ee-8e51-6d4f992cb61b

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/models/node.py
```diff
@@ -1127,6 +1127,9 @@ class DimensionAttributeOutput(SQLModel):
     """
 
     name: str
+    node_name: Optional[str]
```
agorajek, line 1130: I think I agree with adding these attributes, but could you clarify your thinking for adding them here?

agorajek (approved):
So much new UI gears! Thanks @shangyian .

One question in line about the additional attributes of ColumnAttribute.

datajunction-server/datajunction_server/models/node.py
shangyian: So the main reason for adding them here is so that we can make a single call to retrieve common dimensions and get all the info we need to build this UI. Otherwise, we'll need to make multiple calls (as in, we can make a separate call for each dimension node and get their display names/PKs/etc), but that seems wasteful.

agorajek: Sounds good, so any changes on the dimension node should be reflected back here.

PR #875, opened 2023-12-12
Add explicit deps in client + display name to yaml import/export

shangyian:
### Summary

* Adds explicit python client dependencies on `rich` and `pyyaml`
* Adds `display_name` to the yaml import/export functionality

### Test Plan

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-clients/python/datajunction/compile.py
+++ b/datajunction-clients/python/datajunction/compile.py
@@ -52,25 +52,25 @@ def str_presenter(dumper, data):
 yaml.add_representer(str, str_presenter)
 
 
-def _parent_dir(path: Path):
+def _parent_dir(path: Union[str, Path]):
     """
     Returns the parent directory
     """
     return os.path.dirname(path)
 
 
-def _conf_exists(path: Path):
+def _conf_exists(path: Union[str, Path]):
     """
     Returns True if a config exists in the Path
     """
     return os.path.isfile(os.path.join(path, CONFIG_FILENAME))
 
 
-def find_project_root(dir: Optional[str] = None):
+def find_project_root(directory: Optional[str] = None):
     """
     Returns the project root, identified by a root config file
     """
-    checked_dir = dir or os.getcwd()
+    checked_dir = directory or os.getcwd()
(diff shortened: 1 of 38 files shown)


datajunction-clients/python/datajunction/compile.py
```diff
@@ -809,10 +819,7 @@ async def load_data(
                 definition=definition,
                 path=str(path),
             )
-    raise DJClientException(
```
shangyian, line 812: I removed this because in exporting/importing yaml files, they were often in directories that already had other, irrelevant and non-compatible yaml files. This is especially prevalent when running from Jupyter, which would create checkpoint yaml files that could be safely ignored.

datajunction-clients/python/datajunction/compile.py
```diff
@@ -409,11 +419,11 @@ class Project(BaseModel):
     mode: NodeMode = NodeMode.PUBLISHED
 
     @classmethod
-    def load_current(cls):
+    def load_current(cls, directory: Optional[str] = None):
```
shangyian, line 422: Added this largely because it isn't that easy to switch directories when running from a notebook. The only way to successfully do that is to restart Jupyter from a different directory, so seemed more convenient to have the ability to select which directory to load when invoking the client from a notebook.

agorajek (approved):

datajunction-clients/python/datajunction/compile.py
samredai: I added this to prevent overwriting local changes to a project by accidentally importing over it.

datajunction-clients/python/datajunction/compile.py
samredai: I imagined that would a different method we would eventually add. Load current project is meant to detect which project you currently are in and works if you are deeply nested within a sub-directory. What do you think of just making another method called `load_project(...)` that takes a directory path as an argument?

shangyian: @samredai agreed, that definitely makes more sense! I just added a new `load` class method

PR #876, opened 2023-12-12
Speed up getting shared dimensions 

shangyian:
### Summary

This change speeds up the `/metrics/common/dimensions` endpoint by switching to this logic:
1. Group the metrics by their immediate parents
2. Get the dimensions of each immediate parent node
3. Resolve shared dimensions between the parents

It also switches the get metrics call to a single database statement rather than one per metric.

### Test Plan

The existing unit tests for getting common dimensions should all pass without change.

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/metrics.py
+++ b/datajunction-server/datajunction_server/api/metrics.py
@@ -6,7 +6,6 @@ from http import HTTPStatus
 from typing import List, Optional
 
 from fastapi import Depends, HTTPException, Query
-from sqlalchemy.exc import NoResultFound
 from sqlalchemy.sql.operators import is_
 from sqlmodel import Session, select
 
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/metrics.py
```diff
@@ -103,31 +102,28 @@ async def get_common_dimensions(
...
-                )
-            metric_nodes.append(node)
-        except NoResultFound:
+    statement = (
+        select(Node)
+        .where(Node.name.in_(metric))  # type: ignore  # pylint: disable=no-member
+        .where(is_(Node.deactivated_at, None))
+    )
+    metric_nodes = session.exec(statement).all()
+    for node in metric_nodes:
```
agorajek, line 112: Is there a way to avoid this loop-check if we add to the query that we only want Metric nodes?
Ah... I guess this is a way to not fail-quietly and instead tell the user that they gave us a bad metric name.

agorajek (approved):
Nice fix. Just one small question inline.

samredai (approved):

datajunction-server/datajunction_server/api/metrics.py
shangyian: Yeah, I think the idea here was tell the user about all the nodes they supplied that aren't metrics.

Issue #878, opened 2023-12-14
Audit ORM-generated queries


shangyian:
We should audit the queries produced by the ORM as they aren't always the most efficient way to query the database. The following are a list of areas where we aren't using ORM object loading in an efficient way:

* `get_dimensions`: In the `get_dimensions` function, we make numerous calls to the database, at least one for each iteration of the for loop, when the entire function could likely be collapsed into a single recursive CTE.

TBD

PR #879, opened 2023-12-15
Make node graph retrieval more efficient with recursive CTE

shangyian:
### Summary

This change makes getting both node graphs (dimensions and data lineage) more efficient by switching from the current graph traversal approach to a single recursive CTE query. We are now able to get the full DAG for a node with three recursive CTE queries, defined in the following functions: `get_downstream_nodes`, `get_upstream_nodes`, and `get_dimensions`. 

The existing approach for getting the dimensions graph and upstream nodes causes us to issue dozens of calls to the database via retrieval of connected relational objects using the ORM for each layer of the graph traversal, which is inefficient. The new approach just retrieves all nodes in the graph through a single query. 

With this change the latency of the following endpoints drop significantly:
- `/metrics/common/dimensions` (one of the examples I tested with 30 metrics went from ~2s to <500ms)
- `/cubes/{name}`
- `/nodes/{name}/dag`
- `/nodes/{name}/upstreams`

For the dimensions graph CTE, this is the raw SQL query (which I started with and then translated to be SQLAlchemy-compatible):
```
WITH RECURSIVE dimensions_graph AS (
    SELECT
        initial_node.id AS path_start,
        c.name AS col_name,
        dimension_node.id AS path_end,
        initial_node.name || '.' || c.name || ',' || dimension_node.name AS join_path,
        dimension_node.name AS node_name,
        dimension_rev.id AS node_revision_id,
        dimension_rev.display_name AS node_display_name
    FROM noderevision initial_node
    JOIN nodecolumns nc ON initial_node.id = nc.node_id
    JOIN column c ON nc.column_id = c.id
    JOIN node dimension_node ON c.dimension_id = dimension_node.id
    JOIN noderevision dimension_rev 
    ON dimension_rev.version = dimension_node.current_version 
        AND dimension_rev.node_id = dimension_node.id
    WHERE initial_node.id=77

    UNION ALL

    SELECT
        ng.path_start,
        c.name AS col_name,
        next_node.id AS path_end,
        ng.join_path || '.' || c.name || ',' || next_node.name AS join_path,
        next_node.name AS node_name,
        next_rev.id AS node_revision_id,
        next_rev.display_name AS node_display_name
    FROM dimensions_graph ng
    JOIN node current_node ON ng.path_end = current_node.id
    JOIN noderevision current_rev ON current_rev.version = current_node.current_version AND current_rev.node_id = current_node.id
    JOIN nodecolumns current_columns ON current_columns.node_id = current_rev.id
    JOIN column c ON current_columns.column_id = c.id
    JOIN node next_node ON c.dimension_id = next_node.id
    JOIN noderevision next_rev ON next_rev.version = next_node.current_version AND next_rev.node_id=next_node.id
),
dims AS (
    SELECT * from dimensions_graph
)
SELECT 
    dims.node_name, 
    dims.node_revision_id, 
    dims.node_display_name,
    c.name,
    dims.join_path
FROM dims 
JOIN nodecolumns nc ON nc.node_id = dims.node_revision_id
JOIN column c ON nc.column_id = c.id
```

### Test Plan

- [x] PR has an associated issue: #878 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -4,16 +4,20 @@ DAG related functions.
 import itertools
 from typing import Dict, List, Optional, Set, Union
 
-from sqlmodel import Session, select
+from sqlalchemy import and_, or_, select
+from sqlalchemy.orm import aliased
+from sqlmodel import Session, join, select
 
-from datajunction_server.models import Column
+from datajunction_server.models import AttributeType, Column, ColumnAttribute
 from datajunction_server.models.base import NodeColumns
-from datajunction_server.models.node import DimensionAttributeOutput, Node, NodeRevision, NodeRelationship
+from datajunction_server.models.node import (
+    DimensionAttributeOutput,
+    Node,
+    NodeRelationship,
+    NodeRevision,
+)
 from datajunction_server.models.node_type import NodeType
 from datajunction_server.utils import SEPARATOR, get_settings
-from sqlmodel import join
-from sqlalchemy import or_, select, and_
-from sqlalchemy.orm import aliased
 
 settings = get_settings()
 
@@ -44,13 +48,13 @@ def get_dimensions_dag(
                 initial_node.id.label("path_start"),
                 c.name.label("col_name"),
                 dimension_node.id.label("path_end"),
-                (initial_node.name + '.' + c.name + ',' + dimension_node.name).label(
-                    "join_path"
+                (initial_node.name + "." + c.name + "," + dimension_node.name).label(
+                    "join_path",
                 ),
                 dimension_node.name.label("node_name"),
                 dimension_rev.id.label("node_revision_id"),
                 dimension_rev.display_name.label("node_display_name"),
-            ]
+            ],
         )
         .select_from(initial_node)
         .join(NodeColumns, initial_node.id == NodeColumns.node_id)
@@ -73,17 +77,15 @@ def get_dimensions_dag(
                 ng.c.path_start,
                 c.name.label("col_name"),
                 next_node.id.label("path_end"),
-                (ng.c.join_path + '.' + c.name + ',' + next_node.name).label(
-                    "join_path"
+                (ng.c.join_path + "." + c.name + "," + next_node.name).label(
+                    "join_path",
                 ),
                 next_node.name.label("node_name"),
                 next_rev.id.label("node_revision_id"),
                 next_rev.display_name.label("node_display_name"),
-            ]
-        )
-        .select_from(
-            ng
-            .join(current_node, ng.c.path_end == current_node.id)
+            ],
+        ).select_from(
+            ng.join(current_node, ng.c.path_end == current_node.id)
             .join(
                 current_rev,
                 and_(
@@ -100,8 +102,8 @@ def get_dimensions_dag(
                     next_rev.version == next_node.current_version,
                     next_rev.node_id == next_node.id,
                 ),
-            )
-        )
+            ),
+        ),
     )
 
     # Final SELECT statement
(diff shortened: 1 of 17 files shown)


agorajek:
Wow, that's a lot of tables to glue together but I agree this particular use case is good to rewrite. Because we need it often. Thanks @shangyian for doing this!

datajunction-server/datajunction_server/utils.py
```diff
@@ -52,12 +52,13 @@ def get_settings() -> Settings:
...
 
 
+@lru_cache(maxsize=None)
 def get_engine() -> Engine:
     """
     Create the metadata engine.
     """
     settings = get_settings()
-    engine = create_engine(settings.index)
+    engine = create_engine(settings.index, echo=True)
```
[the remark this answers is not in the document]
shangyian: Yep! Although I'm going to remove this before merging. Might be nice to push this up to an server config setting so that we can quickly enable it if we need to debug, although I think having the trace functionality integrated also solves that problem.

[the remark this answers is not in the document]
agorajek: Which "trace functionality" ?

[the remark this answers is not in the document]
shangyian: The opentelemetry sqlalchemy instrumentor will let us trace database queries. We only have that implemented internally though.

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -1,22 +1,244 @@
...
 from datajunction_server.models.node_type import NodeType
 from datajunction_server.utils import SEPARATOR, get_settings
 
 settings = get_settings()
 
 
+def get_dimensions_dag(
+    session: Session,
+    node_revision: NodeRevision,
+    attributes: bool = True,
```
agorajek, line 29: nit: `with_attributes` name would better match the `bool` value.

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -1,22 +1,244 @@
...
+                NodeRevision.display_name,
+                Column.name,
+                Column.type,
+                "join_path",
+            )
+            .where(NodeRevision.id == node_revision.id),
+        )
+    )
+
+    return sorted(
```
agorajek, line 214: nit: consider not sorting here. I think most clients will want to sort on their own, instead of depending on the output of some rest api.

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -1,22 +1,244 @@
...
+            DimensionAttributeOutput(
+                name=f"{row[0]}.{row[2]}",
+                node_name=row[0],
+                node_display_name=row[1],
+                is_primary_key=row[4] == "primary_key",
+                type=str(row[3]),
+                path=row[5].split(",")[:-1] if row[5] else [],
+            )
+            for row in session.exec(final_query).all()
+            if row[5] != ""
```
agorajek, line 225: Could you please add some comments here. At least to remind what `row[5]` is and why are you looking for empty string (instead of None).

agorajek (approved):
@shangyian awesome change. Added few small comments in-line.

datajunction-server/datajunction_server/sql/dag.py
shangyian: Good point! Changing this

datajunction-server/datajunction_server/sql/dag.py
shangyian: Agreed, that makes sense! Although this was part of the original code (likely to deal with tests), so I think I'll fix this in a separate PR to keep things a bit cleaner.

datajunction-server/datajunction_server/sql/dag.py
shangyian: Yep, I added some comments

Issue #880, opened 2023-12-18
Switch from SQLite to Postgres locally


agorajek:
To make sure that Postgres compatibility and performance is tested extensively, since we want to call it as our production DJ database.

shangyian:
One last thing needed here is to switch from SQLite in the unit tests over to Postgres. This will make it so that our unit tests are a bit closer to testing what will actually happen upon deployment with a Postgres metadata database. We can use testcontainers to bring up the Postgres image.

PR #882, opened 2023-12-20
Use postgres for metadata instead of sqlite

shangyian:
### Summary

This changes switches to postgres as the metadata db in the DJ demo setup, replacing sqlite. The motivation behind this change is so that we can actually tailor any performance tuning we do to a specific metadata database, since strategies that work for postgres may not work for sqlite (which will never be used in production).

### Test Plan

Locally + with tests.

- [ ] PR has an associated issue: #880 
- [x] `make check` passes

### Deployment Plan

N/A


samredai (approved):

agorajek (commented):
Looks good. Thanks @shangyian!

I forgot we can use docker for adding Postgres. Docker really simplifies software builds these days.

PR #883, opened 2023-12-22
Remove SQLModel and use SQLAlchemy directly

shangyian:
### Summary

This PR removes our dependency on SQLModel and just declares the database models with SQLAlchemy directly. I've moved all the database models to a separate `database` module, to keep them separated from the Pydantic models under `models`.

This change makes us less dependent on an infrequently updated library, and allows us to upgrade SQLAlchemy to 2.0+, which has better support for async database engines. 

SQLAlchemy 2.0+ allows us to implement the models with `Mapper`, which, along with Pydantic's `from_orm` method, covers most of the functionality that SQLModel was performing for us (i.e., easy conversion between SQLAlchemy and Pydantic). It's also cleaner to not mix the schemas of what's stored in the database with the input/output schemas of the API.

This also prepares us nicely for taking advantage of async database calls, since SQLAlchemy has first-class async support. 

### Test Plan

Locally + unit tests

- [x] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/api/attributes.py
```diff
@@ -121,18 +132,25 @@ def default_attribute_types(session: Session = Depends(get_session)):
...
     )
-    attribute_types = session.exec(statement).all()
+    attribute_types = session.execute(statement).scalars().all()
     for type_ in attribute_types:
-        updated_type = default_attribute_type_names[type_.name].dict(
-            exclude_unset=True,
-        )
-        type_ = type_.update(updated_type)
-        session.add(type_)
+        if type_:  # pragma: no cover
```
samredai, line 137: Why is this no cover needed if the code it replaced was covered?

datajunction-server/datajunction_server/api/health.py
```diff
@@ -44,12 +45,14 @@ async def database_health(session: Session) -> HealthcheckStatus:
             HealthcheckStatus.OK if result == (1,) else HealthcheckStatus.FAILED
         )
         return health_status
-    except Exception:  # pylint: disable=broad-except
-        return HealthcheckStatus.FAILED
+    except Exception:  # pylint: disable=broad-except  # pragma: no cover
```
samredai, line 48: Similar question here. Not a problem to use no cover when needed, I'm just curious why the change from sqlmodel -> sqlalchemy would remove this coverage since I don't see any related changes in the tests.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -278,13 +273,13 @@ def list_all_nodes_with_details(
     return [row for row in results if row.name in approvals]
 
 
-@router.get("/nodes/{name}/", response_model=NodeOutput)
+@router.get("/nodes/{name}/")
```
samredai, line 276: The response still looks like a `NodeOutput` instance so shouldn't `response_model=NodeOutput` be kept here?

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -389,6 +384,7 @@ def create_source(
     node = Node(
         name=data.name,
         namespace=data.namespace,
+        display_name=data.display_name or f"{data.catalog}.{data.schema_}.{data.table}",
```
samredai, line 387: I see the reason for doing this but I'm worried that it might be inconsistent if the user can set a custom display name. If we want to display `f"{data.catalog}.{data.schema_}.{data.table}"` for source nodes in the UI (which I think is a good idea), should we just build that concatenation client side?

datajunction-server/datajunction_server/database/user.py
```diff
@@ -0,0 +1,38 @@
...
+    GOOGLE = "google"
+
+
+class User(Base):  # pylint: disable=too-few-public-methods
+    """Class for a user."""
+
+    __tablename__ = "users"
+
+    id: Mapped[int] = mapped_column(
+        BigInteger().with_variant(Integer, "sqlite"),
```
samredai, line 28: Since we're exclusively using postgres now for the metadata db, do we need this with variant option for sqlite? Is that just for tests? Does it make sense to eventually have the tests use docker to start and init a new postgres instance?

datajunction-server/pyproject.toml
```diff
@@ -48,6 +47,7 @@ dependencies = [
     "types-cachetools>=5.3.0.6",
     "strawberry-graphql>=0.204.0",
     "fastapi-cache2>=0.2.1",
+    "psycopg2>=2.9.9",
```
samredai, line 50: Could we use `psycopg` here? That would get us v3 (https://pypi.org/project/psycopg/#history)

samredai (approved):
Thanks for doing this @shangyian, this is huge! I tried my best to go through it carefully. I left a few comments but nothing major. I'm happy that we're moving aggressively here. :)

datajunction-server/datajunction_server/api/attributes.py
shangyian: Oops, yeah, it is still covered and I just removed the `no cover` label.

datajunction-server/datajunction_server/api/health.py
shangyian: This one actually isn't covered -- I think it's related to how the unit test mocks away the execute: in the SQLModel case the mock caused `session.exec(...)` to raise an error, which would hit this except clause, but in SQLAlchemy, it doesn't, so it just returns a failed healthcheck.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Adding `response_model=NodeOutput` actually causes FastAPI to use the Pydantic model to parse the output twice, which fails because we have some custom logic on `NodeOutput` that "flattens" a node revision into the output model. This flattening logic should be removed, but I'm a bit wary of doing that here because this PR is already quite big.

shangyian: Adding `response_model=NodeOutput` actually causes FastAPI to use the Pydantic model to parse the output twice, which fails because we have some custom logic on `NodeOutput` that "flattens" a node revision into the output model. This flattening logic should be removed, but I'm a bit wary of doing that here because this PR is already quite big.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Hmm, I feel like it might even more confusing if we show a display name in the UI that doesn't match what's stored in the backend? Also this logic is actually just meant to replicate the existing functionality, where source nodes are automatically created with a display name of <catalog>.<schema>.<table> unless the user overrides it with a different display name. We can change this, but I think that's better off in a separate PR. And maybe we should make it so that users can never override the display name of a source node.

datajunction-server/datajunction_server/database/user.py
shangyian: Yeah, we should eventually have the tests run with postgres and not on top of sqlite. This is setup is indeed just for the unit tests.

datajunction-server/pyproject.toml
shangyian: Good point! Let me try rerunning the tests with `psycopg` v3

datajunction-server/datajunction_server/api/health.py
samredai: So is it safe to remove the try except?

datajunction-server/datajunction_server/api/nodes.py
samredai: Ok that's fair--agreed that's better left to a follow up PR

datajunction-server/datajunction_server/api/nodes.py
samredai: > But maybe we should make it so that users can never override the display name of a source node, since I'm not sure it makes sense.

This summarizes my concern well. The purpose of display name feels a bit ambiguous to me. If users can never set it and it's derived from other attributes, then it seems it should be handled in the UI code. If it's something the user is meant to set, then I don't think we should systematically set it at all.

I'm leaning towards source nodes not having a display name at all (or alternatively, it's always ignored), then in the UI wherever we are showing a display name we do a check of the node type. If it's a source node, we define it at runtime as `${catalog}.${schema}.${table}`

datajunction-server/datajunction_server/api/health.py
shangyian: I don't think we should because if the session.execute call fails, the healthcheck should fail rather than returning a 500. We can add a test case for this.

datajunction-server/datajunction_server/database/__init__.py
```diff
@@ -0,0 +1,29 @@
+"""All database schemas."""
```
agorajek, line 1: nit: are these database schemas as in namespaces or as in table structures? I think the latter but the phrase `database schemas` is confusing. I think `table schemas` would be clearer.

agorajek (commented):
@shangyian this looks like a lot of work to rewrite. The SQLAlchemy style looks good imo. But what's better is the knowledge that there is no 3rd library in the middle between our code and the database.

datajunction-server/datajunction_server/api/health.py
```diff
@@ -44,12 +45,14 @@ async def database_health(session: Session) -> HealthcheckStatus:
...
+        return HealthcheckStatus.FAILED  # pragma: no cover
 
 
 @router.get("/health/", response_model=List[HealthCheck])
-async def health_check(session: Session = Depends(get_session)) -> List[HealthCheck]:
+async def health_check(
+    session: Session = Depends(get_session),
+) -> List[HealthCheck]:
```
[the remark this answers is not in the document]
shangyian: If I remember right, this was originally meant to be multiple health checks, i.e., database health check, cache health check etc. But to be honest I don't think our healthchecks need to be this complicated.

shangyian:
re: source node display names, I also agree that source nodes should not have user-configurable display names.

> I am actually not sure if display_name is a good attribute for nodes at all. It does make sense for columns, if you want to pass them to the reporting side. But for nodes... hunch is that description is sufficient. Because I can't think of a place where you would want to show something else to the users... perhaps during selection of Cube nodes available.

@agorajek I think display names for nodes are actually quite useful and we use them when displaying the nodes (not just in the DJ UI, but in other UIs that need to display metrics). They're distinct from descriptions in that descriptions can be very long and explain the details of a particular node, but the display name is just a short, human-readable label.

Issue #853, opened 2023-11-03
Dimensions with Multiple Primary Key Columns


shangyian:
We need to support dimensions with multiple primary key columns.

This is actually more complex than expected due to our current data model for dimension linking. We currently link dimensions to a particular *column*, but if the dimension has multiple PK columns, linking a dimension to a single column is incomplete. One option is to check if a linked dimension has multiple PK columns, and enforce that the user needs to select all the matching columns before they can link the dimension.

PR #885, opened 2024-01-09
Support Complex Dimension Joins

shangyian:
### Summary

This PR adds support for more complex dimension joins, where the joins may have arbitrary custom SQL. The main changes include:
* Dimension link table schema and input/output models
* API endpoint to define a link between a given node and a dimension node
* Addition to the dimension join logic to use these explicitly defined dimension links, as well as those stored on the columns. We may want to consider joining these together so that the logic goes through a single path.
* Modify `get_dimensions` so that it takes into consideration dimensions from the custom SQL dimension links path

### Test Plan

- [x] PR has an associated issue: #853 
- [x] `make check` passes


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -709,6 +712,133 @@ def link_dimension(  # pylint: disable=too-many-arguments
...
+    # Parse the join query and do some basic verification of its validity
+    join_query = parse(f"SELECT 1 FROM {link_input.join_sql}")
+    exc = DJException()
+    ctx = ast.CompileContext(session=session, exception=exc)
+    join_query.compile(ctx)
+
+    # Check that there is a valid join clause
+    if not join_query.select.from_.relations[0].extensions:  # type: ignore
+        raise DJInvalidInputException(
+            f"Provided SQL `{link_input.join_sql}` does contain JOIN clause",
```
CircArgs, line 760: Does *_not_ contain?

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -270,14 +272,22 @@ def join_tables_for_dimensions(
...
-                join_asts = _build_joins_for_dimension(
-                    session,
-                    dim_node,
-                    initial_nodes,
-                    tables,
-                    build_criteria,
-                    required_dimension_columns,
-                )
+                if dim_node in dimension_links and dimension_links[dim_node]:
+                    join_query = parse(f"SELECT 1 FROM {dimension_links[dim_node]}")
```
CircArgs, line 276: This keeps coming up. It shouldn't be too difficult to have dedicated parsers for parts of queries at least the very high-level nodes. Is it worth it though

datajunction-server/datajunction_server/api/nodes.py
CircArgs: or is malformed or something right

CircArgs (approved):

datajunction-server/datajunction_server/api/nodes.py
shangyian: Ah yes, good catch!

datajunction-server/datajunction_server/construction/build.py
shangyian: Hmm, I think it's probably not needed for the time being, since there's a pretty easy workaround.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -709,6 +712,133 @@ def link_dimension(  # pylint: disable=too-many-arguments
...
+        link
+        for link in node.current.dimension_links
+        if link.dimension_id == dimension_node.id
+    ]
+    is_update = False
+
+    if existing_link:
+        # Update the existing dimension link
+        is_update = True
+        dimension_link = existing_link[0]
```
agorajek, line 792: I think this `...[0]` only makes sense if we assume (and validate somewhere) that a node can only be joined to a dimension once. But in reality this is not the case... 

What do you think about adding the "role" attribute to the DimensionLink object and using that the user indicator which join link we are talking about? 

I think it is ok to leave the "role" field as an optional string since most dim links would not need it but during validating we should check if a node has unique dim links as denoted by: dim_node_id + role. Then when we convert all the dimension links to this patter we would get our dim roles "for free".

datajunction-server/datajunction_server/database/dimensionlink.py
```diff
@@ -0,0 +1,61 @@
...
+        "Node",
+        foreign_keys=[dimension_id],
+    )
+
+    # SQL used to join the two nodes
+    join_sql: Mapped[str]
+
+    # Metadata about the join
+    join_type: Mapped[Optional[JoinType]]
+    join_kind: Mapped[JoinKind] = mapped_column(
```
agorajek, line 43: `join_cardinality` is better.

datajunction-server/datajunction_server/database/dimensionlink.py
```diff
@@ -0,0 +1,61 @@
...
+from sqlalchemy import JSON, BigInteger, ForeignKey, Integer
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from datajunction_server.database.connection import Base
+from datajunction_server.database.node import Node, NodeRevision
+from datajunction_server.models.base import sqlalchemy_enum_with_name
+from datajunction_server.models.dimensionlink import JoinKind, JoinType
+
+
+class DimensionLink(Base):  # pylint: disable=too-few-public-methods
```
agorajek, line 13: As I mentioned earlier `role` may be nice to add here.

agorajek (commented):
@shangyian I think I have only one major thing to consider before merging this in. Namely the "role" concept - please see an in-line comment for that. The rest are nits.

datajunction-server/datajunction_server/api/nodes.py
shangyian: This is a great point -- I had forgotten about the dimension roles issue that we discussed previously, and we can incorporate that nicely here. I will update to add the `role` field here.

shangyian:
@agorajek I added `role` and changed to `join_cardinality`. There's probably more interesting things we could do with dimension roles, but I'll save that for subsequent PRs.

agorajek (approved):
Looks good!

PR #894, opened 2024-01-18
Uppercase enums

shangyian:
### Summary

At the moment, the enums stored in the Postgres metadata database are a mixture of lowercase and uppercase values, depending on how the original enum was defined. The default SQLAlchemy enum mapper uses the Python enum's name to store into the database, which in our case is always uppercase, but the original sqlmodel setup stored some of our enums with uppercase values and some with lowercase values. This is now causing issues where Postgres will raise an error if the enum values don't match between cases. 

This PR migrates all database values to uppercase enums, for future consistency and to fix the aforementioned bugs.

### Test Plan

- [x] `make check` passes


agorajek (approved):
Thanks!

PR #901, opened 2024-01-26
Remove the mixture of async endpoints with a sync db connection

shangyian:
### Summary

We should not mix async API endpoints with a synchronous db connection (`psycopg3`). Because our database client does not support async, every database call will introduce blocking calls to the event loop, eventually resulting in the uvicorn workers timing out.

The issue is very similar to the one described here (in their case it was sync API call vs sync database query): https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker/issues/145#issuecomment-1442005465

### Test Plan

N/A

- [x] `make check` passes

### Deployment Plan

Ideally we deploy this right away. Otherwise over time the deployed DJ servers will experience more and more blocking calls, gradually causing workers to time out.

### DIFF
--- a/datajunction-server/datajunction_server/api/access/authentication/basic.py
+++ b/datajunction-server/datajunction_server/api/access/authentication/basic.py
@@ -24,7 +24,7 @@ router = APIRouter(tags=["Basic OAuth2"])
 
 
 @router.post("/basic/user/")
-async def create_a_user(
+def create_a_user(
     email: str = Form(),
     username: str = Form(),
     password: str = Form(),
@@ -63,7 +63,7 @@ async def create_a_user(
 
 
 @router.post("/basic/login/")
-async def login(
+def login(
     form_data: OAuth2PasswordRequestForm = Depends(),
     session: Session = Depends(get_session),
 ):
@@ -89,7 +89,7 @@ async def login(
 
 
 @router.post("/logout/")
-async def logout():
+def logout():
     """
     Logout a user by deleting the auth cookie
     """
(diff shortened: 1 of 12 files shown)


agorajek (approved):
Thanks @shangyian for catching this. @betodealmeida have you seen something like this before?

CircArgs:
@shangyian seeing as the bottleneck has been the orm and db calls, should DJ move to async engines and sessions? 

@samredai and I were working on something and came out to approximately:

```python
def get_engine() -> AsyncEngine:
    """
    Create the metadata engine.
    """
    ...

    def getconn():
        conn = connector.connect(instance_connection_name, "pg8000", user=db_user, password=db_pass, db=db_name)
        return conn

    engine = sqlalchemy.create_engine("postgresql+pg8000://", creator=getconn, future=True)
    return engine
```

as opposed to https://github.com/DataJunction/dj/blob/2df3427788418adfeaad515da501d81bd026e3a4/datajunction-server/datajunction_server/utils.py#L57

shangyian:
@CircArgs yeah, I was investigating this in December and started working on a PR for it, but the necessary changes are actually huge: https://github.com/DataJunction/dj/compare/main...shangyian:dj:use-asyncpg

CircArgs:
@shangyian hmm sqlalchemy strikes again 

> necessary changes are actually huge
...
> if you want to take a stab at it, go for it!

😨

shangyian:
> hmm sqlalchemy strikes again

@CircArgs yeah it's probably also that we weren't using it in the most efficient way... but nonetheless, here we are :( 

> 😨

Yeaaaah unfun

PR #903, opened 2024-01-28
Fix UI tags bug + speed up node DAG

shangyian:
#### Tags Bug

Fix UI bug where updating a node with tags would yield errors. This was due to an issue with setting the right tag values on the form when updating a node (it needs to be set to a list of tag name strings and not a list of objects). 

#### Node DAG 

Change node DAG retrieval query to use `selectinload` for columns rather than `joinedload`. This increases the retrieval speed for dimension nodes by a significant margin (in the tests I ran, the node DAG endpoint is 4-5x faster, yielding a much faster node graph display).

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

N/A but should deploy asap

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -5,7 +5,7 @@ import itertools
 from typing import Dict, List, Optional, Set, Union
 
 from sqlalchemy import and_, func, join, literal, or_, select
-from sqlalchemy.orm import Session, aliased, joinedload
+from sqlalchemy.orm import Session, aliased, joinedload, selectinload
 from sqlalchemy.sql.operators import is_
 
 from datajunction_server.database.attributetype import AttributeType, ColumnAttribute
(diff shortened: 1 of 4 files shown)


samredai (approved):
Thanks @shangyian!

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -30,7 +30,7 @@ def _node_output_options():
     """
     return [
         joinedload(Node.current).options(
-            joinedload(NodeRevision.columns).options(
```
agorajek, line 33: How did you even think of it to try?

agorajek (commented):
Thank you!

datajunction-server/datajunction_server/sql/dag.py
shangyian: @agorajek I read some guidance that indicated `joinedload` was for many-to-one relationships and `selectinload` was for one-to-many or many-to-many relationships. In the case of NodeRevision <-> Columns, it was actually one-to-many (disguised as many-to-many because of the intermediate join table), so I figured it made sense to try out `selectinload`. And it turned out to be much faster.

agorajek: Thanks @shangyian . That makes sense.

PR #905, opened 2024-01-30
Fix issue with search failing when it hits a null node description

shangyian:
### Summary

When someone searches and the search includes nodes with a null description, we get this error in the search UI:
```
Uncaught TypeError: Cannot read properties of null (reading 'length')
    at truncate (Search.jsx:154:16)
    at eval (Search.jsx:235:58)
    at Array.map (<anonymous>)
```

This is can be fixed by checking for null node descriptions and setting a sensible default.

### Test Plan

- [x] `make check` passes


samredai (approved):
Thanks @shangyian!

PR #906, opened 2024-01-30
Improve UI for primary key selection

shangyian:
### Summary

When users add or edit a node, they can specify the primary key of the node, but this is currently done by asking the user to input a comma-separated list of column names. This PR changes it so that we show a dropdown with the available columns as input options:
Screenshot 2024-01-29 at 8 46 09 PM

This makes it so that users will always input a valid primary key, since the columns they select will be confined to those that exist on the node. To achieve this, we do the following:
- Every time a node query gets updated during node creation or update, we save the updated query to the form's values
- When the user focuses on the primary key input, we call the backend api's `/validate` endpoint, which will return a list of columns based on parsing the latest version of the query. We only do this on focus to limit the number of times we call `/validate`, which is relatively expensive.

This sets us up nicely for a similar setup for the required dimensions (metrics only), where we'll want the user to pick from a set of available dimensions rather than freeform type dimensions.

https://github.com/DataJunction/dj/assets/9524628/3fd775c7-dad9-449e-a6db-600cdb01945e

### Test Plan

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-ui/src/app/pages/AddEditNodePage/FormikSelect.jsx
+++ b/datajunction-ui/src/app/pages/AddEditNodePage/FormikSelect.jsx
@@ -12,6 +12,7 @@ export const FormikSelect = ({
   style,
   className = 'SelectInput',
   isMulti = false,
+  onFocus = event => {},
 }) => {
   // eslint-disable-next-line no-unused-vars
   const [field, _, helpers] = useField(formikFieldName);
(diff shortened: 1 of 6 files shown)


CircArgs (approved):

agorajek:
Thanks @shangyian . This is great.

Issue #910, opened 2024-02-02
Redesign the metrics create/edit page


shangyian:
The current metrics create/edit page asks for metrics "SQL", a full SQL statement with both SELECT and FROM, with restrictions that the expression must be an aggregation that sources from a node, but this is often confusing for users. To eliminate confusion about how to structure the SQL, the form could just explicitly ask for:
- An aggregation expression 
- An upstream node

<img width="1223" alt="Screenshot 2024-02-02 at 8 55 15 AM" src="https://github.com/DataJunction/dj/assets/9524628/21e6986d-e3f2-4761-97f7-1434efe0d577">

PR #914, opened 2024-02-04
Alter metrics UI to use upstream + aggregate expression

shangyian:
### Summary

In the metric node create/edit UI, instead of asking for SQL, which is unclear from the user perspective, we instead explicitly ask for an upstream node and an aggregate expression (written in SQL). 

Given that we already have these restrictions in place for metric node SQL (i.e., it must source from one node and there must be an aggregation in the expression), we can just be explicit about these requirements in the UI rather than expecting users to structure it nicely into an ill-defined SQL blob. 

After the changes, this is what the new metric create/edit flow looks like:

Screenshot 2024-02-04 at 12 17 06 AM

The metric node info page also includes the split into upstream + agg expression:

Screenshot 2024-02-04 at 12 21 04 AM

This PR includes some additional changes that help streamline the structure of the add/edit node form page, primarily by splitting out various fields into separate components rather than keeping them in a single file. Some minor server-side changes were also made to support this on the `/metrics` endpoint.

### Test Plan

- [ ] PR has an associated issue: #910 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/models/metric.py
+++ b/datajunction-server/datajunction_server/models/metric.py
@@ -13,6 +13,7 @@ from datajunction_server.models.node import (
     MetricMetadataOutput,
 )
 from datajunction_server.models.query import ColumnMetadata
+from datajunction_server.sql.parsing.backends.antlr4 import parse
 from datajunction_server.transpilation import get_transpilation_plugin
 from datajunction_server.typing import UTCDatetime
 from datajunction_server.utils import get_settings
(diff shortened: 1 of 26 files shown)


agorajek:
Didn't review the code yet, but I just tried this locally and I think this is a great modification. 

Screenshot 2024-02-05 at 11 00 10 AM

Wondering what you think about the following:
1. Making the "FULL NAME *"  field's background feel different. It is a non editable field but its background makes it feel like you can click and edit it, which is confusing. I think that making the color matching the page background could fix that, but you may have a better solution.
2. Limiting the node display name to alphanumerics, space and underscore only. We should probably make this rule match the API somehow.

Both updates could be a separate PR but I think we should add them sooner than later.

agorajek (approved):
Looks good. Just added some comments about two other adjustments.

shangyian:
@agorajek Great points! I actually have a different PR I'm working on that will add in some better namespace/node name checks (i.e., only alphanumeric and _), so I'll tag on these naming related checks there.

PR #915, opened 2024-02-05
Required dimensions, node tab links, node name/namespace validation, 

shangyian:
#### Required Dimensions
- This adds a backend change to support editing required dimensions for a metric. Currently we only allow for setting required dimensions when creating a metric
- Also adds a UI component to set required dimensions when creating or editing a metric.

Screenshot 2024-02-05 at 1 25 54 PM

#### Node Tab Links

Makes each tab in the node info UI accessible under a separate URL, i.e.:
- http://localhost:3000/nodes/default.num_repair_orders/info
- http://localhost:3000/nodes/default.num_repair_orders/columns
- http://localhost:3000/nodes/default.num_repair_orders/graph
etc

#### Node Name/Namespace Validation

This adds a backend validation stage for namespaces to check that they only contain alphanumeric or underscore characters. It also adds a change in the UI so that the full node name does not include non-alphanumeric or underscore characters when it is derived from the display name. The full name field's background color has also been changed to indicate that it isn't modifiable.

Screenshot 2024-02-05 at 1 28 35 PM

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -233,30 +233,26 @@ def get_query(  # pylint: disable=too-many-arguments
 def find_bound_dimensions(
     validated_node: NodeRevision,
     dependencies_map: Dict[NodeRevision, List[ast.Table]],
-) -> Tuple[Set, List[Column]]:
+) -> Tuple[Set[str], List[Column]]:
     """
-    Finds the matched bound dimensions
+    Finds the matched required dimensions
     """
     invalid_required_dimensions = set()
     matched_bound_columns = []
+    required_dimensions_mapping = {}
     for col in validated_node.required_dimensions:
-        names = col.split(".")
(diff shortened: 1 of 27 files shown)


agorajek (approved):
This is great. Will save us ton of trouble.

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -233,30 +233,26 @@ def get_query(  # pylint: disable=too-many-arguments
 def find_bound_dimensions(
```
agorajek, line 233: Do you remember why we call it "bound" in this function name?

shangyian: @agorajek Yeah, I think that used to be an internal name for this, but then for all the user-facing parts we switched it to "required" dimensions. I do prefer required dimensions overall and might just switch them all to use that terminology for consistency.

PR #917, opened 2024-02-08
Switch to 'published' as the default in the node creation UI

shangyian:
### Summary

This PR switches the node creation UI to default to 'published' as the node mode rather than 'draft'. This is largely to avoid some less than ideal UX when someone creates a node with errors in draft mode. 

For example, nodes created in draft mode can have one or more errors that make the node unbuildable (or sometimes undisplayable). However, we don't currently have a mechanism to display what these errors are in the UI, making it difficult for users to fix these errors and promote the node to published. We should eventually add something in the UI that can show these errors and actually allow the use of draft mode in a nice way.

But before we get there, if by default we encourage users to create their nodes in published mode, then they'll get immediate, actionable feedback on errors like this:
Screenshot 2024-02-07 at 4 53 06 PM

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Wow, so many places...

shangyian:
@agorajek haha, luckily most of them were just tests

Issue #856, opened 2023-11-07
Add moving / renaming for nodes


shangyian:
Moving should entail switching the namespace a node is under by deleting the previous node and recreating a new one.

shangyian:
@agorajek I found this old issue and updated it with more details. We can discuss more tomorrow.

shangyian:
Allow for copying multiple nodes together. 
* If some of the nodes are downstream of other nodes, we'll modify the queries of the copied nodes to source from the copied nodes.
* All other attributes remain the same

```
POST /nodes/copy 
{
    "nodes": [
        {"name": "...", "new_name": "..."}
    ]
}
```

In the UI, we can allow users to select nodes they want to copy (including entire namespaces) and copy them to other namespaces/names.

shangyian:
Functionality needed:
* Ability to copy multiple nodes to new node names. The queries on any nodes that depend on each other should be modified during the copy to continue the dependency.
* Ability to rename multiple nodes. The queries on any nodes that depend on each other should be modified during the rename to continue the dependency.
* `cascade` will determine whether or not downstream nodes' queries are modified.

1. Find all nodes that need to be copied or renamed. If `cascade` is true, also find any nodes downstream of these nodes.
2. Build an object to track renaming (i.e., old node name -> new node name) + all the nodes where naming is needed.
3. Copy all the nodes. For any nodes where renaming is needed, parse the query and do the rename.
4. Finish.

PR #919, opened 2024-02-08
Add endpoint for copying nodes

shangyian:
### Summary

This PR adds an API endpoint for copying nodes to a new name:
`POST /nodes/{node_name}/copy?new_name={new_name}` 

The functionality implemented here will copy over most attributes on the original node, like dimension links and column attributes, as well as metric-specific attributes like required dimensions and metric metadata. However, it does not handle copying over materialization (a useful feature that we can build in a follow-up PR).

### Test Plan

Added a test that makes a copy of every node in the roads examples set.

- [x] PR has an associated issue: #856 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -1432,13 +1432,12 @@ def copy_node(
         include_inactive=True,
     )
     if existing_new_node:
-        # TODO: support overwriting deactivated nodes  # pylint: disable=fixme
         if existing_new_node.deactivated_at:
+            hard_delete_node(new_name, session, current_user)
+        else:
             raise DJInvalidInputException(
-                f"A deactivated node with name {new_name} already exists. "
-                "To copy successfully, hard delete the deactivated node first.",
+                f"A node with name {new_name} already exists.",
             )
-        raise DJInvalidInputException(f"A node with name {new_name} already exists.")
 
     # Copy existing node to the new name
     existing_node = get_node_by_name(session, node_name)
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -49,6 +49,7 @@ from datajunction_server.internal.access.authorization import (
 )
 from datajunction_server.internal.nodes import (
     _create_node_from_inactive,
+    copy_to_new_node,
     create_cube_node_revision,
     create_node_revision,
     get_column_level_lineage,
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1402,3 +1403,44 @@ def set_column_partition(  # pylint: disable=too-many-locals
...
+        new_name,
+        raise_if_not_exists=False,
+        include_inactive=True,
+    )
+    if existing_new_node:
+        # TODO: support overwriting deactivated nodes  # pylint: disable=fixme
+        if existing_new_node.deactivated_at:
+            raise DJInvalidInputException(
+                f"A deactivated node with name {new_name} already exists. "
+                "To copy successfully, hard delete the deactivated node first.",
```
agorajek, line 1439: Not sure if we want to tell our users about the hard-delete yet. What if we just do the hard delete for them here? Initially I was thinking that adding a `force` flag could be a good idea, but they could expect that such a flag would also work if there is an existing and active node on the other end. Hence I think we could just do this delete here, unless you think that a node can be deactivated accidentally. Which is the only case where I think this could be risky, but even then... there is a solution. We could simply copy that deactivated node to a dedicated "node cemetery".

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -385,6 +386,95 @@ def save_node(
...
+        parents=old_revision.parents,
+        missing_parents=[
+            MissingParent(name=missing_parent.name)
+            for missing_parent in old_revision.missing_parents
+        ],
+        columns=[col.copy() for col in old_revision.columns],
+        # TODO: availability and materializations are missing here  # pylint: disable=fixme
+        lineage=old_revision.lineage,
+    )
+    new_revision.dimension_links = [
```
agorajek, line 435: What's the reason `dimension_links` are set outside the previous call? This looks similar to the `missing_parents` but maybe I am missing something?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -385,6 +386,95 @@ def save_node(
...
+    new_revision.version = (
+        str(DEFAULT_DRAFT_VERSION)
+        if new_revision.mode == NodeMode.DRAFT
+        else str(DEFAULT_PUBLISHED_VERSION)
+    )
+    new_node.current_version = new_revision.version
+    session.add(new_revision)
+    session.add(new_node)
+
+    # Add a history event recording the copy
```
agorajek, line 450: Nice!

agorajek (approved):
Very cool and indeed does not look that bad. Minor comments inline.

datajunction-server/datajunction_server/internal/nodes.py
shangyian: Good point, I thought it wouldn't work because we had to set `node_revision` to point the new revision, but I think `back_populates` should actually take care of that. Let me give it another try.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Hmm, this is a good point. I guess if a node is already deactivated, then from the user perspective it would make sense to just overwrite that node (which in the backend is a hard-delete followed by a create). I'll add this functionality.

PR #920, opened 2024-02-09
Remove deactivated nodes when listing nodes for a tag + fix metric aggregate expression bug

shangyian:
### Summary

This PR fixes two bugs:
* Deactivated nodes should not show up when listing nodes for a tag
* When we check a metric expression to contain an aggregate, we aren't checking nested functions in the right way, causing valid agg expressions to be rejected

### Test Plan

Added a test for this

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Thanks!

PR #923, opened 2024-02-14
Editing cubes should retain information on tags and node mode

shangyian:
### Summary

This PR fixes a bug where editing cubes (in the UI) does not retain information on tags and node mode. It also changes the /cubes/{cube}/ API endpoint to return all the information needed to power the cube add/edit interface, so that the UI does not need to make two backend calls in order to populate with the right info. There's a similar issue with the `/metrics/{metric}` API that we should also fix.

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Thanks!

PR #924, opened 2024-02-16
Remove unused attributes to not clutter the column attributes space

shangyian:
### Summary

Removes the following system-level column attributes: `event_time`, `effective_time`, `expired_time`. These aren't used by the system in any way, so it's better to remove them until we're ready to implement features that can actually use them. 

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
+1. these attributes are to obscure to keep'em w/o a use case.

PR #929, opened 2024-02-26
On the cube node edit page, enable editing of metrics

shangyian:
### Summary

When editing cube nodes, the metrics selector should be editable rather than disabled.

### Test Plan

Locally

- [x] `make check` passes

### Deployment Plan

ASAP


agorajek (approved):
Thanks!

PR #932, opened 2024-03-01
Switch to node-level dimension links

shangyian:
### Summary

This PR makes the final switch over to node-level dimension links. It does so by:
* Changing the APIs for column-level dimension linking and unlinking to use node-level dimension links
* Changing the UI to display column-level dimension links
* Adding a migration endpoint that we can remove once we are done with migration

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

ASAP. This blocks all other materialization-related changes. Once this PR has been merged, we also can't make new changes without deploying this migration successfully.

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -15,7 +15,6 @@ from sqlalchemy.orm import Session
 from sqlalchemy.sql.operators import is_
 from starlette.requests import Request
 
-from datajunction_server.api.catalogs import UNKNOWN_CATALOG_ID
 from datajunction_server.api.helpers import (
     activate_node,
     deactivate_node,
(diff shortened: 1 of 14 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -860,164 +752,85 @@ def remove_complex_dimension_link(  # pylint: disable=too-many-locals
...
-                session.add(cube)
-                session.add(
-                    status_change_history(
-                        node,
-                        NodeStatus.VALID,
-                        NodeStatus.INVALID,
-                        current_user=current_user,
+
+
+@router.post("/nodes/{name}/migrate", status_code=201)  # pragma: no cover
```
agorajek, line 778: I know this is temporary, but it may feel safer to call if this endpoint is named clearer, e.g. `migrate_dim_link`.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1317,3 +1325,213 @@ def get_cube_revision_metadata(session: Session, name: str):
...
+                "join_sql": link_input.join_on,
+                "join_cardinality": link_input.join_cardinality,
+                "role": link_input.role,
+            },
+            user=current_user.username if current_user else None,
+        ),
+    )
+    session.commit()
+    session.refresh(node)
+    return is_update
```
agorajek, line 1445: I am not sure this Boolean value is a useful return. I think returning nothing would be cleaner (like in `remove_dimension_link`), or returning an Enum (NEW, UPDATED, etc), but this true/false may just catch us by surprise somewhere.

agorajek (approved):
Two small comments, but overall look great.

datajunction-server/datajunction_server/api/nodes.py
shangyian: good point, renaming...

datajunction-server/datajunction_server/internal/nodes.py
shangyian: Makes sense! I think it's easier to just return an `ActivityType` enum, since we already have this enum for the history APIs. Let me make this change

PR #942, opened 2024-03-15
Fix deadlock issue with FastAPI Depends

shangyian:
### Summary

This PR fixes an issue where using FastAPI `Depends` for a database session causes [deadlock issues](https://github.com/tiangolo/fastapi/issues/3205) and a huge increase in the number of database connections as concurrent requests increase. 

### Test Plan

In progress

- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/internal/access/authentication/github.py
+++ b/datajunction-server/datajunction_server/internal/access/authentication/github.py
@@ -42,7 +42,7 @@ def get_github_user(access_token: str) -> Optional[User]:  # pragma: no cover
     ).json()
     if "message" in user_data and user_data["message"] == "Bad credentials":
         return None
-    session = next(get_session())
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/internal/access/authentication/github.py
```diff
@@ -42,7 +42,7 @@ def get_github_user(access_token: str) -> Optional[User]:  # pragma: no cover
     ).json()
     if "message" in user_data and user_data["message"] == "Bad credentials":
         return None
-    session = next(get_session())
+    session = next(get_session())  # type: ignore
```
shangyian, line 45: These two functions don't look like they're actually used. I'm not going to remove them here since I want to keep the changes in this PR contained

agorajek (approved):
Looks good. Thanks @shangyian for this quick fix!

PR #943, opened 2024-03-18
Upgrade FastAPI to the latest version by pinning it to >=0.110.0

shangyian:
### Summary

This PR takes the opportunity to upgrade DJ to the latest version of FastAPI. We're doing this because a number of updates to newer versions of FastAPI may fix the deadlock issue we were seeing with database sessions using `Depends`, namely this fix in [`0.82.0`](https://fastapi.tiangolo.com/release-notes/#0820):
> 🐛 Allow exit code for dependencies with yield to always execute, by removing capacity limiter for them, to e.g. allow closing DB connections without deadlocks.

This is preferable to rolling our own solution. 

The primary changes in this PR are to accommodate the new starlette `TestClient`, which uses `httpx` rather than `requests`. There are also a few changes needed for the updated version of pydantic (although no breaking changes as we haven't upgraded to pydantic>=2).

### Test Plan

Will update as tests are done.

- [ ] `make check` passes

### Deployment Plan

ASAP

### DIFF
--- a/datajunction-server/datajunction_server/api/client.py
+++ b/datajunction-server/datajunction_server/api/client.py
@@ -31,7 +31,7 @@ def client_code_for_creating_node(
     node = get_node_by_name(session, node_name)
 
     # Generic user-configurable node creation params
-    params = NodeOutput.from_orm(node).current.dict(
+    params = NodeOutput.from_orm(node).dict(
         exclude={
             "id",
             "version",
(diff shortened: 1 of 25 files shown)


agorajek (approved):
I agree with your assessment @shangyian . Thanks for finding this in the FastAPI release notes!

PR #944, opened 2024-03-26
FIX Python client signature mistakenly mark description as Optional

anhqle:
`description` is actually required by the API


agorajek:
I don't think I agree... I see `description` as optional at the Node class level: https://github.com/DataJunction/dj/blob/main/datajunction-clients/python/datajunction/nodes.py#L67

What am I missing?

anhqle:
The API schema says description is required https://github.com/DataJunction/dj/blob/3a2d628d5319823885dd08bfc5c1690933304c42/openapi.json#L2806

I'm not blocked by this so feel free to take time if you need to think about the schema. I already switched from using the Python client (which doesn't have all the functionality) to making calls to the API directly.

agorajek:
> The API schema says description is required
> 
> https://github.com/DataJunction/dj/blob/3a2d628d5319823885dd08bfc5c1690933304c42/openapi.json#L2806
> 
> I'm not blocked by this so feel free to take time if you need to think about the schema. I already switched from using the Python client (which doesn't have all the functionality) to making calls to the API directly.

I see. I wonder if this spec is fresh, because I am pretty sure it is autogenerated. Thanks for bringing this up, let's not merge it yet until we know for sure where is the disconnect.

shangyian (approved):
Thanks @anhqle!

shangyian (changes requested):
Upon rereading this -- I was looking at it backwards, and yeah, it's meant to be optional. We should probably regenerate the djopenapi spec.

anhqle:
@shangyian it looks like the server does require `description`? ([server code](https://github.com/DataJunction/dj/blob/127db4f3cdb3461ad959c4fec93b5cff88b29218/datajunction-server/datajunction_server/database/node.py#L382))

DJ UI currently submits a blank `""` description if none is provided.

If it's meant to be Optional then do we just need to change `description` type hint in server code above?

shangyian:
@anhqle That's actually the database schema, which isn't the API's expected input model, although we also default in the input model (see the NodeRevisionBase model) to an empty string. I think changing the expected input on `NodeRevisionBase` to `Optional[str]` makes sense, but I'd rather keep the database schema as-is.

anhqle:
Fixing in #991

PR #945, opened 2024-04-03
FEAT Add support for the complex dimension endpoints to DJService

anhqle:


datajunction-ui/src/app/services/DJService.js
```diff
@@ -613,6 +613,51 @@ export const DataJunctionAPI = {
     );
     return { status: response.status, json: await response.json() };
   },
+
+  addComplexDimensionLink: async function (
+    nodeName,
+    dimensionNode,
+    joinType = 'left',
```
shangyian, line 620: Perhaps these defaults should just be left null, so that they can default to the backend defaults rather than relying on defaults embedded in the frontend? Of course, if we decide that we need different defaults when this is being accessed via the UI vs via an API call, we can revisit this

shangyian (commented):
Thanks for adding these @anhqle! You might need to add tests for these to get the build to pass, and I added one minor comment about defaults, but other than that LGTM

datajunction-ui/src/app/services/DJService.js
anhqle: Fixed

anhqle:
I added some tests, modeled after the tests for `link/unlinkDimension`
These tests don't check for a correct response.

shangyian (approved):
Looks great, thanks for the changes @anhqle!

PR #947, opened 2024-04-05
Add listing nodes with tags plus other minor fix.

agorajek:
### Summary

What the title says. The additional fix is to gracefully handle a situation when you want to register a table as a source node and that table is already registered in your system.

### Test Plan

- [ ] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/datajunction-clients/python/datajunction/_internal.py
+++ b/datajunction-clients/python/datajunction/_internal.py
@@ -26,6 +26,7 @@ from datajunction.exceptions import (
     DJClientException,
     DJNodeAlreadyExists,
     DJTagAlreadyExists,
+    DJTagDoesNotExist,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


datajunction-clients/python/datajunction/_internal.py
```diff
@@ -571,6 +574,22 @@ def _create_tag(
...
+        node_type: Optional[models.NodeType] = None,
+    ):
+        """
+        Retrieves all nodes with a given tag.
+        """
+        response = self._session.get(
+            f"/tags/{tag_name}/nodes"
+            + (f"?node_type={node_type.value}" if node_type else ""),
+        )
+        if response.status_code == 404:
```
shangyian, line 589: If the response is a 404, does that typically mean the tag doesn't exist? If that's the case is it better here to inform the user that the tag doesn't exist rather than defaulting to an empty list of nodes?

shangyian (approved):
Thanks @agorajek! The changes make sense, just one minor comment about handling errors

datajunction-clients/python/datajunction/_internal.py
agorajek: Hmm... I think you may be right. Will fix.

PR #952, opened 2024-04-10
Switch DJ to support Python 3.10 and up only.

agorajek:
### Summary

Getting rid of Py3.8 and 3.9 support, because we'd like to rely on some newer async syntax features.

### Test Plan

- [ ] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/.github/workflows/python-package-daily.yml
+++ b/.github/workflows/python-package-daily.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11.0rc1']
+        python-version: ['3.10', '3.11.0rc1']
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/python-package-daily.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11.0rc1']
+        python-version: ['3.10', '3.11.0rc1']
```
shangyian, line 17: Hmm, I wonder why this one is `3.11.0rc1`

shangyian (approved):

.github/workflows/python-package-daily.yml
agorajek: Oh I thought you'd know. I will remove that to see...

PR #955, opened 2024-04-11
Fix explode function query assembly

shangyian:
### Summary

Fix query string assembly for the `EXPLODE` Spark SQL function. Namely, when we reassemble the AST parts back into a query string, it should be following this pattern:
```
EXPLODE(<array-or-map>) <alias> AS <column list>
```

### Test Plan

Added some tests for this in the `LATERAL VIEW EXPLODE ...` tests.

- [x] `make check` passes

### Deployment Plan

ASAP

### DIFF
--- a/datajunction-server/tests/sql/parsing/test_ast.py
+++ b/datajunction-server/tests/sql/parsing/test_ast.py
@@ -337,20 +337,24 @@ def test_ast_compile_lateral_view_explode3(session: Session):
     """
 
     query = parse(
-        """SELECT a, b, c, d, e
+        """SELECT a, b, c, d, e, v.en
     FROM (SELECT 1 as a, 2 as b, 3 as c) AS foo
     LATERAL VIEW EXPLODE(ARRAY(30, 60)) AS d
-    LATERAL VIEW EXPLODE(ARRAY(40, 80)) AS e;""",
+    LATERAL VIEW EXPLODE(ARRAY(40, 80)) AS e
+    LATERAL VIEW EXPLODE(ARRAY(100, 200)) v AS en;""",
     )
     exc = DJException()
     ctx = ast.CompileContext(session=session, exception=exc)
     query.compile(ctx)
+    assert parse(str(query)) == query
+    assert "LATERAL VIEW EXPLODE(ARRAY(100, 200)) v AS  en" in str(query)
 
     assert query.columns[0].is_compiled()
     assert query.columns[1].is_compiled()
     assert query.columns[2].is_compiled()
     assert query.columns[3].is_compiled()
     assert query.columns[4].is_compiled()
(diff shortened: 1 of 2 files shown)


datajunction-server/tests/sql/parsing/test_ast.py
```diff
@@ -337,20 +337,24 @@ def test_ast_compile_lateral_view_explode3(session: Session):
     """
 
     query = parse(
-        """SELECT a, b, c, d, e
+        """SELECT a, b, c, d, e, v.en
```
agorajek, line 340: This is great @shangyian . I wanted to add a multicolumn LVE to this query but somehow it is failing during parsing. I tested it works ok in Spark:
```
SELECT a, b, c, d, e, v.en, t1.k, t1.v
    FROM (SELECT 1 as a, 2 as b, 3 as c) AS foo
    LATERAL VIEW EXPLODE(ARRAY(30, 60)) AS d
    LATERAL VIEW EXPLODE(ARRAY(40, 80)) AS e
    LATERAL VIEW EXPLODE(ARRAY(100, 200)) v AS en
    -- added line below
    LATERAL VIEW EXPLODE(MAP('a', 1, 'b', 2, 'c', 3)) t1 AS k, v
```

In Spark 3.3 it runs:
```
spark-sql-3.3> SELECT a, b, c, d, e, v.en, t1.k, t1.v
             >     FROM (SELECT 1 as a, 2 as b, 3 as c) AS foo
             >     LATERAL VIEW EXPLODE(ARRAY(30, 60)) AS d
             >     LATERAL VIEW EXPLODE(ARRAY(40, 80)) AS e
             >     LATERAL VIEW EXPLODE(ARRAY(100, 200)) v AS en
             >     LATERAL VIEW EXPLODE(MAP('a', 1, 'b', 2, 'c', 3)) t1 AS k, v;
1	2	3	30	40	100	a	1
1	2	3	30	40	100	b	2
...
```

The error I get is 
```
FAILED tests/sql/parsing/test_ast.py::test_ast_compile_lateral_view_explode3 - datajunction_server.sql.parsing.backends.antlr4.SqlSyntaxError: ('Parse error 16:56:', "mismatched input '(' expecting {<EOF>, ';'}")
```
, but I think we can push your PR in and fix this later.

agorajek (approved):
Left a comment in-line, but let's not make it block this PR.

datajunction-server/tests/sql/parsing/test_ast.py
shangyian: Hmm, yeah, that one is odd, because it looks like it's an antlr parser issue. I think this would parse but isn't valid Spark:
```
LATERAL VIEW EXPLODE(MAP('a', 1, 'b', 2, 'c', 3)) t1 AS (k, v);
```

PR #956, opened 2024-04-11
Hide node Edit and Delete button after deleting a node.

agorajek:
### Summary

Because they should be hidden when the node is deleted. The whole node row could actually go away but since we display a message in that row and I am just starting with React it is what it is.

Also added a confirmation dialog box so we don't delete nodes by accident.

### Test Plan

Tried in the UI and with unit tests. Both work.

- [ ] `make check` passes

### Deployment Plan

n/a


shangyian (approved):
LGTM, thanks for adding this!

Issue #969, opened 2024-04-26
Fix bug where updating nodes that have deactivated/deleted children fail 


shangyian:
When updating nodes, deactivated or deleted children will cause the update to fail during the `propagate_update_downstream` stage, returning a 500.

shangyian:
Partially fixed in https://github.com/DataJunction/dj/pull/966/files#diff-a05c3c3e637b6978560b513567cbe3e855805a33fef38e3243116015ae769fbaR868

PR #970, opened 2024-04-26
Fix bug where updating nodes that have deactivated children fail

shangyian:
### Summary

This PR fixes an issue where updating nodes that have deactivated children will fail due to the deactivated node revisions continuing to live in the parent node's `children`. Because deactivated nodes aren't fully deleted, we should keep the parent - child relationships stored in the database, but we should skip over processing these nodes when propagating updates to downstreams. Not skipping over these was causing node updates to fail with 500s.

This was partially fixed in https://github.com/DataJunction/dj/pull/966/files#diff-a05c3c3e637b6978560b513567cbe3e855805a33fef38e3243116015ae769fbaR868 for updating nodes with deactivated cubes downstream, but this fix also fixes it for downstream deactivated non-cube nodes.

### Test Plan

Added a unit test that covers both the deactivated downstream cube and non-cube test cases.

- [x] PR has an associated issue: #969 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -865,7 +865,7 @@ async def propagate_update_downstream(  # pylint: disable=too-many-locals
                         ),
                     ],
                 )
-                if child_node:  # pragma: no cover
+                if child_node:
                     await update_cube_node(
                         session,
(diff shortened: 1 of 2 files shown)


agorajek (approved):
Such a small change. Thanks for adding the unit test!

PR #971, opened 2024-04-26
Sort history events in descending order + additional validation checks for cubes

shangyian:
### Summary

Sort history events in descending order so that the most recent updates will be at the top of the list in the history diffs UI. This is done server-side because we also limit the number of history events to 100, meaning that the previous UI-side ordering of events was cutting off to only events 1-100 and not [current, current - 100]

Also adds some additional validation checks for cubes:
* Cubes cannot be created if metrics don't exist
* Cubes cannot be created if metrics are invalid

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Makes sense. But I see the limit at 100, not 1k... am I missing something?

shangyian:
@agorajek oops yes, I meant 100 and not 1000 😅

Issue #972, opened 2024-04-28
Add caching for query building


shangyian:
We can cache built queries to speed up both requesting SQL and data. If a node has not been updated (query changed, availability posted, new dimension link added etc), the built query is guaranteed to remain the same, so we can reuse a query that's previously been built. 

There are three types of query building patterns that DJ supports:
* Metrics SQL: query by metrics, dimensions, filters, and limit
* Measures SQL: query by metrics, dimensions, filters, and limit
* Node SQL: query by node name, dimensions, filters, and limit

We can store all such query requests in a database table:
```
queryrequest:
* node_name: str
* metrics: json list of metric names
* dimensions: json list of dimension names
* filters: json list of filters
* engine_name: str
* engine_version: str
* limit: int
* orderby: str
* query: str
* query_type: str, one of [`metrics`, `measures`, `node`]
* valid: bool, whether this request is still valid

** note that only `node` or `metrics` may be populated for any given entry, not both
```

If the node has been updated in some way (query changed, availability posted, new dimension link added etc), asynchronously update any of its and its children's built queries in `queryrequest`.

In some cases, the query request may no longer be valid, in which case we should invalidate the entry in the table.

PR #973, opened 2024-04-29
Save built queries to database to speed up requesting SQL and data

shangyian:
### Summary

This PR implements #972 and saves built queries to a database based on the query request key. The query request key is a combination of:
* `nodes`: these are a list of the nodes (+ versions) that SQL is being requested for, may be a list of metrics or a single transform, metric, or dimension node
* `parents`: a list of the parents of the above nodes and their versions
* `dimensions`: a list of dimension attributes requested (+ node versions) 
* `filters`: a list of filters requested (+ node versions)
* `limit`: the limit set for the query (some query types don't have limit)
* `orderby`: order by clause requested
* `query_type`: one of `metrics`, `measures`, or `node`, depending on the exact `/sql` endpoint it hits

### Test Plan

Locally

- [x] PR has an associated issue: #972
- [x] `make check` passes


datajunction-server/datajunction_server/api/sql.py
```diff
@@ -35,6 +36,13 @@ async def get_measures_sql_for_cube(
     dimensions: List[str] = Query([]),
     filters: List[str] = Query([]),
     *,
+    include_all_columns: bool = Query(
```
agorajek, line 39: Something sneaked in the PR :)

datajunction-server/datajunction_server/database/queryrequest.py
```diff
@@ -0,0 +1,323 @@
+"""Partition database schema."""
```
agorajek, line 1: Copy/pasta?

agorajek (approved):
This is great. Thanks @shangyian .

datajunction-server/datajunction_server/api/sql.py
shangyian: Oops yeah... I also needed to do this and forgot that I was making changes at the same time 😅

PR #974, opened 2024-04-30
Druid Functions Compatibility

shangyian:
### Summary

This PR adds a `dialects` field to the base `Function` class, allowing us to specify what SQL dialects a particular function is compatible with. By default, the base class sets `dialects=[Dialect.SPARK]`, because all of our existing functions are from Spark SQL. Then we annotate all the existing functions that are compatible with Druid SQL, adding a few additional Druid SQL functions as well.

For metric nodes queried using the `GET /metrics/{metric_name}` endpoint, we additionally include `incompatible_druid_functions`, a list of Druid-incompatible functions in the metric expression. Note that this is not 100% accurate at the moment, because we haven't implemented support for all Druid SQL functions in our function registry. I'll leave that for a follow-up PR.

With this addition to the API, we can now display this information in the UI with a warning label:

Screenshot 2024-04-29 at 6 55 17 PM

### Test Plan

- [x] PR has an associated issue: #975 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/functions.py
+++ b/datajunction-server/datajunction_server/sql/functions.py
@@ -398,7 +398,6 @@ def infer_type(
     return col.type  # type: ignore
 
 
-
 class ApproxQuantileDs(Function):
     """
     approx_quantile_ds(col, percentage [, accuracy]) -
@@ -421,6 +420,7 @@ class Array(Function):
     """
     Returns an array of constants
     """
+
     dialects = [Dialect.SPARK, Dialect.DRUID]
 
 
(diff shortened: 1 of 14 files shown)


datajunction-server/datajunction_server/sql/functions.py
agorajek, line 1: Not strictly related to this PR, but I wonder if we would be able one day to have overloaded UDFs with different arguments across different dialects.

agorajek (approved):
Super cool!

datajunction-server/datajunction_server/sql/functions.py
shangyian: Yeah, this is a good point. Maybe the `dialects` declaration should be at the type function declaration level.

Issue #911, opened 2024-02-02
Define what Status means by node type.


agorajek:
We need this to get on the same page. Let's consider the following two aspects:
- metadata / definition / query building
- data / availability / runtime

agorajek:
Internal aspects of nodes health (and which node types applies to):

x - applicable
o - not applicable

Health / Node |  Source | Transform, Metric, Dimension | Cube 
-------------- | ------- | ------------- | -------------
table missing  | x | o | x
query broken <br> (e.g. missing parent node or column)   | o | x | x
dim link / col attr broken <br> (e.g. bad dimension ref or broken join ref) | x | x | o
data not-available | x | x for T w/ mat only | x

In DRAFT mode we should not validate or block any activity. In PUBLISHED mode we should.

agorajek:
Per @shangyian :

> Yeah I agree that should be split into two statuses. Also I think in the UI we should have some hover info boxes that explain what each thing is meant to represent or do so that things are more self-documenting.

artwr:
Agreed ;p

PR #979, opened 2024-05-03
Improve node status display and fix status update bugs

shangyian:
### Summary

This PR makes a number of improvements related to node status.

#### Display detailed node status

This change modifies the node validation endpoint to include a list of errors if the node's status is invalid.

We update the node info UI to (a) always revalidate the node (this can be changed later), and (b) show detailed status information. If a node is valid, we continue to display just the checkmark, but if a node is invalid, we additionally display a list of errors that explain why the node is invalid. This allows users to understand why a node's status is what it is, so that they can fix issues to correct it.

Screenshot 2024-05-04 at 10 38 06 AM

#### Fix node update status propagation bug

We were seeing an issue with node update propagation, where the downstream nodes' statuses were often not updated to the right status. The primary cause of this issue is that we weren't updating the nodes downstream of the updated node in the right order. We need to topologically sort the downstreams, allowing us to update the nodes with no upstream dependencies first, and the nodes with the most upstream dependencies last.

A secondary issue is that we were only updating statuses and not columns. If a node's query changes, it may cause changes to downstream nodes' columns, which also needs to be saved in order for downstream nodes to pick up on them.

### Test Plan

- [ ] PR has an associated issue: #911 #621 
- [ ] `make check` passes


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -813,15 +823,170 @@ async def update_cube_node(  # pylint: disable=too-many-locals
...
+        new_revision.node_id = node.id  # type: ignore
+        session.add(node)
+        session.add(new_revision)
+    await session.commit()
+    await session.refresh(node.current)  # type: ignore
+    await session.refresh(node, ["current"])
+    return node_validator
+
+
+def topological_sort(nodes: List[Node]) -> List[Node]:
```
shangyian, line 942: This was the reason why the node update propagation wasn't working correctly: we need to sort the downstream dependencies in topological order, or it's possible that we'll update a child node before a parent node.

PR #981, opened 2024-05-05
Improve node status display and fix status update bugs

shangyian:
### Summary

This PR makes a number of improvements related to node status.

#### Fix node update status propagation bug

We were seeing an issue with node update propagation, where after a node is updated, its downstreams' statuses were often not updated to the right status. 

The primary cause of this issue is that we weren't updating the downstreams in the right order. We need to **topologically sort** the downstreams, such that if node A queries from node B, B always comes before A in the ordering. We were instead sorting by level order, which is the wrong ordering and may lead to children being processed before parents.

A secondary issue is that we were only updating statuses and not columns. If a node's query changes, it may cause changes to downstream nodes' columns, which also needs to be saved in order for downstream nodes to pick up on them.

#### Display detailed node status

This change modifies the node validation endpoint to include a list of errors if the node's status is invalid.

We update the node info UI to (a) always revalidate the node (this can be changed later), and (b) show detailed status information. If a node is valid, we continue to display just the checkmark, but if a node is invalid, we additionally display a list of errors that explain why the node is invalid. This allows users to understand why a node's status is what it is, so that they can fix issues to correct it.

Screenshot 2024-05-04 at 10 38 06 AM

### Test Plan

- [x] PR has an associated issue: #911 #621 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -41,9 +41,11 @@ from datajunction_server.database.user import User
 from datajunction_server.errors import (
     DJAlreadyExistsException,
     DJDoesNotExistException,
+    DJError,
     DJException,
     DJInvalidInputException,
     DJNodeNotFound,
+    ErrorCode,
 )
 from datajunction_server.internal.engines import get_engine
 from datajunction_server.models import access
@@ -328,6 +330,8 @@ async def resolve_downstream_references(
                 )
 
             downstream_node_revision.status = node_validator.status
+
+            await session.refresh(downstream_node_revision, ["columns"])
             downstream_node_revision.columns = node_validator.columns
             if node_validator.status == NodeStatus.VALID:
                 newly_valid_nodes.append(downstream_node_revision)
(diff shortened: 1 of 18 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -328,6 +330,8 @@ async def resolve_downstream_references(
                 )
 
             downstream_node_revision.status = node_validator.status
+
+            await session.refresh(downstream_node_revision, ["columns"])
```
agorajek, line 334: Oh wow, good catch.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -144,26 +147,43 @@ async def validate_node(
...
         current_user=current_user,
     )
-    return JSONResponse(
-        status_code=HTTPStatus.OK,
-        content={
-            "message": f"Node `{name}` has been set to {status}",
-            "status": status,
-        },
+
+    return NodeStatusDetails(
```
agorajek, line 165: Why the change in the return type?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -830,138 +836,60 @@ async def propagate_update_downstream(  # pylint: disable=too-many-locals
...
-                    for lineage in await get_column_level_lineage(session, new_revision)
-                ]
+    _logger.info("Propagating update of node %s downstream", node.name)
+    downstreams = await get_downstream_nodes(
+        session,
+        node.name,
+        include_deactivated=False,
+        include_cubes=False,
+    )
+    downstreams = topological_sort(downstreams)
```
agorajek, line 846: Thank you @shangyian for catching and fixing this. I completely missed this aspect of revalidating child nodes.

agorajek (approved):
This PR is full of goodness. 

thank_you_kid

datajunction-server/datajunction_server/api/nodes.py
shangyian: The new return type allows us to display the detailed status in the UI -- instead of just returning a `VALID` or `INVALID`, this actually has a list of errors if the status is invalid.

Issue #930, opened 2024-02-29
Add categorical partitions to nodes


shangyian:
For materialization

PR #983, opened 2024-05-07
Add support for categorical partition materialization

shangyian:
### Summary

This PR adds support for setting categorical partitions for materialization.

Setting a categorical partition on a node and subsequently scheduling a materialization will result in a materialization query that includes the categorical partition column as a query parameter. For example, if `state` was set as a categorical partition, then the materialization query would contain a parameterized filter `WHERE state = ${state}`. In the AST, the `${state}` parameter is represented with a generated DJ function that gets substituted at runtime.

This change works for both Spark SQL and Druid ingestion jobs, although performance cannot be guaranteed on the latter. 

#### UI Changes

This PR also makes some additional UI changes to the materialization tab, which will hopefully make it clearer which backfills belong to which materialization types:

Screenshot 2024-05-09 at 7 17 05 AM

### Test Plan

Locally

- [x] PR has an associated issue: #930
- [x] `make check` passes

### Deployment Plan

The `backfill` table's `spec` column will need to be updated. See the alembic migration for more details: 63095c18

### DIFF
--- a/datajunction-server/tests/api/cubes_test.py
+++ b/datajunction-server/tests/api/cubes_test.py
@@ -1618,6 +1618,7 @@ async def test_updating_cube_with_existing_materialization(
     assert result["version"] == "v2.0"
 
     # Check that the query service was called to materialize
+    assert len(query_service_client.materialize.call_args_list) == 3  # type: ignore
     last_call_args = (
         query_service_client.materialize.call_args_list[-1].args[0].dict()  # type: ignore
     )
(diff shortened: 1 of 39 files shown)


datajunction-server/tests/api/cubes_test.py
```diff
@@ -1626,7 +1626,7 @@ async def test_updating_cube_with_existing_materialization(
         == "druid_measures_cube__incremental_time__default.hard_hat.hire_date"
     )
     assert last_call_args["node_name"] == "default.repairs_cube"
-    assert last_call_args["node_version"] == "v2.0"
+    # assert last_call_args["node_version"] == "v2.0"
```
agorajek, line 1629: cleanup leftover

agorajek (approved):
That's a big feature. I did a quick scan, but I'd say let's test it in Prod!

PR #984, opened 2024-05-08
Use the queryrequest cache to serve /data requests + bug fixes

shangyian:
### Summary

The `/data` endpoints also need to build SQL before they can send off requests to the query service for data. These endpoints should also use the `queryrequest` cache where possible.

This PR also fixes some additional issues, including 
* Requiring dimensions when metrics SQL is requested. While we can continue to keep this "at least one dimension" requirement when someone is creating a cube, it should be possible to just request metrics without any dimensions, resulting in a single aggregate value for the metric.
* Not filtering out empty dimensions (arguably this is user error, but easier for the server-side to just filter out these rather than raising an error)
* Fix a bug with defaulting to `name` on the foreign key mapping during sql building when this attribute doesn't exist.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/data.py
+++ b/datajunction-server/datajunction_server/api/data.py
@@ -12,12 +12,10 @@ from sqlalchemy.orm import joinedload, selectinload
 from sse_starlette.sse import EventSourceResponse
 
 from datajunction_server.api.helpers import (
-    assemble_column_metadata,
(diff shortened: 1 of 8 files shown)


agorajek (approved):
Makes sense. Agree with you on the required dimensions here. Although it should be very rare to have these metrics w/o any dims.

PR #985, opened 2024-05-08
FIX python client error not propagating

anhqle:
### Summary

Current code checks for the existence of response with `if exc.response`, which is a semantic bug because 400 responses are actually Falsy.

This PR removes this check (since exc.response should always exist anyway), ultimately allowing the error message to be propagated

### Test Plan

- [x ] `make check` passes
- [x ] `make test` shows 100% unit test coverage


shangyian (approved):
Thanks for fixing this @anhqle!

agorajek (approved):
Thanks!

shangyian:
@anhqle just to check, you have permissions to merge PRs in this repo right?

anhqle:
I think so, will let you know otherwise

PR #993, opened 2024-05-10
Support selecting wildcards in node queries

shangyian:
### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1837,7 +1837,7 @@ async def revalidate_node(  # pylint: disable=too-many-locals,too-many-statement
 
     # Check if any columns have been updated
     existing_columns = {col.name: col for col in node.current.columns}  # type: ignore
-    updated_columns = False
+    updated_columns = len(current_node_revision.columns) != len(node_validator.columns)
```
agorajek, line 1840: nice catch!

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -1119,12 +1119,65 @@ def add_table(self, table: "Table") -> "Wildcard":
...
+                            Column(
+                                name=Name(col.name)
+                                if isinstance(col.name, str)
+                                else col.name,
+                                _table=wildcard_origin,
+                                _type=col.type,
+                            ),
+                        )
+            wildcard_parent.projection.remove(self)
+        else:
```
agorajek, line 1178: nit: it may be easier to invert this conditional:
```
if not isinstance(self.parent, Select):
    return super().compile(ctx)

# and then the main code w/o the need to be indented
```

agorajek (approved):
Nice one! Thank you.

datajunction-server/datajunction_server/sql/parsing/ast.py
shangyian: Good point!

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -241,27 +242,33 @@ async def hard_delete_namespace(
...
+                "true to additionally hard delete the above nodes in this namespace. WARNING:"
                 " this action cannot be undone."
             ),
         )
 
     impacts = {}
-    for node_name in node_names:
-        impacts[node_name] = await hard_delete_node(
-            node_name,
+    for node in reversed(topological_sort(nodes)):
```
shangyian, line 269: While we don't have to do a topological sort here, doing so will mean that we process the deletes for nodes in the namespace in an order that minimizes unnecessary invalidation followed by deletion.

datajunction-server/datajunction_server/internal/validation.py
```diff
@@ -128,7 +129,12 @@ async def validate_node_data(  # pylint: disable=too-many-locals,too-many-statem
                 name=column_name,
                 display_name=labelize(column_name),
                 type=column_type,
-                attributes=existing_column.attributes if existing_column else [],
+                attributes=[
```
shangyian, line 132: This was an annoying issue: the `attributes` field on the `Column` database object is actually a `ColumnAttribute` object and not an `Attribute` object. Since `ColumnAttribute` is the relational link between an `Attribute` and a `Column`, we actually need to recreate this for any copied columns. Otherwise, it'll cause problems downstream since it may reference a column that doesn't exist.

shangyian:
@agorajek yeah, unfortunately I couldn't get one of the unit tests to pass (there was some non-deterministic issue that was hard to track down). I'll take another stab at it soon.

PR #996, opened 2024-05-16
change create_ methods into upsert_

anhqle:
### Summary

This PR changes the `create_somenode` methods into `upsert_somenode`, which create if a node doesn't exist and update if it does exist.

This is useful because users just want to "make the nodes this way" without having to first check for existence. An example use case is writing a notebook with a bunch of upsert, which would be reproducible and idempotent.

The commit so far only changes `create_source` into `upsert_source`. If the approach makes sense I'll make the same changes to other nodes types & and add tests.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -107,7 +107,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
     #
     # Nodes: SOURCE
     #
-    def create_source(  # pylint: disable=too-many-arguments
+    def upsert_source(  # pylint: disable=too-many-arguments
         self,
         name: str,
         catalog: str,
@@ -119,6 +119,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
         primary_key: Optional[List[str]] = None,
         tags: Optional[List[Tag]] = None,
         mode: Optional[models.NodeMode] = models.NodeMode.PUBLISHED,
+        if_exists: Optional[str] = "update",
     ) -> "Source":
         """
         Creates a new Source node with given parameters.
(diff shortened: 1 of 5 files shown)


anhqle:
* **#996** <a href="https://app.graphite.dev/github/pr/DataJunction/dj/996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @anhqle and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

datajunction-clients/python/datajunction/builder.py
```diff
@@ -119,6 +119,7 @@ def create_source(  # pylint: disable=too-many-arguments
         primary_key: Optional[List[str]] = None,
         tags: Optional[List[Tag]] = None,
         mode: Optional[models.NodeMode] = models.NodeMode.PUBLISHED,
+        if_exists: Optional[str] = "update",
```
shangyian, line 122: Should this be an enum instead of a str? Then it'll be a bit clearer what the allowed values for this are without going into the code here

shangyian (approved):
I like this change! Thank you for adding it. Just left a minor comment about using enums.

datajunction-clients/python/datajunction/builder.py
agorajek: I would call the param `update_if_exists` and make it a boolean. And also then... I think the name of the function `create_source` could stay as is.

agorajek (changes requested):
I don't like the name `upsert` but if you all love it then feel free to use it. I just think thank a boolean param like `fail_if_exists` or `update_if_exists` would solve the double functionality in a clean way.

anhqle:
This functionality is added in #1001 instead

PR #999, opened 2024-05-17
Don't use materialized tables in their own materialization job.

agorajek:
### Summary

Main thing here is to use the `for_materialization` flag to indicate that we want to build a node SQL for the purposes of materialization jobs so we don't really want to use any pre-materialized tables. This is potentially risky because it means that we won't use any materialized parent nodes and will go straight the "sources". But this is only for `build_non_cube_materialization_config` so the risk is lower. We can optimize that aspect later.

Minor stuff:
1. Deleted `--cov-report=html` from test commands cause we didn't see the coverage automatically at the end of the run. Now we do.
2. Fixed a lint warning on the Logout button.

### Test Plan

Generated a Spark job for a node and made sure that the full query goes into the workflow even when availability state was present.

- [ ] `make check` passes

### Deployment Plan

nah

### DIFF
--- a/datajunction-clients/python/Makefile
+++ b/datajunction-clients/python/Makefile
@@ -5,7 +5,7 @@ lint:
 	make check
 
 test:
-	pdm run pytest -n auto --cov=datajunction --cov-report=html -vv tests/ --doctest-modules datajunction --without-integration --without-slow-integration ${PYTEST_ARGS}
+	pdm run pytest -n auto --cov=datajunction -vv tests/ --doctest-modules datajunction --without-integration --without-slow-integration ${PYTEST_ARGS}
 
 dev-release:
 	hatch version dev
--- a/datajunction-server/datajunction_server/database/availabilitystate.py
+++ b/datajunction-server/datajunction_server/database/availabilitystate.py
@@ -70,7 +70,8 @@ class AvailabilityState(Base):  # pylint: disable=too-few-public-methods
         """
         Determine whether an availability state is useable given criteria
         """
-        # Criteria to determine if an availability state should be used needs to be added
+        # TODO: we should evaluate this availability state against the criteria. # pylint: disable=fixme
+        #       Remember that VTTS can be also evaluated at runtime dependency.
         return True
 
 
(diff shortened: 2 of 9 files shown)


datajunction-clients/python/Makefile
```diff
@@ -5,7 +5,7 @@ lint:
 	make check
 
 test:
-	pdm run pytest -n auto --cov=datajunction --cov-report=html -vv tests/ --doctest-modules datajunction --without-integration --without-slow-integration ${PYTEST_ARGS}
```
shangyian, line 8: btw, is there a reason the `-cov-report=html` is being removed? (I sometimes glance at it when the build unexpectedly fails in github due to coverage to see what I missed)

datajunction-server/datajunction_server/database/availabilitystate.py
```diff
@@ -70,7 +70,9 @@ def is_available(
         """
         Determine whether an availability state is useable given criteria
         """
-        # Criteria to determine if an availability state should be used needs to be added
+        # If we are materializing, we should not use this availability state
```
shangyian, line 73: Is this build criteria passed down all the way to the other nodes that this node depends on? Or is it only used for this node?

shangyian (commented):
Hmm do you think you could add a test that checks for building materialization SQL for a transform node?

datajunction-clients/python/Makefile
agorajek: I mentioned in the summary. That w/o the HTML option I can see the coverage in text mode at the end of the run now.

datajunction-server/datajunction_server/database/availabilitystate.py
agorajek: Good question. I though it was used by all the nodes, but I couldn't see that in my tests. Need to investigate more...

agorajek: Adjusted with fresh code.

datajunction-clients/python/Makefile
shangyian: Ohhh, I see, makes sense!

datajunction-server/tests/api/materializations_test.py
```diff
@@ -1058,3 +1058,150 @@ async def test_spark_sql_incremental(
             .replace("${country}", "DJ_COUNTRY()"),
         ),
     )
+
+
+@pytest.mark.asyncio
+async def test_spark_with_availablity(
```
shangyian, line 1064: Thanks for adding!

shangyian (approved):
Thanks for fixing this!

PR #1000, opened 2024-05-17
Add more Spark functions

shangyian:
### Summary

This PR adds the following Spark SQL functions:
* [`transform_keys`](https://spark.apache.org/docs/latest/api/sql/index.html#transform_keys)
* [`transform_values`](https://spark.apache.org/docs/latest/api/sql/index.html#transform_values)
* [`map_zip_with`](https://spark.apache.org/docs/latest/api/sql/index.html#map_zip_with)

It also fixes a few bugs:
* Calling `concat` on maps is possible
* `aggregate`'s lambda function can take three or four arguments, depending on whether a `finish` function is provided
* We don't infer types on nested lambda functions correctly, so for the time being we just introduce a workaround with `CAST(... AS ...)` around the nested lambda, allowing the type to be manually specified by the user

### Test Plan

- [x] `make check` passes


agorajek (approved):
Nice!

PR #1009, opened 2024-05-22
Attempt to fix the comparison for struct types.

agorajek:
### Summary

This fixes comparison for struct types that we ingest from our repo. Another way to fix it would be to try to match the other side but I'd like to move on for now since things seems to be working.

### Test Plan

Unit tests plus internal test on a mock source node:
```
create table olek.test_1 (a string, b int, c double, d map<string, string>, e array<int>, s struct<a int, b string>);
```

- [ ] `make check` passes

### Deployment Plan
n/a


shangyian (approved):
Nice!

PR #1010, opened 2024-05-22
Query validation SQL shouldn't use own materialized tables

shangyian:
### Summary

When building SQL for query validation, it should not use its own materialized tables (i.e., the availability state). This PR achieves that by making the SQL builder use the default build criteria that contains the target node name.

This also adds an additional feature to kick off a background task that rebuilds and updates the SQL in the `queryrequest` table after each SQL request, so that subsequen requests will get the most up-to-date SQL.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Awesome.

PR #1012, opened 2024-05-23
Enable dimension values dropdown

shangyian:
### Summary

This change enables a dimension values dropdown for each dimension in the "Add Filters" left pane. This makes it easier for users to figure out what some valid values they can input into each filter field are. Note that we had this feature all along, but it was disabled due to potential performance issues. This doesn't seem to be the case assuming cached dimension nodes.

### Test Plan

Locally

- [x] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Nice. Yay, cache warmup!

PR #1016, opened 2024-05-24
Show source node refreshes in the node history tab

shangyian:
### Summary

This change makes it so that the source node refreshes actually show up in the node history tab.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Thanks!

PR #1017, opened 2024-05-24
Fix two bugs: filter pushdowns + source node refresh

shangyian:
#### Filter Pushdown Bugfix

This PR fixes an issue where filters are not being pushed down correctly when building node SQL. There are two aspects to this issue:
1. We were pushing down filters based on the column name of the dimension attribute rather than the full dimension attr name. This meant that filters were sometimes pushed down to node columns that should not have been filtered on at all.
2. We weren't always applying the filter pushdown to the right `WHERE` clause, sometimes generating invalid filter blocks because they weren't attached to the right place.

#### Source Node Refresh Bugfix

This PR also fixes a source node refresh bug where every time a source node is refreshed, we don't copy the existing dimension links to the new node revision. 

### Test Plan

This change ended up resulting in a fair number of unit tests changes, but this is because we're now keeping each node query's original table aliases. Previously we would overwrite the table aliases with DJ-generated ones, but now if the user's query contains an aliased table, we keep the alias in place. While we're not necessarily worried about readability of DJ-generated queries overall, this can be nice for debugging purposes.

- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -26,6 +26,7 @@ from datajunction_server.api.helpers import (
 from datajunction_server.api.namespaces import create_node_namespace
 from datajunction_server.api.tags import get_tags_by_name
 from datajunction_server.constants import NODE_LIST_MAX
+from datajunction_server.database import DimensionLink
 from datajunction_server.database.attributetype import ColumnAttribute
 from datajunction_server.database.column import Column
 from datajunction_server.database.history import ActivityType, EntityType, History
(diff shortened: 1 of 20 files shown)


datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -997,6 +998,16 @@ async def refresh_source_node(
         schema_=current_revision.schema_,
         table=current_revision.table,
         status=current_revision.status,
+        dimension_links=[
```
agorajek, line 1001: Thanks for catching this!

datajunction-server/datajunction_server/construction/build.py
```diff
@@ -486,27 +457,55 @@ async def _build_tables_on_select(
...
+                isinstance(node_ast.child, ast.Table)
+                and isinstance(tbl, ast.Aliasable)
+                and hasattr(tbl, "alias")
+                and tbl.alias
+            ):
+                node_ast.alias.name = tbl.alias
+
             await node_ast.compile(context)
+
+            # This will push down the filter expression to CTEs, subqueries, and top-level queries.
```
shangyian, line 489: @agorajek This block solves the issue with filter pushdowns not applying to the right `SELECT` clause. Basically, each filter expression applies to some `TableExpression`, and we need to append the filter expressions to the `TableExpression`'s nearest SELECT clause, not the overall SELECT clause. This makes it so that the table aliases are always applied correctly.

PR #1001, opened 2024-05-17
add update functionality to create_ methods

anhqle:
### Summary

Add the ability to `update_if_exists` to `create_X` methods

### Questions

It seems that a Source node can't update its catalog, schema, or table. [code](https://github.com/DataJunction/dj/blob/897787fcf0cfde1a8998ce72cc9b9fd38963d983/datajunction-server/datajunction_server/internal/nodes.py#L1099). If that's true then there's the client Source's [UpdateNode](https://github.com/DataJunction/dj/blob/9a8c1263ce965b27254b666d490a80eedfd68d66/datajunction-clients/python/datajunction/nodes.py#L312) is incorrect

### Test Plan

- [ ] PR has an associated issue: #
- [x] `make check` passes
- [x] `make test` shows 100% unit test coverage

PR #1025, opened 2024-05-28
FIX Remove fields from Source's UpdateNode that aren't actually update-able

anhqle:
### Summary

Fixing the bug discussed/discovered in #1001 

### Test Plan

- [x] `make check` passes


shangyian (approved):
Makes sense, thanks!

agorajek (approved):
Nice. Thank you @anhqle

PR #1029, opened 2024-05-30
Fix cube elements ordering bug

shangyian:
### Summary

Fixes an issue where the metrics in a cube don't follow the order specified when they're passed in. For consistency, we should keep track of the ordering of metrics in a cube and use that ordering when returning a list of metrics for the cube.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Nice!

PR #1030, opened 2024-05-31
Speed up sql building by short-circuiting the compilation for CAST(...)

shangyian:
### Summary

We can speed up node SQL building for some cases by short-circuiting the compilation process for `CAST(...)` expressions, since the type of the expression is clearly determined by what the cast data type is set to.

### Test Plan

N/A

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
The low hanging fruits! :)

PR #1038, opened 2024-06-05
Copying node should rewrite dimension link join SQL

shangyian:
### Summary

When a node is copied to a new node name, we should rewrite all of its dimension links' join clauses based on the copied node's new name. Otherwise we'll create copied nodes with invalid dimension links.

### Test Plan

Locally

- [x] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Looks good. This was hard to guess before hand.

PR #1043, opened 2024-06-07
Fix two SQL building bugs

shangyian:
### Summary

This fixes two bugs during SQL building:
* Fix a bug where writing a node query that joins against the same node twice would not generate one of the joins
* Fix a bug where any node query that uses UNION ALL would not return with the right parentheses around the query components before and after the union

### Test Plan

- [ ] `make check` passes


agorajek (approved):
wow...

PR #1047, opened 2024-06-11
Fix updating cubes with the Python client

shangyian:
### Summary

Updating existing cubes doesn't work with the Python client because the `GET /nodes/<node name>/` endpoint doesn't return the right metadata for `metrics` and `dimensions`. Instead, we also make a call to `/cubes/<node name>`, which does provide that extra metadata.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -90,7 +90,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
         type_: models.NodeType,
         name: str,
         data: Dict,
-        update_if_exists: bool = False,
+        update_if_exists: bool = True,
     ) -> "Node":
         """
         Create or update a new node
@@ -112,6 +112,13 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
 
         try:
             existing_node_dict = self._get_node(name)
+            if type_ == models.NodeType.CUBE:
+                cube_dict = self._get_cube(name)
+                # This check is for the unit tests, which don't raise an exception
+                # for >= 400 status codes
+                if "name" in cube_dict:
+                    existing_node_dict["metrics"] = cube_dict["cube_node_metrics"]
+                    existing_node_dict["dimensions"] = cube_dict["cube_node_dimensions"]
         except DJClientException as e:  # pragma: no cover # pytest fixture doesn't raise
             if re.search(r"node .* does not exist", str(e)):
                 existing_node_dict = None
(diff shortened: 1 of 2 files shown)


datajunction-clients/python/datajunction/builder.py
```diff
@@ -112,6 +112,11 @@ def create_node(
 
         try:
             existing_node_dict = self._get_node(name)
+            if type_ == models.NodeType.CUBE:
+                cube_dict = self._get_cube(name)
+                if "name" in cube_dict:
```
agorajek, line 117: This check seems unnecessary. Or am I missing something?

anhqle: `_get_node` and `_get_cube` hit two different API endpoints, `/nodes/` and `/cubes/` respectively. Only the latter returns the necessary fields, specifically `metrics` and `dimensions`

datajunction-clients/python/datajunction/_internal.py
```diff
@@ -301,7 +301,12 @@ def _update_node(self, node_name: str, update_input: models.UpdateNode):
         """
         Call node update API with attributes to update.
         """
-        return self._session.patch(f"/nodes/{node_name}/", json=update_input.dict())
+        response = self._session.patch(f"/nodes/{node_name}/", json=update_input.dict())
+        if not response.status_code < 400:
+            raise DJClientException(
```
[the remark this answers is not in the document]
shangyian: Hmm, this is a good point. Let me actually remove this part, because ideally this change should just be on the test client.

datajunction-clients/python/datajunction/_internal.py
```diff
@@ -301,7 +301,12 @@ def _update_node(self, node_name: str, update_input: models.UpdateNode):
         """
         Call node update API with attributes to update.
         """
-        return self._session.patch(f"/nodes/{node_name}/", json=update_input.dict())
+        response = self._session.patch(f"/nodes/{node_name}/", json=update_input.dict())
+        if not response.status_code < 400:
```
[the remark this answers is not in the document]
shangyian: I removed this entire block. But the longer-term fix should be in the unit tests' mocking of RequestsSessionWithEndpoint

datajunction-clients/python/datajunction/builder.py
agorajek: I understand that, but I don't understand that if `self._get_cube(name)` returns w/o failure and returns some dict, do we really need to check if that dict has a key called "name" ?

shangyian: Yeah, I think I only added this because the unit tests don't raise an error. I've removed it since it works for the actual client.

agorajek: The things we do for unit tests :)

agorajek (approved):

PR #1048, opened 2024-06-11
FIX create_cube doesn't hit the right end point to retrieve existing cube node

anhqle:
### Summary

`create_cube` should hit the `/cubes/` endpoint, which returns the `metrics` and `dimensions` fields that are required to instantiate a `Cube` instance.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -111,7 +111,10 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
         }
 
         try:
-            existing_node_dict = self._get_node(name)
+            if type_ == models.NodeType.CUBE:
+                existing_node_dict = self._get_cube(name)
+            else:
+                existing_node_dict = self._get_node(name)
         except DJClientException as e:  # pragma: no cover # pytest fixture doesn't raise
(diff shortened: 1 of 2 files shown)


datajunction-clients/python/datajunction/builder.py
```diff
@@ -126,7 +132,17 @@ def create_node(
...
             # update
             if update_if_exists:
-                existing_node = node_cls(dj_client=self, **existing_node_dict)
+                # For Cube, DJ server returns metrics & dimensions under slightly different names
+                #   that need to be re-mapped here
+                if type_ == models.NodeType.CUBE:
+                    existing_node = Cube(
+                        dj_client=self,
+                        metrics=existing_node_dict.get("cube_node_metrics"),
+                        dimensions=existing_node_dict.get("cube_node_dimensions"),
```
anhqle, line 141: Does the `/cubes/` endpoint intentionally return a dict with fields like `cube_node_metrics` instead of `metrics`? 
If yes then this bit of ugly code is necessary and I'll write test to cover it.
If not then I can put up a PR for DJServer to rename the fields.

anhqle:
Close in favor of https://github.com/DataJunction/dj/pull/1047

Issue #1062, opened 2024-06-18
Generate notebook for exporting/moving nodes


shangyian:
We can generate a notebook to export multiple nodes with the Python client.

An "export" in this context means that we will generate all of the Python code needed to replicate a particular node:
- Basic create call
- Set tags on node
- Set column-level attributes
- Link dimensions on node
- Replicate materializations

**Export Cube**

This functionality will export a cube and all of the nodes in its data graph and dimensions graph into a notebook. The namespaces of all the nodes will be replaceable so that the nodes can be copied to a different namespace.

**Export Namespace**

This functionality will export all of the nodes in namespace, along with the nodes in each node's data graph and dimensions graph. The namespaces of all the nodes will be replaceable so that the nodes can be copied to a different namespace.

PR #1070, opened 2024-06-20
Generate notebook for exporting cubes / namespaces

shangyian:
### Summary

This PR adds a feature to generate a notebook for exporting cubes and namespaces. For cubes, the generated notebook will include all nodes referenced by the cube, i.e., all metrics and their upstream transforms. By default this export excludes sources and dimensions, but these can be included with a flag. For namespaces, the generated notebook includes all nodes in the namespace.

In the exported notebook, users can modify `NAMESPACE_MAPPING` to map current namespaces to new ones and rerun the notebook to recreate the nodes in those new namespaces. For each exported node with a query, we will modify the query to replace the node names with the namespace mapping as an f-string. This does mean that queries with comments will have the comments stripped in the exported versions.

These changes also switch to using jinja templates for the generated client code snippets. This makes it easier to read, as generating them through string manipulation was difficult to maintain. 

### Test Plan

Ran locally

- [x] PR has an associated issue: #1062 
- [ ] `make check` passes

### Deployment Plan

N/A


datajunction-server/tests/api/client_test.py
```diff
@@ -1,96 +1,146 @@
...
+
+def trim_trailing_whitespace(string: str) -> str:
+    """Trim trailing whitespace on each line"""
+    return "\n".join([chunk.rstrip() for chunk in string.split("\n")]).strip()
+
 
 @pytest.mark.asyncio
-async def test_generated_python_client_code_new_metric(client_with_roads: AsyncClient):
+async def test_generated_python_client_code_new_source(
+    module__client_with_roads: AsyncClient,
```
shangyian, line 34: @agorajek I switched all of these to use the `module__*` AsyncClients rather than the default ones. This one's fairly easy since the entire test should be read-only

shangyian: Never mind, I'm going to remove this from this PR in favor of the one that does this for all tests

shangyian:
@agorajek and @samredai going to merge this now to get it in, but feel free to leave comments and I'll address them in a follow-up PR

Issue #1059, opened 2024-06-18
Proposal: Collections


samredai:
Currently, the UI Explore view is a top level view that includes all nodes in the system. The search helps but the hierarchical namespace explorer side tab becomes harder to navigate as you approach 1k+ nodes. This is a proposal for a lightweight metadata concept of a node "Collection".
- Flat logical groupings of nodes (no nesting of collections)
- Nodes can be added to multiple collections
- Deleting a collection will not delete any nodes or namespaces
- UI homepage shows all collections, including a default "All Nodes" collection that essentially takes you to the current home page with an explore view of all nodes
- Each node page will have an "Add to Collection" button
- Collection pages will have an "Add Node" button

**Implementation**
I don't think we need to over complicate the implementation. I'm open to suggestions but I'm thinking a `collections` table that contains high level info on the collection (id, name, description, created_by, created_at, deleted_at) and a `collection_nodes` table that includes a mapping of collection IDs to node IDs.

shangyian:
Collections as a concept is definitely a good idea! Do you think we can reuse `tags` for saving collections in the backend though? The `tag_type` can just be `collection`. 

It would be nice if we could auto-create a collection of each user's nodes, i.e., something like `Yian's Nodes` that encompasses all of the nodes that user created. Then we can just display that on their landing page by default. And of course a user can create other collections as well, which would also show up there, but this would remove a manual step for many of them.

samredai:
I'm sure we can make it work using tags since they're pretty flexible, but I think this might make the metadata more messy in the long run since the relationship is reversed. A critical node could exist in 30 collections created by different users and having 30 tags on that node feels less intuitive than having each of those collections maintain their own pointer directly to that node. Also I suspect we'll want to continue to add more metadata around collections and I wouldn't want to shoehorn this into the tags construct. What do you think?

As for a "My Nodes" collection, I agree I think that would be super useful. I don't know that it would even need to be implemented via a collection though since it warrants a dedicated endpoint and maybe even a dedicated page in the UI. Would it be a cheap runtime query to pull all the nodes for the current user since that's stored on the node's metadata?

shangyian:
[...]

PR #1071, opened 2024-06-20
Collections API

samredai:
### Summary

This enables creating collections via the API and adding one or more nodes to them. A collection has a unique name and a description for now but we can expand it to capture more metadata later. Nodes are added to collections via a relationship table and a node can be added to multiple collections. Deleting a collection does not delete any of the nodes in it.

### Test Plan

- [x] PR has an associated issue: #1059 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/collection.py
+++ b/datajunction-server/datajunction_server/api/collection.py
@@ -1,0 +1,126 @@
+"""
+Collection related APIs.
+"""
+import logging
+from http import HTTPStatus
+from typing import List, Any
+
+from fastapi import Depends, HTTPException
+from fastapi.responses import JSONResponse
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import selectinload
+from sqlalchemy.sql.operators import is_
+
+from datajunction_server.api.helpers import (
+    get_collection_by_name,
+)
+from datajunction_server.database.collection import Collection
+from datajunction_server.database.node import Node
+from datajunction_server.errors import DJException
+from datajunction_server.internal.access.authentication.http import SecureAPIRouter
+from datajunction_server.models.collection import CollectionInfo, CreateCollection
+from datajunction_server.utils import get_session, get_settings
+
+_logger = logging.getLogger(__name__)
+settings = get_settings()
+router = SecureAPIRouter(tags=["collections"])
+
+
+@router.post("/collections/", response_model=CollectionInfo, status_code=201)
+async def create_a_collection(
+    data: CreateCollection,
+    *,
+    session: AsyncSession = Depends(get_session),
+) -> CollectionInfo:
+    """
+    Create a Collection
+    """
+    try:
+        await get_collection_by_name(session, data.name)
+    except DJException:
+        pass
+    else:
+        raise HTTPException(
+            status_code=HTTPStatus.CONFLICT,
+            detail=f"Collection already exists: `{data.name}`",
+        )
+
+    collection = Collection(
+        name=data.name,
+        description=data.description,
+    )
+    session.add(collection)
+    await session.commit()
+    await session.refresh(collection)
+
+    return CollectionInfo.from_orm(collection)
+
+@router.delete("/collections/{name}", status_code=HTTPStatus.OK)
+async def delete_a_collection(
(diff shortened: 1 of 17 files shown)


shangyian (commented):
Looks good overall! The only thing I think we should consider carefully is whether a collection's `name` should be a unique value or if collections can be tracked by `id` instead.

datajunction-server/datajunction_server/database/collection.py
```diff
@@ -0,0 +1,60 @@
...
+    A collection of nodes
+    """
+
+    __tablename__ = "collection"
+
+    id: Mapped[int] = mapped_column(
+        BigInteger().with_variant(Integer, "sqlite"),
+        primary_key=True,
+    )
+    name: Mapped[Optional[str]] = mapped_column(String, default=None)
```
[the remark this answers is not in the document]
samredai: Great catch, yes I think a unique constraint is the better of the two options you described. Updating now, thanks!

samredai:
@shangyian @agorajek this is ready for review

datajunction-server/datajunction_server/database/collection.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    A collection of nodes
+    """
+
+    __tablename__ = "collection"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[Optional[str]] = mapped_column(String, default=None, unique=True)
+    description: Mapped[Optional[str]] = mapped_column(String, default=None)
+    nodes: Mapped[List[Node]] = relationship(
```
agorajek, line 26: Since we store a list of Nodes in here, why would we need a separate `collectionnodes` table?

datajunction-server/tests/api/collections_test.py
```diff
@@ -0,0 +1,208 @@
...
+
+class TestDataForNode:
+    """
+    Test ``POST /collections/``.
+    """
+
+    @pytest.mark.asyncio
+    async def test_creating_a_collection(
+        self,
+        client_with_account_revenue: AsyncClient,
```
agorajek, line 17: I know this may add some work to your PR, but it would be nice to use the module scope fixtures on the new code already. It is easier to modify the tests when you wrote them than for a new person to learn how they are designed. No pressure. 

If you decide to do so, the patter is simple... use the same fixture but with the `module__` prefix (it should exist already) and try to make it work with your tests.

agorajek (approved):
+1 on the id/name design change. It seems like a low hanging fruit to make the collections more flexible w/o much effort and complexity, unlike with the node case.

Thanks for the quick PR after coming back!

Few small comments inline.

datajunction-server/datajunction_server/database/collection.py
shangyian: `collectionnodes` is the association table, since each collection can have many nodes and each node can have many collections.

shangyian (approved):

datajunction-server/datajunction_server/api/collection.py
```diff
@@ -0,0 +1,154 @@
...
+            detail=f"Cannot add nodes to collection, no nodes found: `{data}`",
+        )
+    collection.nodes.extend(nodes)  # type: ignore
+    session.add(collection)
+    await session.commit()
+    await session.refresh(collection)
+    return Response(status_code=HTTPStatus.NO_CONTENT)
+
+
+@router.post(
```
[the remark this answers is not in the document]
samredai: The problem there is that delete requests can't have a body and the post version of that endpoint is already taken. I'm open to any other suggestions though.

datajunction-server/tests/api/collections_test.py
```diff
@@ -0,0 +1,208 @@
+# pylint: disable=too-many-lines
+"""
+Tests for the collections API.
+"""
+import pytest
+from httpx import AsyncClient
+
+
+class TestDataForNode:
```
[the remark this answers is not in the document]
samredai: Oh right! Thanks, fixed

datajunction-server/tests/api/collections_test.py
samredai: That's right, I forgot about this new pattern. Thanks @agorajek!

PR #1077, opened 2024-06-27
Raise error if a temporal partition was not specified for a Druid materialization

shangyian:
### Summary

When someone sets up Druid materialization for a cube node, we need a timestamp column to use for ingestion to Druid. We can use the user-specified temporal partition for this, and if this is not supplied by the user, we should raise an error instead of silently failing.

Additionally, we were previously falling back to any timestamp column on the cube node (if one was available), but I think this behavior is confusing and removed it. It's better for users to be explicit about which timestamp column they want to use.

### Test Plan

- [x] `make check` passes

### DIFF
--- a/datajunction-query/docker/druid_spec.json
+++ b/datajunction-query/docker/druid_spec.json
@@ -24,7 +24,7 @@
         "type": "uniform",
         "queryGranularity": "HOUR",
         "rollup": true,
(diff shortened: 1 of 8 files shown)


agorajek (approved):
Sounds good. I was just working on the materializaton_test.py but I'll move to the next one since you are touching it here.

PR #1084, opened 2024-07-03
Use module-level fixtures for sql tests

shangyian:
### Summary

Switch to using module-level fixtures for sql tests. There were a few tests in here that made significant modifications to nodes, which I left on the non-module-level fixtures, as resetting them was not easy. 

### Test Plan

- [ ] PR has an associated issue
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/tests/api/access_test.py
+++ b/datajunction-server/tests/api/access_test.py
@@ -17,7 +17,7 @@ class TestDataAccessControl:  # pylint: disable=too-few-public-methods
     @pytest.mark.asyncio
     async def test_get_metric_data_unauthorized(
         self,
-        module__client_with_basic,
+        module__client_with_examples: AsyncClient,
     ) -> None:
         """
         Test retrieving data for a metric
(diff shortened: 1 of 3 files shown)


agorajek:
Thanks for working on this... I know how hard it is to reset the nodes sometimes so don't try too hard. From time to time I cut some corners in order to just move the PR ahead.

Looks like you still getting some errors:
```
FAILED tests/api/sql_test.py::test_sql_with_filters_orderby_no_access - RuntimeError: Caught handled exception, but response already started.
FAILED tests/api/sql_test.py::test_union_all - KeyError: 'sql'
FAILED tests/api/sql_test.py::test_multiple_joins_to_same_node - KeyError: 'sql'
FAILED tests/api/sql_test.py::test_cross_join_unnest - KeyError: 'sql'
FAILED tests/api/sql_test.py::test_lateral_view_explode - KeyError: 'sql'
FAILED tests/api/sql_test.py::test_get_sql_for_metrics_failures - assert <HTTPStatus.FORBIDDEN: 403> == 200
```

shangyian:
@agorajek I think I figured out a fix for those tests -- it was just that the override function for `validate_access` wasn't being unset, which meant that after `test_sql_with_filters_orderby_no_access` runs, the rest of the tests would fail due to being denied access to nodes.

However... I think I'm seeing a bigger issue which is that this change isn't having a huge effect on how long these tests take to run.

agorajek:
@shangyian yes the `validate_access` is a common issue I had to deal with either through a reset or through putting these tests to a separate file.

shangyian:
@agorajek this is finally ready 😄

agorajek (approved):
This is great! 1h!

PR #1085, opened 2024-07-03
Allow docs dropdown in UI with .env configured urls

samredai:
### Summary

Small change to the UI to allow expanding the docs dropdown with additional custom URLs. The purpose of this is so that internal deployments can add additional links, like URLs to internal docs.

As an example, if you set this in the .env
```env
REACT_APP_DOCS_SITES='{"Open-Source": "https://www.datajunction.io/", "Red Button": "https://clicktheredbutton.com/"}'
```

The UI would look like this:

https://github.com/DataJunction/dj/assets/43911210/75f3bc52-36ef-4c86-accb-988a1648b1fa

If you do not set this variable at all, the dropdown will default to just having the "Open-Source" item that links to the OSS docs.

### Test Plan

Ran it locally.

- [ ] `make check` passes

### DIFF
--- a/datajunction-ui/src/app/pages/Root/__tests__/index.test.jsx
+++ b/datajunction-ui/src/app/pages/Root/__tests__/index.test.jsx
@@ -39,7 +39,7 @@ describe('<Root />', () => {
         'href',
         '/sql',
       );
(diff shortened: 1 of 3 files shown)


agorajek (approved):
Neat

shangyian (approved):

PR #1086, opened 2024-07-05
Fix issue with lateral view explode sql rendering

shangyian:
### Summary

`LATERAL VIEW EXPLODE(...) abc AS a1, b1` should not have parentheses around `a1, b1`, but the current sql rendering does put parens around them, which results in an invalid Spark SQL query.

### Test Plan

- [ ] `make check` passes


anhqle (approved):

anhqle:
Thank you! Please give me a ping when merged + deployed.

Issue #1064, opened 2024-06-19
Passing authorization Parameters


bazooka720:
We use trino (interact via REST API of trino)  and trying to understand how we can pass on params like user name/pwd (or other auth params)  and other params trino support when sending query to execute. Below is link to examples but it was not clear how it would work? (Trino through ranger serves as our RBAC )
https://github.com/bazooka720/dj/tree/main/datajunction-query#executing-queries

samredai:
Hi @bazooka720! The query service is a pluggable piece of the architecture (talked about a little in the docs [here](https://datajunction.io/docs/0.1.0/deploying-dj/query-service/)), meaning you can implement the query service however you'd like including any sort of authorization. Then the core API will connect to it using any dependency injected query service client that has the expected spec (see [here](https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/api/data.py#L177)). So a typical enterprise setup today would use the pure OSS version of the core datajunction service with an internal custom query service and query service client that plugs into the core service.

I'd really like to see a robust OSS offering of a query service that's fully integrated to work with Trino that people can just configure, deploy, and use out of the box, but I don't think we're there yet. Let me know if you want to collaborate on something like that!

bazooka720:
Thanks @samredai. Just to confirm, currently you don't support Trino? (In the picture i saw trino but was not sure). Is IBIS a potential option

samredai:
To be completely accurate, the query service implementation that exists in the repo today does support sqlalchemy and allows you to pass in any sqlalchemy uri with any additional parameters to the sqlalchemy engine initialization. I've made a note to document this more clearly and add it to the repo.

What I was really trying to get at is that there is still work to be done to make it a richer experience. For example, you mentioned you're using RBAC via Ranger which implies you'll need user info to flow through to the queries submitted to the query service. We support plugging in authentication to the main service but the OSS query service does not have any support for various authorization schemes. The OSS query service assumes all queries will be submitted to Trino using a generic set of credentials that were provided for initializing the sqlalchemy engine instance.

bazooka720:
I see now. New to DJ, I really love the approach (albeit i have so many questions and trying to understand how it's all coming together). Regarding the authorization, yes...we would ideally need more of dynamic approach as our current interaction with Trino itself is via its REST API and every query sent is authorized. Is this something feasible? One alternative for us is to deploy via isolated serverless may be?

samredai:
[...]

PR #1090, opened 2024-07-08
Add Trino to DJQS, including an optional trino docker compose profile

samredai:
### Summary

This adds the official trino client to DJQS to allow sqlalchemy to process `trino://....` URIs. It also includes a couple of bug fixes related to passing in connection args. Lastly, it adds an optional trino container to the docker compose setup that only runs when you include `--profile trino` to the docker compose up command. The official trino docker container comes with the tpch catalog and the sf1 schema that has some nice tables we can use as part of a demo for modeling over trino tables.

### Test Plan

- [x] PR has an associated issue: #1064 
- [x] `make check` passes


datajunction-query/.env
```diff
@@ -0,0 +1,3 @@
+CONFIGURATION_FILE="/code/config.djqs.yml"
+DEFAULT_REFLECTION_ENGINE=trino
```
shangyian, line 2: Can this be simplified to `DEFAULT_ENGINE`?

shangyian (approved):
Nice! This looks pretty contained too 🥳

datajunction-query/.env
samredai: This attribute already existed from a previous PR, I just set it here for the demo environment. It's used in a couple of places so I'll follow up with a separate PR to change this everywhere.

shangyian: Ah ok, it's fair to just leave it then!

Issue #1037, opened 2024-06-04
Query for complex dimension join fails because it doesn't SELECT enough columns 


anhqle:
Example: Join a `fact` Transform with a `allocation` Dimension with a complex dimension link, e.g.

```
fact.event_timestamp BETWEEN allocation.window_start and allocation.window_end
```

The constructed query looks like
```
SELECT
FROM (SELECT * FROM source1) AS fact
JOIN (SELECT ... FROM source2) AS allocation
ON fact.event_timestamp BETWEEN allocation.window_start and allocation.window_end
```

It seems that DJ tries to be smart and only select what's needed inside of `SELECT ... FROM source2`. However, it's not smart enough while parsing the complex join condition to realize that `window_start` and `window_end` are needed. This causes an invalid SQL, namely `Column allocation.window_start and allocation.window_end` not found.`

PR #1091, opened 2024-07-08
Fix optimization that tries to select only the columns needed when joining in dimension

shangyian:
### Summary

This PR fixes a bug with the optimization that tries to select only the columns needed when joining in a dimension node. We fail to include the columns referenced in the filter clauses, meaning that the subsequent queries aren't built correctly. 

### Test Plan

Locally

- [x] PR has an associated issue: #1037
- [x] `make check` passes

### Deployment Plan

N/A


samredai (approved):

agorajek (approved):
Side note... server tests down to 55 min !

PR #1093, opened 2024-07-09
Add access for node and measures sql to the Python client

shangyian:
### Summary

This adds the ability to retrieve node SQL and measures SQL using the Python client. This is helpful for anyone who is debugging their nodes.

### Test Plan

Added parallelism for the Python client tests by passing in `-n auto` when running client tests. This reduces the runtime of the client tests from 28min to 14min cc @agorajek 

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-clients/python/datajunction/client.py
+++ b/datajunction-clients/python/datajunction/client.py
@@ -162,14 +162,45 @@ class DJClient(_internal.DJClient):
         filters: Optional[List[str]] = None,
         engine_name: Optional[str] = None,
         engine_version: Optional[str] = None,
+        measures: bool = False,
     ):
         """
-        Builds SQL for one or more metrics with the provided dimensions and filters.
+        Builds SQL for one or more metrics with the provided group by dimensions and filters.
         """
+        endpoint = "/sql/"
+        if measures:
+            endpoint = "/sql/measures/"
         response = self._session.get(
-            "/sql/",
+            endpoint,
             params={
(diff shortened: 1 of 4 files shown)


datajunction-clients/python/datajunction/client.py
```diff
@@ -180,6 +184,33 @@ def sql(  # pylint: disable=too-many-arguments
...
+        """
+        Builds SQL for a node with the provided dimensions and filters.
+        """
+        response = self._session.get(
+            f"/sql/{node_name}",
+            params={
+                "dimensions": dimensions or [],
+                "filters": filters or [],
+                "engine_name": engine_name or self.engine_name,
+                "engine_version": engine_version or self.engine_version,
```
samredai, line 207: Versions for engines are optional but this line would prevent us from selecting an engine where engine_version="" since `"" or self.engine_version` would always resolve to `self.engine_version`. I ran into this problem with similar logic in the query service and fixed it in the last PR (see [these](https://github.com/DataJunction/dj/blob/main/datajunction-query/djqs/api/tables.py#L37-L46) lines). I was trying to use engine_name of "default" and engine_version of "" and couldn't.

samredai (approved):

datajunction-clients/python/datajunction/client.py
shangyian: Good catch! Fixing this now

PR #1095, opened 2024-07-10
Retrieve Cube SQL from `/sql/{node}`

shangyian:
### Summary

This PR adds the ability to retrieve SQL for a cube using `/sql/{node}`, since this endpoint is meant to work for all nodes, including cube nodes. When we get a request for cube SQL, we save both a node-type query request and a metrics-type query request, so that if either query request type comes in again, we can reuse the same generated SQL.

### Test Plan

- [ ] `make check` passes


samredai (approved):
Thanks @shangyian!

PR #1096, opened 2024-07-10
Add users endpoints for retrieving nodes by user

shangyian:
### Summary

This PR adds two `/users` endpoints, one for retrieving all users and the other for retrieving nodes for a given user.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/users.py
+++ b/datajunction-server/datajunction_server/api/users.py
@@ -1,0 +1,62 @@
+"""
+User related APIs.
+"""
+
+from typing import List
+
+from fastapi import Depends, Query
+from sqlalchemy import distinct, select
+from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import joinedload, selectinload
+
+from datajunction_server.database.column import Column
+from datajunction_server.database.history import ActivityType, EntityType, History
+from datajunction_server.database.node import Node, NodeRevision
+from datajunction_server.internal.access.authentication.http import SecureAPIRouter
+from datajunction_server.models.node import NodeMinimumDetail
+from datajunction_server.utils import get_session, get_settings
+
+settings = get_settings()
+router = SecureAPIRouter(tags=["users"])
+
+
+@router.get("/users/{username}", response_model=List[NodeMinimumDetail])
+async def list_nodes_by_username(
+    username: str,
+    *,
+    session: AsyncSession = Depends(get_session),
+    activity_types: List[str] = Query([ActivityType.CREATE, ActivityType.UPDATE]),
+) -> List[NodeMinimumDetail]:
+    """
+    List all nodes created by the user
+    """
+    statement = select(distinct(History.entity_name)).where(
+        (History.user == username)
+        & (History.entity_type == EntityType.NODE)
+        & (History.activity_type.in_(activity_types)),
+    )
+    result = await session.execute(statement)
+    nodes = await Node.get_by_names(
+        session=session,
+        names=list(set(result.scalars().all())),
+        options=[
+            joinedload(Node.current).options(
+                selectinload(NodeRevision.cube_elements)
+                .selectinload(Column.node_revisions)
+                .options(
+                    selectinload(NodeRevision.node),
+                ),
+            ),
+        ],
+    )
+    return [node.current for node in nodes]
+
+
+@router.get("/users", response_model=List[str])
+async def list_users(session: AsyncSession = Depends(get_session)) -> List[str]:
+    """
+    List all users
+    """
+    statement = select(distinct(History.user))
+    result = await session.execute(statement)
+    return list({user for user in result.scalars().all() if user})
--- a/datajunction-server/datajunction_server/api/main.py
+++ b/datajunction-server/datajunction_server/api/main.py
@@ -42,6 +42,7 @@ from datajunction_server.api import (
     nodes,
     sql,
     tags,
+    users,
 )
 from datajunction_server.api.access.authentication import whoami
 from datajunction_server.api.attributes import default_attribute_types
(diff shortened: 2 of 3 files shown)


samredai (approved):
@shangyian you mentioned it's not ideal to use the History table for this and I agree. It might be useful to add a `TODO` somewhere and maybe we can open an issue on the implementation we eventually want.

The trickiest part is that the specific DJ implementation handles user auth by injecting the `get_user` and so the open source code can't guarantee that a user will be found in the `users` table, even though it exists. I wonder if we can stack the dependencies (like in [this](https://fastapi.tiangolo.com/tutorial/dependencies/sub-dependencies/#second-dependency-dependable-and-dependant) example in the fastapi docs). We could have an `upsert_get_user` dependency that saves the user to the database if they don't exist yet, and then just calls the `get_user` dependency. Custom auth implementations would be able to continue injecting `get_user` without any change, but the OSS logic would handle saving the user to the database if they don't exist, always ensuring that the core DJ service has a user entity it can tie things to. What do you think?

shangyian:
@samredai actually I'm going to hold off on merging this, because I think internally we can get our implementation to match the expected OSS functionality of storing users in the `users` table, and therefore the `GET /users` endpoint should just list the items in that table. I mostly implemented this as a shortcut solution, but we don't need it right away. (And if we do need this shortcut, the internal API should just bypass the open source one)

samredai:
@shangyian that makes sense. What I was getting at though is that outside of the actual auth logic, there is some logic that will be identical across all implementations, even those with custom auth. Specifically I'm referring to after the user is authenticated when we simply add them to the users table if they don't exist already. Since everyone implementation will do that in exactly the same way, we can include that as common logic that the project handles. Since we're unblocked and have some time, let me put a quick example PR up and we can see if it makes sense.

shangyian:
@samredai but is that actually a strategy we want to generalize? I think the flow that makes sense to me is having users register through some flow (in this case we have two options implemented), and creating new users in that way. I'm not sure that we should just be randomly creating a user that was injected. Instead it seems like we should reject access to DJ if the user doesn't already exist (which we didn't add before since that would break us internally)

samredai:
> Instead it seems like we should reject access to DJ if the user doesn't already exist (which we didn't add before since that would break us internally)

This would still be true if we order the dependency chain that way--the user would need to go through authentication first before we run the upsert. In other words, we'd never add a user to the user table that doesn't exist.

shangyian:
> This would still be true if we order the dependency chain that way--the user would need to go through authentication first before we run the upsert. In other words, we'd never add a user to the user table that doesn't exist.

Ah ok, I see what you mean. It'll be convenient to have this option here, but I'm still wondering if it makes sense to make each user go through an explicit registration process rather than an implicit one as a part of the authentication phase.

samredai:
@shangyian I've opened an implementation of what I was describing. I'm not in a rush to get it merged and it would be good to have a few eyes on it to see if we want to go this way. We should discuss it in the next community sync.

> but I'm still wondering if it makes sense to make each user go through an explicit registration process rather than an implicit one as a part of the authentication phase.

I might be misunderstanding here but whether a deployment is configured to make users go through a registration process (like a sign up form) or to just automatically get detected by some other system is completely up to the implementers to decide as they can today. What I'm proposing doesn't change any of that, it simply catches the scenario where a custom auth implementation forgets to also save a representative user record for DJ to have and use for lots of things. If a custom auth implementation goes the extra mile and saves the user to the `users` table, then this new logic will effectively be a no-op and just pass through the user object.

shangyian:
> What I'm proposing doesn't change any of that, it simply catches the scenario where a custom auth implementation forgets to also save a representative user record for DJ to have and use for lots of things. If a custom auth implementation goes the extra mile and saves the user to the users table, then this new logic will effectively be a no-op and just pass through the user object.

Right, I'm just wondering if we're better off assuming (and recommending) that the custom auth implementation should save the user to the users table. It seems like that's a cleaner way to do it, although of course having what you proposed as a backup is still useful

agorajek (approved):
One day we should have a created_by attribute but it looks good for now.

datajunction-server/datajunction_server/api/users.py
```diff
@@ -0,0 +1,91 @@
...
+from datajunction_server.internal.access.authentication.http import SecureAPIRouter
+from datajunction_server.models.node import NodeMinimumDetail
+from datajunction_server.models.user import UserActivity, UserOutput
+from datajunction_server.utils import get_current_user, get_session, get_settings
+
+settings = get_settings()
+router = SecureAPIRouter(tags=["users"])
+
+
+@router.get("/users/me", response_model=UserOutput)
```
samredai, line 25: Isn't this is a duplicate of the `/whoami` endpoint? If we add this here we should remove that one.

datajunction-server/datajunction_server/api/users.py
```diff
@@ -0,0 +1,91 @@
...
+                .options(
+                    selectinload(NodeRevision.node),
+                ),
+            ),
+        ],
+    )
+    return [node.current for node in nodes]
+
+
+@router.get("/users", response_model=List[UserActivity])
```
samredai, line 69: Should we split this out into two endpoints? The activity count is useful but I'd imagine most requests for a list of users won't need the activity count so the join seems wasteful to do all the time. Maybe we can have `/users` and `/users/activity`.

shangyian: Agreed, let me do a follow-up PR here

datajunction-server/datajunction_server/api/users.py
shangyian: Good point, I'll remove this one

PR #1100, opened 2024-07-11
Make add namespace popover less confusing 

shangyian:
### Summary

This PR makes the add namespace popover less confusing with the following changes:
* Pre-populating the form field with the namespace that the user is currently in, if any
* Changing the icon from an edit icon to an add icon

Screenshot 2024-07-10 at 8 01 49 PM

### Test Plan

Locally

- [x] PR has an associated issue: #1097 
- [x] `make check` passes

### Deployment Plan

N/A


samredai (approved):
This makes sense!

PR #1101, opened 2024-07-12
Upsert users who are authenticated but not stored in the user table

samredai:
### Summary

This adds a `get_current_user_and_upsert` dependable that is also a dependent that relies on the `get_current_user` dependable that already existed (Am I the only one confused by this terminology?). Essentially all of this dependency magic turns into a middleware chain where:
- First the `get_current_user` dependency runs as it does today. That means any internal deployments that have injected their own authentication middleware that overrides this dependency will still work exactly as it does today.
- Next on the chain is the `get_current_user_and_upsert` function which will do a simple upsert to insert the user into the DJ metadata `users` table, if it does not already exist (based on the username)

Below describes what will happen in a few of the typical auth scenarios:

#### Scenario 1: Basic Auth (no change)
- When a user goes through the basic auth signup flow, the user is inserted into the DJ metadata `users` table.
- When the user logs in, they're authenticated and the encrypted cookie is stored
- On authenticated requests, the cookie is processed by the backend and the user is retrieved from the `users` table
- The `get_current_user_and_upsert` is essentially a no-op since the user already exists in the `users` table

#### Scenario 2: Google OAuth (no change)
- A user signs in with google through the google oauth flow
- If it's the user's first time, they're inserted into the `users` table
- An encrypted cookie is stored on the client
- On authenticated requests, the cookie is processed by the backend and the user is retrieved from the `users` table
- The `get_current_user_and_upsert` is essentially a no-op since the user already exists in the `users` table

#### Scenario 3: `get_current_user` override to use custom auth managed by an external system (only setup impacted by this PR)
- `get_current_user` returns a `User` object instantiated from data retrieved from some system external to DJ
- `get_current_user_and_upsert` runs to guarantee the user is stored in the DJ deployments `users` table

It's true that in Scenario 3, the internal logic in the `get_current_user` override should handle the upsert on it's own, but there's no guarantee of that since the `User` instance could simply be constructed in memory based on the propagated user identity and never saved to the database. This PR makes it so that custom auth implementations never have to concern themselves with making sure they save the user to the DJ metadata database, the open source logic will automatically handle that, guaranteeing that we always have a DJ user entity to tie things to in the system.

### Test Plan

Added some tests and tested it in the local docker environment

- [ ] `make check` passes

### Deployment Plan

Shouldn't effect existing deployments nor require any changes

### DIFF
--- a/datajunction-server/datajunction_server/utils.py
+++ b/datajunction-server/datajunction_server/utils.py
@@ -261,7 +261,7 @@ async def get_current_user(request: Request) -> Optional["User"]:
     return None  # pragma: no cover
 
 
-async def get_current_user_and_upsert(
+async def get_and_update_current_user(
     session: AsyncSession = Depends(get_session),
     current_user: Optional["User"] = Depends(get_current_user),
 ) -> Optional["User"]:
--- a/datajunction-server/datajunction_server/utils.py
+++ b/datajunction-server/datajunction_server/utils.py
@@ -8,11 +8,13 @@ import re
 from functools import lru_cache
 
 # pylint: disable=line-too-long
-from typing import TYPE_CHECKING, AsyncIterator, List, Optional
+from typing import AsyncIterator, List, Optional
 
 from dotenv import load_dotenv
+from fastapi import Depends
 from rich.logging import RichHandler
 from sqlalchemy import AsyncAdaptedQueuePool
+from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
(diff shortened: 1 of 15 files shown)


shangyian (approved):
I think this makes sense. We might as well have a decent backup plan in case the auth implementation doesn't save a user to the users table!

datajunction-server/datajunction_server/utils.py
```diff
@@ -261,4 +261,32 @@ async def get_current_user(request: Request) -> Optional["User"]:
     return None  # pragma: no cover
 
 
+async def get_current_user_and_upsert(
```
agorajek, line 264: This looks good. But why not use more english here? Say `get_and update_current_user` ?

agorajek (approved):
This looks good. Would you consider copying your PR summary to some dev doc page?

Also made one naming suggestion in line.

datajunction-server/datajunction_server/utils.py
samredai: Updated!

samredai:
> Would you consider copying your PR summary to some dev doc page?

@agorajek good suggestion, I added a section at the end of the Authentication page.

samredai:
Related to PR #1096

PR #1107, opened 2024-07-15
Add find nodes query to GraphQL

shangyian:
### Summary

This PR adds a `findNodes` query to GraphQL, which allows users to find nodes based on the following input parameters:
* `fragment`: A fragment of a node name to search for
* `names`: Filters to nodes with the requested node names
* `nodeTypes`: Filters to nodes with the requested node types
* `tags`: Filters to nodes tagged with the requested tags

For the REST API, we structured the output based on the endpoint - for example, using the `/cubes` endpoint to retrieve a cube node would return extra metadata fields pertaining to cubes, using the `/nodes/{node}/dag` endpoint would return only the amount of node metadata needed to power the graph view in the UI. However, with GraphQL, we can use a single output structure that looks very similar to the actual database schemas, and users can request the specific fields they need.

We also optimize the database query used to retrieve nodes so that we only perform joins to retrieve fields if the user requested them. For REST, our strategy was simply to create different endpoints, each of which would resolve to optimized database queries, but this setup allows for a much higher degree of customization from the user.

Note that the setup doesn't use the [experimental Pydantic support from Strawberry](https://strawberry.rocks/docs/integrations/pydantic) for a few reasons:
* The structure of the GraphQL schemas are significantly different from the REST API outputs.
* Using Pydantic schemas does not allow us to customize the GraphQL resolver that builds the database query for retrieving the node (the pydantic-sqlalchemy integration will always result in us retrieving everything, which is expensive)

### Test Plan

Tested locally + added unit tests

- [ ] `make check` passes


shangyian:
@CircArgs it's for these two reasons:
> Note that the setup doesn't use the [experimental Pydantic support from Strawberry](https://strawberry.rocks/docs/integrations/pydantic) for a few reasons:
> 
> * The structure of the GraphQL schemas are significantly different from the REST API outputs.
> * Using Pydantic schemas does not allow us to customize the GraphQL resolver that builds the database query for retrieving the node (the pydantic-sqlalchemy integration will always result in us retrieving everything, which is expensive)

CircArgs:
> @CircArgs it's for these two reasons:
> 
> > Note that the setup doesn't use the [experimental Pydantic support from Strawberry](https://strawberry.rocks/docs/integrations/pydantic) for a few reasons:
> > 
> > * The structure of the GraphQL schemas are significantly different from the REST API outputs.
> > * Using Pydantic schemas does not allow us to customize the GraphQL resolver that builds the database query for retrieving the node (the pydantic-sqlalchemy integration will always result in us retrieving everything, which is expensive)

Sorry! I skipped that part😅

CircArgs (approved):

PR #1109, opened 2024-07-16
Fix SQL building with filters

shangyian:
### Summary

This fixes an issue where building SQL for a transform with filters turns any linked dimensions into an `INNER JOIN`.
For example, when we request SQL for a transform node with a filter, the generated SQL looks like this:
```
SELECT ...
FROM transform
LEFT JOIN dimension ON ...
WHERE transform.colA = 1 AND transform.colB = 'b'
```

Instead, we should see if the particular filters can be pushed down to the transform subquery and apply them there if possible. Only in cases where it's not possible to apply filters within the transform subquery should we pull it back into the outer layer. The new query structure looks like this:

```
(
SELECT ...
FROM transform
WHERE transform.colA = 1 AND transform.colB = 'b'
)
LEFT JOIN dimension ON ...
```

### Test Plan

- [x] PR has an associated issue: #1102 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -214,6 +214,7 @@ async def get_query(  # pylint: disable=too-many-arguments
         limit=limit,
         build_criteria=build_criteria,
         access_control=access_control,
+        top_level=True,
     )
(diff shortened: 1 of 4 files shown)


agorajek:
Real cool. This will make DB engines more happy.

agorajek (approved):

PR #1113, opened 2024-07-22
Require current_user everywhere, no longer optional

samredai:
### Summary

This changes the typing and some logic everywhere to make user required. After this PR, user is no longer optional and any auth implementation will at a minimum have to pass through a `User` instance.

Running DJ without auth will now require a custom override of the auth logic that injects a `User` instance. Also when you're using the client, if the server just uses basic auth you can either provided the username and password in the client code or optionally you can set `DJ_USER` and `DJ_PWD` as environment variables. This isn't too hard but I've opened issue #1118 #1119 to keep track of adding docs for both of these.

### Test Plan

Updated existing tests.

- [ ] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/_internal.py
+++ b/datajunction-clients/python/datajunction/_internal.py
@@ -2,6 +2,7 @@
 
 # pylint: disable=redefined-outer-name, import-outside-toplevel, too-many-lines
 import logging
+import os
 import platform
 import warnings
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, TypedDict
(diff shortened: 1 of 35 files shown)


shangyian (approved):
LGTM! At some point we should have a different version of the Modeling the Roads Example Database notebook that leverages the Python client instead of direct API calls that showcases this.

agorajek (approved):
This is great! Thanks @samredai

samredai:
> LGTM! At some point we should have a different version of the Modeling the Roads Example Database notebook that leverages the Python client instead of direct API calls that showcases this.

Agreed! I've also been wanting to add a separate test job that goes through the notebook end to end and it could replace the integration tests entirely. I'll open an issue to track that.

Issue #1111, opened 2024-07-19
SQL generated for `EXPLODE` missing a parenthesis


anhqle:
In SparkSQL,

- EXPLODE requires a parenthesis around the col names, i.e. `EXPLODE(my_map) AS (col1, col2)`, but
- LATERAL VIEW EXPLODE is only valid without a parenthesis, i.e. `LATERAL VIEW EXPLODE(my_map) AS col1, col2`

PR #1086 removes the parenthesis, which makes the 2nd case work but not the 1st.

PR #1114, opened 2024-07-22
Bugfix, make sure EXPLODE includes parenthesis around column list

samredai:
### Summary

This fixes a bug where EXPLODE would not surround columns lists when rendering. For example, it would render in the projection as `EXPLODE(my_map) AS col1, col2` when it should actually be `EXPLODE(my_map) AS (col1, col2)`. Thanks to @anhqle for catching this.

### Test Plan
Added a test that explicitly checks for this syntax when using `EXPLODE`.

- [x] PR has an associated issue: #1111 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -2296,11 +2296,15 @@ class FunctionTable(FunctionTableExpression):
             if self.column_list
             else ""
         )
-        column_list_str = (
-            f"({cols})"
-            if self.name.name.upper() in ("UNNEST", "EXPLODE")
-            else f"{cols}"
-        )
+
+        column_parens = False
+        if self.name.name.upper() == "UNNEST" or (
+            self.name.name.upper() == "EXPLODE"
+            and not isinstance(self.parent, LateralView)
+        ):
+            column_parens = True
+
+        column_list_str = f"({cols})" if column_parens else f"{cols}"
         args_str = f"({', '.join(str(col) for col in self.args)})" if self.args else ""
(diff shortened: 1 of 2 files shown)


agorajek (approved):
Nice catch!

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -2296,9 +2296,15 @@ def __str__(self) -> str:
             if self.column_list
             else ""
         )
-        column_list_str = (
-            f"{cols}" if self.name.name.upper() != "UNNEST" else f"({cols})"
-        )
+
+        column_parens = False
+        if self.name.name.upper() == "UNNEST" or (
```
shangyian, line 2301: Hmm, does this still work for `LATERAL VIEW EXPLODE`, which is only valid without parentheses, i.e. `LATERAL VIEW EXPLODE(somemap) AS col1, col2`?

samredai: Yes it does still work in that case. If it's an `EXPLODE` function call I check whether the immediate parent in the ast is a `LateralView`:
```py
self.name.name.upper() == "EXPLODE" and not isinstance(self.parent, LateralView)
```
so it would include parens when it's just an `EXPLODE` of a map in a projection but would not include parens if it's an `EXPLODE` as part of a `LATERAL VIEW EXPLODE` in a from clause.

shangyian: Sounds good, thanks!

shangyian (approved):

PR #1115, opened 2024-07-23
feature: Dynamic auth params

sadath-12:
### Summary

Adding dynamic auth params from dj server straight to query service

- [ ] `make check` passes

### DIFF
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -28,10 +28,6 @@ from djqs.typing import ColumnType, Description, SQLADialect, Stream, TypeEnum
 
 _logger = logging.getLogger(__name__)
 
-QUERY_USER = None
-QUERY_PASSWORD = None
-HTTP_SCHEME = None
-
 
 def get_columns_from_description(
     description: Description,
@@ -83,15 +79,12 @@ def create_trino_engine(engine: Engine, headers: Optional[Dict[str, str]]) -> "E
     path = parsed_uri.path.lstrip("/")
     catalog, schema = path.split("/", 1) if "/" in path else (path, None)
 
-    global QUERY_USER, QUERY_PASSWORD, HTTP_SCHEME
+    query_user = headers.get("QUERY_USER") or engine.extra_params["user"]
+    query_password = headers.get("QUERY_PASSWORD") or engine.extra_params["password"]
+    http_scheme = headers.get("TRINO_HTTP_SCHEME") or engine.extra_params["http_scheme"]
 
-    # Override user and password from headers if available
-    if headers and "QUERY_USER" in headers and "QUERY_PASSWORD" in headers:
-        QUERY_USER = headers.get("QUERY_USER")
-        QUERY_PASSWORD = headers.get("QUERY_PASSWORD")
-
-    if HTTP_SCHEME == "https":
-        engine_str = f"{scheme}://{QUERY_USER}:{QUERY_PASSWORD}@{host}:{port}/{catalog}/{schema}?protocol={HTTP_SCHEME}"
+    if http_scheme == "https" and query_user and query_password:
+        engine_str = f"{scheme}://{query_user}:{query_password}@{host}:{port}/{catalog}/{schema}?protocol={http_scheme}"
         return create_engine(engine_str)
     else:
         engine_str = engine.uri
@@ -120,11 +113,6 @@ def run_query(
 
     parsed_uri = urlparse(engine.uri)
     scheme = parsed_uri.scheme
-
-    global QUERY_USER, QUERY_PASSWORD, HTTP_SCHEME
-    QUERY_USER = engine.extra_params["user"]
-    QUERY_PASSWORD = engine.extra_params["password"]
-    HTTP_SCHEME = "https"
 
     if engine.type == EngineType.DUCKDB:
         conn = (
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -88,9 +88,6 @@ def run_query(
         .where(Engine.version == query.engine_version),
     ).one()
 
-    parsed_uri = urlparse(engine.uri)
-    scheme = parsed_uri.scheme
-
     query_server = headers.get("QUERY_ENGINE_SERVER")
 
     if query_server:
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -5,6 +5,9 @@ Query related functions.
 import logging
 import os
 from datetime import datetime, timezone
+from http import HTTPStatus
+from http.client import HTTPException
+from pathlib import PurePosixPath
 from typing import Dict, List, Optional, Tuple
 from urllib.parse import urlparse
 
@@ -76,8 +79,14 @@ def create_trino_engine(engine: Engine, headers: Optional[Dict[str, str]]) -> "E
     scheme = parsed_uri.scheme
     host = parsed_uri.hostname
     port = parsed_uri.port
-    path = parsed_uri.path.lstrip("/")
-    catalog, schema = path.split("/", 1) if "/" in path else (path, None)
+    path_parts = PurePosixPath(parsed_uri.path).parts
+    if len(path_parts) < 2:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail="Cannot create a Trino engine, catalog or schema missing",
+        )
+    catalog = path_parts[1] if len(path_parts) > 1 else None
+    schema = path_parts[2] if len(path_parts) > 2 else None
 
     query_user = headers.get("QUERY_USER") or engine.extra_params["user"]
     query_password = headers.get("QUERY_PASSWORD") or engine.extra_params["password"]
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -85,7 +85,7 @@ def run_query(
         .where(Engine.version == query.engine_version),
     ).one()
 
-    query_server = headers.get("QUERY_ENGINE_SERVER")
+    query_server = headers.get("SQLALCHEMY_URI")
 
     if query_server:
         sqla_engine = create_engine(query_server)
--- a/datajunction-query/djqs/engine.py
+++ b/datajunction-query/djqs/engine.py
@@ -5,7 +5,8 @@ Query related functions.
 import logging
 import os
 from datetime import datetime, timezone
-from typing import List, Tuple
+from typing import Dict, List, Optional, Tuple
+from urllib.parse import urlparse
 
 import duckdb
 import snowflake.connector
@@ -25,7 +26,14 @@ from djqs.models.query import (
 )
 from djqs.typing import ColumnType, Description, SQLADialect, Stream, TypeEnum
 
+# Determine the base directory relative to the script
+
+
 _logger = logging.getLogger(__name__)
+
+QUERY_USER = None
+QUERY_PASSWORD = None
+HTTP_SCHEME = None
 
 
 def get_columns_from_description(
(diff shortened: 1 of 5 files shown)


datajunction-query/config.djqs.yml
```diff
@@ -5,13 +5,22 @@ engines:
...
       location: /code/docker/default.duckdb
+  # - name: trino
+  #   version: 451
+  #   type: sqlalchemy
+  #   uri: trino://domain:443/tpch/sf1
+  #   extra_params:
+  #     http_scheme: https
+  #     user: admin
+  #     password: 123
+  #     auth_type: dynamic
```
samredai, line 16: These commented lines were added by accident I'm assuming.

samredai (commented):
Thanks @sadath-12! This looks good to me but I left a few comments. You'll also need to run `make lint` in the `datajunction-query/` directory and commit those changes so that the linters/formatters will pass.

datajunction-query/config.djqs.yml
sadath-12: oh I added them intentionally to give an example

sadath-12:
@samredai I want to get user and pass from session of the current logged in user under query service code but it seems like we need to pass USER model importing from server which throws error  . I'm working on python project for first time so not sure how modules , dependencies .. etc work . So

datajunction-query/djqs/engine.py
```diff
@@ -100,7 +116,11 @@ def run_query(
         cur = conn.cursor()
 
         return run_snowflake_query(query, cur)
-    sqla_engine = create_engine(engine.uri, connect_args=engine.extra_params)
+    
+    if engine.type == EngineType.SQLALCHEMY:
```
samredai, line 120: Not sure this is correct. If the engine type is sqlalchemy, it doesn't necessarily mean it's using trino. If you want to call this specific trino engine init, I would inspect the `uri`.

datajunction-query/config.djqs.yml
samredai: I think it would be better to put that in the docs, maybe in the authentication section you can start a section on auth between the query service and query engines, starting with Trino. Also I see `auth_type` here but don't see where that's used in the code.

samredai:
> @samredai I want to get user and pass from session of the current logged in user under query service code but it seems like we need to pass USER model importing from server which throws error . I'm working on python project for first time so not sure how modules , dependencies .. etc work . So

The user object on the server is the user that's logged into datajunction. It's possible that the username and password is the same as the one defined in Trino, but that's not guaranteed for every implementation, right? I'm not sure of the details of your specific setup but maybe you should have a dedicated user for your query service that has user impersonation privileges that it applies (`SET SESSION AUTHORIZATION username`) according to the user value passed from the main server to the query server.

datajunction-query/djqs/engine.py
```diff
@@ -65,6 +65,22 @@ def get_columns_from_description(
...
+    password = engine.extra_params['password']
+    scheme, rest = uri.split("://")
+    host_port, catalog_schema = rest.split("/", 1)
+    catalog , schema = catalog_schema.split('/')
+    host, port = host_port.split(":")
+    if scheme == "https":
+        engine_str = f"trino://{user}:{password}@{host}:{port}/{catalog}/{schema}?protocol={scheme}"
+        return create_engine(engine_str)
+    else:
+        return create_engine(engine.uri,engine.extra_params)
```
shangyian, line 83: There can probably be a check here to make sure that the engine.uri actually starts with `trino` right?

shangyian (commented):
I wonder if this would be clearer if in cases where the URI is not provided, we can use the Engine's `dialect` field to build the URI. That would make providing the URI optional, and instead we can construct the right DB URI based on the configured engine's dialect and extra params.

sadath-12:
> > @samredai I want to get user and pass from session of the current logged in user under query service code but it seems like we need to pass USER model importing from server which throws error . I'm working on python project for first time so not sure how modules , dependencies .. etc work . So
> 
> The user object on the server is the user that's logged into datajunction. It's possible that the username and password is the same as the one defined in Trino, but that's not guaranteed for every implementation, right? I'm not sure of the details of your specific setup but maybe you should have a dedicated user for your query service that has user impersonation privileges that it applies (`SET SESSION AUTHORIZATION username`) according to the user value passed from the main server to the query server.

sadath-12:
> > @samredai I want to get user and pass from session of the current logged in user under query service code but it seems like we need to pass USER model importing from server which throws error . I'm working on python project for first time so not sure how modules , dependencies .. etc work . So
> 
> The user object on the server is the user that's logged into datajunction. It's possible that the username and password is the same as the one defined in Trino, but that's not guaranteed for every implementation, right? I'm not sure of the details of your specific setup but maybe you should have a dedicated user for your query service that has user impersonation privileges that it applies (`SET SESSION AUTHORIZATION username`) according to the user value passed from the main server to the query server.

Yes , that's why I provided ```auth_type``` field . If its dynamic then the creds of server and query service is same else it will pick up from config for query service

samredai:
> Yes , that's why I provided `auth_type` field . If its dynamic then the creds of server and query service is same else it will pick up from config for query service .

What I'm worried about here is telling someone to make sure they use the password "abc" when they're creating their datajunction user because the password for their other user account in Trino is "abc". That seems like it might cause a few problems. Is it a trusted connection between your query server deployment and your main dj server deployment? If so, then when the main server makes requests to the query server we can pass along the encrypted JWT. When the query server decrypts it, it will get the username but the password will be hashed and probably useless with respect to Trino. The query service can then create a connection to Trino for that user. The remaining open question then is where does the query service get the Trino username and password from? The only two options I can think of is the query service database has all of the Trino username and passwords, but that's not risk-free and probably annoying to maintain. Which leads me to believe that [Trino Session Authorization](https://trino.io/docs/current/sql/set-session-authorization.html) is most likely what we need here.

> So the problem was more about referencing the modules . Should we first remove the dedups of database models from server query service or we create user model again in query service now

I don't think we need to maintain DRY across the services. If we go with session authorization, then the query service won't be storing any user info in the database and also won't be sending user data, so we shouldn't need any sort of `User` pydantic model or ORM model.

sadath-12:
Hi @samredai  @shangyian  . Confirming the request auth headers approach .

samredai:
> Hi @samredai @shangyian . Confirming the request auth headers approach , to proceed further

This is the right approach but I noticed there are many areas where the query service client is used and sometimes passed into functions to be used in logic tucked into the call stack. So I opened PR #1120 to apply the same approach here but to the entire codebase. If you rebase this after that's merged, then we can focus this PR on getting the Trino logic in the query service situated.

datajunction-query/djqs/engine.py
```diff
@@ -66,9 +66,30 @@ def get_columns_from_description(
...
 
 
+def create_trino_engine(engine: Engine, headers: Optional[Dict[str, str]]) -> "Engine":
+    """
+    Create a SQLAlchemy engine for Trino.
+    """
+    uri = engine.uri
+    user = engine.extra_params["user"]
+    password = engine.extra_params["password"]
+    scheme, rest = uri.split("://")
```
shangyian, line 76: Can you use `urlparse` here instead? https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse 

Then extracting the `scheme`, `host`, and `port` from the uri can be much simplified: 
```
from urllib.parse import urlparse
parsed = urlparse(engine.uri)
print(parsed.scheme, parsed.port, parsed.hostname)
```

datajunction-query/djqs/engine.py
```diff
@@ -66,9 +66,30 @@ def get_columns_from_description(
     return columns
 
 
+def create_trino_engine(engine: Engine, headers: Optional[Dict[str, str]]) -> "Engine":
+    """
+    Create a SQLAlchemy engine for Trino.
+    """
+    uri = engine.uri
+    user = engine.extra_params["user"]
```
shangyian, line 74: Can these arguments to `extra_params` be pulled out to a set of constants at the top of the file? It'll be easier to refer to them and also see what the expected keys in `extra_params` are.

shangyian (commented):
Overall the other changes look good, I think you just have to rebase on the latest main and run `make check` to get the linters to pass.

sadath-12:
Looks like you took care of passing headers everywhere so I resetted from main and just modified query service to add that on trino

datajunction-query/djqs/engine.py
```diff
@@ -66,22 +71,61 @@ def get_columns_from_description(
...
     engine = session.exec(
         select(Engine)
         .where(Engine.name == query.engine_name)
         .where(Engine.version == query.engine_version),
     ).one()
+
+    parsed_uri = urlparse(engine.uri)
+    scheme = parsed_uri.scheme
+
+    global QUERY_USER, QUERY_PASSWORD, HTTP_SCHEME
```
samredai, line 124: Why are these global variables?

datajunction-query/djqs/engine.py
```diff
@@ -66,22 +71,61 @@ def get_columns_from_description(
...
+    scheme = parsed_uri.scheme
+    host = parsed_uri.hostname
+    port = parsed_uri.port
+    path = parsed_uri.path.lstrip("/")
+    catalog, schema = path.split("/", 1) if "/" in path else (path, None)
+
+    global QUERY_USER, QUERY_PASSWORD, HTTP_SCHEME
+
+    # Override user and password from headers if available
+    if headers and "QUERY_USER" in headers and "QUERY_PASSWORD" in headers:
```
samredai, line 89: Instead of doing this, you can just do `query_user = headers.get("QUERY_USER")` and `query_password = headers.get("QUERY_PASSWORD")` and then later when you use them you can check if they're `None`.
```py
    if HTTP_SCHEME == "https" and query_user and query_password:
        engine_str = f"{scheme}://{QUERY_USER}:{QUERY_PASSWORD}@{host}:{port}/{catalog}/{schema}?protocol={HTTP_SCHEME}"
        return create_engine(engine_str)
    else:
        engine_str = engine.uri
        return create_engine(engine_str, connect_args=engine.extra_params)
```

datajunction-query/djqs/engine.py
```diff
@@ -66,22 +71,61 @@ def get_columns_from_description(
...
         .where(Engine.version == query.engine_version),
     ).one()
+
+    parsed_uri = urlparse(engine.uri)
+    scheme = parsed_uri.scheme
+
+    global QUERY_USER, QUERY_PASSWORD, HTTP_SCHEME
+    QUERY_USER = engine.extra_params["user"]
+    QUERY_PASSWORD = engine.extra_params["password"]
+    HTTP_SCHEME = "https"
```
samredai, line 127: Instead of hard-coding this here, can you include that in the request headers and then parse them out here? `http_scheme = headers.get("TRINO_HTTP_SCHEME")`

samredai:
@sadath-12 you still need to pass the request headers from the fastapi endpoint into the internal functions [here](https://github.com/DataJunction/dj/blob/main/datajunction-query/djqs/api/queries.py#L102), [here](https://github.com/DataJunction/dj/blob/main/datajunction-query/djqs/api/queries.py#L150) and [here](https://github.com/DataJunction/dj/blob/main/datajunction-query/djqs/api/queries.py#L155) so that it makes it to the logic that instantiates the trino connection. So `request_headers = dict(request.headers)` then pass that into the functions.

sadath-12:
thanks @samredai

datajunction-query/djqs/engine.py
```diff
@@ -66,22 +67,53 @@ def get_columns_from_description(
...
+    """
+    Create a SQLAlchemy engine for Trino.
+    """
+    # Parsing the URI
+    parsed_uri = urlparse(engine.uri)
+    scheme = parsed_uri.scheme
+    host = parsed_uri.hostname
+    port = parsed_uri.port
+    path = parsed_uri.path.lstrip("/")
+    catalog, schema = path.split("/", 1) if "/" in path else (path, None)
```
samredai, line 80: Can you explain this string parsing here? What is the format of `engine.uri` you're expecting?

sadath-12: I expect the following fields to be available 
```python
ParseResult(scheme='trino', netloc='knorket-abc-trino.kloudsoft.co:443', path='/tpch/sf1', params='', query='', fragment='')
```

samredai: I see. The string manipulation will be hard to debug when it hits an error. What do you think about making a PosixPath and then just checking the number of parts? So you could replace:
```py
    path = parsed_uri.path.lstrip("/")
    catalog, schema = path.split("/", 1) if "/" in path else (path, None)
```
with:
```py
    from pathlib import PurePosixPath
    ...
    path_parts = PurePosixPath(parsed_uri.path).parts
    if len(path_parts) < 2:
        raise HTTPException(
            status_code=HTTPStatus.BAD_REQUEST,
            detail="Cannot create a Trino engine, catalog or schema missing",
        )
    catalog = path_parts[1] if len(path_parts) > 1 else None
    schema = path_parts[2] if len(path_parts) > 2 else None
```

samredai:
@sadath-12 I have a question about the implementation you're aiming for here. Are you really planning to path `QUERY_USER` and `QUERY_PASSWORD` as plain values in the header? Can you tell me a bit more about your Trino deployment? Is there a JWT in the headers that can be used instead?

sadath-12:
@samredai for now not using jwt , when needed will open the pr for sure . Also added trino url as wanted to communicate with multiple trino instances

samredai:
> @samredai for now not using jwt , when needed will open the pr for sure . Also added trino url as wanted to communicate with multiple trino instances

Got it. Do you control the way these header values are set? I'm wondering if we should just cut to the chase and have a `URI` request header value that you can set to whatever you want. What do you think?

sadath-12:
> > @samredai for now not using jwt , when needed will open the pr for sure . Also added trino url as wanted to communicate with multiple trino instances
> 
> Got it. Do you control the way these header values are set? I'm wondering if we should just cut to the chase and have a `URI` request header value that you can set to whatever you want. Then this doesn't need to be Trino specific and we could still have a single `create_engine` function. What do you think?

well ya , thats a clean way tho

datajunction-query/djqs/engine.py
```diff
@@ -69,20 +68,28 @@ def get_columns_from_description(
...
     _logger.info("Running query on catalog %s", query.catalog_name)
+
     engine = session.exec(
         select(Engine)
         .where(Engine.name == query.engine_name)
         .where(Engine.version == query.engine_version),
     ).one()
-    if engine.type == EngineType.DUCKDB:
+
+    query_server = headers.get("QUERY_ENGINE_SERVER")
```
samredai, line 88: I think `QUERY_ENGINE_SERVER` should be renamed to something more explicit like `SQLALCHEMY_URI`.

samredai (changes requested):
@sadath-12 this is very close, just left one comment about the header arg name. I'm sure we'll eventually also want to allow passing extra information like connect_args via request headers but we can handle that in the future with something like a base64 encoded json object.

samredai (approved):

samredai:
Thanks @sadath-12!

PR #1116, opened 2024-07-23
Node SQL build

shangyian:
### Summary

This is a fairly large change that revamps how we generate node SQL.

The primary rationale behind this is that our existing SQL building does repeated traversals of the dimensions graph, along with multiple calls to `compile(...)` at different points in the call tree, all of which are expensive and unnecessary. It is also difficult to untangle most of this, including figuring out which dimensions/filters can be pushed down, which dimensions/filters require joins, and so on. This rewrite tries to minimize the graph traversal and clean up the SQL generation strategy. 

For some complex queries, the query build time is significantly reduced, from 5+ seconds to <1sec.

#### New Endpoint

This PR adds a different API endpoint for retrieving measures SQL that uses the new build method: `GET /sql/measures/v2`.

This is to allow for a slower migration process from the existing SQL endpoints to this one. This measures SQL endpoint has also been modified to return separate measures queries for each of the upstream nodes, rather than a single joined query, which would not have made sense if the requested metrics had two upstream nodes each at different grains.

#### SQL Building Strategy

Below is an outline of the new SQL building strategy:

##### Building Node SQL (handled in `build_ast`)

1. Recursively turn node references into query ASTs. These are replaced with materialized tables where possible (i.e., source nodes will always be replaced with a materialized table), but otherwise we generate the SQL of each upstream node reference.
2. Apply any filters that can be pushed down to each referenced node's AST.
3. Initialize the final query with the node's query AST added to it as a CTE.

##### Dimension Node Joins (handled in `build_node`)

1. For any dimensions or filters requested for the node, we parse each and determine if a join is needed to bring in the dimension or filter. We keep track of all the necessary dimension joins in a dict that maps dimension nodes to join paths. This removes the need for unnecessary dimensions graph traversal.
2. For each of the necessary dimension joins, build the dimension node's query in the same manner as above, recursively replacing any node references and pushing down requested filters where possible. 
3. Add the dimension node's query AST to the final query as a CTE.

##### Final Query (handled in `build_node`)

1. Build the final query using the various CTEs. This does all the joins between the node query AST and the dimension nodes' ASTs using the join logic from the dimension links.
2. Add all requested dimensions to the final select.
3. At this point, we should not need to add any additional filters because all of the filters will have been done on either the original node AST or one of the joined in dimensions’ ASTs.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -266,6 +266,7 @@ async def build_node(  # pylint: disable=too-many-arguments
         node_ast.select.where,
         *pushdown_filters,
     )
+    print("node_ast after adding filters", node_ast)
 
     ###################
     ## Build Node CTEs
@@ -347,9 +348,11 @@ async def build_node(  # pylint: disable=too-many-arguments
         )
         final_ast.select.projection.extend(dimensions_columns)
 
-    # Add remaining join filters to the where clause
-    for filter_ast in join_filters:
-        filter_ast = replace_filter_dimension_refs(filter_ast, dimension_node_joins)
+        # Replace dimensions referenced in filters
+        for filter_ast in join_filters:
+            filter_ast = replace_filter_dimension_refs(filter_ast, link, dimension_node_joins)
+
+    # Add remaining filters to the where clause
     final_ast.select.where = combine_filter_conditions(
         final_ast.select.where,
         *join_filters,
@@ -513,30 +516,14 @@ def build_requested_dimensions_columns(
 ):
     dimensions_columns = []
     for dim in requested_dimensions:
-        full_dimension_attr = FullColumnName(dim)
-        dim_node = full_dimension_attr.node_name
-        # Check if it's a foreign key reference on one of the dimensions
-        foreign_key_column_name = None
-        if dim in link.foreign_keys_reversed:
-            foreign_key_column_name = FullColumnName(
-                link.foreign_keys_reversed[dim],
-            ).column_name
-
-        if dim_node in dimension_node_joins:
-            for col in dimension_node_joins[dim_node].node_query.select.projection:
-                if col.alias_or_name.name == full_dimension_attr.column_name or (
-                    foreign_key_column_name
-                    and col.alias_or_name.identifier() == foreign_key_column_name
-                ):
-                    dimensions_columns.append(
-                        ast.Column(
-                            name=ast.Name(col.alias_or_name.name),
-                            # **({"alias": ast.Name(amenable_name(dim))} if full_dimension_attr.column_name in final_columns else {}),
-                            alias=ast.Name(amenable_name(dim)),
-                            _table=dimension_node_joins[dim_node].node_query,
-                            _type=col.type,
-                        ),
-                    )
+        replacement = build_dimension_attribute(
+            dim,
+            dimension_node_joins,
+            link,
+            alias=amenable_name(dim),
+        )
+        if replacement:
+            dimensions_columns.append(replacement)
     return dimensions_columns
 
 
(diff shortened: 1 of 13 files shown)


shangyian:
@agorajek @samredai @anhqle this is ready for review. The good news is that this doesn't break or change any of the endpoints that already exist, so we can merge this in without any issues and test completely separately from the existing flows.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -0,0 +1,937 @@
...
+    5. Add each dimension node's query AST to the final query as a CTE.
+    6. Build the final query using the various CTEs. This does all the joins between the node
+    query AST and the dimension nodes' ASTs using the join logic from the dimension links.
+    7. Add all requested dimensions to the final select.
+    8. Add order by and limit to the final select (TODO)
+    """
+    if access_control:
+        access_control.add_request_by_node(node)
+
+    # Set the dialect by finding available engines for this node, or default to Spark
```
agorajek, line 230: Comment does not seem to match the code.

agorajek (approved):
This is super cool. Thanks @shangyian ! I like the fact that you coded it on the side, because I think we may need to reconsider the all-in filter pushdown strategy. As much as I like it for performance reasons I think this may cause some unexpected results eventually. Ideally I think it should only be done:
- if we see the guarantee correctness
- or if the user explicitly says so (a.k.a SQL hints)

datajunction-server/datajunction_server/construction/build_v2.py
shangyian: Ahh yes, let me remove that!

shangyian:
> if the user explicitly says so (a.k.a SQL hints)

@agorajek fwiw, I think the dimension links are somewhat this already, as in, we will only push down the filter if there's a linked dimension for it -- we won't try to parse the node SQL to figure out if we can push down filters to specific columns. That I agree is super error-prone. The dimension linking is essentially a user-provided SQL hint. Or do you think we should allow users to control the pushdown at SQL build time?

agorajek:
> > if the user explicitly says so (a.k.a SQL hints)
> 
> @agorajek fwiw, I think the dimension links are somewhat this already, as in, we will only push down the filter if there's a linked dimension for it -- we won't try to parse the node SQL to figure out if we can push down filters to specific columns. That I agree is super error-prone. The dimension linking is essentially a user-provided SQL hint. Or do you think we should allow users to control the pushdown at SQL build time?

Yes... I understand that the dimension linking is a great proxy for passing the filters, but what if we have multiple nodes with the same dimension linked. I think there may be some situations where this filter push-down may be undesired... but until we see a good case it will be a theoretical concern of mine. I think the least we could do now is to add a flag to turn the filter push-down off completely. This would give us some leeway when we run into a bad use case. The next step later on (if necessary) could be to add explicit hints to build time where one could specify filters with the desired node name to apply on.

shangyian:
@agorajek yeah that makes sense. I'll add a flag to control whether or not the filter pushdown should happen (might be a separate PR though). This may be difficult for users to control for any automated integrations though, like I'm not sure how users who are accessing things thru a dashboard would actually control this.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -0,0 +1,938 @@
...
+    needed to bring in the dimension or filter. Keep track of all the necessary dimension
+    joins in a dict that maps dimension nodes to join paths.
+    4. For each of the necessary dimension joins, build the dimension node's query in the
+    same manner as above, recursively replacing any node references and pushing down requested
+    filters where possible.
+    5. Add each dimension node's query AST to the final query as a CTE.
+    6. Build the final query using the various CTEs. This does all the joins between the node
+    query AST and the dimension nodes' ASTs using the join logic from the dimension links.
+    7. Add all requested dimensions to the final select.
+    8. Add order by and limit to the final select (TODO)
```
samredai, line 225: Looking at this build strategy has me thinking--should we use the [builder pattern](https://refactoring.guru/design-patterns/builder/python/example) here since there are a series of complicated steps we need to do in order? A few benefits I can think of is:
- It will make the order of operations easier to understand by just looking at the method chain that happens in the build
- It will let us keep a shared context for the build instance where we can store warnings/errors
- When a particular section of the build fails, the error at the top of the chain can explicitly say that, like `DJBuildException("Failed to build dimension queries")

shangyian: Great idea, this pattern would help! I'll do another round of refactoring now that we have some basic tests in place.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -0,0 +1,938 @@
...
+    table = None
+    can_use_materialization = (
+        build_criteria and node.name != build_criteria.target_node_name
+    )
+    if node.type == NodeType.SOURCE:
+        table_name = (
+            f"{node.catalog.name}.{node.schema_}.{node.table}"
+            if node.schema_ == "iceberg"
+            else f"{node.schema_}.{node.table}"
+        )
```
samredai, line 906: I didn't know we store this `schema_` value on nodes. Is it true that only iceberg tables require the catalog name? Wouldn't hive tables or any table queried via Trino also need the catalog name included?

datajunction-server/datajunction_server/construction/build_v2.py
samredai: Awesome! The example in that link does a bit more than I think we'd need to do here like create an abstract base class and stuff. I think a single builder class that carries all of these step functions and a `build` method that calls them in the right order (and catches and re-raises errors for each step) is good enough. Then the return type can just be an immutable fully built query.

shangyian: Oh yeah, we don't need that. I was just going to add a `QueryBuilder` class that stores errors and allows for building each of the arguments to this function separately (i.e., `with_orderby`, or `with_limit`, neither of which I've implemented at the moment).

datajunction-server/datajunction_server/construction/build_v2.py
shangyian: We only store `schema_` for source nodes, as far as I can tell. And yes, I recall that there's a weird edge case where if the schema is `iceberg`, it needs the catalog name or the Spark query will fail to run (this is copied from our existing build fyi).

Issue #1092, opened 2024-07-09
Add dynamic query passing 


sadathknorket:
Let's say we use Trino query engine , since we have auth setup in that for multi-user . Would like the credentials to pass through DJ straight to Trino so auth happens there

sadathknorket:
looks like we create different session for each user ?

PR #1120, opened 2024-07-25
forward request headers when requests are made to query service

samredai:
### Summary

This updates every query service client request to include the headers that came in with the request to the main server. It also updates all usage of the query service client to make sure the headers are passed in.

cc: @sadath-12

### Test Plan

Updated tests

- [ ] PR has an associated issue: #1092 #1064
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/cubes.py
+++ b/datajunction-server/datajunction_server/api/cubes.py
@@ -3,9 +3,9 @@
 Cube related APIs.
 """
 import logging
-from typing import Annotated, List, Optional
+from typing import List, Optional
 
-from fastapi import Depends, Header, Query, Request
+from fastapi import Depends, Query, Request
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from datajunction_server.api.helpers import get_catalog_by_name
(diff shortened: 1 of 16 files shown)


agorajek (approved):
That was fast. Thanks @samredai !

datajunction-server/datajunction_server/api/cubes.py
```diff
@@ -139,9 +141,10 @@ async def get_cube_dimension_values(  # pylint: disable=too-many-locals
         submitted_query=translated_sql.sql,
         async_=async_,
     )
+    request_headers.update({"Cache-Control": cache_control})
```
shangyian, line 144: The cache control headers should already be in `request_headers` if they were passed in. So I wonder if this can be simplified by removing `cache_control: Annotated[str, Header()] = ""` from this method and just extracting the cache control values from `request.headers`, if it's even needed here? If it's not needed here and the sole purpose is just to pass over the headers to the query service, then we can just pass `request.headers` over instead of specifically pulling out cache control values.

shangyian (approved):

datajunction-server/datajunction_server/api/cubes.py
samredai: Great catch! I completely missed that but yes, from what I can tell we don't do anything with the cache control headers in the main service so these are just to pass along.

PR #1126, opened 2024-07-29
Explore filters

shangyian:
### Summary

This PR adds a filters bar to the Explore view's header so that anyone viewing nodes through the default Explore UI can filter nodes by type, tags, and user.

Screenshot 2024-07-29 at 7 38 06 AM

### Test Plan

Locally

- [x] PR has an associated issue: #1052 
- [ ] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/database/namespace.py
```diff
@@ -117,15 +125,17 @@ async def list_nodes(
...
+                version=row.current_version,
                 type=row.type,
-                status=row.status,
-                mode=row.mode,
-                updated_at=row.updated_at,
+                status=row.current.status,
+                mode=row.current.mode,
+                updated_at=row.current.updated_at,
+                tags=row.tags,
+                edited_by=list({entry.user for entry in row.history if entry.user}),
```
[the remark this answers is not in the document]
shangyian: Yeah, that might be more future-proof, but I was actually thinking that at some point we should just switch the UI over to the GraphQL API, which would then allow the UI to just retrieve the specific fields it cares about, and the GraphQL setup would handle deciding which joins it wants the DB to make. I can add in what you're describing, but it feels like reimplementing what we already have in GraphQL (i.e., only retrieve X field, and only do the DB join if X field is requested).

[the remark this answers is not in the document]
samredai: Ok that's a fair point.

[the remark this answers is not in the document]
shangyian: I added an option to the list nodes for a namespace endpoint to only retrieve history if `with_edited_by` is set to true for the time being.

datajunction-server/datajunction_server/api/namespaces.py
```diff
@@ -144,12 +144,21 @@ async def list_nodes_in_namespace(
         default=None,
         description="Filter the list of nodes to this type",
     ),
+    with_edited_by: bool = Query(
+        default=False,
+        description="Whether to include a list of users who edited each node",
```
agorajek, line 149: Could you pls clarify a use case for this param? I can seem to put a finger on a good one...

datajunction-server/datajunction_server/database/node.py
```diff
@@ -222,9 +223,24 @@ class Node(Base):  # pylint: disable=too-few-public-methods
...
+    history: Mapped[List[History]] = relationship(
+        primaryjoin="History.entity_name==Node.name",
+        order_by="History.created_at",
+        foreign_keys="History.entity_name",
+    )
+
     def __hash__(self) -> int:
         return hash(self.id)
 
+    @hybrid_property
```
agorajek, line 235: Wow. What is this?

datajunction-server/datajunction_server/database/node.py
```diff
@@ -222,9 +223,24 @@ class Node(Base):  # pylint: disable=too-few-public-methods
...
+    def edited_by(self) -> List[str]:
+        """
+        Editors of the node
+        """
+        return list(  # pragma: no cover
```
agorajek, line 240: Sigh... :)

agorajek (approved):
lovin-it

datajunction-server/datajunction_server/api/namespaces.py
samredai: From this thread: https://github.com/DataJunction/dj/pull/1126#discussion_r1695571189

samredai (approved):

datajunction-server/datajunction_server/database/node.py
shangyian: It's a sqlalchemy object property: https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html#sqlalchemy.ext.hybrid.hybrid_property

PR #1128, opened 2024-07-31
Refactor node query building v2 to use the builder pattern

shangyian:
### Summary

This change refactors node query building v2 to use the builder pattern. We create a `QueryBuilder` class, which is designed to:
* Allow for chaining of method calls to layer in requests for filters, dimensions, order by clauses etc in an incremental manner.
* Incorporate a validation process to check the feasibility of the query based on requested dimensions and filters, as well as access controls. All errors encountered during query construction are collected and raised in the final stage of the build process.

I've also changed the output of the measures SQL endpoint to include both the node name and its version, so that there is no ambiguity about what the generated SQL was for. This output model is captured by the new `GeneratedSQL` class.

```
The changes are actually quite minimal, but the github diffs don't seem able to reflect them properly.
Indenting makes it go haywire...
```

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -584,7 +584,7 @@ class QueryBuilder:  # pylint: disable=too-many-instance-attributes,too-many-pub
             dim_node = dimension_attr.node_name
             if dim_node == self.node_revision.name:
                 continue
-            self.add_request_by_node_name(dim_node)
+            await self.add_request_by_node_name(dim_node)
             if dim_node not in dimension_node_joins:
                 join_path = await dimension_join_path(
                     self.session,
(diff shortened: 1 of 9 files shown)


datajunction-server/datajunction_server/models/sql.py
```diff
@@ -0,0 +1,55 @@
...
+    @root_validator(pre=False)
+    def transpile_sql(  # pylint: disable=no-self-argument
+        cls,
+        values,
+    ):
+        """
+        Transpiles SQL to the specified dialect with the configured transpilation plugin.
+        If no plugin is configured, it will just return the original generated query.
+        """
+        settings = get_settings()
```
agorajek, line 47: @samredai recently mentioned that we should be depending on this function rather than importing it directly, but I may be missing some context here.

agorajek (approved):
@shangyian this is so good! Not too mention that we now return things from the building functions and not just do it as side effects ;)

Left one question in-line.

datajunction-server/datajunction_server/models/sql.py
samredai: Yeah that's right, I realized this when I was working on some stuff in the query service. Anywhere we just call it like this without using `Depends` won't get replaced when we try to replace it with dependency injection. I'm not sure how deep the call stack is to get here but it might be better to just make `sql_transpilation_library` an argument to the function and then pass it in from the endpoint function that uses `settings: Settings = Depends(get_settings)`.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -179,202 +184,398 @@ async def get_measures_query(  # pylint: disable=too-many-locals
...
+        4. For each of the necessary dimension joins, build the dimension node's query in the
+        same manner as above, recursively replacing any node references and pushing down requested
+        filters where possible.
+        5. Add each dimension node's query AST to the final query as a CTE.
+        6. Build the final query using the various CTEs. This does all the joins between the node
+        query AST and the dimension nodes' ASTs using the join logic from the dimension links.
+        7. Add all requested dimensions to the final select.
+        8. Add order by and limit to the final select (TODO)
+        """
+        physical_table = get_table_for_node(
```
samredai, line 374: This would be cleaner as a property.
```py
@property
def physical_table(self):
    return get_table_for_node(
        self.node_revision,
        build_criteria=self._build_criteria,
    )
```

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -179,202 +184,398 @@ async def get_measures_query(  # pylint: disable=too-many-locals
...
+            self.node_revision,
+            build_criteria=self._build_criteria,
+        )
+        if physical_table and not self._filters and not self.dimensions:
+            physical_table.parenthesized = False
+            return ast.Query(
```
samredai, line 380: It makes sense to me why this would return early here but it would be nice to know that the rest of the steps outlined in the docstring were skipped. If a `QueryBuilder` instance has an attribute that can be marked as complete, maybe we can complete the rest of the chain, even if only just to see that the build complete early and log a single line "Skipping doing x, query build complete".

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -179,202 +184,398 @@ async def get_measures_query(  # pylint: disable=too-many-locals
...
+                from_=ast.From(relations=[ast.Relation(node_alias)]),  # type: ignore
+            ),
+            ctes=[*node_ast.ctes, node_ast],
+        )
+        node_ast.ctes = []
+        node_ast = node_ast.to_cte(node_alias)
+        cte_mapping[self.node_revision.name] = node_ast
+
+        # Start building the dimension joins and adding them to the CTEs
+        for _, dimension_join in dimension_node_joins.items():
```
samredai, line 443: This feels like it would go nicely in a build step method on this class. It looks like it's the only place `dimension_node_joins` defined above is used so you can even wrap line 399 into this. Then here it would just be a single self-documenting line along the lines of `self.add_dimensions_as_ctes()`

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -179,202 +184,398 @@ async def get_measures_query(  # pylint: disable=too-many-locals
...
+                dimensions_columns = build_requested_dimensions_columns(
+                    requested_dimensions,
+                    join_path[-1],
+                    dimension_node_joins,
+                )
+                final_ast.select.projection.extend(dimensions_columns)
+
+        # Mark any remaining requested dimensions that don't need a join with
+        # their canonical dimension names
+        for dim_name in self.dimensions:
```
samredai, line 495: Same as my other comment, I think this can be wrapped into a build step method called something like `self.set_canonical_dimension_aliases()`...or a better named method :)

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -179,202 +184,398 @@ async def get_measures_query(  # pylint: disable=too-many-locals
...
-                dimension_node_query,
-                final_ast,
-                link.dimension.name,
+        )
+
+        # Find all dimension node joins necessary for the requested dimensions and filters
+        dimension_node_joins = await self.find_dimension_node_joins()
+
+        ###################
+        ## Build Node CTEs
```
samredai, line 402: I like that this comment highlights this step, but why not make it self documenting by converting this to a build step method and just calling `self.build_node_ctes()` here?

samredai (commented):
Thanks for doing this @shangyian! I left a few comments, mainly around the `build` method. In summary, I'm wondering if we can make the order of each step to be sort of self-documenting in the build function. I'm picturing something like:
```py
class QueryBuilder:
    ...
    def build(self) -> Query:
        self.prepare_some_stuff()
        self.find_some_things()
        self.do_some_joins()
        self.add_order_by()
        self.add_limit()
        self.validate_access()
        if self.errors:
            raise DJQueryBuildException(errors=self.errors)
        return self.ast
```

datajunction-server/datajunction_server/construction/build_v2.py
shangyian: @samredai I'd originally avoided the self-documenting submethods because I was trying to avoid the pattern of modifying arguments that are passed into functions. However, I do agree that it'll be easier to read, so what I ended up doing was creating a set of attributes on the `QueryBuilder` that will be modified as the query gets built (so that it's not a surprise that some arguments will be changed), and then just return the final query at the end of `build()`.

datajunction-server/datajunction_server/models/sql.py
shangyian: Hmm the issue is that this is a pydantic model's `root_validator`, so we can't actually add any extra arguments here, since it's pydantic that is making the call to this method. I think what we can do instead is add `sql_transpilation_library ` to the pydantic model itself and then use that to modify the object in the `root_validator`.

datajunction-server/datajunction_server/construction/build_v2.py
samredai: Makes sense! Thanks @shangyian

PR #1129, opened 2024-07-31
Do not pass on encoding-related headers to the query service

shangyian:
### Summary

We should not pass on encoding-related headers to the query service. 

If we pass a header like:
```
Accept-Encoding: gzip, deflate
```
The DJ core API will need to be able to process a gzipped response from the query service. Given that we're not ready to do this, this change just ignores those headers when doing a headers pass-through.

### Test Plan

- [ ] `make check` passes


samredai (approved):

agorajek (approved):
Thanks for catching.

PR #1138, opened 2024-08-09
Fix cube dimensions

shangyian:
### Summary

There's a bug with creating cubes that have dimension attributes that share the same column name but come from different dimension nodes. This is due to a "group by" we do on the column name that shouldn't be done.

### Test Plan

Locally

- [x] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Thanks!

PR #1140, opened 2024-08-12
schema: multi-tenancy

sadath-12:
### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/alembic/env.py
+++ b/datajunction-server/alembic/env.py
@@ -65,7 +65,7 @@ def run_migrations_offline():
     script output.
 
     """
-    url = settings.index
+    url = settings.index 
     context.configure(
         url=url,
         target_metadata=target_metadata,
@@ -85,13 +85,14 @@ def run_migrations_online():
     and associate a connection with the context.
 
     """
-    connectable = create_engine(settings.index)
+    connectable = create_engine(settings.index,connect_args={"options": "-csearch_path=ex2"})
 
     with connectable.connect() as connection:
         context.configure(
             connection=connection,
             target_metadata=target_metadata,
             render_as_batch=True,
+            schema_translate_map={None: "ex2"}
         )
 
         with context.begin_transaction():
--- a/datajunction-server/datajunction_server/utils.py
+++ b/datajunction-server/datajunction_server/utils.py
@@ -91,7 +91,9 @@ class DatabaseSessionManager:
                 "keepalives_interval": settings.db_keepalives_interval,
                 "keepalives_count": settings.db_keepalives_count,
             },
-        )
+        ).execution_options(
+            schema_translate_map={None: self.schema}
+    )        
 
         async_session_factory = async_sessionmaker(
             bind=self.engine,
@@ -114,23 +116,24 @@ class DatabaseSessionManager:
 
 
 @lru_cache(maxsize=None)
-def get_session_manager() -> DatabaseSessionManager:
+def get_session_manager(request: Request) -> DatabaseSessionManager:
     """
     Get session manager
     """
     session_manager = DatabaseSessionManager()
+    session_manager.schema = request.headers.get("user")
     session_manager.init_db()
     return session_manager
 
 
 @lru_cache(maxsize=None)
-def get_engine() -> AsyncEngine:
+def get_engine(schema: str) -> AsyncEngine:
     """
     Create the metadata engine.
     """
     settings = get_settings()
     engine = create_async_engine(
-        settings.index,
+        settings.index, 
         future=True,
         echo=settings.db_echo,
         pool_pre_ping=settings.db_pool_pre_ping,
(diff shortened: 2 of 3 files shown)


samredai (changes requested):

datajunction-server/alembic/env.py
```diff
@@ -85,13 +85,14 @@ def run_migrations_online():
     and associate a connection with the context.
 
     """
-    connectable = create_engine(settings.index)
+    connectable = create_engine(settings.index,connect_args={"options": "-csearch_path=ex2"})
```
shangyian, line 88: Are the options passed in here something that you can make configurable through `.env` instead? It's also not clear why you would want to hardcode this to ex2.

datajunction-server/alembic/env.py
```diff
@@ -85,13 +85,14 @@ def run_migrations_online():
...
 
     with connectable.connect() as connection:
         context.configure(
             connection=connection,
             target_metadata=target_metadata,
             render_as_batch=True,
+            schema_translate_map={None: "ex2"}
```
shangyian, line 95: I'm not sure this is needed? Since the search_path is already set to ex2.

sadath-12:
I have updated the implementation to only route to different schema and remove creation of schema code as discussed with @samredai

datajunction-query/djqs/utils.py
```diff
@@ -56,11 +57,18 @@ def get_metadata_engine() -> Engine:
...
     Per-request session.
     """
+    schema = request.headers.get("tenant")
     engine = get_metadata_engine()
+    settings = get_settings()
+
+    if schema:
+        engine = engine.execution_options(schema_translate_map={None: schema})
+
+    settings.customSchema = request.headers.get("new_tenant")
```
samredai, line 71: @sadath-12 can you help me understand why you need to modify the settings instance directly like this? Also, since you're pulling schema from the "tenant" header, you don't need this settings property at all anymore, right?

sadath-12: Hi @samredai  , this is to tell dj to switch to the schema when we hit register schema endpoint (created internally) so dj can run migrations on it on runtime or else it will have to be restarted

samredai: Thanks @sadath-12 I understand the intent here more. I don't think modifying the application settings is the right approach though. Instead we should parameterize the alembic migration scripts (I think I recall @shangyian proposing this). Then in your custom endpoint logic, you can programmatically set that schema parameter using the context.

ockhamlabs:
Wondering if this is close to finalization? Anything we can do to help ?

samredai:
@ockhamlabs sorry for the late reply here. Let's revisit this after this refactor for the query service goes in which should make this ultimately much easier to achieve. I'm planning to catch up with @sadath-12 soon, either during the community meeting or async.

sadath-12:
(after talk with @samredai )
 Note: Whenever we switch across schemas make sure u use existing connection pool

samredai:
@sadath-12 super sorry for losing track of this. I think we're close to figuring this out--this PR is doing the right thing by parsing the tenant value from the request header, but let's save any of the changes in the datajunction-query directory or to the alembic/env.py file for a follow up PR. If you could update this PR to just include the changes to datajunction_server/config.py and datajunction_server/utils.py, we can iterate on that and get this PR merged pretty quickly.

datajunction-server/datajunction_server/utils.py
```diff
@@ -115,17 +121,20 @@ async def close(self):
...
 @lru_cache(maxsize=None)
-def get_session_manager() -> DatabaseSessionManager:
+def get_session_manager(request: Optional[Request] = None) -> DatabaseSessionManager:
     """
     Get session manager
     """
     session_manager = DatabaseSessionManager()
+    session_manager.schema = request.headers.get("tenant")
+    settings = get_settings()
+    settings.customSchema = request.headers.get("new_tenant")
```
samredai, line 131: @sadath-12 from our discussions I remember you telling me that you had to do this so that the alembic script would run properly when you init a new tenant's schema, but since [this](https://github.com/DataJunction/dj/commit/504f198a555b5ef020b35d01ef1ae5c19adf2cad#diff-515aae121be36eeb90e22b78bc3051a911772a4565f2e825def4f17c5da4a4ac) commit the alembic scripts are now configurable to target a specific schema. This means in any of your custom endpoints you could do something like this:

```py
from alembic.config import Config
from alembic import command

...
SCHEMA = request.headers.get("tenant") # For example, "tenant1"
if SCHEMA:
    alembic_cfg = Config()
    alembic_cfg.set_main_option("script_location", "./alembic")
    alembic_cfg.cmd_opts = {"x": [f"uri=postgresql+psycopg://username:password@server:5432/{SCHEMA}"]}
    command.upgrade(alembic_cfg, "head")
...
```

Which means you don't need to add this customSchema to the settings class at all.

PR #1142, opened 2024-08-13
Add created by and edited by fields to both the REST API and GraphQL query

shangyian:
### Summary

This change exposes `created_by` and adds `edited_by` to both the REST API and the GraphQL findNodes query. For the GraphQL query, the join against the history table is only done when `edited_by` is requested, and the `created_by` attribute is only joined in when it is requested. 

There were a few issues with SQLAlchemy managing the join correctly without explicitly specifying the `foreign_keys` attribute on the ORM object. 

### Test Plan

Added a test

- [x] PR has an associated issue: #1144
- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/scalars/node.py
+++ b/datajunction-server/datajunction_server/api/graphql/scalars/node.py
@@ -19,6 +19,7 @@ from datajunction_server.database.dimensionlink import (
     JoinCardinality as JoinCardinality_,
 )
 from datajunction_server.database.dimensionlink import JoinType as JoinType_
+from datajunction_server.database.node import Node as DBNode
 from datajunction_server.database.node import NodeRevision as DBNodeRevision
 from datajunction_server.models.node import NodeMode as NodeMode_
 from datajunction_server.models.node import NodeStatus as NodeStatus_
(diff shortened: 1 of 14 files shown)


samredai:
@shangyian I just opened PR #1147 that will add a direct created_by_id relationship on a few entities including nodes. Can you take a look when you have a chance? If that looks good, this PR can just handle pulling the edited by list which I agree should come from the history table like you have it here.

datajunction-server/datajunction_server/api/graphql/scalars/node.py
```diff
@@ -182,3 +183,10 @@ class Node:  # pylint: disable=too-few-public-methods
 
     tags: List[Tag]
     created_by: User
+
+    @strawberry.field
+    def edited_by(self, root: "DBNode") -> List[str]:
+        """
+        The users who edited this node
+        """
+        return list({entry.user for entry in root.history})
```
agorajek, line 192: Should we distinctify this list before returning?

agorajek (approved):
Thanks @shangyian . This looks good.

datajunction-server/datajunction_server/api/graphql/scalars/node.py
shangyian: Oh it's actually already distinct because of the set comprehensive `{ ... }`

PR #1145, opened 2024-08-14
rename tables in djqs

### LINKED ISSUE #1143: Rename Query Service metadata tables to not conflict with core service table names
This would allow using the same postgres instance and schema as a metadata db for both the core and query service. cc: @sadath-12

sadath-12:
### Summary

closes #1143 

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-query/alembic/env.py
+++ b/datajunction-query/alembic/env.py
@@ -8,9 +8,9 @@ from logging.config import fileConfig
 from sqlmodel import SQLModel, create_engine
 
 from alembic import context
-from djqs.models.catalog import Catalog
-from djqs.models.engine import Engine
-from djqs.models.query import Query
+from djqs.models.catalog import QSCatalog
+from djqs.models.engine import QSEngine
+from djqs.models.query import QSQuery
 from djqs.utils import get_settings
 
 settings = get_settings()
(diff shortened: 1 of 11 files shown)


shangyian (commented):
I think you'll want to run `make check` (linters) and `make test`. Some of the tests are failing due to the naming change from `Catalog` -> `QSCatalog`

samredai:
Thanks @sadath-12! I'll review this soon and also pull it locally to try it out.

> One original assumption was that it would call the core API to post changes rather than touching the metadata db, but it can now just make changes in the metadata database directly.

I'm hoping we can stick with this assumption and avoid having one of the services directly modify metadata for another service. It's nice that this enables sharing a single postgres db for the metadata but it's still worth allowing them to be separate postgres instances if someone wants to use it that way.

samredai:
@sadath-12 take a look at #1168 when you have a chance. This removes the requirement to do this rename and greatly simplifies the query service. Now it only relies on a single `query` table (which does not conflict with the main service). Now, based on how you want to deploy it, it can share a database instance with the core service or run on an entirely separate one.

I know this PR was meant to unblock #1140 which is also achieved by the new PR. Let's revisit that one after this refactor is merged. It should be much easier to make the changes necessary to get a good multi-tenancy setup working for you.

PR #1146, opened 2024-08-15
include sql_columns output in /cube/{cube_name} endpoint

samredai:
### Summary

This adds a `sql_columns` attribute that derives the names that will be found in the SQL generated for a cube. This will let you know the schema when calling `/cube/{cube_name}` (without calling `/sql/{cube_name}` to find that out).

There is an existing `columns` field but that pulls the actual column entities as defined on those base nodes. It's a small nuanced difference where with metrics we automatically alias metric expression with the `_DOT_` notation, so the name matches the column in the final dataset, however for dimensions, the column entity is the name as it exists on the actual dimension node which is fully determined by the SQL contained in that node (as it should be). It didn't seem right to change any of that so I added this `sql_columns` that gets right to the point of providing the columns that will be found in the SQL projection.

### Test Plan

Added a test for the new function to make sure the names are being derived properly.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -72,7 +72,7 @@ from datajunction_server.models.node import (
     UpdateNode,
 )
 from datajunction_server.models.node_type import NodeType
-from datajunction_server.naming import from_amenable_name
+from datajunction_server.naming import amenable_name, from_amenable_name
 from datajunction_server.service_clients import QueryServiceClient
 from datajunction_server.sql.dag import (
     get_downstream_nodes,
@@ -1399,11 +1399,8 @@ async def derive_sql_column(
     if cube_element.type == "metric":
         query_column_name = cube_element.name
     elif cube_element.type == "dimension":
-        query_column_name = (
-            f"{cube_element.node_name}{SEPARATOR}{cube_element.name}".replace(
-                SEPARATOR,
-                "_DOT_",
-            )
+        query_column_name = amenable_name(
+            f"{cube_element.node_name}{SEPARATOR}{cube_element.name}",
         )
     return ColumnOutput(
         name=query_column_name,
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1388,6 +1389,29 @@ async def column_lineage(
...
 
 
+async def derive_sql_column(
+    cube_element: CubeElementMetadata,
+) -> ColumnOutput:
+    """
+    Derives the column name in the generated Cube SQL based on the CubeElement
+    """
+    query_column_name = None
+    if cube_element.type == "metric":
```
agorajek, line 1399: Is this the best way to check for type "metric" ?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1388,6 +1389,29 @@ async def column_lineage(
...
+        query_column_name = cube_element.name
+    elif cube_element.type == "dimension":
```
agorajek, line 1401: ditto

agorajek (commented):
Makes total sense to me. Just one small finding in-line.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1388,6 +1389,29 @@ async def column_lineage(
...
+        query_column_name = (
+            f"{cube_element.node_name}{SEPARATOR}{cube_element.name}".replace(
+                SEPARATOR,
+                "_DOT_",
```
shangyian, line 1405: Can this part be extracted out into a constant (or actually I think it's already a constant) so that if we do decide to change this value, we can just change the constant?

datajunction-server/datajunction_server/internal/nodes.py
samredai: I think so...I don't like to match on strings either but at this point in the logic we're just working with the pydantic model that just has metadata (dicts, strings, and ints).

datajunction-server/datajunction_server/internal/nodes.py
samredai: I forgot we have this `amenable_name` function that will take care of this entirely so I switched to using that.

agorajek (approved):
Clarified my old inline comment.

Issue #1127, opened 2024-07-30
Ability to register a view / sql as a source node


vcknorket:
Ability to register a view / sql as a source node will really help unlock and transition existing org artifacts to quickly onboard into DJ. Kind of equivalent of "brain-dump" of organizational semantic knowledge :) This is linked with https://github.com/DataJunction/dj/issues/1117

agorajek:
Similarly to suggestion from #1117 we could add the following client method:

```
dj_client.create_source_from_query(
  query,      # (required) without the CREATE VIEW prefix, because we want to know the name
  view_name,  # (required)
  node_name,  # (optional)
  db-engine,  # (optional)
)
```

The above method would:
1. Runs the CRATE OR REPLACE VIEW view_name ... on the db-engine
2. Creates a new source node:
  - if `node_name` is not provided it can be inferred from the `view_name`
  - DJ will store the query as an attribute on the node, but it will not use it for anything other than showing it to the user [*]
  - DJ will create the source node based on the view name and will pull the schema from the the DB
  
[*] later on, we could extend this by adding the ability to edit, re-run the view creation and refreshing the node on demand.

agorajek:
Yes, I will put this on my list for next week.

ockhamlabs:
Curious what's the ETA for this?

agorajek:
PR is ready... https://github.com/DataJunction/dj/pull/1153

agorajek:
This work has been completed.

vcknorket:
nice!
________________________________

PR #1153, opened 2024-08-30
Add register_view as another way to create Source nodes.

agorajek:
### Summary

This adds the ability for DJ to create Source nodes based on database views, but with the extra option of creating the view first. This does not mean that DJ will maintain that view. Only that we use the query param verbatim to build a view and if it is successful we create a source node on top of it. DJ will also store the view SQL but will not do anything with it for now. The feature of allowing this query SQL to be updated could be a future feature.

### Test Plan

Unit tests.

- [x] PR has an associated issue: #1127 
- [x] `make check` passes

### Deployment Plan

n/a

### DIFF
--- a/datajunction-clients/python/README.md
+++ b/datajunction-clients/python/README.md
@@ -350,6 +350,7 @@ List of all available DJ builder methods:
   - source(self, node_name: str)
   - create_source( ..., mode: Optional[NodeMode] = NodeMode.PUBLISHED)
   - register_table( catalog: str, schema: str, table: str)
+  - register_view( catalog: str, schema: str, view: str, query: str, replace: bool = False)
 
   ### nodes: transform
   - transform(self, node_name: str)
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -10,6 +10,7 @@ from datajunction.exceptions import (
     DJClientException,
     DJNamespaceAlreadyExists,
     DJTableAlreadyRegistered,
+    DJViewAlreadyRegistered,
 )
 from datajunction.nodes import (
     Cube,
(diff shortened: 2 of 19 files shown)


samredai:
@agorajek, this is AWESOME! I have thoughts on the location of the view registering logic though. Wouldn't it be better if that logic is all encapsulated within the query service?
- Request comes to the core service to create a view-based source node
- The core service calls `query_service_client.create_view(...)`
- The query service then does `or_replace = "OR REPLACE" if replace else ""; query = f"CREATE {or_replace} VIEW {view} AS {query}"` (right now you are doing this in the core service instead)
- Then, if successful, the query service returns a 200 along with the schema and then the core service just has create the source node definition with that schema.

The two main reasons I'm advocating for this is 1) It's a single call to the query service instead of two calls and 2) if someone wants to customize how exactly they register views at their company, they don't need to add code to the OSS project and can instead just customize the query service.

agorajek:
@samredai your suggestion is good. I'll try to update the code.

shangyian:
Yeah, +1 to moving  he "create view" part of this to the query service. It'll make it easier to adjust for any database engines that don't create views in the standard way.

agorajek:
@shangyian . I am going to move the view creation to the query service in a separate PR because @samredai is currently refactoring the query service.

agorajek:
@samredai and @shangyian this PR is ready if we want to hold on with moving the query generation to query service. @samredai let me know if you care either way.

samredai:
@agorajek I just opened the DJQS refactor PR #1168 but I'll leave it up to you if you'd rather merge this first than migrate the register view logic over to the refactored query service.

samredai (approved):

agorajek:
I'll merge this and move the CREATE VIEW sql to djqs later.

PR #1156, opened 2024-09-03
Increase NODE_LIST_MAX to 10000 and set cache expiration to every 3m

samredai:
### Summary

This improves the DJ search to allow for more nodes to be included in the index (10,000 instead of 1,000) which should be performant enough now that the index endpoint is cached. I also set the cache interval for that endpoint to 3 minutes.

### Test Plan

Ran everything locally with `docker compose --profile=demo up` and test the search with a few thousand nodes.

- [ ] `make check` passes


samredai:
Closing in favor of #1157

Issue #1158, opened 2024-09-05
Add the ability to always build down to source and ignore materialized tables


shangyian:
This should be added to the following endpoints:
* `GET /sql/<node>`
* `GET /sql/?metrics=...`
* `GET /sql/measures/v2?metrics=...`

In the `QueryBuilder` class, we can add a builder option like `use_materialized_table`, which can then be passed down into subsequent function calls like `build_ast` to not do the call to `get_table_for_node` for non-source nodes (which uses the materialized tables, if any).

shangyian:
fyi @skastel in case you want to poke at this one

samredai:
Since materializing tables is really a form of caching, I wonder if this should be implemented using cache directives where a request with `max-age=0` in the cache headers would then set `use_materialized_table=False` when building the query.

skastel:
Using `max-age=0` to bypass the materialized table feels like a hidden feature, rather than intentional behavior to me. I would prefer an explicit flag to avoid confusion and keep future maintainers clear on the intended behavior for clients.

shangyian:
Yeah, I also prefer that we add an explicit flag in the API call, because caching can refer to many things, including the actual caching of the generated queries themselves, which we also do in the DJ backend.

PR #1161, opened 2024-09-06
Adding 'use_materialized' flag to sql endpoints

skastel:
### Summary

To address https://github.com/DataJunction/dj/issues/1158 this adds an `use_materialized` flag to the `sql` endpoints so that clients can set to `False` and request SQL that bypasses any materialized data. By default, no behavior is changing, the default value is still `True` for this flag.

### Test Plan

Unit tests cover this code already, focus is on making sure

- [x] PR has an associated issue: #1158
- [x] `make check` passes

### Deployment Plan
Nothing special.

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -991,14 +991,19 @@ async def build_ast(  # pylint: disable=too-many-arguments,too-many-locals
         for ref_expr in reference_expressions:
 
             # Try to find a materialized table attached to this node, if one exists.
-            physical_table = cast(
-                Optional[ast.Table],
-                get_table_for_node(
-                    referenced_node,
-                    build_criteria=build_criteria,
-                ),
-            )
-            if not physical_table or not use_materialized:
+            physical_table = None
+            if use_materialized:
+                print(f"Checking for physical node: {referenced_node.name}")
+                physical_table = cast(
+                    Optional[ast.Table],
+                    get_table_for_node(
+                        referenced_node,
+                        build_criteria=build_criteria,
+                    ),
+                )
+
+            if not physical_table:
+                print(f"No physical node for {referenced_node.name}")
                 # Build a new CTE with the query AST if there is no materialized table
                 if referenced_node.name not in ctes_mapping:
                     node_query = parse(cast(str, referenced_node.query))
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -993,7 +993,7 @@ async def build_ast(  # pylint: disable=too-many-arguments,too-many-locals
             # Try to find a materialized table attached to this node, if one exists.
             physical_table = None
             if use_materialized:
-                print(f"Checking for physical node: {referenced_node.name}")
+                logger.debug("Checking for physical node: %s", referenced_node.name)
                 physical_table = cast(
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -981,7 +998,7 @@ async def build_ast(  # pylint: disable=too-many-arguments,too-many-locals
                     build_criteria=build_criteria,
                 ),
             )
-            if not physical_table:
+            if not physical_table or not use_materialized:
```
shangyian, line 1001: Can this logic actually be moved slightly earlier, so that we don't retrieve the physical table at all if `use_materialized` is set to `False`?

shangyian (commented):
Thanks for adding this @skastel! Could you add an additional unit test for the use_materialized=False case, when a materialized table exists for the node?

agorajek:
Looks fine to me, but I am still wondering what would be the use case for not using materialized tables?

datajunction-server/datajunction_server/construction/build_v2.py
skastel: Done!

skastel:
> Thanks for adding this @skastel! Could you add an additional unit test for the use_materialized=False case, when a materialized table exists for the node?

Just added a unit test for this @shangyian

skastel:
> Looks fine to me, but I am still wondering what would be the use case for not using materialized tables?

@agorajek By allowing a client to get the full definition of the node/measure/metric and bypassing any materializations the client can see the underlying data sources. We'll likely use this in a few places where we're trying to do our own materializations and caching in a parallel headless BI system, but relying on DJ for definitions.

agorajek:
> > Looks fine to me, but I am still wondering what would be the use case for not using materialized tables?
> 
> @agorajek By allowing a client to get the full definition of the node/measure/metric and bypassing any materializations the client can see the underlying data sources. We'll likely use this in a few places where we're trying to do our own materializations and caching in a parallel headless BI system, but relying on DJ for definitions.

Sad panda... but that makes sense.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -974,14 +991,19 @@ async def build_ast(  # pylint: disable=too-many-arguments,too-many-locals
...
-            physical_table = cast(
-                Optional[ast.Table],
-                get_table_for_node(
-                    referenced_node,
-                    build_criteria=build_criteria,
-                ),
-            )
+            physical_table = None
+            if use_materialized:
+                print(f"Checking for physical node: {referenced_node.name}")
```
shangyian, line 996: Could you use `logger` here instead of having a print statement?

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -974,14 +991,19 @@ async def build_ast(  # pylint: disable=too-many-arguments,too-many-locals
...
+                physical_table = cast(
+                    Optional[ast.Table],
+                    get_table_for_node(
+                        referenced_node,
+                        build_criteria=build_criteria,
+                    ),
+                )
+
             if not physical_table:
+                print(f"No physical node for {referenced_node.name}")
```
shangyian, line 1006: And same here

datajunction-server/datajunction_server/construction/build_v2.py
skastel: oops! good catch, fixed.

PR #1167, opened 2024-09-09
Add default catalog setting for DJ nodes 

shangyian:
### Summary

In some cases, DJ nodes are created (in draft mode) that don't have any valid upstream nodes, thus making it so that we can't infer an appropriate catalog for the node. We seem to default to `catalog_id = 0` for these invalid nodes, which causes issues since this catalog is never updated after the node becomes valid. This PR adds a default catalog for nodes that don't have an inferrable catalog at creation time.

### Test Plan

Locally

- [ ] `make check` passes

### Deployment Plan

ASAP

### DIFF
--- a/datajunction-server/datajunction_server/config.py
+++ b/datajunction-server/datajunction_server/config.py
@@ -91,6 +91,8 @@ class Settings(
     # Interval in seconds with which to expire caching of any indexes
     index_cache_expire = 60
 
+    default_catalog_id: int = 0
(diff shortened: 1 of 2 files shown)


samredai (approved):
Thanks for hunting this bug down @shangyian! I vaguely remember this logic, it might have been a TODO that got lost along the way :(

agorajek (approved):

PR #1168, opened 2024-09-11
Refactor DJQS to remove `sqlmodel` and `pydantic` and just rely on `psycopg` for metadata db queries

samredai:
### Summary

This makes a number of changes to the OSS datajunction query service (`djqs`) to resolve a number of indirect side-effects as well as simplify the entire library.
- Catalogs and engines defined in `config.djqs.[yaml|yml|toml|json]` are read into memory on app startup and cached. Also related, the CRUD endpoints for creating and connecting catalogs and engines were removed. **Why**: These don't change that often and the added complexity of maintaining the crud endpoints and backend metadata tables doesn't seem worth it...I'll take the blame for that idea :)
- The single remaining `query` table is now purely interacted with using the `psycopg` library and the SQL composition utility. That means this PR gets rid of `sqlmodel` and `pydantic` as direct dependencies and also uses `psycopg` to submit async queries to the backend metadata database instead of `sqlalchemy`. **Why**: There was a web of dependency complications that didn't seem worth it to maintain a backend with a single table--can't move to sqlalchemy 2 because pydantic was pinned because sqlmodel was only recently compatible which means we had to update the sqlmodel definitions etc etc etc. Keep in mind this still gets structural validations for our endpoints using python's builtin dataclasses. Fastapi does something with them under the hood that uses pydantic, but it's much better as a transitive dependency.
- The `Settings` class has been updated a bit, primarily to get rid of the dependency on pydantic, however it also loads the catalogs and engines from the config file (cached via the lru_cache on `get_settings()`) and has helper methods `find_engine` and `find_catalog` that's used throughout the library.
- The tests now use testcontainers to launch a postgres instance, similar to how the `datajunction_server` already does.

### Test Plan

Tests and linters pass locally but might have to make tweaks to have the github actions run smoothly.

- [ ] `make check` passes


samredai:
@sadath-12, @vcknorket

samredai:
murray

datajunction-server/alembic/env.py
```diff
@@ -65,9 +64,9 @@ def run_migrations_offline():
     script output.
 
     """
-    url = settings.index
+    x_args = context.get_x_argument(as_dictionary=True)
```
samredai, line 67: @sadath-12 this `-x uri=` param will let you run the migration script for different tenants.
```
alembic -x uri="postgresql+psycopg://username:password@server:5432/tenant1" upgrade head
alembic -x uri="postgresql+psycopg://username:password@server:5432/tenant2" upgrade head
# ...etc
```

...or in python

```py
from alembic.config import Config
from alembic import command

DATABASE = "tenant1"

alembic_cfg = Config()
alembic_cfg.set_main_option("script_location", "./alembic")
alembic_cfg.cmd_opts = {"x": [f"uri=postgresql+psycopg://username:password@server:5432/{DATABASE}"]}
command.upgrade(alembic_cfg, "head")
```

datajunction-query/djqs/api/queries.py
```diff
@@ -49,32 +51,24 @@
...
-    session: Session = Depends(get_session),
     settings: Settings = Depends(get_settings),
     request: Request,
     response: Response,
+    postgres_pool: AsyncConnectionPool = Depends(get_postgres_pool),
     background_tasks: BackgroundTasks,
-    body: Any = Body(...),
+    body: Any = Body(
+        ...,
+        example={
```
agorajek, line 65: Nice!

datajunction-query/.coveragerc
```diff
@@ -2,7 +2,9 @@
 [run]
 branch = True
 source = dj
-# omit = bad_file.py
+omit =
+    djqs/config.py
```
agorajek, line 6: Why do we need to omit these files?

agorajek (approved):
@samredai this is awesome. Your summary is super clear. I can't wait to remove pydantic from the client as well.

datajunction-query/alembic/env.py
```diff
@@ -28,7 +23,7 @@
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-target_metadata = SQLModel.metadata
+target_metadata = None  # pylint: disable=invalid-name
```
shangyian, line 26: In this case, shouldn't the target metadata point to the SQLAlchemy base metadata?

datajunction-query/djqs/config.py
```diff
@@ -3,94 +3,202 @@
...
+class EngineType(Enum):
     """
-    Configuration for the query service
+    Supported engine types
     """
 
-    name: str = "DJQS"
-    description: str = "A DataJunction Query Service"
-    url: str = "http://localhost:8001/"
+    DUCKDB = "duckdb"
```
shangyian, line 27: Should Spark be included here as well?

datajunction-query/djqs/config.py
```diff
@@ -3,94 +3,202 @@
...
 
-    # Configuration file for catalogs and engines
-    configuration_file: Optional[str] = None
+    def __init__(  # pylint: disable=too-many-arguments,too-many-locals,dangerous-default-value
+        self,
+        name: Optional[str] = "DJQS",
+        description: Optional[str] = "A DataJunction Query Service",
+        url: Optional[str] = "http://localhost:8001/",
+        index: Optional[str] = "postgresql://dj:dj@postgres_metadata:5432/dj",
+        alembic_uri: Optional[
```
shangyian, line 74: Curious as to why there's a separation between alembic_uri and index?

datajunction-query/djqs/config.py
```diff
@@ -3,94 +3,202 @@
...
+        catalogs: Optional[List[CatalogInfo]] = None,
+    ):
+        self.name: str = os.getenv("NAME", name or "")
+        self.description: str = os.getenv("DESCRIPTION", description or "")
+        self.url: str = os.getenv("URL", url or "")
+
+        # SQLAlchemy URI for the metadata database.
+        self.index: str = os.getenv("INDEX", index or "")
+
+        self.alembic_uri: str = os.getenv("ALEMBIC_URI", alembic_uri or "")
```
shangyian, line 96: It seems like this could default to self.index if it's not set here?

datajunction-query/djqs/db/postgres.py
```diff
@@ -0,0 +1,138 @@
...
+        progress: float = 0.0,
+        executed_query: str = None,
+        scheduled: datetime = None,
+        started: datetime = None,
+        finished: datetime = None,
+    ):
+        """
+        Save metadata about a query
+        """
+        self.inserts.append(
```
shangyian, line 70: Can this just use the SQLAlchemy ORM directly, given that the models are already there and upserting a row like this would be trivial?

shangyian (approved):

datajunction-query/.coveragerc
samredai: The logic in here is thoroughly covered since we use `get_settings()` all over the place, but the missing coverage is because we now directly use environment variables in config.py (instead of relying on the pydantic abstraction which does the same under the hood). I was weighing writing a bunch of tests to make sure `os.getenv` works but that seemed silly so I opted to just exclude this settings class and the DJUnknownCatalog exception. Let me know if this makes you nervous and I can write some more tests for these.

datajunction-query/djqs/config.py
samredai: From what I remember, we removed spark from here once we got rid of the spark session container in the OSS setup. The idea being that if someone extends this open source implementation to include spark support then they can add it to this enum for their implementation.

datajunction-query/djqs/config.py
samredai: Thanks for this, I completely forgot that this can be removed now. It was initially different because alembic uses a sqlalchemy uri where the index value now gets passed directly into psycopg. The format is a little different where slqalchemy needs the `....+psycopg...` included.

Just updated this to removed `alembic_uri` since that's now set via a `--uri=....` parameter to the alembic command. This also helps with @sadath-12's requirement for multi-tenancy so he doesn't have to hack the Settings object. Another nice side-effect is that the alembic script now has no dependencies on the djqs library.

datajunction-query/djqs/config.py
samredai: Fixed by removing alembic_uri, it's not actually needed anymore.

datajunction-query/alembic/env.py
samredai: The interaction between the server and the metadata library now only uses psycopg, so no sqlalchemy engines or models at all meaning there's nothing in sqlalchemy base metadata to use (we still use sqlalchemy to let people configure connections for retrieving data from a catalog). It might seem like a tradeoff since we have to write the migration scripts by hand instead of relying on the sqlalchemy metadata but I actually find the SQL statements in the versioning to be a bit easier to understand and this backend DB is so simple that I don't think it will be an issue.

Issue #1050, opened 2024-06-14
Add full support for reference links


shangyian:
We can reuse the existing setup with column-level dimension links for this purpose.

PR #1172, opened 2024-09-19
Reference links

shangyian:
### Summary

This change adds support in the backend and Python client for reference links. We already have documentation on [reference links](https://datajunction.io/docs/0.1.0/data-modeling/dimension-links/#reference-link), so this PR brings us up-to-date with the docs.

The reference link implementation reuses our original dimension link database schema: on the `columns` table, we store reference dimension links using the table's `dimension_id` and `dimension_column` columns. This setup means that each column on a node is allowed a *single* reference dimension. 

The following features (from the checklist in #1050) are implemented:
- An API endpoint to create/update a reference link
- An API endpoint to remove a reference link
- Python client support for creating / updating / removing reference links
- Updates to the `get_dimensions` functionality based on the presence of reference links in the dimensions graph

### Test Plan

- [ ] PR has an associated issue: #1050 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -5,7 +5,7 @@ Node related APIs.
 import logging
 import os
 from http import HTTPStatus
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from fastapi import BackgroundTasks, Depends, Query, Response
 from fastapi.responses import JSONResponse
@@ -67,9 +67,11 @@ from datajunction_server.internal.validation import validate_node_data
 from datajunction_server.models import access
 from datajunction_server.models.attribute import AttributeTypeIdentifier
 from datajunction_server.models.dimensionlink import (
+    DimensionLinkType,
+    JoinLinkInput,
     JoinType,
     LinkDimensionIdentifier,
-    LinkDimensionInput,
+    ReferenceLinkInput,
 )
 from datajunction_server.models.node import (
     ColumnOutput,
@@ -874,7 +876,7 @@ async def link_dimension(
                 " These column types are incompatible and the dimension cannot be linked",
             )
 
-    link_input = LinkDimensionInput(
+    link_input = JoinLinkInput(
         dimension_node=dimension,
         join_type=JoinType.LEFT,
         join_on=(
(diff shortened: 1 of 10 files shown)


datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -781,14 +790,18 @@ async def dimension_join_path(
...
-    if dimension.startswith(node.name):
-        for col in node.columns:  # pragma: no cover
-            # Decide if we should restrict this to only columns marked as dimensional
-            # await session.refresh(col, ["attributes"]) TODO
-            # if col.is_dimensional():
-            #     ...
-            if f"{node.name}.{col.name}" == dimension:
-                return []
+    for col in node.columns:  # pragma: no cover
+        # Decide if we should restrict this to only columns marked as dimensional
```
samredai, line 794: I'm ok with restricting this only to columns marked as dimensional. It's reasonable that users will want some control of this and it's easy enough to mark a ton of columns as dimensional if that's what they want.

samredai (approved):
LGTM, thanks @shangyian! Left some commentary but nothing blocking.

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -910,10 +910,130 @@ async def link_dimension(
...
+    node = await Node.get_by_name(session, node_name, raise_if_not_exists=True)
+    dim_node = await Node.get_by_name(session, dimension_node, raise_if_not_exists=True)
+    if dim_node.type != NodeType.DIMENSION:  # type: ignore
+        raise DJInvalidInputException(
+            message=f"Node {node.name} is not of type dimension!",  # type: ignore
+        )
+
+    # The target and dimension columns should both exist
+    target_column = await get_column(session, node.current, node_column)  # type: ignore
+    dim_column = await get_column(session, dim_node.current, dimension_column)  # type: ignore
```
agorajek, line 935: This may be unnecessary, but I am wondering if we should also make sure that the `dim_column` is not the only PK of this dim node. If we don't check it, we are risking that the user may also create a regular dimension link from this column to the same or different dim node. It's a bit of an F&R space but there is a risk of redundancy and design mismatches if we allow PK columns here imo.

datajunction-server/datajunction_server/construction/build_v2.py
agorajek: This is a similar conundrum to my question about the PK dim columns. I feel like we should start with more limitation early on (to keep things clean) and open up later if needed. Going to other way may be harder.

agorajek (approved):
I like it when the docs move faster than the features. This is rare... :)

Feel free to push this as-is or adjust the two checks Sam and I identified.

datajunction-server/datajunction_server/api/nodes.py
shangyian: Hmm yeah, this is probably a good idea. In general I think having more restrictions for users will be better in the process of creating the dimensions graph. However, I'm going to punt this to a separate PR

datajunction-server/datajunction_server/construction/build_v2.py
shangyian: Yeah, I agree, I'll do this in a follow-up PR though, since this change wasn't actually about this problem.

PR #1173, opened 2024-09-20
Speed up SQL building by assigning columns to their table origins in parallel

shangyian:
### Summary

This PR speeds up the SQL building process by trying to match columns to their table origins in parallel. 

The current way we find table sources for a given column is inefficient -- we process each column serially, and we always search up the graph to find possible table sources for each column, which results in many unnecessary tree traversals, one per column.

With these new changes, we can process each column in parallel, and we search "down" the graph, which is to say that we now start from the query, determine the possible table sources for that query, and then we find all columns in the query, and try to assign columns to table sources in parallel. 

### Test Plan

All existing tests should pass

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -2692,7 +2692,6 @@ class Query(TableExpression, UnNamed):
                         table_options[idx] = cte_mapping[option.name.name]
                 await table_options[idx].compile(ctx)
 
-            tp = ThreadPoolExecutor()
             expressions_to_compile = [
                 self.select.projection,
                 self.select.group_by,
@@ -2711,12 +2710,13 @@ class Query(TableExpression, UnNamed):
                         col for expr in expression for col in expr.find_all(Column)
                     ]
 
-            list(
-                tp.map(
-                    _compile,
-                    [(col, table_options) for col in columns_to_compile],
-                ),
-            )
+            with ThreadPoolExecutor() as executor:
+                list(
+                    executor.map(
+                        _compile,
+                        [(col, table_options) for col in columns_to_compile],
+                    ),
+                )
 
         for child in self.children:
             if child is not self and not child.is_compiled():
--- a/datajunction-server/datajunction_server/sql/parsing/ast.py
+++ b/datajunction-server/datajunction_server/sql/parsing/ast.py
@@ -2650,6 +2650,29 @@ class Query(TableExpression, UnNamed):
         if self._is_compiled:
             return
 
+        def _compile(info: Tuple[Column, List[TableExpression]]):
+            """
+            Given a list of table sources, find a matching origin table for the column.
+            """
+            col, table_options = info
+            matching_origin_tables = 0
+            for option in table_options:
+                namespace = col.namespace[0].name if col.namespace else None
+                table_alias = option.alias.name if option.alias else None
+                if not namespace or (namespace and namespace == table_alias):
+                    result = option.add_reference_column(col)
+                    if result:
+                        matching_origin_tables += 1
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -1226,7 +1227,16 @@ async def add_ref_column(
                     "add Column ref without a compilation context.",
                 )
             await self.compile(ctx)
+        return self.add_reference_column(column)
 
+    def add_reference_column(
```
agorajek, line 1232: nit: `add_column_reference` is more accurate, no?

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -2640,6 +2650,74 @@ async def compile(self, ctx: CompileContext):
...
+        def _compile(info: Tuple[Column, List[TableExpression]]):
+            """
+            Given a list of table sources, find a matching origin table for the column.
+            """
+            col, table_options = info
+            matching_origin_tables = 0
+            for option in table_options:
+                namespace = col.namespace[0].name if col.namespace else None
+                table_alias = option.alias.name if option.alias else None
+                if not namespace or (namespace and namespace == table_alias):
```
agorajek, line 2662:
```suggestion
                if namespace is None or namespace == table_alias:
```

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -2640,6 +2650,74 @@ async def compile(self, ctx: CompileContext):
...
+                    result = option.add_reference_column(col)
+                    if result:
+                        matching_origin_tables += 1
+                        col._is_compiled = True
```
agorajek, line 2666: If this internal flag is set because `result = option.add_reference_column(col)` was succesfull, then shouldn't we just set it as part of the `add_reference_column` function call?

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -2640,6 +2650,74 @@ async def compile(self, ctx: CompileContext):
...
+            columns_to_compile = []
+            for expression in expressions_to_compile:
+                if expression and not isinstance(expression, list):
+                    columns_to_compile += list(expression.find_all(Column))
+                if isinstance(expression, list):
+                    columns_to_compile += [
+                        col for expr in expression for col in expr.find_all(Column)
+                    ]
+
+            with ThreadPoolExecutor() as executor:
```
agorajek, line 2713: Oh interesting... this should def speed things up. But I wonder if these executions can still use some partial caching or is every one of them going to repeat the same work for columns originating from the same table. I guess it depend what optimizations are done in `_compile()`.

agorajek (approved):
Nice, I am sure this will have a positive perf effect. Although CPUs will get hotter :)
I wonder what is the default thread pool size, but I am sure I can google it...

datajunction-server/datajunction_server/sql/parsing/ast.py
shangyian: The problem is that it's not clear to me that all that's needed to compile `col` is calling `option.add_column_reference`. We're using it as a shortcut here, but I'm not sure that other calls that use this same function are using it this way?

datajunction-server/datajunction_server/sql/parsing/ast.py
shangyian: The primary optimization from the changes to `_compile` is the fact that we now start with a parent that already exists, so we're not going to every child and then traversing up the tree to find the parents, which actually consumes quite a bit of time.

Issue #1175, opened 2024-09-24
Support `ORDER BY` for new SQL build


shangyian:
The new SQL builder should support `ORDER BY` clauses

PR #1176, opened 2024-09-24
Add `ORDER BY` support to new SQL build

shangyian:
### Summary

Add `ORDER BY` support to the new SQL build. This change also adds the `orderby` option for the `/sql/measures/v2` endpoint, allowing users to specify order by clauses like `default.hard_hat.state DESC`.

### Test Plan

- [x] PR has an associated issue: #1175 
- [x] `make check` passes


agorajek (approved):
Nice.

PR #1177, opened 2024-09-24
Fix bug with handling inline table queries

shangyian:
### Summary

This PR fixes an issue with queries that use inline tables, where they aren't parsed properly into the query AST. An example of such a query:
```
SELECT
  tab.source 
FROM 
  VALUES ('a'), ('b'), ('c') AS tab(source)
```

While other inline table queries were working, those with only one column per row, like the example above, were failing. This is due to the way the parsed structure would return a nested list in the cases where there was more than one column per row, but an unnested list for the example above. This change handles the special case.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/parsing/backends/antlr4.py
+++ b/datajunction-server/datajunction_server/sql/parsing/backends/antlr4.py
@@ -293,13 +293,13 @@ def _(ctx: sbp.InlineTableContext):
     alias, columns = visit(ctx.tableAlias())
 
     # Generate default column aliases if they weren't specified
-    args = args[0] if isinstance(args[0], list) else args
+    col_args = args[0] if isinstance(args[0], list) else args
     inline_table_columns = (
-        [ast.Column(col, _type=value.type) for col, value in zip(columns, args)]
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/sql/parsing/backends/antlr4.py
```diff
@@ -287,12 +293,13 @@ def _(ctx: sbp.InlineTableContext):
     alias, columns = visit(ctx.tableAlias())
 
     # Generate default column aliases if they weren't specified
+    col_args = args[0] if isinstance(args[0], list) else args
```
samredai, line 296: I'm unsure here but if this instead did `col_args = [args[0]] if not isinstance(args[0], list) else args[0]` to guarantee this is always a list, would that require less changes elsewhere?

samredai (approved):
LGTM but I left one question

datajunction-server/datajunction_server/sql/parsing/backends/antlr4.py
shangyian: Actually I made a change slightly further down that simplified the changes in the ast:
```
        values=[[value] if not isinstance(value, list) else value for value in args],
```

Issue #1178, opened 2024-09-26
Support dimension links with `CROSS JOIN` as the join type


shangyian:
While we allow the `CROSS` join type to be passed in as a dimension link join type, we don't actually support it because we don't allow for the join `ON` clause to be empty.

PR #1179, opened 2024-09-26
Support dimension links with `CROSS JOIN` 

shangyian:
### Summary

This change supports using `CROSS JOIN` as the join type when creating a dimension join link. While we allow the `CROSS` join type to be passed in when creating dimension links, we don't actually support this join type because we don't allow for the join ON clause to be empty (typical in cross joins).

### Test Plan

- [x] PR has an associated issue: #1178 
- [x] `make check` passes


anhqle:
> While we allow the CROSS join type to be passed in when creating dimension links, we don't actually support this join type because we don't allow for the join ON clause to be empty (typical in cross joins).

hmm, so what new functionality are we gaining here?

shangyian:
@anhqle this is making it so that dimension links with join type `CROSS` actually work when generating SQL (so the generated SQL will have the cross join). This setup could mean that things like cross join to some window could be done only optionally when requested through dimension links.

PR #1180, opened 2024-09-26
Caching and cache invalidation for the expensive dimension dag computation

samredai:
### Summary

This adds a dependency injected application cache and uses it in the dimensions dag logic. This is to help ease the burden of the expensive recursive query required to compute that. Here is an outline of a few things happening in this PR:
- There's now a very simple `CacheInterface` abstract base class that has a `get` and `set` method as well as a `get_cache` dependency that returns an implementation.
- The OSS implementation just uses `cachelib.SimpleCache` but I've added some docs to the doc site on how to implement a custom cache so someone could easily switch to `cachelib.redis.RedisCache`, for example, if they wanted to.
- The OSS implementation honors `no-cache` in the headers and it does this by returning `None` from the cache dependency. In all of the logic, cache is only used if the dependency returns a cache instance so this works nicely. (I also recommended in the docs that custom implementations also honor this header)
- The `get_dimensions` function is the only place I've added logic that adds to the cache where it caches the dimension dag with a cache key that's specific to the node name, node version, and the type of data we're caching (in this case, "dimensions"). There's a helper function to generate the cache key which looks like `"node_name_here@node_version_here:dimensions"`. In the future we can cache other things besides dimensions and then add additional values to the `CachedValueType` enum, for example "sql" which would then be cached as `"node_name_here@node_version_here:sql"`
- Cache invalidation for the dimension dag happens whenever a new dimension link is performed on that specific node. That means in the following endpoints:
  - POST "/nodes/{name}/columns/{column}/"
  - DELETE "/nodes/{name}/columns/{column}/"
  - POST "/nodes/{node_name}/link/"
  - DELETE "/nodes/{node_name}/link/"

There's a lot of nuance here on how to avoid certain user experiences so I'm hoping we can continue to discuss and iterate on this.

### Test Plan

Wrote a few tests and in a couple I mocked the cache dependency to ensure the right calls were being made.

- [ ] `make check` passes


datajunction-server/datajunction_server/api/metrics.py
```diff
@@ -96,13 +98,20 @@ async def list_metric_metadata() -> MetricMetadataOptions:
 
 @router.get("/metrics/{name}/", response_model=Metric)
 async def get_a_metric(
-    name: str, *, session: AsyncSession = Depends(get_session)
+    name: str,
+    *,
+    session: AsyncSession = Depends(get_session),
+    application_cache: Optional[CacheInterface] = Depends(get_cache),
```
agorajek, line 104: Why `application_cache` and not simply `cache`?

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -842,14 +847,26 @@ async def link_dimension(
...
     )
+    if application_cache:
+        application_cache.set(
+            key=get_cache_key_for_node(
+                name=node.name,  # type: ignore
+                version=node.current_version,  # type: ignore
+                cached_value_type=CachedValueTypes.DIMENSIONS,
+            ),
+            value=None,
+            timeout=0,
```
agorajek, line 858: Maybe expose this `0` as a constant somewhere in the header to make it easier to find and adjust later?

datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
```diff
@@ -0,0 +1,34 @@
...
+        super().get(key)
+        return self.cache.get(key)
+
+    def set(self, key: str, value: Any, timeout: int = 3600) -> None:
+        """Cache a value in the simple cache"""
+        super().set(key, value, timeout)
+        self.cache.set(key, value, timeout=timeout)
+
+
+def get_cache(request: Request) -> Optional[CacheInterface]:
```
agorajek, line 30: Why `-> Optional[CacheInterface]` instead of `-> Optional[Cache]` ?

agorajek (approved):
This is great. Super clear and easy to extend further. Nice docs as well!

docs/content/0.1.0/docs/deploying-dj/caching.md
```diff
@@ -0,0 +1,107 @@
+---
```
shangyian, line 1: Yay thanks for the documentation!

datajunction-server/datajunction_server/api/metrics.py
samredai: I initially had that but it conflicted with the `fastapi_cache.decorator.cache` that we use as `@cache` on the server endpoints which is just caching responses based on the request signature. It made me realize we may get these caches confused in the future so went for a more explicit `application_cache` variable name so we know this is the dependency injected cache to do caching/cache-invalidation within the application logic.

datajunction-server/datajunction_server/api/nodes.py
samredai: Hmm, I don't think we would ever change this value though. This is effectively deleting the entry in the cache by setting the value to `None` and the timeout to 0 so that it expires immediately. I could have added a `.delete(...)` method to the cache interface but I chose to keep that as simple as possible with just the `get` and `set`. What I'm saying is that if it was a constant it would basically be `ZERO=0` which isn't that helpful.

datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
samredai: The main reason is that there are a number of situations where we want to disable the cache, the most obvious one is when `no-cache` is in the headers. I wanted to leave this in total control of the dependency injected function though (someone may want to extend that with very intricate logic, such as cache for requests from app A but not from app B). In order to keep that complexity out of the OSS codebase in the future, all of the code is written as "if a cache is returned, it is used", so that anyone in their own dependency can write their own rules on when to return the cache and when not to.

Wrt to OSS, not returning a cache currently only happens when `no-cache` is found in the cache headers.

datajunction-server/datajunction_server/internal/caching/interface.py
```diff
@@ -0,0 +1,35 @@
+"""
+Caching interface and logging wrapper
+"""
+import logging
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+
+class CacheInterface(ABC):
```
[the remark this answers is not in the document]
samredai: Let me know if I'm misunderstanding but do you mean require that an implementation of this [Backend](https://github.com/long2ice/fastapi-cache/blob/main/fastapi_cache/types.py#L25-L40) class is returned from the dependency function and then we create a `FastAPICache` instance from it? I was under the impression that fastapi-cache was purely for caching request-based responses but this PR is trying to solve a case that spans many different request signatures.

datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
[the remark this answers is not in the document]
samredai: I didn't add it as a global app dependency but I add it using `Depends` to the specific endpoints where we need it. For example:
```py
@router.post("/nodes/{name}/columns/{column}/", status_code=201)
async def link_dimension(
    name: str,
    column: str,
    dimension: str,
    dimension_column: Optional[str] = None,  # pylint: disable=unused-argument
    session: AsyncSession = Depends(get_session),
    current_user: User = Depends(get_and_update_current_user),
    application_cache: Optional[CacheInterface] = Depends(get_cache),
) -> JSONResponse:
```
But you made me realize that I'm not instantiating a single cache instance for the dependency to return and instead am returning a new one every time 🤦‍♂️ . Just committed a 2 line change to fix that!

samredai:
@agorajek @shangyian I found one more case where we need to invalidate the cache. When we delete (deactivate) or hard delete a dimension node, we have to invalidate the cache for any nodes that were linked to it. Since we don't call the link/unlink endpoints as part of that process, this PR won't handle that. Any thoughts on the best strategy here? These are the two endpoints in question (and see the single failing test).
```py
@router.delete("/nodes/{name}/")
async def delete_node(
    name: str,
    *,
    session: AsyncSession = Depends(get_session),
    current_user: User = Depends(get_and_update_current_user),
):
    """
    Delete (aka deactivate) the specified node.
    """
    await deactivate_node(session=session, name=name, current_user=current_user)
    return JSONResponse(
        status_code=HTTPStatus.OK,
        content={"message": f"Node `{name}` has been successfully deleted."},
    )


@router.delete("/nodes/{name}/hard/", name="Hard Delete a DJ Node")
async def hard_delete(
    name: str,
    session: AsyncSession = Depends(get_session),
    current_user: User = Depends(get_and_update_current_user),
) -> JSONResponse:
    """
    Hard delete a node, destroying all links and invalidating all downstream nodes.
    This should be used with caution, deactivating a node is preferred.
    """
    impact = await hard_delete_node(
        name=name,
        session=session,
        current_user=current_user,
    )
    return JSONResponse(
        status_code=HTTPStatus.OK,
        content={
            "message": f"The node `{name}` has been completely removed.",
            "impact": impact,
        },
    )
    ```

samredai:
> An alternative though is...should we allow deleting a dimension that is still being linked to?
I decided to try this route and added commit [160da7a](https://github.com/DataJunction/dj/pull/1180/commits/160da7a1aa2de5a039a544e7ecd6bbf0db1ba057). It's easy to revert if we decide to go another way but this prevent deactivating a dimension node that other nodes have links to while still allowing a hard delete. We always have big danger warnings around hard deletes so it doesn't seem so bad that it could lead to some stale caching until ttls are hit.

So the one behavior that I can think of that we'd be living with for now is:
- A dimension dag is retrieved for a transform and cached
- Subsequent requests for the dimension dag for that transform and version is pulled from the cache
- A dimension node in that dag is hard deleted (without unlinking from the transform)
- Subsequent requests still use the (now stale) cached dimension dag until the TTL or 'no-cache' headers

Let's discuss this fully in the next sync.

datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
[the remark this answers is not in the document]
samredai: PR #1183

datajunction-server/datajunction_server/api/metrics.py
agorajek: The risk of confusing them is non-trivial and could lead to some unexpected behavior.

PR #1183, opened 2024-09-30
fastapi cache on dimension dag endpoints

samredai:
### Summary

This is a much less complex alternative to PR #1180 that uses `fastapi-cache` (which was already setup) and avoids any granular cache invalidation, doing a blanket cache on the endpoints instead. The tradeoff here is that users could hit cached data if they're bouncing between "builder" operations and calling any of the cached endpoints that perform the expensive dimension dag generation.

I really do like the idea of avoiding the web of complexity that comes with cache invalidation in a graph system and so I'm really scratching my head here trying to figure out if this PR and maybe a follow-up PR that adds modest rate limiting is all we need to provide the performance boost we need without degrading the user experience in any meaningful way.

### Test Plan

Updated tests, mainly to set cache-control headers.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/metrics.py
+++ b/datajunction-server/datajunction_server/api/metrics.py
@@ -6,6 +6,7 @@ from http import HTTPStatus
 from typing import List, Optional
 
 from fastapi import Depends, HTTPException, Query
+from fastapi_cache.decorator import cache
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import selectinload
@@ -95,6 +96,7 @@ async def list_metric_metadata() -> MetricMetadataOptions:
 
 
 @router.get("/metrics/{name}/", response_model=Metric)
+@cache(expire=settings.index_cache_expire)
 async def get_a_metric(
     name: str, *, session: AsyncSession = Depends(get_session)
 ) -> Metric:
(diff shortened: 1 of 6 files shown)


samredai:
@netlify /rebuild

shangyian (approved):
Yeah, I also like this approach for its simplicity, since we can take our time to add in the right cache invalidation strategies. I suspect there's going to be a lot of edge cases...

However, I'm wondering if the Python Builder client should switch to using `no-cache` by default, since the target audience of that is mostly users who are making one-off manual changes and then making a few read requests shortly after to verify their changes. For that group of users, seeing the results immediately would be beneficial. It could also be configurable at the client level, of course.

agorajek (approved):
I like this PR better. I assume we still need to override this implementation internally?

samredai:
> I suspect there's going to be a lot of edge cases...

Definitely. I got complete test coverage on the other PR but it involved handling some of the edge cases I found and I'm pretty confident there's more that I haven't found.

> I'm wondering if the Python Builder client should switch to using `no-cache` by default

I was actually thinking we'd go the other direction and require they explicitly disable the cache with something like a simple "disable_cache=True" arg, since it's easier to switch the default later to no cache than it would be in reverse. Also, if the cache is disabled by default I don't know that most users would care to enable it to unburden the backend.

shangyian:
> I don't know that most users would care to enable it to unburden the backend.

Yeah, perhaps not in the OSS client, but I think in our case internally, we don't really see this issue from users of the client (whose usage patterns are not particularly automated), so I think it's better to just optimize there for user experience by defaulting to no caching. But it's fine if we don't add this into OSS

PR #1189, opened 2024-10-02
Make sure that we detect cycles in the dimensions traversal queries

shangyian:
### Summary

This change adds cycle detection to the recursive queries used to traverse the dimensions graph. We do so by adding `CYCLE node_revision_id SET is_cycle USING path` to the recursive CTE, which is recommended in [the Postgres docs](https://www.postgresql.org/docs/16/queries-with.html#QUERIES-WITH-CYCLE).

We also add a `depth` parameter to the API call, so that clients can make less expensive queries if they only need a few levels and not the entire graph. The default `depth` is set to 30, which is reasonable for the vast majority of cases. Note that if we actually have circular loops, keeping the cycle detection from above will make it significantly faster for the query to return, since it will stop traversing immediately after detecting a cycle.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -168,20 +168,24 @@ async def get_upstream_nodes(
         )
 
     dag = (
-        select(
-            NodeRelationship.child_id,
-            NodeRevision.id,
-            NodeRevision.node_id,
+        (
+            select(
+                NodeRelationship.child_id,
+                NodeRevision.id,
+                NodeRevision.node_id,
+            )
+            .where(NodeRelationship.child_id == node.current.id)
+            .join(Node, NodeRelationship.parent_id == Node.id)
+            .join(
+                NodeRevision,
+                (Node.id == NodeRevision.node_id)
+                & (Node.current_version == NodeRevision.version),
+            )
         )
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -301,6 +303,7 @@ async def get_dimensions_dag(  # pylint: disable=too-many-locals
         )
         .where(initial_node.id == node_revision.id)
     ).cte("dimensions_graph", recursive=True)
+    dimensions_graph = dimensions_graph.suffix_with("CYCLE node_revision_id SET is_cycle USING path")
```
agorajek: This is a nice SQL way to define cycle detection.

agorajek (commented):
Looks great. Wonder how it works... 

Should we add this to other methods here or is `get_dimensions_dag` the only place that needs it?

samredai (approved):
This is great, thanks a ton @shangyian. I say we move forward with these and monitor it for a bit. Even when it's not enough to bog down the server, we do see the occasional dropped connection that I suspect will completely go away after this PR--which would signal this fix is all we really need for this specific problem.

agorajek (approved):

PR #1193, opened 2024-10-04
Cleanup YAML project code.

agorajek:
### Summary

Few things in this PR:
1. Added `update_if_exists` flag to namespace and tag creation.
2. Made the above flag default to False everywhere. It used to be True on the node, which I think is not great default. cc @anhqle 
3. Added system project namespace cleanup when run inside validation.

### Test Plan

Unit tests and my local examples.

- [ ] `make check` passes

### Deployment Plan
Asap.

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -11,6 +11,7 @@ from datajunction.exceptions import (
     DJClientException,
     DJNamespaceAlreadyExists,
     DJTableAlreadyRegistered,
+    DJTagAlreadyExists,
     DJViewAlreadyRegistered,
 )
 from datajunction.nodes import (
@@ -42,7 +43,11 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
             raise DJClientException(f"Namespace `{namespace}` does not exist.")
         return Namespace(namespace=namespace, dj_client=self)
 
-    def create_namespace(self, namespace: str) -> "Namespace":
+    def create_namespace(
+        self,
+        namespace: str,
+        update_if_exists: bool = False,
+    ) -> "Namespace":
         """
         Create a namespace with a given name.
         """
@@ -51,7 +56,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
             timeout=self._timeout,
         )
         json_response = response.json()
-        if response.status_code == 409:
+        if response.status_code == 409 and not update_if_exists:
             raise DJNamespaceAlreadyExists(json_response["message"])
         return Namespace(namespace=namespace, dj_client=self)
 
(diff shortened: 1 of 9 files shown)


datajunction-clients/python/datajunction/builder.py
```diff
@@ -227,7 +232,7 @@ def create_source(  # pylint: disable=too-many-arguments
         description: Optional[str] = None,
         columns: Optional[List[models.Column]] = None,
         primary_key: Optional[List[str]] = None,
-        tags: Optional[List[Tag]] = None,
+        tags: Optional[List[str]] = None,
```
shangyian, line 235: Nice, this will definitely make it easier for people to use

datajunction-clients/python/datajunction/compile.py
```diff
@@ -680,40 +667,43 @@ def _deploy_nodes(
...
                 if isinstance(
                     node_config.definition,
                     (TransformYAML, DimensionYAML, MetricYAML),
                 ):
                     rendered_node_config.definition.query = render_prefixes(
-                        node_config.definition.query,
+                        rendered_node_config.definition.query or "",
                         prefix,
                     )
+                # pre-fix the tags
```
shangyian, line 680: Is this essentially like a namespace for tags? I wonder if we should just explicitly give tags a similar kind of namespaced treatment

shangyian (approved):
> Made the above flag default to False everywhere. It used to be True on the node, which I think is not great default.

@agorajek From what I've seen, a lot of people seem to rely on this flag being `true`... of course we can announce that this is a breaking change, but I'm wondering if that's just the default behavior that people expect or prefer?

datajunction-clients/python/datajunction/compile.py
agorajek: Yes, that's right. Although on the 2nd look I think this code will prevent us from using any existing tags in these nodes.

agorajek:
> Changes largely LGTM!
> 
> Only one comment:
> 
> > Made the above flag default to False everywhere. It used to be True on the node, which I think is not great default.
> 
> @agorajek From what I've seen, a lot of people seem to rely on this flag being `true`... of course we can announce that this is a breaking change, but I'm wondering if that's just the default behavior that people expect or prefer?

@shangyian I am worried that the current setup (created_node --> update_if_exists=True) will lead to silent overrides. I think I could do the following:
- keep this change, `create_node( ..., update_if_exists: default False)` and 
- add a new method to the client `create_or_update_node( ..., update_if_exists: default True)` 
This should make it easy for users to pick the expected behavior easily. cc @anhqle 

Thoughts?

anhqle (approved):
I'm good with changing the default of `update_if_exists` to False.

In that case, it'd be nice to tell user about the `update_if_exists` functionality when they run into node-already-exists-error.

agorajek:
> I'm good with changing the default of `update_if_exists` to False.
> 
> In that case, it'd be nice to tell user about the `update_if_exists` functionality when they run into node-already-exists-error.

Yes, we have that already, example:
```
datajunction.exceptions.DJClientException: Node `default.account_type_table` already exists. Use `update_if_exists=True` to update the node.
```

PR #1195, opened 2024-10-07
Minor SQL build fixes

shangyian:
### Summary

This PR has two fixes to the SQL building:
* The `upstream_tables` list for built SQL should only include source nodes. Including non-source-nodes leads to issues downstream when we try to search for the physical table.
* When searching the dimensions graph during SQL building, we should use breadth-first and not depth-first search. Otherwise we'll end up performing a more expensive join to pull in a dimension when we could've used the shortest path.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -10,7 +10,7 @@ import re
 import time
 import uuid
 from http import HTTPStatus
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple, cast
 
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -727,6 +727,9 @@ async def build_sql_for_multiple_metrics(  # pylint: disable=too-many-arguments,
         assemble_column_metadata(col)  # type: ignore
         for col in query_ast.select.projection
     ]
+    upstream_tables = [tbl for tbl in query_ast.find_all(ast.Table) if tbl.dj_node]
+    for tbl in upstream_tables:
+        await session.refresh(tbl.dj_node, ["availability"])
     return (
         TranslatedSQL(
             sql=str(query_ast),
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -739,7 +739,7 @@ async def build_sql_for_multiple_metrics(  # pylint: disable=too-many-arguments,
                 # It must be either a Source node or some type of Materialized node.
                 # Even if there is no availability state for it right now, this qery may be built
                 # for future execution, so we don't need to check for availability state here.
-                if tbl.dj_node
+                if tbl.dj_node and tbl.dj_node.type == NodeType.SOURCE
```
agorajek, line 742: I see what you are saying... but this seems to differ from the comment lines 740-741. No?

shangyian: Hmm, I don't think we should include nodes with no availability state in this list of upstream tables though. I added a check to include nodes with a materialization. 

I guess the real question is the purpose of this list -- we're currently using it as a way of determining the list of upstream physical tables that this generated query references, but perhaps we shouldn't use it that way?

datajunction-server/tests/api/dimension_links_test.py
```diff
@@ -503,7 +503,7 @@ async def test_link_complex_dimension_with_role(
 FROM default_DOT_events
 LEFT JOIN default_DOT_users
   ON default_DOT_events.user_id = default_DOT_users.user_id
-  AND default_DOT_events.event_start_date BETWEEN default_DOT_users.snapshot_date AND CAST(DATE_ADD(CAST(default_DOT_users.snapshot_date AS DATE), 10) AS INT)
+  AND default_DOT_events.event_start_date = default_DOT_users.snapshot_date
```
agorajek, line 506: Any ideas why did this unit test changed?

agorajek (approved):
Looks good! Just one question in-line.

datajunction-server/tests/api/dimension_links_test.py
shangyian: This is most likely from the join order changing for dimensions discovery, but also the old unit tests don't look quite right either

PR #1198, opened 2024-10-09
Requesting metrics SQL should work for metrics that have different pa…

shangyian:
…rents as long as they share the same dimensions

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/sql.py
+++ b/datajunction-server/datajunction_server/api/sql.py
@@ -124,7 +124,6 @@ async def build_and_save_node_sql(  # pylint: disable=too-many-locals
     # If it's a cube, we'll build SQL for the metrics in the cube, along with any additional
     # dimensions or filters provided in the arguments
     if node.type == NodeType.CUBE:
-        await session.refresh(node.current, ["cube_node_dimensions"])
         dimensions = list(
             OrderedDict.fromkeys(node.current.cube_node_dimensions + dimensions),
         )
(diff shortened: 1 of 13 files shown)


datajunction-server/datajunction_server/construction/build.py
```diff
@@ -270,7 +270,41 @@ async def build_metric_nodes(
 
         await parent_ast.compile(context)
         measures_queries.append(parent_ast)
-    return measures_queries[0]
+
+    # Join together the transforms on the shared dimensions and select all
+    # requested metrics and dimensions in the final select projection
+    base_query = measures_queries[0]
+    for parent_query in measures_queries[1:]:
```
agorajek, line 277: These parent_queries are already not identical, right?

datajunction-server/datajunction_server/api/sql.py
```diff
@@ -124,6 +124,7 @@ async def build_and_save_node_sql(  # pylint: disable=too-many-locals
     # If it's a cube, we'll build SQL for the metrics in the cube, along with any additional
     # dimensions or filters provided in the arguments
     if node.type == NodeType.CUBE:
+        await session.refresh(node.current, ["cube_node_dimensions"])
```
agorajek, line 127: What's this for?

agorajek (approved):
Thanks for quick fix. This used to work for the build v1, right?

Issue #1192, opened 2024-10-04
Cleanup Python client YAML project.


agorajek:
Mostly to account for some recent DJ client changes.
And to add some validation cleanup.

PR #1199, opened 2024-10-09
Cleanup update_if_exists vs skip_if_exists.

agorajek:
### Summary

I noticed that in the case of create_namespace() and create_tag() the argument `update_if_exists` was misleading. Adjusted it to `skip_if_exists` . The update if exist functionality is still possible through the get()/save() pattern.

Also found one more issue with the _deploy_namespaces() method of the project compile.
It was erroring out because of existing namespaces, example:
```
  users.olek.project1.xp.source-controlled     tag         Tag users.olek.project1.xp.source-controlled successfully created (or updated)
  users.olek.project1                          namespace   {'message': 'Node namespace `users.olek.project1` already exists'}
```

### Test Plan

Unit test plus test deployments of my example YAML project.

Now getting a warning if the namespace exists:
```
                                                                                                                      XP Nodes
                                                                                                     Deployment for Prefix: users.olek.project1

  Name                                         Type        Message
 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  users.olek.project1.xp.source-controlled     tag         Tag users.olek.project1.xp.source-controlled successfully created (or updated)
  users.olek.project1                          namespace   Namespace users.olek.project1 already exists
  users.olek.project1.xp                       namespace   Namespace users.olek.project1.xp already exists
  users.olek.project1.xp.foo-bar               dimension   Node users.olek.project1.foo-bar successfully created (or updated)
...
```

- [x] PR has an associated issue: #1192 
- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/__about__.py
+++ b/datajunction-clients/python/datajunction/__about__.py
@@ -1,4 +1,4 @@
 """
 Version for Hatch
 """
-__version__ = "0.0.1a68.dev0"
+__version__ = "0.0.1a68"
--- a/datajunction-clients/python/datajunction/__about__.py
+++ b/datajunction-clients/python/datajunction/__about__.py
@@ -1,4 +1,4 @@
 """
 Version for Hatch
 """
-__version__ = "0.0.1a68"
+__version__ = "0.0.1a68.dev0"
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -487,7 +487,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
         description: Optional[str],
         tag_metadata: Dict,
         tag_type: str,
-        skip_if_exists: bool = False,
+        update_if_exists: bool = False,
     ) -> Tag:
         """
         Create a tag with a given name.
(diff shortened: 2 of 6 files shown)


datajunction-clients/python/datajunction/builder.py
```diff
@@ -502,7 +502,7 @@ def create_tag(  # pylint: disable=too-many-arguments
         try:
             self._create_tag(tag=new_tag)
         except DJTagAlreadyExists as exc:
-            if not update_if_exists:
+            if not skip_if_exists:
```
shangyian, line 505: How come for tags it's this  flag? Like I could imagine a need to update the tag if it exists, say if the metadata json blob on the tag changes, or the display name changes

shangyian (approved):

datajunction-clients/python/datajunction/builder.py
agorajek: You are right... and there is a way to do that through tag's `get` and `save` methods. Just these `create_<something>` had a different approach. It was just too much to fix in one PR.

datajunction-clients/python/datajunction/__about__.py
```diff
@@ -1,4 +1,4 @@
 """
 Version for Hatch
 """
-__version__ = "0.0.1a68"
+__version__ = "0.0.1a68.dev0"
```
agorajek, line 4: oh shoot

datajunction-clients/python/datajunction/builder.py
agorajek: It wasn't that bad after all. Changed `create_tag()` to update_if_exists.

PR #1200, opened 2024-10-11
Fix retrieving cube SQL failure

shangyian:
### Summary

There are two bugfixes in this PR:

(1) In some cases, retrieving cube SQL fails with this error:
```
greenlet_spawn has not been called; can't call await_only() here. Was IO attempted in an unexpected place?
```
This is because the node object loaded from the database did not have cube metadata on it pre-loaded, like `cube_elements`.

(2) We should refresh the metrics SQL cache every time it's requested. Otherwise when we deploy changes to the SQL build, we won't keep the cache refreshed with the changes.

### Test Plan

Tested locally

- [x] `make check` passes

### Deployment Plan

ASAP

### DIFF
--- a/datajunction-server/datajunction_server/api/sql.py
+++ b/datajunction-server/datajunction_server/api/sql.py
@@ -124,6 +124,10 @@ async def build_and_save_node_sql(  # pylint: disable=too-many-locals
     # If it's a cube, we'll build SQL for the metrics in the cube, along with any additional
     # dimensions or filters provided in the arguments
     if node.type == NodeType.CUBE:
+        node = cast(
+            Node,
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/sql.py
```diff
@@ -124,6 +124,10 @@ async def build_and_save_node_sql(  # pylint: disable=too-many-locals
     # If it's a cube, we'll build SQL for the metrics in the cube, along with any additional
     # dimensions or filters provided in the arguments
     if node.type == NodeType.CUBE:
+        node = cast(
```
agorajek, line 127: nice trick!

agorajek (approved):

PR #1209, opened 2024-10-21
Added cache interface and cachelib + noop implementations

samredai:
### Summary

This breaks out just the cache interface from #1180 and leaves out any use of the dependency injection anywhere. I've also added a `NoOpCache` implementation of the `CacheInterface` which makes it nicer to use the dependency while also keeping it optional (A `NoOpCache` has the same methods so no need to do a null check for the cache everywhere).

I also added the caching docs page that describes how this works and how someone can implement their own custom cache dependency.

### Test Plan

Wrote some tests to get full coverage.

- [ ] `make check` passes

### Deployment Plan

Deploying this has no effect because it's not used anywhere yet.

### DIFF
--- a/datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
+++ b/datajunction-server/datajunction_server/internal/caching/cachelib_cache.py
@@ -1,0 +1,43 @@
+"""
+Cachelib-based cache implementation
+"""
+from typing import Any, Optional
+
+from cachelib import SimpleCache
+from fastapi import Request
+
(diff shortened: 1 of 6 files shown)


agorajek (approved):
This is super clear. Love it. Let's get it in and start using.

shangyian (approved):
LGTM! Thanks for adding

agorajek:
@samredai let's not loose this one from the sight. When you have a moment try to test it and merge.

PR #1214, opened 2024-10-24
Add a dictionary of links to AvailabilityState.

agorajek:
### Summary

This feature allows to integrate our AvailabilityStates with external tools, such as data visualizations. For example... when a new availability state is added or updated, we can submit additional urls that can be shown on the Materialization tab of the node.

This PR adds:
- the API call to save the additional links information,
- the adjustment to the UI to show the `links` together with the Materialized Datasets

Example:
Screenshot 2024-10-24 at 10 55 16 AM

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/database/availabilitystate.py
+++ b/datajunction-server/datajunction_server/database/availabilitystate.py
@@ -3,7 +3,7 @@
 
 from datetime import datetime, timezone
 from functools import partial
-from typing import List, Optional
+from typing import Any, Dict, List, Optional
 
 import sqlalchemy as sa
 from sqlalchemy import JSON, DateTime, ForeignKey
@@ -31,6 +31,7 @@ class AvailabilityState(Base):  # pylint: disable=too-few-public-methods
     table: Mapped[str]
     valid_through_ts: Mapped[int] = mapped_column(sa.BigInteger())
     url: Mapped[Optional[str]]
+    links: Mapped[Optional[Dict[str, Any]]] = mapped_column(JSON, default=dict)
 
     # An ordered list of categorical partitions like ["country", "group_id"]
     # or ["region_id", "age_group"]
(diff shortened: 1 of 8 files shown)


samredai (approved):
This is awesome and is going to really unlock the seamless end to end dev flow we've been working towards. Huge thanks @agorajek. One thing I was wondering is if we should expect cases where someone wants to add links but not necessarily attach it to the materialization for a node. For example if a metric node has a doc that might be useful to link out to. I think it's enough to suggest they just put it in the metric description but just wanted to get your thoughts on that.

datajunction-server/datajunction_server/database/availabilitystate.py
```diff
@@ -31,6 +31,7 @@ class AvailabilityState(Base):  # pylint: disable=too-few-public-methods
     table: Mapped[str]
     valid_through_ts: Mapped[int] = mapped_column(sa.BigInteger())
     url: Mapped[Optional[str]]
```
shangyian, line 33: I wonder if we can actually remove `url` now, since that was just a single link, but with this change we'll have the option for multiple, and all named as well. Although there will be a database migration involved if we do remove.

shangyian (approved):
LGTM, thanks for adding this!

agorajek:
> This is awesome and is going to really unlock the seamless end to end dev flow we've been working towards. Huge thanks @agorajek. One thing I was wondering is if we should expect cases where someone wants to add links but not necessarily attach it to the materialization for a node. For example if a metric node has a doc that might be useful to link out to. I think it's enough to suggest they just put it in the metric description but just wanted to get your thoughts on that.

I think we will want to have "links" in various places. This one is specific to the Availability State / MaterializedDataset, hence I added it on that entity. In the case of other links I think we should place them as close to the proper entity as possible. Because metric nodes (an all nodes) have a description that's a good place to add a doc link, but I can see having a generic links field on the nodes as well. This way the integrations with other tools are easier to pull off.

datajunction-server/datajunction_server/database/availabilitystate.py
agorajek: I was thinking about it. And I noticed that the top level `url` attribute is typically used as a link to a canonical catalog for a particular dataset, which could make sense to be separate. But then it should be called `catalog_url` so maybe I should just rename it, if you agree with this idea?

agorajek: Actually `dataset_url` sounds more intuitive than `catalog_url`. What do you think?

agorajek: I'll skip this for now.

Issue #1213, opened 2024-10-22
Edit node page should have back navigation to namespace


shangyian:
We should update the navbar at the top of the edit node page so that you can click back to the namespace the node is in once you're done creating or editing it.

PR #1217, opened 2024-10-28
Create/edit node pages should have a navigation bar that goes back to the right namespace

shangyian:
### Summary

This change makes it so that both the create and edit node pages will have a namespace navigation bar that goes back to the right namespace. For example:

Screenshot 2024-10-28 at 8 16 39 AM

### Test Plan

- [x] PR has an associated issue: #1213 
- [x] `make check` passes


agorajek (approved):
Thanks!

Issue #1220, opened 2024-11-06
Python client not hydrating items as dataclasses


shangyian:
For example, the expected type of `type(col)` here should be `Column`:
```
source_node = dj.source("source.node")
[type(col) for col in source_node.columns]
```

However, it is currently a `Dict`. This means that we're initializing a `Source` object with `columns` of type `Dict` rather than `Column`, causing downstream issues for anyone using the client and expecting a particular type.

PR #1221, opened 2024-11-06
Fix python client model class hydration issue

shangyian:
### Summary

This fixes an issue where the python client's model classes weren't hydrating as dataclasses, but as dictionaries. For example, the expected type of `type(col)` here should be `Column`:
```
source_node = dj.source("source.node")
[type(col) for col in source_node.columns]
```

This fixes things by adding a custom serializer from dictionaries to DJ model dataclasses: `SerializableMixin`. This mixin includes a `from_dict` method that will examine the structure of the dataclass and recursively convert a dictionary to the dataclass.

### Test Plan

- [x] PR has an associated issue: #1220
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-clients/python/datajunction/_base.py
+++ b/datajunction-clients/python/datajunction/_base.py
@@ -1,3 +1,4 @@
+"""Base mixins for client dataclasses."""
 from dataclasses import fields, is_dataclass
 from typing import TYPE_CHECKING, Any, Dict, Type, TypeVar, Union, get_args, get_origin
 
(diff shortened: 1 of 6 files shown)


datajunction-clients/python/datajunction/_base.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import TYPE_CHECKING, Any, Dict, Type, TypeVar, Union, get_args, get_origin
+
+if TYPE_CHECKING:  # pragma: no cover
+    from datajunction.client import DJClient
+
+
+T = TypeVar("T")
+
+
+class SerializableMixin:  # pylint: disable=too-few-public-methods
```
agorajek, line 12: This could be part of _internal module, but no big deal either way.

agorajek (approved):
Thank you for the quick fix!

datajunction-clients/python/datajunction/_base.py
shangyian: Yeah, I originally had it there but it felt like that file got really long. This piece is pretty generic whereas that module is now mostly client models.

Issue #830, opened 2023-10-20
notifications


shangyian:
* A node should have an owner
* A node can have a watchlist
* A user can have a notifications list 
* Whenever a change is made that affects downstream nodes, we should add a notification to all users on each downstream's watchlist

PR #1225, opened 2024-11-14
A get_notifier dependency for sending notifications

samredai:
### Summary

This adds a `get_notifier` dependency for plugging in a custom notification mechanism.

### Test Plan

Wrote a test to test the default dependency which just logs the message.

- [x] PR has an associated issue: #830 
- [x] `make check` passes

### DIFF
--- a/docs/content/0.1.0/docs/deploying-dj/notifications.md
+++ b/docs/content/0.1.0/docs/deploying-dj/notifications.md
@@ -3,67 +3,55 @@ weight: 60
 title: Notifications
 ---
 
-In DataJunction, notifications play a vital role in keeping users informed about various activity, such as changes in
-the availability state of nodes. This section discusses how notifications are managed within DataJunction and how you
-can implement a custom notification solution using FastAPI's dependency injection.
+In DataJunction, notifications are crucial for keeping users informed about various activities, such as changes in the
+state of nodes. This document outlines how notifications are managed within DataJunction and how you can implement a
+custom notification solution using FastAPI's dependency injection.
 
-### How Notifications are Used
+## How Notifications are Used
 
 DataJunction uses notifications to alert users about significant events, such as updates to nodes or changes in their
-availability state. By handling notifications within the OSS project, DataJunction provides an opinionated take on where
-and when notifications should be sent.
+state. By handling notifications within the OSS project, DataJunction provides an opinionated take on where and when
+notifications should be sent.
 
-### Default Notification Implementation
+## Default Notification Implementation
 
-Out of the box, DataJunction includes a simple placeholder dependency that simply logs notifications.
+Out of the box, DataJunction includes a simple placeholder dependency that logs notifications. This implementation is
+designed to be straightforward and easily replaceable with a custom solution.
 
-Here's a brief look at this placeholder notification implementation:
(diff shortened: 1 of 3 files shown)


docs/content/0.1.0/docs/deploying-dj/notifications.md
```diff
@@ -0,0 +1,93 @@
...
+
+### How Notifications are Used
+
+DataJunction uses notifications to alert users about significant events, such as updates to nodes or changes in their
+availability state. By handling notifications within the OSS project, DataJunction provides an opinionated take on where
+and when notifications should be sent.
+
+### Default Notification Implementation
+
+Out of the box, DataJunction includes a simple placeholder dependecny that simply logs notifications.
```
agorajek, line 18: `s/dependecny/dependency`

docs/content/0.1.0/docs/deploying-dj/notifications.md
```diff
@@ -0,0 +1,93 @@
...
+
+You can implement a custom notification system by using FastAPI's dependency injection and injecting a `get_notifier`
+dependency. The custom notifier must handle the `notify` method, which processes the notification messages.
+
+#### Implementing a Custom Notifier
+
+To implement a custom notifier, create a function that processes notification messages and use FastAPI's dependency
+injection to inject your custom notifier.
+
+Here's an example of a custom notifier implementation:
```
agorajek, line 61: This may be silly, and not needed for seasoned engineers but I asking for a friend... :) Do you mind adding a note/sentence on where this change should happen?

agorajek (approved):
Thanks @samredai . This is great. I'll start calling this in places...

docs/content/0.1.0/docs/deploying-dj/notifications.md
samredai, line 18:
```suggestion
Out of the box, DataJunction includes a simple placeholder dependency that simply logs notifications.
```

docs/content/0.1.0/docs/deploying-dj/notifications.md
samredai: Great point, let me add the line for actually overriding the dependency in the app!

samredai: Done

samredai:
@agorajek thanks for our convo earlier, it gave me some really good ideas. I noticed that we already have [these](https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/database/history.py) enums for storing history events and I think it would be neat if we can reuse it for notifications. I updated this PR to do just that. The nice part about this is in areas where we save a history event (which we do for availability states), we can just reuse that event object without changing anything, so basically just `notify(event: History)`. Of course this still leaves us with the flexibility of creating an event just for notifications for cases where we don't need to store the history event but just want to send a notification.

agorajek:
> @agorajek thanks for our convo earlier, it gave me some really good ideas. I noticed that we already have [these](https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/database/history.py) enums for storing history events and I think it would be neat if we can reuse it for notifications. I updated this PR to do just that ....

This makes sense!

Issue #1226, opened 2024-11-21
Better errors when requesting SQL for metrics


agorajek:
Some feedback on error handling for DJ
- When passing metrics doesn't exist, we get
`{"request":"GET https://.../sql/?metrics=foo.bar.baz&use_materialized=True","message":"'NoneType' object has no attribute 'current'"}%`
- Passing a node that exists, but is not a metric, gives me different messages. Example, passing a cube gives me:
`{"message":"","errors":[{"code":204,"message":"","debug":null,"context":""}],"warnings":[]}`

PR #1227, opened 2024-11-22
Better input validation on get_sql_for_metrics().

agorajek:
### Summary

We did not properly check the input to `/sql` api. This PR does a better job, example:

1. When metrics arg has a non-existing node:
    - before: `'NoneType' object has no attribute 'current'`
    - after: `datajunction_server.errors.DJNodeNotFound: A node with name `foo.bar.baz` does not exist.`
2. When metrics argument has a non-metric node:
    - before: `{"message":"","errors":[{"code":204,"message":"","debug":null,"context":""}],"warnings":[]}`
    - after: `datajunction_server.errors.DJInvalidInputException: All nodes must be of metric type, but some are not: system.logs.cube (cube).`

### Test Plan

Locally.

- [x] PR has an associated issue: #1226 
- [x] `make check` passes

### Deployment Plan

auto

### DIFF
--- a/datajunction-server/datajunction_server/api/sql.py
+++ b/datajunction-server/datajunction_server/api/sql.py
@@ -4,6 +4,7 @@ SQL related APIs.
 """
 import logging
 from collections import OrderedDict
+from http import HTTPStatus
 from typing import List, Optional, Tuple, cast
 
 from fastapi import BackgroundTasks, Depends, Query
@@ -18,6 +19,7 @@ from datajunction_server.api.helpers import (
 from datajunction_server.database import Engine, Node
 from datajunction_server.database.queryrequest import QueryBuildType, QueryRequest
 from datajunction_server.database.user import User
+from datajunction_server.errors import DJInvalidInputException
 from datajunction_server.internal.access.authentication.http import SecureAPIRouter
 from datajunction_server.internal.access.authorization import validate_access
 from datajunction_server.internal.engines import get_engine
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/sql.py
```diff
@@ -378,6 +380,20 @@ async def get_sql_for_metrics(  # pylint: disable=too-many-locals
         base_verb=access.ResourceRequestVerb.READ,
     )
 
+    # make sure all metrics exist and have correct node type
+    nodes = [
+        await Node.get_by_name(session, node, raise_if_not_exists=True)
```
samredai, line 385: There's a `Node.get_by_names` method that lets you pass in a list and queries for them all at once right? Then you can just compare the counts.

samredai (commented):
Is that error `'NoneType' object has no attribute 'current'` coming from here? https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/api/helpers.py#L365-L380

That `validate_cube` function gets called somewhere along the way and I wonder if there's a way we can fix it in there by just breaking up the call so we can check if all the nodes exist first before loading all the relationships. Something like this:

```py
    metric_nodes: List[Node] = await Node.get_by_names(
        session,
        metric_names,
        options=[],
        include_inactive=False,
    )

    # Compare the counts here to make sure all of the metric names exist

    # Then only load the relationships once we know there will always be a Node.current
    metric_nodes = sorted(
        await session.execute(
            select(Node)
            .where(Node.id.in_([node.id for node in metric_nodes]))
            .options(
                    selectinload(NodeRevision.columns).options(
                        selectinload(Column.node_revisions),
                    ),
                    joinedload(NodeRevision.catalog),
                    selectinload(NodeRevision.parents),
                ),
            )
        ).scalars().all(),
        key=lambda x: metrics_sorting_order.get(x.name, 0),
    )
    ```

agorajek:
> Is that error `'NoneType' object has no attribute 'current'` coming from here? https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/api/helpers.py#L365-L380

Nope. It was coming from here: https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/database/queryrequest.py#L313

datajunction-server/datajunction_server/api/sql.py
agorajek: I didn't know... yes I could probably use it here.

samredai (approved):

samredai:
> > Is that error `'NoneType' object has no attribute 'current'` coming from here? https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/api/helpers.py#L365-L380
> 
> Nope. It was coming from here: https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/database/queryrequest.py#L313

I see. I don't want to hold this PR up but I'm worried about having to perform the query for the metric nodes twice--once to make sure they exist, and then again when we're generating the sql. I suspect there's a way to just make the second time we call it robust enough to raise the node not found error like you did here.

PR #1234, opened 2024-12-02
Add measures metadata for metrics

shangyian:
### Summary

The detailed context and explanation of the broader problem we're trying to solve is in [the issue](https://github.com/DataJunction/dj/issues/1223). To briefly summarize here:

For any set of metrics and dimensions, we can generate measures SQL, which is used to materialize a dataset that can serve those metrics and dimensions in a more efficient manner. Today, the generated measures SQL is non-optimized - the measures are computed at the level of the metrics' upstream transforms, without considering any downstream aggregation or grain optimizations.

We can make the generated measures SQL significantly more efficient by breaking down metrics into pre-aggregated (but still further aggregatable) **measures**. This PR provides appropriate appropriate metadata for a given metric on its measures, as a first step towards better measures SQL.

image

* **Metrics** are combinations of one or more measures with various aggregation or formulas applied (e.g., `SUM(sales_amount) * 100.0`, `AVG(revenue)`, `SUM(clicks) / SUM(impressions)`).
* **Measures** are components used to build metrics (e.g., `sales_amount`, `revenue`, `user_count`), along with aggregation and other metadata:
```
measures:
- name: sales_amount_sum
  expression: amount
  aggregation: SUM
  rule: FULL
- name: user_id_count_distinct
  expression: "distinct user_id"
  aggregation: COUNT
  rule: LIMITED
```

### Test Plan

- [x] PR has an associated issue: #1223
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/decompose.py
+++ b/datajunction-server/datajunction_server/sql/decompose.py
@@ -1,0 +1,98 @@
+from pydantic import BaseModel
+from datajunction_server.sql import functions as dj_functions
+from datajunction_server.sql.parsing.backends.antlr4 import ast, parse
+
+
+class Measure(BaseModel):
+    """
+    Measures are components used to build metrics (e.g., sales_amount, revenue, user_count).
+    """
+
+    name: str
+    expression: str  # A SQL expression for defining the measure
+    aggregation: str
+
+
+class MeasureExtractor:
+    def __init__(self):
+        """Register handlers for aggregation functions"""
+        self.handlers = {
+            dj_functions.Sum: self._generic_additive_agg,
+            dj_functions.Count: self._generic_additive_agg,
+            dj_functions.Max: self._generic_additive_agg,
+            dj_functions.Min: self._generic_additive_agg,
+            dj_functions.Avg: self._avg,
+        }
+
+    def extract_measures(self, metric_query: str) -> tuple[list[Measure], ast.Query]:
+        """
+        Decomposes the metric query into its constituent aggregatable measures and
+        constructs a SQL query derived from those measures.
+        """
+        query_ast = parse(metric_query)
+        measures = []
+
+        for func in query_ast.find_all(ast.Function):
+            handler = self.handlers.get(func.function())
+            if handler:
+                func_measures = handler(func)
+                if func_measures:
+                    self._update_ast(func, func_measures)
+                measures.extend(func_measures)
+
+        return measures, query_ast
+
+    def _generic_additive_agg(self, func) -> list[Measure]:
+        arg = func.args[0]
+        measure_name = "_".join([str(col) for col in arg.find_all(ast.Column)])
+        return [
+            Measure(
+                name=measure_name,
+                expression=str(arg),
+                aggregation=func.name.name,
+            ),
+        ]
+
+    def _avg(self, func) -> list[Measure]:
+        arg = func.args[0]
+        measure_name = "_".join([str(col) for col in arg.find_all(ast.Column)])
+        return [
+            Measure(
+                name=measure_name,
+                expression=str(arg),
+                aggregation=dj_functions.Sum,
+            ),
+            Measure(
+                name="count",
+                expression="1",
+                aggregation=dj_functions.Count,
+            ),
+        ]
+
+    def _update_ast(self, func, measures: list[Measure]):
+        """
+        Updates the query AST based on the measures derived from the function.
+        """
+        if func.function() == dj_functions.Avg:
+            func.parent.replace(
+                from_=func,
+                to=ast.BinaryOp(
+                    op=ast.BinaryOpKind.Divide,
+                    left=ast.Function(
+                        ast.Name("SUM"),
+                        args=[ast.Column(ast.Name(measures[0].name))],
+                    ),
+                    right=ast.Function(
+                        ast.Name("COUNT"),
+                        args=[ast.Column(ast.Name(measures[1].name))],
+                    ),
+                ),
+            )
+        elif func.function() == dj_functions.Count:
+            func.name.name = "SUM"
+            func.args = [ast.Column(ast.Name(measure.name)) for measure in measures]
+        else:
+            func.args = [ast.Column(ast.Name(measure.name)) for measure in measures]
+
+
+extractor = MeasureExtractor()
--- a/datajunction-server/datajunction_server/sql/decompose.py
+++ b/datajunction-server/datajunction_server/sql/decompose.py
@@ -77,8 +77,8 @@ class MeasureExtractor:
         for idx, func in enumerate(query_ast.find_all(ast.Function)):
             dj_function = func.function()
             handler = self.handlers.get(dj_function)
-            if handler:
-                if (func_measures := handler(func, idx)) and dj_function.is_aggregation:
+            if handler and dj_function.is_aggregation:
+                if func_measures := handler(func, idx):  # pragma: no cover
                     MeasureExtractor.update_ast(func, func_measures)
                 measures.extend(func_measures)
 
--- a/datajunction-server/datajunction_server/sql/decompose.py
+++ b/datajunction-server/datajunction_server/sql/decompose.py
@@ -75,10 +75,10 @@ class MeasureExtractor:
         measures = []
 
         for idx, func in enumerate(query_ast.find_all(ast.Function)):
-            handler = self.handlers.get(func.function())
+            dj_function = func.function()
+            handler = self.handlers.get(dj_function)
             if handler:
-                func_measures = handler(func, idx)
-                if func_measures:
+                if (func_measures := handler(func, idx)) and dj_function.is_aggregation:
                     MeasureExtractor.update_ast(func, func_measures)
                 measures.extend(func_measures)
 
--- a/datajunction-server/datajunction_server/sql/decompose.py
+++ b/datajunction-server/datajunction_server/sql/decompose.py
@@ -65,7 +65,7 @@ class MeasureExtractor:
             if handler:
                 func_measures = handler(func, idx)
                 if func_measures:
-                    self.update_ast(func, func_measures)
+                    MeasureExtractor.update_ast(func, func_measures)
                 measures.extend(func_measures)
 
         return measures, query_ast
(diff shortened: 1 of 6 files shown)


datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+                    type=Aggregability.FULL
+                    if func.quantifier != "DISTINCT"
+                    else Aggregability.LIMITED,
+                ),
+            ),
+        ]
+
+    def update_ast(self, func, measures: list[Measure]):
+        """
+        Updates the query AST based on the measures derived from the function.
```
agorajek, line 122: How does this function make any updates to `self`? Is it by updating the `func` pointers?

agorajek (commented):
@shangyian this is so good. Very clear and easy to follow. Minor questions in line.

I can't wait to use this in Cube's materialization and collapsing the 2 styles into one!

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+            [str(col) for col in arg.find_all(ast.Column)] + [func.name.name.lower()],
+        )
+        return [
+            Measure(
+                name=f"{measure_name}_{idx}",
+                expression=f"{func.quantifier} {arg}" if func.quantifier else str(arg),
+                aggregation=func.name.name.upper(),
+                rule=AggregationRule(
+                    type=Aggregability.FULL
+                    if func.quantifier != "DISTINCT"
```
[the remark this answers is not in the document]
shangyian: Good point, let me add this

datajunction-server/datajunction_server/sql/decompose.py
shangyian: Yeah, you're right, it actually doesn't make any changes to `self` and just updates `func`. Let me refactor this method to be a `staticmethod` instead.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class Aggregability(StrEnum):
+    """
+    Type of allowed aggregation for a given measure.
+    """
+
+    FULL = "full"
+    LIMITED = "limited"
+    NONE = "none"
```
anhqle, line 16: What's an example of "limited aggregability"? `COUNT DISTINCT` is categorized as "limited" below, but I can't think of a way to aggregate `COUNT DISTINCT`

anhqle (approved):
Great PR! Really enjoyable & elegant to read.

Initially, I was envisioning composing metric expression from user-input measures. This PR's approach of parsing arbitrary, user-input metric expression into measures is more general at the cost of greater complexity. Thanks to you I think the solution here wasn't that crazy complex.

Plus, the experimentation UI side may force users to construct metric expression out of constrained measures anyway, so the risk of us parsing the metric expression wrong is further reduced.

Thanks again!

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def extract_measures(self, metric_query: str) -> tuple[list[Measure], ast.Query]:
+        """
+        Decomposes the metric query into its constituent aggregatable measures and
+        constructs a SQL query derived from those measures.
+        """
+        query_ast = parse(metric_query)
+        measures = []
+
+        for idx, func in enumerate(query_ast.find_all(ast.Function)):
```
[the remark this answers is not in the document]
shangyian: Actually these are already just aggregate functions, because at the moment they're limited to the ones registered under `self.handlers`, but let me add a check anyway to future-proof this.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+class MeasureExtractor:
+    """
+    Extracts aggregatable measures from a metric definition and generates SQL
+    derived from those measures.
+    """
+
+    def __init__(self):
+        """Register handlers for aggregation functions"""
+        self.handlers = {
+            dj_functions.Sum: self._generic_additive_agg,
```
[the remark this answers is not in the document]
shangyian: Makes sense! Changed the name

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+                if func_measures:
+                    self.update_ast(func, func_measures)
+                measures.extend(func_measures)
+
+        return measures, query_ast
+
+    def _generic_additive_agg(self, func, idx) -> list[Measure]:
+        arg = func.args[0]
+        measure_name = "_".join(
+            [str(col) for col in arg.find_all(ast.Column)] + [func.name.name.lower()],
```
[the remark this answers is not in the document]
shangyian: In general no, it's only meant to support aggregation functions like `SUM(a)`, `MAX(b)` etc. I wouldn't expect them to take multiple arguments. Maybe a better name for this function is `_simple_associative_agg`.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+                aggregation=func.name.name.upper(),
+                rule=AggregationRule(
+                    type=Aggregability.FULL
+                    if func.quantifier != "DISTINCT"
+                    else Aggregability.LIMITED,
+                ),
+            ),
+        ]
+
+    def _avg(self, func, idx) -> list[Measure]:
```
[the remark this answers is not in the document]
shangyian: I don't think so. This should only be mapped to `AVG`

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    FULL = "full"
+    LIMITED = "limited"
+    NONE = "none"
+
+
+class AggregationRule(BaseModel):
+    """
+    Measures are components used to build metrics (e.g., sales_amount, revenue, user_count).
```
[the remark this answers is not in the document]
shangyian: Thanks, edited the docstring. Also added a separate docstring for the aggregation rule that explains how `LIMITED` and `level` are meant to work in conjunction.

datajunction-server/datajunction_server/sql/decompose.py
shangyian: Let's consider a metric definition like `COUNT(DISTINCT user_id)`. This can be decomposed into a single measure with expression `DISTINCT user_id` that can be aggregatable if the measure is calculated at least at the `user_id` level, but not if it's calculated at some other level (e.g., `country`).

Technically, if the aggregability type is `LIMITED`, the `level` should be specified to highlight the level at which the measure needs to be aggregated to in order to support the specified aggregation function. I haven't added this yet because I'm not sure how useful this is in practice.

So in the particular example of `COUNT(DISTINCT user_id)`, it can be decomposed into:
```
    - name: users_count_distinct
      expression: DISTINCT user_id
      aggregation: COUNT
      rule:
        type: LIMITED
        level: ["user_id"]
```

agorajek: Not saying this always makes sense, but I can see one doing some additional aggs (say AVG()) on top of count_distinct results over some particular dimensions. And if your data is MECE over some dimensions you could even do SUM() on top of it.

shangyian: Yeah, agreed, although it does also get super complicated to manage or indicate through metadata.

datajunction-server/datajunction_server/sql/decompose.py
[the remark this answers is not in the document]
anhqle: If we don't expect to handle multiple arguments, do we need/want `arg.find_all(ast.Column)`? Should we either find the first Column only, or raise when multiple are found?

datajunction-server/datajunction_server/sql/decompose.py
anhqle: > This can be decomposed into a single measure with expression DISTINCT user_id that can be aggregatable if the measure is calculated at least at the user_id level, but not if it's calculated at some other level (e.g., country).

OK, I see the case for "limited aggregability" now. 
But I think the example above is backward. Let's say we want to COUNT DISTINCT user_id globally. It _is_ possible to first `COUNT DISTINCT user_id GROUP BY country`, then `SUM(count_distinct_user_id)`. Essentially, GROUP BY any grain that is as or more coarse than `user_id` is allowed.

Is that right?

agorajek: > Essentially, GROUP BY any grain that is as or more coarse than user_id is allowed.

Only if the `grain` is `1:many` with the `user_id`, otherwise you may be double counting.

shangyian: @anhqle if you did `COUNT DISTINCT user_id GROUP BY country` first, then you would lose the ability to distinguish between overlapping user_ids across countries, and end up with double-counted results.

datajunction-server/datajunction_server/sql/decompose.py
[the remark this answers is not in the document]
shangyian: Actually, we're already only picking the first argument to the aggregation function - see line 74: `arg = func.args[0]`. However, we still need `arg.find_all(ast.Column)`, because that will find *all* referenced columns in the argument, which might not be just a single column.

For example, if you have `SUM(view_secs + clicks)`, there is a single argument to`SUM` which is `view_secs + clicks`, and `arg.find_all(ast.Column)` will return both of the columns referenced in that single argument: `[ast.Column("view_secs"), ast.Column("clicks")]`. We're then using the name of both columns to construct the measure name. 

Alternatively, we could name the measures something much less intuitive (e.g., just increase a counter and have `m0`, `m1`, `m2` etc), but I figured that actually naming them something sensible would be helpful to users parsing the metadata.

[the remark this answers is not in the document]
anhqle: Makes sense! Thank you for explaining!

datajunction-server/datajunction_server/sql/decompose.py
anhqle: Yeah, my unstated assumption is that user:country is many:one.

In any case, it's complicated to track that relationship as we discussed above. So in practice, "limited aggregability" only works when the measure is aggregated to _exactly_ `level`.

PR #1238, opened 2024-12-04
Only refresh SQLAlchemy ORM objects when needed

shangyian:
### Summary

This helps speed up the SQL building process by only refreshing SQLAlchemy ORM objects when the attribute is not already loaded. After the migration to async database requests, we had to explicitly load attributes when they were needed. However, we were overdoing this by consistently calling `session.refresh(...)` on attributes that had already been loaded, thereby reloading attributes unnecessarily and making many calls to the database. This change will reduce calls to the database during SQL building, making it faster.

### Test Plan

No additional tests, but existing ones should pass.

- [ ] `make check` passes


agorajek (approved):
Looks good. I see you are also piggybacking some caching for SQL...

shangyian:
@agorajek haha yes, I did a second round of profiling this morning and realized that we can further speed up the query building by just caching some of the antlr SQL parsing calls.

PR #1240, opened 2024-12-06
Add 'hidden' as a column attribute type

anhqle:
### Summary

Add a `hidden` attribute type for columns on **dimension** nodes that are not useful for end users. For example, timestamp is technically a dimension of a fact, but it's rare that end users would want to filter or group by a timestamp.

Client application can use this `hidden` attribute to skip surfacing certain columns.

### Test Plan

- [x] `make check` passes


shangyian (approved):
Nice, thank you!

PR #1241, opened 2024-12-06
Add derived measures to GraphQL 

shangyian:
### Summary

This change makes the derived measures metadata from #1234 available to query via GraphQL. For example:
```
query ListNodes($namespace: String, $nodeTypes: [NodeType!], $tags: [String!], $editedBy: String, $after: String, $limit: Int) {
  findNodesPaginated(
    namespace: $namespace
    nodeTypes: $nodeTypes
    tags: $tags
    editedBy: $editedBy
    limit: $limit
    after: $after
  ) {
    edges {
      node {
        name
        type
        current {
          extractedMeasures {
            measures {
              name
              expression
              aggregation
              rule {
                type
              }
            }
            derivedQuery
            derivedExpression
          }
        }
      }
    }
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPrevPage
    }
  }
}
```

### Test Plan

- [x] PR has an associated issue: #1223 
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/scalars/metricmetadata.py
+++ b/datajunction-server/datajunction_server/api/graphql/scalars/metricmetadata.py
@@ -5,8 +5,10 @@ from typing import Optional
 import strawberry
 
 from datajunction_server.models.node import MetricDirection as MetricDirection_
+from datajunction_server.sql.decompose import Aggregability as Aggregability_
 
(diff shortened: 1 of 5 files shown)


agorajek (approved):
We talked about having the `derivedQuery` and `derivedExpression` instead, but I assume we can change this later.

PR #1242, opened 2024-12-06
Generate measures SQL pre-aggregated to the requested dimensions

shangyian:
### Summary

This PR adds the option to generate `preaggregated` measures SQL for a set of metrics + dimensions. This means that DJ will generate SQL for the requested metrics, aggregated as much as possible to the requested dimensions, using the extracted measures from the requested metrics.

Exactly how aggregated the measures SQL is depends on the measures in question. For example, a measure with `SUM` as the aggregation can always be further aggregated upon, but a measure with `COUNT DISTINCT` as the aggregation cannot be. For the latter, we'll continue to fallback to the previously generated measures SQL, which was at the level of the metric's parent node.

The final metric aggregation(s) can then be performed on the output of this "intermediate" stage. This also sets us up nicely for derived metrics support in the future.

#### Example

Suppose that we have the following nodes (one transform, two metrics):
```
- name: default.repair_orders_fact
  type: transform
  query: | 
    SELECT
      repair_orders.repair_order_id,
      repair_orders.dispatcher_id,
      repair_order_details.price * repair_order_details.quantity AS total_repair_cost,
      repair_orders.dispatched_date - repair_orders.order_date AS time_to_dispatch
    FROM default.repair_orders repair_orders
    JOIN default.repair_order_details repair_order_details ON repair_orders.repair_order_id = repair_order_details.repair_order_id

- name: default.total_repair_cost
  type: metric
  query: SELECT sum(total_repair_cost) FROM default.repair_orders_fact

- name: default.avg_time_to_dispatch
  type: metric
  query: SELECT avg(cast(time_to_dispatch as int)) FROM default.repair_orders_fact
```

A user can request measures SQL for both metrics, pre-aggregated to the `repair_orders.dispatcher_id` dimension with:
```
GET /sql/measures/v2?metrics=default.total_repair_cost&metrics=default.avg_time_to_dispatch&dimensions=repair_orders.dispatcher_id
```

DJ will return SQL that selects all of the necessary measures for each of the metrics (for `total_repair_cost`, it will select the `SUM(total_repair_cost)` measure, for `avg_time_to_dispatch`, it will select these two measures: `SUM(cast(time_to_dispatch as int))`, `COUNT(1)`), grouped by the requested dimension `repair_orders.dispatcher_id`. 

### Test Plan

Locally

- [x] PR has an associated issue: #848   
- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/api/sql.py
+++ b/datajunction-server/datajunction_server/api/sql.py
@@ -91,7 +91,7 @@ async def get_measures_sql_for_cube_v2(
         get_measures_query,
     )
 
-    metrics = list(OrderedDict.fromkeys(set(metrics)))
+    metrics = list(OrderedDict.fromkeys(metrics))
     measures_query = await get_measures_query(
         session=session,
         metrics=metrics,
--- a/datajunction-server/datajunction_server/api/sql.py
+++ b/datajunction-server/datajunction_server/api/sql.py
@@ -51,7 +51,13 @@ async def get_measures_sql_for_cube_v2(
     dimensions: List[str] = Query([]),
     filters: List[str] = Query([]),
     orderby: List[str] = Query([]),
-    preagg: bool = False,
+    preaggregate: bool = Query(
+        False,
+        description=(
+            "Whether to pre-aggregate to the requested dimensions so that "
+            "subsequent queries are more efficient."
+        ),
+    ),
     *,
     include_all_columns: bool = Query(
         False,
@@ -99,7 +105,7 @@ async def get_measures_sql_for_cube_v2(
         include_all_columns=include_all_columns,
         sql_transpilation_library=settings.sql_transpilation_library,
         use_materialized=use_materialized,
-        preagg=preagg,
+        preaggregate=preaggregate,
     )
     return measures_query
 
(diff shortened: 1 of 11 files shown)


datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -154,7 +183,7 @@ def update_ast(func, measures: list[Measure]):
                         args=[ast.Column(ast.Name(measures[0].name))],
                     ),
                     right=ast.Function(
-                        ast.Name("COUNT"),
+                        ast.Name("SUM"),
```
agorajek, line 186: I think `COUNT` was correct... oh wait, no this is to aggregate the components of the expression, right? And to aggregate count's we would still use a sum.

agorajek (approved):
This looks good. Just a question in line.

datajunction-server/datajunction_server/api/sql.py
```diff
@@ -51,6 +51,13 @@ async def get_measures_sql_for_cube_v2(
     dimensions: List[str] = Query([]),
     filters: List[str] = Query([]),
     orderby: List[str] = Query([]),
+    preaggregate: bool = Query(
```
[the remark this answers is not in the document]
shangyian: It looks like this was the old way of declaring query parameters for FastAPI endpoints: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#alternative-old-query-as-the-default-value 

We can migrate this to the newer `Annotated` structure.

datajunction-server/datajunction_server/sql/decompose.py
shangyian: Yeah, I think in this case we actually have to use `SUM`, because we need to sum together all the counts for it to be correct.

PR #1243, opened 2024-12-06
Add is_hidden to DimensionAttributeOutput

anhqle:
### Test Plan

- [ ] `make check` passes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -80,7 +80,7 @@ repos:
 #  - id: reorder-python-imports
 #    args: [--application-directories=.:src]
 - repo: https://github.com/hadialqattan/pycln
-  rev: v2.1.7 # Possible releases: https://github.com/hadialqattan/pycln/tags
+  rev: v2.4.0 # Possible releases: https://github.com/hadialqattan/pycln/tags
   hooks:
   - id: pycln
     args: [--config=pyproject.toml]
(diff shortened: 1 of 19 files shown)


shangyian:
@anhqle This information should already be available on each column's `attributes` list. You should be able to parse it from the client-side. It looks like this (and is available via GraphQL too):
```
{"columns": [
   {
      "name":"some_dim",
      "display_name":"Some Dim",
      "type":"int",
      "attributes": [
         {
            "attribute_type": {
               "namespace": "system",
               "name": "hidden"
            }
         }
      ],
   },
   ...
]}
```

You can just

shangyian (commented):
@anhqle This information should already be available on each column's `attributes` list, and is accessible via GraphQL too. You can parse it from the client-side, or alternatively, add some python client helper functions for easier access there. It looks like this (and is available via GraphQL too):
```
{"columns": [
   {
      "name":"some_dim",
      "display_name":"Some Dim",
      "type":"int",
      "attributes": [
         {
            "attribute_type": {
               "namespace": "system",
               "name": "hidden"
            }
         }
      ],
   },
   ...
]}
```

anhqle:
The client application is using the `/metrics/common/dimensions/` endpoint to get all the dimensions info, so I'd like DJ to return `is_hidden` to be returned in that same call. Returning all of the attributes also works too.

Can we change the endpoint to do that, or you'd suggest making a separate call to get the `hidden` attribute?

shangyian:
@anhqle Ahh yes, that makes total sense! Sorry I misunderstood in my original comment.

shangyian:
@anhqle fyi, I pushed a fix to the unit tests + added a new one for this attribute type. Also changed the implementation slightly to use an enum for all system-managed column attributes.

Issue #1246, opened 2024-12-10
Support using dimension attributes from reference links on dimension nodes


shangyian:
We should support using dimension attributes that are sourced from reference links on dimension nodes. At the moment we only support single-hop reference links, rather than multi-hop ones (e.g., when they're from one dimension node to a second dimension node).

PR #1247, opened 2024-12-10
Support using dimension attributes from reference links between two dim nodes in SQL building

shangyian:
### Summary

This change adds support for using dimension attributes that come from reference links in SQL building. Specifically, these reference links are from one dimension node to another. We weren't discovering these in the process of finding available dimensions to build SQL with.

### Test Plan

- [ ] PR has an associated issue: #1246
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -1292,7 +1292,9 @@ async def dimension_join_path(
                 )
                 for link in col.dimension.current.dimension_links:
                     if (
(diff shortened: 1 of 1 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -498,11 +497,6 @@ async def validate_cube(  # pylint: disable=too-many-locals
             message=("Metrics and dimensions must be part of a common catalog"),
         )
 
-    await validate_shared_dimensions(
```
shangyian, line 501: This validation happens later anyway, so skipping it here. This makes the sql build process faster and avoids the issue of dimension attributes that come from second-order reference links being skipped.

PR #1249, opened 2024-12-10
YAML deploy needs to handle dimension link deletion for all links not…

shangyian:
… represented in YAML file

### Test Plan

- [x] PR has an associated issue: #1248 
- [x] `make check` passes

### DIFF
--- a/datajunction-clients/python/datajunction/nodes.py
+++ b/datajunction-clients/python/datajunction/nodes.py
@@ -65,6 +65,27 @@ class Namespace(ClientEntity):  # pylint: disable=protected-access
 
 
 @dataclass
+class NodeName(SerializableMixin):
+    """Node name"""
+
+    name: str
+
+
+@dataclass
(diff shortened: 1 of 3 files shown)


datajunction-clients/python/datajunction/nodes.py
```diff
@@ -420,13 +421,35 @@ def validate(self) -> str:
...
+    name: str
+
+
+@dataclass
+class DimensionLink(SerializableMixin):
+    """
+    Dimension join links
+    """
+
+    dimension: NodeName
```
agorajek, line 437: Why not simply a `dimension: str` ?

agorajek (approved):
Just one question in line.

datajunction-clients/python/datajunction/nodes.py
shangyian: Unfortunately the format of ORM-based dataclass serialization means that wouldn't work out of the box.

PR #1251, opened 2024-12-14
YAML Config Parity

shangyian:
### Summary

This change adds several additional YAML configuration features.

#### Generated Node Deploy Order

In the export process, we don't generate a topologically sorted node deployment order, which can cause issues if the nodes in a namespace have dependencies on each other.

However, for the notebook export, we already have a function to perform a topological sort on nodes, which gives us the right deployment order. We can apply the same sort here and then use the ordered list to auto-populate the `build.priority` node list.

#### Column-Level Settings

This adds support for configuring various column-level settings via YAML, including `attributes` and `display_name`. This fields can be optionally configured by adding the following section to a node definition config:
```
columns:
- name: is_filtered
  attributes:
  - dimension
  display_name: Is Filtered
- name: is_clicked
  attributes:
  - dimension
  display_name: 'Clicked?'
```

#### Export Parity for YAML

The YAML export did not support all of the features that were supported by the YAML deploy functionality. This adds parity for the following features:
* Tags
* Column attributes
* Column display name
* Primary keys on all non-dimension node types (previously we only supported the PK export on dimension nodes).
* The required dimensions, unit, and direction fields on metric nodes

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -26,6 +26,7 @@
 )
 from datajunction_server.models.node import NodeMinimumDetail
 from datajunction_server.models.node_type import NodeType
+from datajunction_server.sql.dag import topological_sort
```
agorajek, line 29: Oh, nice!

agorajek (approved):
Thanks for filling in the blanks. There was a lot missing!

PR #1252, opened 2024-12-14
Fix recursion error from deepcopy

shangyian:
### Summary

Using `cached_parse` for the node query would sometimes cause `RecursionError: maximum recursion depth exceeded` errors. This error comes from the deepcopy, which fails with a recursion error on compiled node queries. 

### Test Plan

Reproduced the error locally 

- [x] `make check` passes

### Deployment Plan

ASAP


agorajek:
So do we still get some cacheing here or is there none after this fix?

PR #1255, opened 2024-12-19
Make measures derived from metrics unique based on the expression

shangyian:
### Summary

The measures derived from metric expressions should be unique based on the measures' expressions. We can achieve this by generating a hash based on the measure's SQL expression, and appending it to the end of the measure's descriptive name. 

In the original implementation, we just looped through all possible measures on a metric expression and assigned a new index for each possible measure, but this sometimes results in collisions, where the generated measure name is the same as another, but the expression is actually different.

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -122,10 +129,13 @@ def _simple_associative_agg(self, func, idx) -> list[Measure]:
         measure_name = "_".join(
             [str(col) for col in arg.find_all(ast.Column)] + [func.name.name.lower()],
         )
+        expression = f"{func.quantifier} {arg}" if func.quantifier else str(arg)
```
agorajek, line 132: I know its not important now but maybe add an underscore in `expression = f"{func.quantifier}_{arg}"` for a good measure in case we use this label in some other way later on.

agorajek (approved):
Nice catch. I had a feeling the previous label was looking too good :)

datajunction-server/datajunction_server/sql/decompose.py
shangyian: Actually this expression is the SQL expression, so adding an underscore here would make it invalid. The `measure_name` is the string we use as a label, and that does already have an underscore.

Issue #1261, opened 2025-01-07
Better categorize the various exception types


shangyian:
We have a lot of `DJException` littered across the server, but debugging of issues would be easier if we categorize them more cleanly, especially between user errors and server errors.

PR #1262, opened 2025-01-07
Better categorize the various exceptions into user vs system errors

shangyian:
### Summary

This PR goes through most of the instances where we raise `DJException` and tries to categorize them into more useful exception categories. This makes it easier to distinguish between user errors and system errors for debugging or alerting purposes.

### Test Plan

- [x] PR has an associated issue: #1261 
- [x] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/api/metrics.py
```diff
@@ -121,7 +121,7 @@ async def get_common_dimensions(
     """
     Return common dimensions for a set of metrics.
     """
-    errors = []
+    user_input_errors = []
```
agorajek, line 124: Seems like `input_errors` would be enough, no?

agorajek (approved):
Nice cleanup! Two comments in-line.

PR #1263, opened 2025-01-08
GQL common dimensions query + dim attr output shape

shangyian:
### Summary

There are two core changes here:
1. Add the common dimensions query to GraphQL.
2. Change the dimension attribute output shape -- instead of having separate fields like `is_primary_key`, `is_hidden` for each of the different column-level attributes, we just have a single `properties` field that is a list of all of them (e.g., `properties: ["primary_key", "hidden"]`).

### Test Plan

- [ ] `make check` passes


anhqle:
Having a single field makes sense. 

1. Why didn't you keep using the name `attributes`, and opted for `properties` instead?
2. Are we in the process of gradually replacing all endpoints with GQL? Should the client app that current calls `/nodes/{node_name}/dimensions` migrate to the new GQL in this PR to fetch dimensions?

agorajek (approved):
All good, but wondering if the `properties` or `attributes` field won't raise too much expectations, since they seem to be all boolean. If we want to keep them boolean then perhaps `flags` or `bool_attributes` would be a better name.

shangyian:
@agorajek I've been thinking of them more as column-level tags, but `attributes` was the original name we settled on, but it now appears confusing hence why I renamed it... But I'm certainly open to switching it from `properties` to something else. 

@anhqle 
> Why didn't you keep using the name attributes, and opted for properties instead?

It's actually because `attribute` is already a field on the `DimensionAttributeOutput` object, and I thought using `attributes` would be confusing. 

> Are we in the process of gradually replacing all endpoints with GQL? Should the client app that current calls /nodes/{node_name}/dimensions migrate to the new GQL in this PR to fetch dimensions?

I like to think of the GQL endpoints as complementary to the REST ones. I'm adding these for consistency, but I don't think there's a need to migrate unless you want to take advantage of the flexibility that GQL provides. And by the client app, are you referring to the Python client? I would prefer that one stays REST, since it doesn't need the flexible querying of GQL.

PR #1264, opened 2025-01-10
Only generate preaggregated measures SQL where possible

shangyian:
### Summary

We should only generate preaggregated measures SQL if the metrics requested are compatible. For example, if the requested metric has measures with `Aggregability.LIMITED` or `Aggregability.NONE`, we should not generate preaggregated measures SQL.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Looks good. Are you planning to use the grain property right away?

shangyian:
@agorajek I wasn't planning on using grain straight away, but it'll be a useful piece of metadata to store. However, I think the more important part of this PR is the one that defaults back to the original transform-level dataset if it's unable to pre-aggregate based on the derived measures. If we don't handle that correctly, then generating this dataset could be completely wrong.

PR #1271, opened 2025-01-15
Seed initial data for running with docker compose

shangyian:
### Summary

Running DJ with the docker compose setup should seed the database with relevant data, including an initial user `dj`, along with the roads demo database. This is done by:
* Running `datajunction-server/init_data/seed.sql`, which inserts the initial `dj` user if one isn't already created
* Running the `Modeling the Roads Example Database.ipynb` notebook, which will create all roads demo DJ nodes. This notebook also needed some updating based on several API changes we've made.

### Test Plan

Tested locally

- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Thank you!

Issue #1275, opened 2025-01-17
Make sure there is no WHERE clause and a single FROM clause in Metric node argument.


agorajek:
And explain that we want metric nodes to use the IF() filters if needed and be a single column expressions w/o any WHERE clauses.

agorajek:
Solved with https://github.com/DataJunction/dj/pull/1276 by @shangyian

PR #1276, opened 2025-01-17
Add metric SQL validation

shangyian:
### Summary

This PR adds validation checks for metric SQL when creating a metric:
* Only a single output column
* No `WHERE` clause

### Test Plan

- [x] PR has an associated issue: #1275 
- [x] `make check` passes


agorajek (approved):
Thanks! I bet we will need to add a similar check for GROUP BY and LIMIT and maybe ORDER BY as well at some point.

shangyian:
@agorajek Good point! Let me update this PR to cover those cases too

PR #1278, opened 2025-01-19
Generate faster SQL by removing unnecessary validation

shangyian:
### Summary

We can generate faster measures SQL by removing expensive and unnecessary validation. For SQL generation, we use the same validation checks as for cube creation, but the same validation essentially gets run twice for generating SQL. Since we're already going to traverse the dimensions graph to bring each requested dimension into the generated SQL build, there's no need for a redundant validate shared dimensions prior to starting the build.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Nice one! So this is about not calling `validate_cube` in getting the measures/metric query, right?

shangyian:
@agorajek yep, it turns out that call is actually quite expensive!

PR #1280, opened 2025-01-21
SQL Generation Speed: Store pickled query AST

shangyian:
### Summary

This PR speeds up SQL generation quite significantly by storing and loading pickled query ASTs rather than recompiling a node query every time we need it. The primary changes include:

* A new `query_ast` column on the `noderevision` table to store compressed pickled ASTs.
* Every time a transform or dimension node is created, updated, or revalidated, we refresh the stored query AST.
* When we need to load a compiled query AST, instead of parsing and compiling, which can be expensive for large/complex queries, we unpickle the object from the database and use that instead.

Our queries are structured in a way where we can easily reuse these compiled query ASTs across many query builds. 

### Test Plan

Local testing + unit tests.

- [x] PR has an associated issue: #1286 
- [x] `make check` passes

### Deployment Plan

We should do some additional testing prior to release.

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -14,7 +14,7 @@ from typing import Dict, List, Optional, Set, Tuple, cast
 
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
-from sqlalchemy.orm import joinedload, selectinload
+from sqlalchemy.orm import defer, joinedload, selectinload
 from sqlalchemy.sql.operators import and_, is_
 
 from datajunction_server.construction.build import (
@@ -441,9 +441,7 @@ async def check_metrics_exist(session: AsyncSession, metrics: list[str]) -> list
             metrics,
             options=[
                 joinedload(Node.current).options(
-                    selectinload(NodeRevision.columns).options(
-                        selectinload(Column.node_revisions),
-                    ),
+                    selectinload(NodeRevision.columns),
                     joinedload(NodeRevision.catalog),
                     selectinload(NodeRevision.parents),
                 ),
@@ -480,9 +478,8 @@ async def check_dimension_attributes_exist(
             dimension_node_names,
             options=[
                 joinedload(Node.current).options(
-                    selectinload(NodeRevision.columns).options(
-                        joinedload(Column.node_revisions),
-                    ),
+                    selectinload(NodeRevision.columns),
+                    defer(NodeRevision.query_ast),
                 ),
             ],
         )
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -3,7 +3,6 @@
 import collections
 import logging
 import re
-import traceback
 from dataclasses import dataclass
 from functools import cached_property
 from typing import Any, DefaultDict, Dict, List, Optional, Tuple, Union, cast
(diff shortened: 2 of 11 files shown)


datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -480,9 +478,8 @@ async def check_dimension_attributes_exist(
             dimension_node_names,
             options=[
                 joinedload(Node.current).options(
-                    selectinload(NodeRevision.columns).options(
```
shangyian, line 483: These extra joins brought in are fairly expensive and not used in SQL buidling

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -217,6 +217,9 @@ async def get_measures_query(  # pylint: disable=too-many-locals
                 measure_columns.append(expr)
                 expr.set_semantic_type(SemanticType.MEASURE)  # type: ignore
         await parent_ast.compile(context)
+        dependencies, _ = await parent_ast.extract_dependencies(
```
shangyian, line 220: By extracting dependencies at this stage, we avoid calling another `query.compile()` call on the final query, which is expensive and without much benefit (all it does is add type inference).

datajunction-server/datajunction_server/sql/parsing/ast.py
```diff
@@ -2861,29 +2863,3 @@ async def extract_dependencies(
     @property
     def type(self) -> ColumnType:
         return self.select.type
-
-    async def build(  # pylint: disable=R0913,C0415
```
shangyian, line 2865: legacy query build that is now completely unused

datajunction-server/alembic/versions/2025_01_19_1808-9650f9b728a2_add_query_ast_for_noderevision.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = "9650f9b728a2"
+down_revision = "70904373eab3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("noderevision", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("query_ast", sa.PickleType(), nullable=True))
```
agorajek, line 25: Wow, SQL alchemy has an answer for any type these days...

datajunction-server/datajunction_server/construction/build_v2.py
agorajek: That one fallout of imprecise function names. Every time I see `build` or `compile` I need to go back to the source and remind myself what it means.

agorajek (approved):
Awesome code!

datajunction-server/alembic/versions/2025_01_19_1808-9650f9b728a2_add_query_ast_for_noderevision.py
shangyian: Yeah, this one is convenient!

datajunction-server/datajunction_server/construction/build_v2.py
shangyian: Agreed... `compile` is kind of like a "hydrate query with metadata" stage, and `build` actually changes the structure of the query. If you can come up with a better function name that represents the concept happy to rename compile! 😅

PR #1284, opened 2025-01-23
Endpoints and data model for saving user notification subscriptions

samredai:
### Summary

This adds a `POST /notifications/subscribe` and a `GET /notifications` endpoint that allows users to set a preference for notifications on any set of `ActivityType`s for an `EntityType`. Those are existing enums that we use for storing events to the history table. That's intentional since the idea is that any dependency injected notifier can simply match up these saved notification preferences to the event being passed through the notify dependency and act accordingly.

### Test Plan

Added some tests and tested locally in the docker setup (`docker compose up`).

- [ ] `make check` passes


datajunction-server/datajunction_server/api/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+from datajunction_server.database.notification_preference import NotificationPreference
+from datajunction_server.database.user import User
+from datajunction_server.internal.access.authentication.http import SecureAPIRouter
+from datajunction_server.utils import get_and_update_current_user, get_session
+
+router = SecureAPIRouter(tags=["notifications"])
+_logger = logging.getLogger(__name__)
+
+
+def get_notifier():
```
samredai, line 21: This was just moved over from notification.py which I renamed to notification**s**.py to be more consistent with the other router files.

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
```diff
@@ -0,0 +1,84 @@
...
+# revision identifiers, used by Alembic.
+revision = "0e428c1bb9bc"
+down_revision = "70904373eab3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "notification_preferences",
```
agorajek, line 23: nice! we need more tables with underscores in names :)

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
```diff
@@ -0,0 +1,84 @@
...
+        sa.Column(
+            "id",
+            sa.BigInteger().with_variant(sa.Integer(), "sqlite"),
```
agorajek, line 26: What does the "sqlite" variant mean?

datajunction-server/alembic/versions/2025_01_23_0229-e865e6d53e6c_indexes_and_unique_constraints_for_.py
```diff
@@ -0,0 +1,35 @@
...
+# revision identifiers, used by Alembic.
+revision = "e865e6d53e6c"
+down_revision = "0e428c1bb9bc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("notification_preferences", schema=None) as batch_op:
+        batch_op.create_index("ix_entity_name", ["entity_name"], unique=False)
```
agorajek, line 23: Unless I am missing some of your thinking here...

I don't think we need separate indexes on `entity_name` and `entity_type`, because:
1. the unique_constraint defined on these columns should come with an implicit index on both columns.
2. and because index on `entity_type` will likely be not very effective because of the low cardinality

I would also flip the order of the columns in the unique constraint to `["entity_name", "entity_type"]`, which will effectively serve as an index on the `entity_name` itself.

datajunction-server/datajunction_server/database/notification_preference.py
```diff
@@ -0,0 +1,63 @@
...
+
+class NotificationPreference(Base):  # pylint: disable=too-few-public-methods
+    """
+    User notification preferences for a specific entity and activity type.
+    """
+
+    __tablename__ = "notification_preferences"
+    __table_args__ = (
+        UniqueConstraint("entity_type", "entity_name", name="uix_entity_type_name"),
+        Index("ix_entity_name", "entity_name"),
```
agorajek, line 33: This may need to be adjusted if you make changes in alembic.

But the fact that this framework requires us to define database model kind of twice is painful ;)

datajunction-server/datajunction_server/api/notifications.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def notify(event: History):
+        """Send a notification for an event"""
+        _logger.debug("Sending notification for event %s", event)
+
+    return notify
+
+
+@router.post("/notifications/subscribe")
+async def subscribe_to_notifications(
```
agorajek, line 32: What about removing notifications? Maybe we should change this to `update_notifications` which would wipe them out if the args are empty or add a separate `delete_notifications` ?

agorajek (approved):
Thanks for adding this. Two main comments in-line:
- one about the indexes
- another about the deletion

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
samredai: You're lucky I didn't name it `user_notification_preferences` 😂

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
samredai: This is just telling alembic to use an Integer if sqlite is the backend. It automatically generated this file and I just left it because we left it in all of the previous alembic migration files.

datajunction-server/alembic/versions/2025_01_23_0229-e865e6d53e6c_indexes_and_unique_constraints_for_.py
samredai: This is great! Thanks Olek, this all makes sense to me. Let me update it.

datajunction-server/datajunction_server/database/notification_preference.py
samredai: Oh I didn't write anything in the alembic file. I just wrote this here and then when I generate the revision it inspects this sqlalchemy class and then generates the equivalent logic in the alembic file. I know, my head is spinning too, lol.

datajunction-server/datajunction_server/api/notifications.py
samredai: Oh right, how could I have forgotten that. It's like when you can subscribe to a server with 2 clicks on your phone but if you want to cancel they route through 43 different call centers and then make you mail them a handwritten letter.

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
```diff
@@ -0,0 +1,84 @@
...
+                "QUERY",
+                "TAG",
+                name="notification_entitytype",
+            ),
+            nullable=False,
+        ),
+        sa.Column("entity_name", sa.String(), nullable=False),
+        sa.Column(
+            "activity_types",
+            sa.ARRAY(
```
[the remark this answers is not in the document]
samredai: I used the enum in the sqlalchemy class but alembic pulled the values and expanded them like this. I tried to reason through why and I think it's because if we update the enum and add/remove values, we would actually need to generate another alembic migration to modify the enum in the database. Does that sound right or am I missing something?

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
agorajek: Actually I was serious. I think most of our tables are worddelimiterless.

datajunction-server/datajunction_server/database/notification_preference.py
agorajek: Oh, I keep forgetting this is the source. Nice!

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
```diff
@@ -0,0 +1,84 @@
...
+    op.create_table(
+        "notification_preferences",
+        sa.Column(
+            "id",
+            sa.BigInteger().with_variant(sa.Integer(), "sqlite"),
+            nullable=False,
+        ),
+        sa.Column(
+            "entity_type",
+            sa.Enum(
```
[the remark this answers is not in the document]
agorajek: +1

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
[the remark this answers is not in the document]
agorajek: Oh you are correct... I think this is the only way, cause these values have to be passed to the database.

datajunction-server/alembic/versions/2025_01_22_1638-0e428c1bb9bc_notification_preferences_table.py
shangyian: Haha, yeah, I think it started off that way due to the default table name generation pattern if you don't explicitly spell out the table name, and that has since calcified...

Issue #1287, opened 2025-01-23
Include alembic migrations in server releases


shangyian:
We should include the alembic migrations in datajunction-server releases so that they can be applied elsewhere.

PR #1288, opened 2025-01-23
Include alembic migrations as part of datajunction-server release

shangyian:
### Summary

This includes alembic migrations as a part of the `datajunction-server` so that we can automate the process of running database upgrades.

### Test Plan

- [x] PR has an associated issue: #1287 
- [x] `make check` passes


agorajek (approved):
Makes sense.

datajunction-server/datajunction_server/alembic/env.py
```diff
@@ -84,21 +88,21 @@ def run_migrations_online():
...
         )
 
-        with context.begin_transaction():
-            context.run_migrations()
+        with alembic.context.begin_transaction():
+            alembic.context.run_migrations()
 
 
-if context.is_offline_mode():
+if alembic.context.is_offline_mode():
```
samredai, line 105: How does this top level if statement work now that it's part of the library? Whenever this is imported wouldn't that mean the migration runs? Is that what we want? What I'm trying to think through is what happens when you're deploying this to a cluster and the containers are all trying to migrate on startup at the same time.

agorajek: Hmm... that's a good point.

shangyian: This is never imported though, as far as I can tell. It's just used for configuring alembic, but we don't use it anywhere in the actual server code.

Issue #1289, opened 2025-01-23
Add a node level custom json field.


agorajek:
To store any arbitrary user defined json value.

agorajek:
This was done in https://github.com/DataJunction/dj/pull/1291

PR #1291, opened 2025-01-24
Add custom metadata field to node revision.

agorajek:
### Summary

So we can use it for custom integrations.

### Test Plan

Added unit test to create and update the new field.

- [ ] PR has an associated issue: #1289 
- [ ] `make check` passes

### Deployment Plan

auto

### DIFF
--- a/datajunction-clients/python/datajunction/builder.py
+++ b/datajunction-clients/python/datajunction/builder.py
@@ -355,6 +355,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
         primary_key: Optional[List[str]] = None,
         tags: Optional[List[str]] = None,
         mode: Optional[models.NodeMode] = models.NodeMode.PUBLISHED,
+        custom_metadata: Optional[Dict] = None,
         update_if_exists: bool = False,
     ) -> "Transform":
         """
@@ -371,6 +372,7 @@ class DJBuilder(DJClient):  # pylint: disable=too-many-public-methods
                 "primary_key": primary_key,
                 "tags": tags,
                 "mode": mode,
+                "custom_metadata": custom_metadata,
             },
             update_if_exists=update_if_exists,
         )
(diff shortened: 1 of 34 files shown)


datajunction-server/alembic/versions/2025_01_24_0020-bec3296d7537_add_custom_metadata_field_to_nodes.py
```diff
@@ -0,0 +1,28 @@
+"""Add custom metadata field to nodes.
```
shangyian, line 1: Haha, we're gonna need a lot of rebasing depending on who merges their PR first ;)

shangyian (approved):

samredai (approved):
Feel free to merge this first, I need to regenerate the migrations in my PR anyway after I change the indexes.

datajunction-server/alembic/versions/2025_01_24_0020-bec3296d7537_add_custom_metadata_field_to_nodes.py
agorajek: You can go first. I forgot to add this to the YAML project gears.

agorajek:
> @agorajek just a general question. Why would this metadata be at the node revision level and not at the node level?

I was debating this question and I thought that having a version controlled field could be useful for integration, but the more I think about it the more it feels like its probably too much. I can change this to node level... @shangyian any opinion here?

shangyian:
Yeah, I think having this at the noderevision level makes sense, in case folks want to keep track of different versions of their node metadata.

agorajek:
My concern wasn't performance since I know Postgres can handle large fields in different ways. But even though I made this field version controlled I just couldn't come up with a good example of such use. I am not sure someone would ever care what was the previous value in here.

agorajek:
After some discussion I arrived at the following plan:
1. Add custom_metadata field to node revision.
2. Make a new  "minor" revision when the field is updated.

Adding this to the client and YAML project before I merge the  PR...

PR #1292, opened 2025-01-24
Update dependencies based on python >=3.10

shangyian:
### Summary

Updated and cleaned up a bunch of dependencies. Also restricted the server to be run with python 3.10+, rather than 3.8+, given that we only run tests for 3.10+.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
Makes sense. Thanks!

PR #1299, opened 2025-01-31
[fix] bug in how AVG is decomposed into measures

anhqle:
### Test Plan

I edited `decompose_test.py`

- [x] PR has an associated issue: #1300 1300
- [x] `make check` passes

Note: `make check` passed locally after making a bunch of small edits. I'm not sure why it's failing here, and why the changes need to be made to unrelated alembic files in the first place. Suggestions welcome!


shangyian (approved):
Thanks for fixing this 🎉

PR #1308, opened 2025-02-12
Fix Pydantic model definition for valid OpenAPI schema

shangyian:
### Summary

The extra comma in the pydantic field `description` meant that it was a tuple rather than a str, causing an OpenAPI schema validation problem. 

### Test Plan

- [x] `make check` passes


agorajek:
Yay, all these pylint comments going away.

PR #1309, opened 2025-02-12
Add additional database indexes for performance

shangyian:
### Summary

This migration adds missing indexes across several tables to improve query performance, particularly for recursive CTE queries and other common join patterns in the metadata database.

Specifically, indexes were added on foreign keys and frequently filtered columns across the following tables: `cube`, `dimensionlink`, `nodecolumns`, and `noderelationship`. 

#### Benchmark Results (100 runs each) on `/sql/measures`

| Configuration | Average Duration (ms) | Std Dev (ms) | 
| -- | -- | -- |
| Before (w/o indexes) | 3035.37 | 528.03 | 
| After (with indexes) | 2712.10 | 503.81 |

#### Benchmark Results (100 runs each) on `/metrics/common/dimensions`

| Configuration | Average Duration (ms) | Std Dev (ms) | 
| -- | -- | -- |
| Before (w/o indexes) | 4566.99 | 494.04 |
| After (with indexes) | 4415.48 | 479.55 |

A two-sample t-test confirms the difference is statistically significant for both endpoints.

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/database/attributetype.py
```diff
@@ -87,7 +87,10 @@ class ColumnAttribute(Base):
     """
 
     __tablename__ = "columnattribute"
-    __table_args__ = (UniqueConstraint("attribute_type_id", "column_id"),)
+    __table_args__ = (
+        UniqueConstraint("attribute_type_id", "column_id"),
+        Index("idx_columnattribute_column_id", "column_id"),
```
agorajek, line 92: I think that if the earlier UniqueConstraint had the column order switched then we wouldn't need this new one at all.
If it's easier to add then to change the other one then go for it.

datajunction-server/datajunction_server/database/availabilitystate.py
```diff
@@ -83,6 +83,7 @@ class NodeAvailabilityState(Base):
     """
 
     __tablename__ = "nodeavailabilitystate"
+    __table_args__ = (Index("idx_nodeavailabilitystate_node_id", "node_id"),)
```
agorajek, line 86: +1

datajunction-server/datajunction_server/database/measure.py
```diff
@@ -37,6 +37,7 @@ class Measure(Base):  # type: ignore
     columns: Mapped[List["Column"]] = relationship(
         back_populates="measure",
         lazy="joined",
+        order_by="Column.name",
```
agorajek, line 40: Are you sure we need this?

datajunction-server/datajunction_server/database/node.py
```diff
@@ -68,6 +68,10 @@ class NodeRelationship(Base):
     """
 
     __tablename__ = "noderelationship"
+    __table_args__ = (
```
agorajek, line 71: ++1

datajunction-server/datajunction_server/database/node.py
```diff
@@ -259,6 +255,23 @@ class Node(Base):
         foreign_keys="History.entity_name",
     )
 
+    __table_args__ = (
```
agorajek, line 258: Why all this?

datajunction-server/datajunction_server/database/partition.py
```diff
@@ -29,6 +29,9 @@ class Partition(Base):  # type: ignore
     """
 
     __tablename__ = "partition"
+    __table_args__ = (
+        Index("idx_partition_column_id", "column_id", postgresql_using="btree"),
```
agorajek, line 33: I think btree is the default, so probably no need to specify it everywhere.

datajunction-server/datajunction_server/database/attributetype.py
shangyian: I ended up removing this because I'm not sure it makes much of a difference

datajunction-server/datajunction_server/database/measure.py
shangyian: We don't, removed!

datajunction-server/datajunction_server/database/node.py
shangyian: I removed these changes

datajunction-server/datajunction_server/database/partition.py
shangyian: Agreed, also removed these indexes because I don't think they make a difference for the specific slow endpoints I'm looking at

shangyian:
@agorajek I updated this with a smaller subset of indexes, which do show a stat sig increase in endpoint speed (see the updated PR description for details).

PR #1310, opened 2025-02-13
run tests on main branch after merging PR

samredai:
### Summary

This makes a small change to the test workflow so that tests run after a PR is merged to the main branch. In general I think this is good practice since multiple PRs can be merged very close to each other without a rebase and pass on their own but fail when combined in the main branch. I noticed this because the badge in the README shows as "no status" because it only shows tests run on the main branch.

badge

### Test Plan

- [ ] `make check` passes


shangyian (approved):

agorajek:
Thanks. This is great!

PR #1315, opened 2025-02-18
Update java client

samredai:
### Summary

This updates the java client to add methods to: login, create nodes, link dimensions, and get SQL as well as updates the gradle wrapper version and the integration tests.

### Test Plan

Manually ran the client integration tests.

- [ ] `make check` passes

### DIFF
--- a/datajunction-clients/java/src/main/java/io/datajunction/models/requests/Catalog.java
+++ b/datajunction-clients/java/src/main/java/io/datajunction/models/requests/Catalog.java
@@ -7,7 +7,7 @@ import lombok.Data;
 public class Catalog {
     private String name;
 
-    public Catalog(@JsonProperty("name") String name) {
+    public Catalog(String name) {
         this.name = name;
     }
 }


agorajek (approved):
Thanks! Btw... how do you run these unit tests? Maybe we should also add the call to https://github.com/DataJunction/dj/blob/main/.github/workflows/test.yml#L15 .

samredai:
> Thanks! Btw... how do you run these unit tests? Maybe we should also add the call to https://github.com/DataJunction/dj/blob/main/.github/workflows/test.yml#L15 .

Hmm, that's a good idea. The only thing is it requires starting up the entire demo docker compose setup since these don't have any unit tests where we mock stuff for example. But the whole thing still runs in under 5 mins so that should still be ok, right? [Here's](https://github.com/DataJunction/dj/blob/main/.github/workflows/client-integration-tests.yml#L109-L113) how it's run with gradle in the client integration test workflow.

PR #1318, opened 2025-02-23
Attach availability to materialization

shangyian:
### Test Plan

- [ ] `make check` passes


samredai (approved):

agorajek:
This sounds like a very good way forward even though I realize there may be some fallout. Let's talk through this during our next OSS sync. We should make a list of items that will need to be addressed and plan for implementation.

PR #1322, opened 2025-02-25
Add better metric metadata to materializations

shangyian:
### Summary

This adds better metric metadata to the saved materialization config:
* Exposes the metric expression rather than the full query (saves downstream consumers from having to parse out the expression)
* Includes each metric's display name for each materialization

### Test Plan

- [x] PR has an associated issue: #1321 #1320
- [x] `make check` passes


agorajek (approved):
Nice. Thanks!

PR #1326, opened 2025-03-03
Decompose ANY_VALUE in metric definitions into measures

shangyian:
### Summary

This handles decomposition of metric definitions with `ANY_VALUE` as the aggregation function.

### Test Plan

- [x] `make check` passes


agorajek (approved):
Thanks!

PR #1331, opened 2025-03-07
Fix aggregation function check for metric definition

shangyian:
### Summary

We should check all children of the metric query for aggregation functions before raising an error.

### Test Plan

Locally

- [x] PR has an associated issue: #1330
- [ ] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
Looks good.

PR #1333, opened 2025-03-09
Process every history event through notifier depnedency after saving

samredai:
### Summary

Currently saving history events and notifying are two separate calls, but in all cases they should be one right after the other (even if the notification dependency chooses not to send out any notifications for that history event type). Instead of adding calls to the notifier everywhere we save history, this PR couples the two together within a dependency function called `save_history` that also passes the event through the notifier dependency to handle it after it's saved.

### Test Plan

`make test`

- [ ] `make check` passes


agorajek (approved):
This is it!

PR #1335, opened 2025-03-11
Normalize all column names to be lowercase

shangyian:
### Summary

All column names should be case insensitive. In many (most?) SQL varieties, column names are not case sensitive. This means that we should always normalize the stored column names in the metadata db to be lower case. 

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -1473,7 +1473,7 @@ async def column_lineage(
         col
         for col in query_ast.select.projection
         if (  # pragma: no cover
-            col != ast.Null() and col.alias_or_name.name == column_name  # type: ignore
+            col != ast.Null() and col.alias_or_name.name.lower() == column_name.lower()  # type: ignore
         )
     ][0]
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1224,7 +1224,9 @@ async def create_new_revision_from_existing(
         type=old_revision.type,
         columns=[
             Column(
-                name=column_data.name,
+                name=column_data.name.lower()
+                if node.type != NodeType.METRIC
+                else column_data.name,
```
samredai, line 1229: I'm curious why we wouldn't do it for metric nodes too. Is it just because we don't have to? Is there a case to be made for consistency?

samredai (approved):

datajunction-server/datajunction_server/internal/nodes.py
shangyian: We introduce the `DOT` separator ourselves for metrics, which already has the upper case so it wouldn't make sense to normalize for those as well.

PR #1340, opened 2025-03-14
Break out logic for getting a user's notification preferences to a separate function

samredai:
### Summary

This moves the logic in the endpoint for getting a user's notification preferences out to a separate function because it's useful to use it in other places. The main place top of mind is in injected notification dependencies that need to look up what the current user's preferences are.

### Test Plan

`make test`

- [ ] `make check` passes

### DIFF
--- a/docs/content/0.1.0/docs/deploying-dj/notifications.md
+++ b/docs/content/0.1.0/docs/deploying-dj/notifications.md
@@ -67,10 +67,9 @@ app.dependency_overrides[get_notifier] = get_custom_notifier
 
 ## Example Usage
 
-Here's how the `get_notifier` dependency is used within the application to send a notification:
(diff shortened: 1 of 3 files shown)


agorajek (approved):
This is great! Do you know if we have a doc page to clarify these `internal` hooks ?

docs/content/0.1.0/docs/deploying-dj/notifications.md
```diff
@@ -81,4 +80,33 @@ notify(event)
...
+- ENGINE
+- LINK
+- MATERIALIZATION
+- NAMESPACE
+- NODE
+- PARTITION
+- QUERY
+- TAG
+
+### Activity types
```
samredai, line 103: The docs still look up to date but I decided to list out the entity types and activity types at the end for convenience. I find myself looking these enums up in the code enough times that it would be useful to have on the docs site. :P

Issue #1244, opened 2024-12-06
Add column-level description


anhqle:
We currently have node description (equivalent to a table description in the ware house). Column level description would also be useful to explain, e.g. how a dimension should be used

PR #1343, opened 2025-03-17
Add column description

anhqle:
### Summary

Add optional description to column. This allows client application to adds more details about how a column should be used to end users.

```
curl -X PATCH 'http://localhost:8000/nodes/foo.bar/columns/one/description/?description=blahdescription3'
```

### The use of Claude Code

I wanted to test [Claude Code](https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview) and used this feature as a test case. The first commit is entirely written by Claude, with the following prompt:

```
DJ has a concept called node. Right now there’s node-level description, but I want column-level description as well. (Each node can have multiple columns). Work on both the front end and back end changes
```

*Positive*: This turns out to be a best case scenario for Claude as there is already a "set column display name" feature that's highly similar. So the use of Claude Code is very labor saving in this case, as it was able to identify all the relevant places that need a change (without me having to read through the codebase). The only major of hallucination is the Alembic migration file. 

*Negative*: 
1. Surprisingly, asking Claude Code to update the unit tests to include the new "description" field was hard. It keeps `grepping` and then intermittently edits the files. I end up manually edit the tests.
2. The UI code is a little buggy, but testing the UI reveals the issue quickly. I'll admit that I can't judge if the UI code is decent or not.

### Test Plan

- [x] PR has an associated issue: #1244 
- [x] `make check` passes

### Deployment Plan

N/A


agorajek:
Looks good, but I think you need to remove the Draft mode to have all tests executed.

datajunction-server/datajunction_server/database/column.py
```diff
@@ -39,6 +39,7 @@ class Column(Base):  # type: ignore
         insert_default=lambda context: labelize(context.current_parameters.get("name")),
     )
     type: Mapped[Optional[ColumnType]] = mapped_column(ColumnTypeDecorator)
+    description: Mapped[Optional[str]] = mapped_column()
```
shangyian, line 42: Should this be `mapped_column(String, default=None)` instead?

anhqle: I think nullability will be derived from the `Optional` type?

"In the absence of both of these parameters, the presence of typing.Optional[] within the [Mapped](https://docs.sqlalchemy.org/en/20/orm/internals.html#sqlalchemy.orm.Mapped) type annotation will be used to determine nullability, where typing.Optional[] means NULL, and the absence of typing.Optional[] means NOT NULL."

https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html

shangyian: Ah true, this is not strictly required

datajunction-server/datajunction_server/api/nodes.py
```diff
@@ -1645,6 +1645,48 @@ async def set_column_display_name(
...
+    """
+    node = await Node.get_by_name(
+        session,
+        node_name,
+        options=[joinedload(Node.current)],
+    )
+    column = await get_column(session, node.current, column_name)  # type: ignore
+    column.description = description
+    session.add(column)
+    await save_history(
```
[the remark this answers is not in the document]
anhqle: Makes sense. I'll switch the order for this method and for the `set_column_display_name` one as well

anhqle:
The UI code is failing lint (e.g. missing unique key prop), but that is happening on `main` too, unrelated to this PR I believe.

shangyian (approved):
LGTM, thanks for addressing some of the other issues.

By the way, the UI build is failing due to a lack of test coverage (unrelated to linters):
```
Jest: "global" coverage threshold for statements (87%) not met: 86.86%
```

agorajek (approved):
We should fill in the UI unit tests to 100% at some point.

PR #1344, opened 2025-03-17
Enable metrics to reference available dimensions in definition

shangyian:
### Summary

This PR enables metrics to reference available dimensions in the metric's SQL definition. This enables flexibility to model the dimensions graph in the right way -- e.g., leveraging the dimension link structure to support optionally joining in dimensions as needed, rather than directly building it into a transform node. 

For example, it should be possible to write metric definitions like the following, assuming that `default.repair_orders_fact` is linked to `default.municipality_dim` in its dimensions graph:
```sql
SELECT
  COUNT(DISTINCT default.municipality_dim.contact_name)
FROM default.repair_orders_fact
```

The above setup means that you don't have to write the join between `default.repair_orders_fact` and `default.municipality_dim` directly into the query for `default.repair_orders_fact`. DJ will perform the join when the metric is requested, and will not do so when other metrics from this transform are requested.

### Test Plan

Tested locally and wrote a series of unit tests.

- [ ] PR has an associated issue:
- [ ] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -159,12 +159,13 @@ async def get_measures_query(
     matcher = re.compile(column_name_regex)
 
     # Find any dimensions referenced in the metric definitions and add to requested dimensions
-    for parent_node, children in common_parents.items():  # type: ignore
-        for metric in children:
-            metric_ast = parse(metric.query)
-            for ref in metric_ast.find_all(ast.Column):
-                if "." in ref.identifier():
-                    dimensions.append(ref.identifier())
+    dimensions.extend(
+        ref.identifier()
+        for metrics in common_parents.values()
+        for metric in metrics
+        for ref in parse(metric.query).find_all(ast.Column)
+        if SEPARATOR in ref.identifier().rsplit(SEPARATOR, 1)[0]
+    )
 
     dimensions_without_roles = [matcher.findall(dim)[0][0] for dim in dimensions]
 
@@ -1108,7 +1109,7 @@ class CubeQueryBuilder:
         for metric in self.metric_nodes:
             metric_ast = parse(metric.current.query)
             for ref in metric_ast.find_all(ast.Column):
-                if "." in ref.identifier():
+                if SEPARATOR in ref.identifier():
                     self.add_dimension(ref.identifier())
 
         measures_queries = await self.build_measures_queries()
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -159,15 +159,16 @@ async def get_measures_query(
     matcher = re.compile(column_name_regex)
 
     # Find any dimensions referenced in the metric definitions and add to requested dimensions
-    dimensions.extend(
+    dimensions_in_metric_defs = set(
         ref.identifier()
         for metric in metric_nodes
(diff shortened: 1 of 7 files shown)


agorajek:
What happens (or how the join happens) when your metric definition references a dimension that has no join path to any of the metric upstream nodes?

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -154,6 +154,19 @@ async def get_measures_query(
...
+
+    # Find any dimensions referenced in the metric definitions and add to requested dimensions
+    dimensions.extend(
+        ref.identifier()
+        for metric in metric_nodes
+        for ref in parse(metric.current.query).find_all(ast.Column)
+        if (
+            (parts := ref.identifier().rsplit(SEPARATOR, 1))
+            and len(parts) > 1
+            and SEPARATOR in parts[0]
```
anhqle, line 166: Does this mean the metric expression has to be explicit about naming that an identifier is a dimension, e.g. `namespace.dim.day` in
`COUNT_DISTINCT(IF(measure > 0, namespace.dim.day_number, NULL))`?

I can see the motivation for this. It makes it efficient to know exactly which dimension node needs to be brought in (instead of DJ having to trawl through the dimension graph to find where a generically named `day_number` dimension is located).

The downside is that the metric expression is now coupled with the implementation details of how the dimension is sourced. Changing the dimension means having to edit the metric expression. And analytics / experimentation metrics can't share the same metric node (if the concept of `day_nbr` differs between the two use case).

I think I'm ultimately landing on agreeing with the current implementation. Separating out the analytics vs experimentation metrics into different node is probably a good thing for clarity.

shangyian: In the case of the exact metric we're talking about -- the metric only makes sense in an experimentation context, because the relative days concept doesn't exist with an experiment. But every other metric that doesn't need that context can still be useful in both analytics and experimentation I think.

shangyian:
@agorajek the setup here won't raise an error at metric creation, but will raise an error when trying to generate SQL for the metric. We can also raise an error at metric creation if that's cleaner.

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -372,6 +387,19 @@ def build_preaggregate_query(
...
 
+def get_dimensions_referenced_in_metrics(metric_nodes: list[Node]) -> List[str]:
+    """
+    Returns a list of dimensions referenced in the metric nodes' query definitions.
+    """
+    dimensions = set()
+    for metric in metric_nodes:
+        metric_ast = parse(metric.current.query)
+        for ref in metric_ast.find_all(ast.Column):
+            if SEPARATOR in ref.identifier().rsplit(SEPARATOR, 1)[0]:
```
agorajek, line 398: Is this the best way to identify a dimension column?

agorajek (approved):
This will introduce a lot of flexibility in defining metrics. But also probably a bit slower SQL generation. Something to keep in mind. Thanks @shangyian for pushing this.

I feel like we should come up with a good marketing content for this feature. I realized that the concept of "required dimensions" could now be even more confusing for unaware users. Perhaps the following could work:
- "(metric) required dimensions" : dimension(s) required to be displayed for context when presenting values of a given metric.
- "(metric) referenced dimensions" :  dimensions referenced in the metric definition, not required for context, but required for metric computation.

PR #1345, opened 2025-03-19
Fix SQL column metadata

shangyian:
### Summary

When building SQL and then providing column metadata for the built SQL, there were cases where we weren't assigning the node name correctly to the semantic entity for a given column. 

### Test Plan

- [x] `make check` passes

### Deployment Plan

N/A


agorajek (approved):
I assume no unit tests exists to prove it?

shangyian:
@agorajek I'll see if I can add some. I think this is specifically for cases where there are uppercase/lowercase column outputs.

PR #1353, opened 2025-04-04
Get all users subscribed to a specific notification

samredai:
### Summary

This adds the reverse of an existing feature, that is, it returns the users who are subscribed to a specific notification.

### Test Plan

Added a test for this.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/notifications.py
+++ b/datajunction-server/datajunction_server/api/notifications.py
@@ -14,7 +14,10 @@ from datajunction_server.database.notification_preference import NotificationPre
 from datajunction_server.database.user import User
 from datajunction_server.errors import DJDoesNotExistException
 from datajunction_server.internal.access.authentication.http import SecureAPIRouter
-from datajunction_server.internal.notifications import get_notification_preferences
+from datajunction_server.internal.notifications import (
+    get_entity_notification_preferences,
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/api/notifications.py
```diff
@@ -121,3 +124,19 @@ async def get_user_notification_preferences(
...
+    session: AsyncSession = Depends(get_session),
+) -> JSONResponse:
+    """Get users for the given notification preference"""
+    notification_preferences = await get_entity_notification_preferences(
+        session=session,
+        entity_name=entity_name,
+        entity_type=entity_type,
+    )
+    users = [perf.user.username for perf in notification_preferences]
+    return JSONResponse(content=users)
```
shangyian, line 142: Can these have typed responses rather than just `JSONResponse`? So in this case, something like:
```
@router.get("/notifications/users")
async def get_users_for_notification(
    entity_name: str,
    entity_type: EntityType,
    session: AsyncSession = Depends(get_session),
) -> list[str]:
     ...
```

shangyian (approved):

datajunction-server/datajunction_server/api/notifications.py
agorajek: +1

PR #1356, opened 2025-04-06
Add formatting fields for metric display precision

shangyian:
### Summary

This PR introduces three optional metadata fields to support numeric formatting for metrics:

* `significant_digits`: Specifies the number of significant digits to display, if set.
* `min_decimal_exponent`: Values with exponents below this threshold will be displayed in scientific notation.
* `max_decimal_exponent`: Values with exponents above this threshold will be displayed in scientific notation.

These fields provide more granular control over how metric values are presented in UIs, especially for values that are very small or very large.

### Test Plan

Locally

- [x] PR has an associated issue: #1355
- [x] `make check` passes


datajunction-server/datajunction_server/models/node.py
```diff
@@ -473,8 +476,11 @@ class MetricMetadataInput(BaseModel):
     Metric metadata output
     """
 
-    direction: Optional[MetricDirection]
-    unit: Optional[str]
+    direction: MetricDirection | None
+    unit: str | None
```
agorajek, line 480: `str` or `Unit` ?

agorajek (approved):
Good idea.

datajunction-server/datajunction_server/models/node.py
shangyian: In this case it has to be `str` because pydantic won't interpret a complex enum like `Unit` correctly

PR #1363, opened 2025-04-11
Handle measures SQL generation where for measures columns with reserved characters in the name

shangyian:
### Summary

In cases where we have a node with column names that contain special sequences like `_DOT_`, we weren't including these columns in the generated measures SQL, even when they were actually referenced in the metric definition. This is because the logic for including the referenced columns was not taking into account that there could be column names that could contain `_DOT_`. 
 
### Test Plan

- [ ] `make check` passes


agorajek (approved):
Thanks!

PR #1366, opened 2025-04-16
Cache metrics list endpoint

samredai:
### Summary

This uses the cache dependency to cache the metrics list endpoint. It isn't necessarily an endpoint that needs optimization but this is a low stakes endpoint that I'm using as the first endpoint that uses caching.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/metrics.py
+++ b/datajunction-server/datajunction_server/api/metrics.py
@@ -5,7 +5,7 @@ Metric related APIs.
 from http import HTTPStatus
 from typing import List, Optional
 
-from fastapi import Depends, HTTPException, Query
+from fastapi import BackgroundTasks, Depends, HTTPException, Query
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import selectinload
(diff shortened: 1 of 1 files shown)


datajunction-server/datajunction_server/api/metrics.py
```diff
@@ -69,17 +71,23 @@ async def list_metrics(
...
     List all available metrics.
     """
-    return await list_nodes(
-        node_type=NodeType.METRIC,
-        prefix=prefix,
-        session=session,
-        current_user=current_user,
-        validate_access=validate_access,
-    )
+    metrics = cache.get("metrics")
```
shangyian, line 80: Can we set this up so that the `cache` here is optional? And thus if it's configured, we can check against it, but if it's not configured, we'll just fallback to the regular behavior.

shangyian: Ah never mind, I see that this would fallback to a `NoOpCache`. LGTM then!

shangyian (approved):

datajunction-server/datajunction_server/api/metrics.py
```diff
@@ -69,17 +71,23 @@ async def list_metrics(
     validate_access: access.ValidateAccessFn = Depends(
         validate_access,
     ),
+    cache: Cache = Depends(get_cache),
+    background_tasks: BackgroundTasks,
```
shangyian, line 75: Can this also include a `cache_control` header so that the client can control whether they want the cached value or not? E.g.:
```
Cache-Control: no-cache
Cache-Control: max-age=60
```

shangyian: That makes sense, thanks!

samredai:
Sorry for letting this PR linger. I need to find some time to find out why running this new test alone passes but running it with the rest of the tests makes this same single test fail. I suspect it's because the tests are running in parallel and so the cache-miss scenario is not happening. Should be a quick fix 🤞

PR #1379, opened 2025-05-14
Change enum on history events to str

samredai:
### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/alembic/versions/2025_05_14_1545-cb7368a4bfae_change_enum_columns_to_string_type.py
+++ b/datajunction-server/datajunction_server/alembic/versions/2025_05_14_1545-cb7368a4bfae_change_enum_columns_to_string_type.py
@@ -1,0 +1,42 @@
+"""Change enum columns to string type
+
+Revision ID: cb7368a4bfae
+Revises: 51547dcccb10
+Create Date: 2025-05-14 15:45:43.734154+00:00
+
+"""
+# pylint: disable=no-member, invalid-name, missing-function-docstring, unused-import, no-name-in-module
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = 'cb7368a4bfae'
+down_revision = '51547dcccb10'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/alembic/versions/2025_05_14_1545-cb7368a4bfae_change_enum_columns_to_string_type.py
```diff
@@ -0,0 +1,102 @@
...
+down_revision = "51547dcccb10"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("history", schema=None) as batch_op:
+        batch_op.alter_column(
+            "entity_type",
+            existing_type=postgresql.ENUM(
```
shangyian, line 25: Should this also drop the enum types from Postgres? Maybe with something like:
```
batch_op.execute("DROP TYPE IF EXISTS entitytype")
batch_op.execute("DROP TYPE IF EXISTS activitytype")
```

datajunction-server/datajunction_server/database/history.py
```diff
@@ -35,14 +33,14 @@ class History(Base):
         BigInteger().with_variant(Integer, "sqlite"),
         primary_key=True,
     )
-    entity_type: Mapped[Optional[EntityType]] = mapped_column(
-        Enum(EntityType),
+    entity_type: Mapped[Optional[str]] = mapped_column(
+        String(20),
```
agorajek, line 37: Why only 20?

agorajek (commented):
@samredai this looks good, but I think you should try running this locally and post the outcome on the PR. Just left one comment inline.

datajunction-server/datajunction_server/database/history.py
samredai: No reason, I just thought that the postgres default of infinite length seemed unnecessary for what will just be a string representation of an enum value. Should I make this larger? Or maybe don't specify the length at all?

samredai:
> @samredai this looks good, but I think you should try running this locally and post the outcome on the PR. Just left one comment inline.

Good idea, let me populate it with enum values locally then run a migration and see what happens to the data.

agorajek:
@samredai is this still on your radar? That seems like a good idea.

PR #1381, opened 2025-05-15
Fix build_preaggregate_query bug.

agorajek:
### Summary

Main: fix a bug in query generation for metrics referencing columns that need to go to the group by clause, example `COUNT(DISTINCT col1)`.

Piggyback: add Spark dialect to APPROX_COUNT_DISTINCT.

### Test Plan

With some manual runs, but ideally we should add a unit test.

- [ ] `make check` passes

### Deployment Plan
Auto

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -297,9 +297,13 @@ def build_preaggregate_query(
     # Find all required GROUP BY columns based on each of the measure's aggregation rules.
     # If the measure supports full aggregation, there are no required group-by columns, but if it
     # supports limited aggregation, we need to aggregate to the specified level.
+    colum_name_to_alias = {
+        col.name.identifier(): (col.alias or col.name).identifier()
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -297,9 +297,12 @@ def build_preaggregate_query(
     # Find all required GROUP BY columns based on each of the measure's aggregation rules.
     # If the measure supports full aggregation, there are no required group-by columns, but if it
     # supports limited aggregation, we need to aggregate to the specified level.
+    dim_colum_name_to_alias = {
+        str(col.name): str(col.alias) or str(col.name) for col in dimensional_columns
```
shangyian, line 301: I think you can replace this with `col.identifier()` or `col.name.identifier()`. So maybe something like:
```
        col.name.identifier(): (col.alias or col.name).identifier() for col in dimensional_columns
```

shangyian (approved):

PR #1382, opened 2025-05-16
[WIP] Add query parameters to DJ...

agorajek:
### Summary

Shall we...? 

Reason: some queries run much better if you can plug in a value at run-time anywhere in the query.

### Test Plan

- [ ] `make check` passes


shangyian:
Definitely! We can chat about this more next week, but so far I added:
* A `QueryParameter` class to the AST that stores the parameter name and quote style
* Logic that does the query parameter replacement when building SQL. I kept this simple for now and made it so that when an equality filter corresponds to a query parameter, we replace the query param with the filter value. This means that the query params will have types that come from the filter.
* We can add an option to pass in free-form query parameters, but we'll want to think through how that works (we can assume String type for simplicity, but it's possible the entire SQL endpoint may need to change from `GET` to `POST`).

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -589,6 +608,20 @@ async def build(self) -> ast.Query:
...
+        # Replace any parameters in the final AST with their values
+        for param in self.final_ast.find_all(ast.QueryParameter):  # type: ignore
+            if param.name in self.parameters and param.parent:
+                param.parent.replace(param, self.parameters[param.name])
+            else:
+                raise DJQueryBuildException(
+                    errors=[
+                        DJQueryBuildError(
+                            code=ErrorCode.MISSING_PARAMETER,
+                            message=f"Missing parameter: {param.name}",
```
agorajek, line 620: `s/Missing parameter/Missing value for parameter`

shangyian: good idea, done

shangyian:
@agorajek separately, it looks like the build is failing due to a linting issue. I fixed that here: https://github.com/DataJunction/dj/pull/1383

datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -889,6 +888,24 @@ async def find_dimension_node_joins(
                     dimension_node_joins[dim_node].requested_dimensions.append(dim)
         return dimension_node_joins
 
+    @classmethod
+    def normalize_query_param_value(cls, param: str, value: ast.Value | Any):
```
agorajek, line 892: I am not sure we need this... let's chat.

datajunction-server/datajunction_server/api/helpers.py
```diff
@@ -201,6 +201,8 @@ async def get_query(
     engine: Optional[Engine] = None,
     access_control: Optional[access.AccessControlStore] = None,
     use_materialized: bool = True,
+    query_parameters: Optional[Dict[str, str]] = None,
+    ignore_errors: bool = True,
```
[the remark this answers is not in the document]
shangyian: This is just mirroring the default behavior -- in some cases the errors are more like warnings about some type being incompatible.

shangyian (approved):

PR #1385, opened 2025-05-21
Speed up the recursive dimensions graph CTE query

shangyian:
### Summary

The existing dimensions graph query is a recursive CTE that is increasingly slow on even fairly small graphs. I had a closer look at the recursive CTE query we're using, and it looks like it's exhaustively tracing the graph in a way that finds every possible path to a dimension node, rather than just picking up the shortest paths.

This PR changes things so that the dimensions graph function does the following:
1. Discovers all dimension nodes in the graph via a recursive CTE (this is the same structure as the node upstreams or downstreams query and is reasonably fast).
2. Uses each dimension node's columns to augment with reference links.
3. Augments with any local dimensions from the original node.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/sql/dag.py
+++ b/datajunction-server/datajunction_server/sql/dag.py
@@ -227,6 +227,39 @@ async def get_upstream_nodes(
     ]
 
 
+async def build_reference_link(
+    session: AsyncSession,
+    col: Column,
+    path: list[str],
+    role: list[str] | None = None,
+) -> DimensionAttributeOutput | None:
+    """
+    Builds a reference link dimension attribute output for a column.
+    """
+    if not (col.dimension_id and col.dimension_column):
+        return None
+
+    await session.refresh(col.dimension, ["current"])
+    await session.refresh(col.dimension.current, ["columns"])
+
+    dim_cols = col.dimension.current.columns
+    if dim_col := next(
+        (dc for dc in dim_cols if dc.name == col.dimension_column),
+        None,
+    ):
+        return DimensionAttributeOutput(
+            name=f"{col.dimension.name}.{col.dimension_column}" + f"[{'->'.join(role)}]"
+            if role
+            else "",
+            node_name=col.dimension.name,
+            node_display_name=col.dimension.current.display_name,
+            properties=dim_col.attribute_names(),
+            type=str(col.type),
+            path=path,
+        )
+    return None
+
+
 async def get_dimension_attributes(
     session: AsyncSession,
     node_name: str,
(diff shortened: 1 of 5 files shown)


datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -226,6 +227,202 @@ async def get_upstream_nodes(
...
+            name=f"{col.dimension.name}.{col.dimension_column}" + f"[{'->'.join(role)}]"
+            if role
+            else "",
+            node_name=col.dimension.name,
+            node_display_name=col.dimension.current.display_name,
+            properties=dim_col.attribute_names(),
+            type=str(col.type),
+            path=path,
+        )
+    return None
```
agorajek, line 259: Not needed in Python.

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -226,6 +227,202 @@ async def get_upstream_nodes(
...
+        )
+        .cte("dimensions", recursive=True)
+        .suffix_with(
+            "CYCLE node_id SET is_cycle USING path",
+        )
+    )
+
+    paths = dag.union_all(
+        select(
+            dag.c.node_revision_id,
```
agorajek, line 390: What is `.c` ?

agorajek (commented):
I see you just added `local_dimensions`. This is super cool. I am going to test it locally as well and then respond again.

datajunction-server/datajunction_server/sql/dag.py
shangyian: Yeah, I just had an annoying linter flag this function for not having returns in every case 😱

datajunction-server/datajunction_server/sql/dag.py
shangyian: It's how you reference a column on a SQLAlchemy query or table object

agorajek: Wow. What a shortcut... Show how much I know ORMs :)

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -226,6 +227,200 @@ async def get_upstream_nodes(
...
+    local_dimensions = [
+        DimensionAttributeOutput(
+            name=f"{node.name}.{col.name}",
+            node_name=node.name,
+            node_display_name=node.current.display_name,
+            properties=col.attribute_names(),
+            type=str(col.type),
+            path=[],
+        )
+        for col in node.current.columns
```
agorajek, line 341: Should we skip some numerical columns here, or just let the user hurt themselves?

datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -226,6 +227,200 @@ async def get_upstream_nodes(
...
+    ]
+    local_dimensions = [
```
agorajek, line 343: Oh never mind, here is the filtering I asked about above. We could probably merge these two statements, cause there may be some unused work in the first one. But up to you.

datajunction-server/tests/api/dimension_links_test.py
```diff
@@ -326,13 +326,19 @@ async def test_link_complex_dimension_without_role(
...
-        ("default.users.residence_country", ["default.events"]),
-        ("default.users.snapshot_date", ["default.events"]),
-        ("default.users.user_id", ["default.events"]),
-    ]
+    assert sorted(
+        [(attr["name"], attr["path"]) for attr in response.json()],
+        key=lambda x: x[0],
+    ) == sorted(
+        [
+            ("default.users.account_type", ["default.events", "default.users"]),
```
agorajek, line 334: I am trying to figure out why the `default.users` item was added here. Any ideas?

agorajek (approved):
This is so much better. I think the stages of building the dimension list are much clearer now.

datajunction-server/datajunction_server/sql/dag.py
shangyian: What do you mean by numerical columns?

PR #1386, opened 2025-05-21
Fix `show_deleted` query param for /materializations, rename it `include_all`

samredai:
### Summary

When calling `/nodes/{node_name}/materializations?show_deleted=true`, it wasn't returning all of the materializations for a given node because it only looks at the current version (via `node.current`). This fixes that by pulling all materializations for all node revisions, only when that query param is enabled, otherwise the behavior remains the same.

I also renamed it to `include_all` since previous versions of the node could have materializations that aren't deleted, but still can only be accessed via this flag, so viewing deleted materializations is just a subset of why this would be used. I don't think this query param is used anywhere so changing this shouldn't break anything, but let me know if I missed somewhere it is being used.

### Test Plan

Added `test_include_all_materialization_configs`.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/materializations.py
+++ b/datajunction-server/datajunction_server/api/materializations.py
@@ -314,7 +314,7 @@ async def list_node_materializations(
                             request_headers=request_headers,
                         ),
                     )
-    else:  # Get just the active materializations on the current node revision
+    else:  # Get just the materializations on the current node revision
         for materialization in node.current.materializations:  # type: ignore
             if not materialization.deactivated_at or show_inactive:  # pragma: no cover
                 materializations.append(
--- a/datajunction-server/datajunction_server/api/materializations.py
+++ b/datajunction-server/datajunction_server/api/materializations.py
@@ -305,7 +305,7 @@ async def list_node_materializations(
     if include_all_revisions:  # Get all materializations for all node revisions
         for node_revision in node.revisions:  # type: ignore
             for materialization in node_revision.materializations:
-                if materialization.deactivated_at or show_inactive:
+                if not materialization.deactivated_at or show_inactive:
                     materializations.append(
                         get_materialization_info(
                             query_service_client=query_service_client,
--- a/datajunction-server/datajunction_server/api/materializations.py
+++ b/datajunction-server/datajunction_server/api/materializations.py
@@ -305,7 +305,9 @@ async def list_node_materializations(
     if include_all_revisions:  # Get all materializations for all node revisions
         for node_revision in node.revisions:  # type: ignore
             for materialization in node_revision.materializations:
-                if not materialization.deactivated_at or show_inactive:
+                if (
+                    not materialization.deactivated_at or show_inactive
+                ):  # pragma: no cover
                     materializations.append(
                         get_materialization_info(
                             query_service_client=query_service_client,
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/materializations.py
```diff
@@ -288,27 +288,48 @@ async def list_node_materializations(
...
+                    get_materialization_info(
+                        query_service_client=query_service_client,
+                        node_revision=node_revision,
+                        materialization=materialization,
+                        request_headers=request_headers,
+                    ),
+                )
+    else:  # Get just the active materializations on the current node revision
+        for materialization in node.current.materializations:  # type: ignore
+            if not materialization.deactivated_at:  # pragma: no cover
```
agorajek, line 314: I think this code now would not give you the option of seeing just the current materialization regardless of their status. I wonder if it would be cleaner to have two params on this whole method:
- `show_deleted` as before (although I think `show_inactive` sounds better)
- `include_all_revisions` as a param to scan previos node states.

I would lean to longer params/names that are obvious, rather than shorter once that look nice.

samredai: I see what you're saying. So `show_deleted` would basically ignore the `deactivated_at` value and `include_all_revisions` would iterate over all of the node revisions to pull the materialization. Let me update this.

samredai: Updated 👍

agorajek: Yes, i think this now supports all the query options. Thanks!

agorajek (approved):
Just some unit tests to fix...

datajunction-server/datajunction_server/api/materializations.py
```diff
@@ -286,29 +286,47 @@ async def list_node_materializations(
...
-                **info.dict(),
-            )
-            materializations.append(materialization)
+    if include_all_revisions:  # Get all materializations for all node revisions
+        for node_revision in node.revisions:  # type: ignore
+            for materialization in node_revision.materializations:
+                if (
+                    not materialization.deactivated_at or show_inactive
+                ):  # pragma: no cover
+                    materializations.append(
```
shangyian, line 311: Can this part be pulled into a separate function that operates on `node_revision`, whether it's current or an older revision?

datajunction-server/datajunction_server/api/materializations.py
```diff
@@ -286,29 +286,47 @@ async def list_node_materializations(
...
-                **materialization_config_output.dict(),
-                **info.dict(),
-            )
-            materializations.append(materialization)
+    if include_all_revisions:  # Get all materializations for all node revisions
+        for node_revision in node.revisions:  # type: ignore
+            for materialization in node_revision.materializations:
+                if (
+                    not materialization.deactivated_at or show_inactive
+                ):  # pragma: no cover
```
shangyian, line 310: I think there is coverage for this line right?

shangyian (approved):

datajunction-server/datajunction_server/api/materializations.py
samredai: Good idea, done!

datajunction-server/datajunction_server/api/materializations.py
samredai: You're right, I'm not sure why I put this here. 🤔

Issue #869, opened 2023-11-26
Proposal for cube-centric consumption endpoints


samredai:
The client experiences have naturally become split, having a client type (call it the "reader" persona) for those wanting to pull metadata on nodes or pull sql/data and those who want to contribute towards building the DJ graph  (call it the "builder" persona). We've largely thought of reader functionality as a subset of builder functionality, in other words, readers need to view a set of things and builders need to view/edit/create those things.

![image](https://github.com/DataJunction/dj/assets/43911210/ec6dcbce-d5a7-4317-b3df-51fb50e4f1b4)

This is largely correct but I think the CRUDish feeling of the "reader" functionality is not inviting to a user who isn't interested in the DJ internals but just wants to know what data is available for consumption and how can they get it. This is largely what the cube construct is meant to support and I would roughly say that data is available for consumption if:
- A cube is defined that brings a set of compatible metrics and dimensions together
- That cube can return data with very low latency (it's either materialized to a performant datastore or generates a query on the underlying source tables that's already very performant)
- The data is fresh with the acceptable threshold being controlled by the `max-stale` value in the cache control headers

This proposal is to extend the reader functionality by extending the client and adding some facilitating endpoints to the backend API that extends the cube router to provide a rich "cube only" set of features.
![image](https://github.com/DataJunction/dj/assets/43911210/35b1511a-cdc4-4f92-b0de-95de370634c2)

A few examples would be:
- `GET /cubes/` - List available cubes with all sorts of options to filter on cube attributes like freshness, materialization types, or even providing a set of metrics/dimensions to only see cubes that contain them as a subset
- `GET /cubes/{cube_name}/data/` - A cube focused data endpoint that honors all [cache control headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
- `POST /cubes/utilization/` - A callback endpoint to report utilization of a cube. This is a flexible endpoint that viz integrations can optionally call to self-report usage of a cube
- ...many more

samredai:
@agorajek using the client interface was probably a bit of a distraction but the core of what I'm proposing is that we extend the api by adding a number of `/cubes/{functionality}/` endpoints. 

We do this today for metrics and dimensions, i.e. the `/metrics/common/dimensions/` and `/dimensions/common/metrics/` endpoints so I didn't expect it to be too controversial to do something similar for cubes. I just wanted to get everyone's thoughts.

agorajek:
[...]

PR #1391, opened 2025-05-27
Add /cubes endpoint to get metadata on all cubes

samredai:
### Summary

This adds a `/cubes` endpoint that retrieves the latest revision for each cube along with all of the metadata that would be returned by `/cubes/{name}`. The `/cubes/{name}` endpoint is unchanged aside from a little refactoring to support the other endpoint.

I've also added an optional `available_in_catalog` that gets only the cubes that are available in a specific catalog.

### Test Plan

Added a test

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/cubes.py
+++ b/datajunction-server/datajunction_server/api/cubes.py
@@ -55,7 +55,7 @@ router = SecureAPIRouter(tags=["cubes"])
 async def get_all_cubes(
     *,
     session: AsyncSession = Depends(get_session),
-    available_in_catalog: Optional[str] = Query(
+    catalog: Optional[str] = Query(
         None,
         description="Filter to include only cubes available in a specific catalog",
     ),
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/api/cubes.py
```diff
@@ -48,6 +51,24 @@
 router = SecureAPIRouter(tags=["cubes"])
 
 
+@router.get("/cubes", name="Get all Cubes")
+async def get_all_cubes(
+    *,
+    session: AsyncSession = Depends(get_session),
+    catalog: Optional[str] = Query(
```
shangyian, line 58: Might be fine for now, but would be nicer if we can add pagination in here too, since with cube revision metadata for all cubes, this payload could become quite large.

shangyian (approved):

datajunction-server/datajunction_server/api/cubes.py
samredai: Good call! I'll make a follow-up PR and I added issue #1392 to not lose track.

samredai:
#869

PR #1396, opened 2025-06-03
Add option to delete materializations by node version.

agorajek:
### Summary

What the title says plus:
- add version field to the History table,
- added `deactivated_at` to the output schema of the Materialization 

TODO: (in another PR) add a page/tab of all node Materializations by node version and an ability to review/delete them.

### Test Plan

Manually 
- [ ] `make check` passes

### Deployment Plan

Auto.


datajunction-server/datajunction_server/api/materializations.py
```diff
@@ -316,51 +316,96 @@ async def list_node_materializations(
...
-    node = await Node.get_by_name(session, node_name)
+
+    # find the node revision to deactivate the materialization for
+    node_revision = None
+    if node_version:
+        node = await Node.get_by_name(
+            session,
+            node_name,
+            options=[
+                joinedload(Node.revisions).options(
```
shangyian: Can this part just ask the database to query for the single revision `node_version` that it needs? This setup with `joinedload(Node.revisions)` will end up loading all revisions for the node, which could be a lot.

Maybe something like:
```
stmt = (
    select(NodeRevision)
    .options(*NodeRevision.default_load_options())
    .where(NodeRevision.name == node_name, NodeRevision.version == node_version)
)
result = await session.execute(stmt)
node_revision = result.scalars().first()

if not node_revision:
    raise DJDoesNotExistException(...)
```

agorajek: Thanks! I'll try that.

shangyian (approved):
LGTM! Thanks for adding

PR #1397, opened 2025-06-04
Column metadata should include origin for columns

shangyian:
### Summary

At the moment, the `ColumnMetadata` provided by the measures SQL query includes several fields: `node`, `column`, and `semantic_entity`. At the moment,  `node` and `column` are just split-up versions of `semantic_entity`, which is not helpful for downstream consumers, since it's providing the exact same metadata that can be derived.

Instead, it would be more helpful if `node` and `column` actually represented the origin column of the parent node(s), where possible. This PR makes that switch.
 
### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/tests/api/sql_v2_test.py
+++ b/datajunction-server/tests/api/sql_v2_test.py
@@ -89,7 +89,7 @@ async def fix_dimension_links(module__client_with_roads: AsyncClient):
                 {
                     "column": "dispatcher_id",
                     "name": "default_DOT_dispatcher_DOT_dispatcher_id",
-                    "node": "default.dispatcher",
+                    "node": "default.repair_orders_fact",
                     "semantic_entity": "default.dispatcher.dispatcher_id",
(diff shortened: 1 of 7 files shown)


datajunction-server/tests/api/sql_v2_test.py
```diff
@@ -89,7 +89,7 @@ async def fix_dimension_links(module__client_with_roads: AsyncClient):
                 {
                     "column": "dispatcher_id",
                     "name": "default_DOT_dispatcher_DOT_dispatcher_id",
-                    "node": "default.dispatcher",
+                    "node": "default.repair_orders_fact",
```
shangyian, line 92: The reason this is being renamed is that the origin of `dispatcher_id` is actually `default.repair_orders_fact.dispatcher_id`. The `semantic_entity` reflects something separate.

shangyian:
@agorajek I did run into issues... will explain separately since it's from a downstream consumer of this metadata.

PR #1399, opened 2025-06-05
Enable cubes to be created with no dimensions

shangyian:
### Summary

This relaxes the restrictions on cubes so that cubes can be created with with no dimension attributes. There is no reason why dimensions have to be included.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -419,7 +419,9 @@ async def validate_cube(
             dimensions.append(columns[column_name_without_role])
 
     if require_dimensions and not dimensions:
-        raise DJInvalidInputException(message="At least one dimension is required")
+        raise DJInvalidInputException(  # pragma: no cover
+            message="At least one dimension is required",
+        )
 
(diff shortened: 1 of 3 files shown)


agorajek:
> @agorajek yeah, that's what I meant -- it's a single row of metric values. Conversely, I'm actually not sure how a cube without metrics would work.

OK got it.

A cube w/o metrics is a table with dimension values only. It can tell you the unique counts of for all dimensions included, and you can even slice and dice these implicit metrics.

agorajek (approved):

shangyian:
@agorajek that's interesting -- do you think we should enable such a dimensions-only cube as well? So you can either have a cube with no metrics or no dimensions (but not nothing at all, obviously).

agorajek:
> @agorajek that's interesting -- do you think we should enable such a dimensions-only cube as well? So you can either have a cube with no metrics or no dimensions (but not nothing at all, obviously).

Yes, I like this idea. Especially when using data access tools that can show implicit metrics about the dimensions of a given dataset.

PR #1401, opened 2025-06-13
Add custom_metadata and downstream nodes to GraphQL api.

agorajek:
### Summary

Two things in this PR:
- added `custom_metadata` field to GraphQL node output,
- added new query `downstream_nodes` to GraphQL api.

### Test Plan

Unit tests.

- [ ] `make check` passes

### Deployment Plan

Auto.

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/main.py
+++ b/datajunction-server/datajunction_server/api/graphql/main.py
@@ -8,7 +8,10 @@ from fastapi import Request
 from strawberry.fastapi import GraphQLRouter
 from strawberry.types import Info
 from datajunction_server.api.graphql.queries.catalogs import list_catalogs
-from datajunction_server.api.graphql.queries.dag import common_dimensions
+from datajunction_server.api.graphql.queries.dag import (
+    common_dimensions,
+    downstream_nodes,
+)
 from datajunction_server.api.graphql.queries.engines import list_engines, list_dialects
 from datajunction_server.api.graphql.queries.nodes import (
     find_nodes,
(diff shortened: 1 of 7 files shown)


datajunction-server/datajunction_server/api/graphql/main.py
```diff
@@ -110,6 +113,10 @@ class Query:
         resolver=log_resolver(common_dimensions),
         description="Get common dimensions for one or more nodes",
     )
+    downstream_nodes: list[Node] = strawberry.field(
```
shangyian, line 116: What do you think about potentially putting this as a `children` (or `downstreams`) attribute on the `Node` entity? I ask this because then we don't need a separate GraphQL query to get this info, and then you can always explore a `Node`'s downstreams.

agorajek: I don't know GraphQL well enough to have a strong opinion. But my thinking was that... this query (to find all downstream nodes) is expensive so I would prefer if people made it consciously, rather than add an attribute to `findNodes` and potentially affect the system with slow queries and maybe even have this data pulled out and not be used.

shangyian: Hmm, I think if we're worried about the query being expensive, putting it in a separate query doesn’t really protect us? I wonder if we can instead add some further protections like an argument to `downstreams` to control the depth fetched, so that even if someone does unintentionally fetch it on a `Node`, they will only fetch the most immediate layer of downstreams:
```
downstreams(depth: int = 1)
```

agorajek: Yeah, that makes sense. But let's add it later, if that's ok with you?

shangyian: Yeah, that works for me!

shangyian (approved):

PR #1403, opened 2025-06-16
Manage GraphQL DB session via middleware

shangyian:
### Summary

There is an issue in the current GraphQL database session management, where we currently use `Depends(get_session)` in `get_context()` to provide a database session for each request. However, this does not work well because Strawberry's `get_context()` is meant to be a regular function that returns a dict, with no support for teardown/generator-based context. That means that we get errors like the following when servicing GraphQL requests:
```
Can't reconnect until invalid transaction is rolled back.  Please rollback() fully before proceeding
```

This change adds a `GraphQLSessionMiddleware` to handle database session creation, commit/rollback, and cleanup for each GraphQL request. This middleware only applies to /graphql endpoint, so REST dependency behavior is kept unchanged. It ensures consistent db session management without manual handling in each resolver.

### Test Plan

Added unit tests

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/main.py
+++ b/datajunction-server/datajunction_server/api/graphql/main.py
@@ -66,6 +66,9 @@ def log_resolver(func):
 
 
 async def get_context(request: Request):
+    """
+    Provides the context for graphql requests
+    """
     return {
         "session": request.state.db,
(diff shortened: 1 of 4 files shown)


agorajek (approved):
Makes sense.

PR #1404, opened 2025-06-16
Fix hashing on measure decomposition.

agorajek:
### Summary

Because we only used the metric expression to create a measure hash we saw conflicts when the same looking metric (say `COUNT(*)`) was defined on different tables. This fix adds the FROM clause to the hashing so we still get to reuse the measure on the same parent nodes, but we get a different one if they come from a different node.

### Test Plan

Unit tests.

- [ ] `make check` passes

### Deployment Plan

Auto.

### DIFF
--- a/datajunction-server/tests/sql/decompose_test.py
+++ b/datajunction-server/tests/sql/decompose_test.py
@@ -24,7 +24,7 @@ def test_simple_sum():
     measures, derived_sql = extractor.extract()
     expected_measures = [
         MetricComponent(
-            name="sales_amount_sum_a1b27bc7",
+            name="sales_amount_sum_b5a3cefe",
             expression="sales_amount",
             aggregation="SUM",
             rule=AggregationRule(type=Aggregability.FULL),
@@ -32,7 +32,7 @@ def test_simple_sum():
     ]
     assert measures == expected_measures
     assert str(derived_sql) == str(
-        parse("SELECT SUM(sales_amount_sum_a1b27bc7) FROM parent_node"),
+        parse("SELECT SUM(sales_amount_sum_b5a3cefe) FROM parent_node"),
     )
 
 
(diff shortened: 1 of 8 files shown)


agorajek:
Sorry, found a bug in my code....

datajunction-server/tests/sql/decompose_test.py
```diff
@@ -265,13 +279,13 @@ def test_rate():
...
         MetricComponent(
-            name="clicks_sum_c9e9e0fc",
+            name="clicks_sum_6b583421",
             expression="clicks",
             aggregation="SUM",
             rule=AggregationRule(type=Aggregability.FULL),
         ),
         MetricComponent(
-            name="impressions_sum_87e980e6",
+            name="impressions_sum_6b583421",
```
shangyian: Should the `MetricComponent` name also change to include part of the `FROM` clause? That might help with disambiguating one count from another when debugging (I know that the hash is the true disambiguation factor, but it's much harder to read compared to a label).

shangyian (approved):
Would it be possible to also add a unit test for decomposition that uses two different `FROM` tables?

datajunction-server/tests/sql/decompose_test.py
agorajek: Hmm. I am not sure if I understand your request. I missed a lot of unit test adjustments with my first commit. Added just now...

agorajek:
This PR should be ready to go now.

datajunction-server/tests/sql/decompose_test.py
agorajek: Aha. I see what you mean. I feel like this name could be getting long, but maybe we could include the from as an attribute of the MetricComponent ?

shangyian: Yeah, that makes sense! I imagine that just adding something like `origin` onto MetricComponent is enough? But that can be a follow-up PR

PR #1405, opened 2025-06-18
Minor adjustments around Roads Example notebook.

agorajek:
### Summary

Plus some client code improvements and one config param fix wrt max db connections, because I was hitting an issue when running the notebook - after we added the GraphQL middleware.

### Test Plan

unit tests + executing the notebook successfully in `docker compose up --profile demo`

- [ ] `make check` passes

### Deployment Plan

auto


shangyian:
FYI I think the database connections issue (which is indeed from the GraphQL middleware) gets fixed by https://github.com/DataJunction/dj/pull/1430

datajunction-clients/python/datajunction/admin.py
```diff
@@ -53,26 +58,36 @@ def add_engine(
...
-            },
-            timeout=self._timeout,
-        )
-        json_response = response.json()
+        params = {
+            "name": f"{name}",
+            "version": f"{version}",
+        }
+        if uri:  # pragma: no cover
+            params["uri"] = f"{uri}"
```
shangyian, line 81: minor: this can just be `uri`

datajunction-clients/python/datajunction/admin.py
```diff
@@ -53,26 +58,36 @@ def add_engine(
...
+        if dialect:  # pragma: no cover
+            params["dialect"] = f"{dialect}"
```
shangyian, line 83: same here, it can just be `dialect`

datajunction-clients/python/datajunction/admin.py
```diff
@@ -21,19 +21,24 @@ def get_catalog(self, name: str) -> dict:
...
-        )
-        json_response = response.json()
+        try:
+            response = self._session.post(
+                "/catalogs/",
+                json={"name": f"{name}"},
+                timeout=self._timeout,
+            )
+        except DJClientException as exc:  # pragma: no cover
+            if skip_if_exists and "already exists" in str(exc):
```
shangyian, line 39: Should this have some message saying that it already exists so that users are informed?

datajunction-clients/python/datajunction/admin.py
```diff
@@ -53,26 +58,36 @@ def add_engine(
...
+            params["dialect"] = f"{dialect}"
+
+        try:
+            response = self._session.post(
+                "/engines/",
+                json=params,
+                timeout=self._timeout,
+            )
+        except DJClientException as exc:  # pragma: no cover
+            if skip_if_exists and "already exists" in str(exc):
```
shangyian, line 92: Similar to above: should this have a message for the user?

datajunction-clients/python/datajunction/builder.py
```diff
@@ -460,7 +474,7 @@ def create_metric(  # pylint: disable=too-many-arguments
                 "mode": mode,
             },
             update_if_exists=update_if_exists,
-        )
+        )  # type: ignore
```
shangyian, line 477: I don't think the `# type: ignore` is doing anything on this line, so it can be removed

datajunction-clients/python/datajunction/builder.py
```diff
@@ -494,7 +508,7 @@ def create_cube(  # pylint: disable=too-many-arguments
                 "tags": tags,
             },
             update_if_exists=update_if_exists,
-        )
+        )  # type: ignore
```
shangyian, line 511: Similar here, can be removed

shangyian (approved):
LGTM! Just added a few minor comments

datajunction-clients/python/datajunction/admin.py
agorajek: yes, but only if I change the default value to empty string otherwise we get:
```
>>> a = None
>>> f"{a}"
'None'
```

datajunction-clients/python/datajunction/admin.py
agorajek: ditto

datajunction-clients/python/datajunction/builder.py
agorajek: actually it is needed here for the return type "mismatch"

datajunction-clients/python/datajunction/builder.py
agorajek: Here it is needed because Cube is not matching the `create_node` function return type of any node.

PR #1415, opened 2025-06-19
Fix test action.

agorajek:
### Summary

Our PR test action should only run when there is a change to test. The filters and matrix should take care of this mechanism but it stopped working... trying to get that working again.

Please don't ask me why but `dorny/paths-filter@v3` is very hard to tame. It was very hard to make it do what I want. but I think it is working now.  

### Test Plan

Lotsa manual tests and made sure only the affected components are tested.

- [ ] `make check` passes

### Deployment Plan

Auto.


shangyian (approved):
Thanks for fixing this!

PR #1417, opened 2025-06-20
Add reader database cluster support

shangyian:
### Summary

This PR adds support for using a reader database cluster (for the metadata database) for read-only operations, along with the default writer database cluster for all other operations. It improves scalability by offloading read traffic from the writer database, and it sets up support for read/write-aware session management across both REST and GraphQL interfaces. The reader cluster configs are optional, so users who don't have a read replica setup can just set the writer cluster.

* For REST, we automatically detect when there is a `GET` request and route to the reader session.
* For GraphQL, we automatically detect when a request is not a mutation and route to the reader session (at the moment this is all GraphQL requests since we don't support mutations).

#### Settings

The settings for configuring a metadata database are now done via the nested `writer_db` and `reader_db` fields, each with a `DatabaseConfig` model to support detailed connection settings like `pool_size`, `max_overflow`, `connect_timeout`, etc.

`.env` files should be updated to include `READER_DB__URI` and other nested settings if reader DB is used:
```
# Writer DB (required)
WRITER_DB__URI=postgresql+psycopg://dj:dj@postgres_metadata:5432/dj
WRITER_DB__POOL_SIZE=20
WRITER_DB__MAX_OVERFLOW=20
WRITER_DB__POOL_TIMEOUT=10
WRITER_DB__CONNECT_TIMEOUT=5
WRITER_DB__POOL_PRE_PING=true
WRITER_DB__ECHO=false
WRITER_DB__KEEPALIVES=1
WRITER_DB__KEEPALIVES_IDLE=30
WRITER_DB__KEEPALIVES_INTERVAL=10
WRITER_DB__KEEPALIVES_COUNT=5

# Reader DB (optional)
READER_DB__URI=postgresql+psycopg://dj:dj@postgres_metadata:5432/dj
READER_DB__POOL_SIZE=10
READER_DB__MAX_OVERFLOW=10
READER_DB__POOL_TIMEOUT=5
READER_DB__CONNECT_TIMEOUT=5
READER_DB__POOL_PRE_PING=true
READER_DB__ECHO=false
READER_DB__KEEPALIVES=1
READER_DB__KEEPALIVES_IDLE=30
READER_DB__KEEPALIVES_INTERVAL=10
READER_DB__KEEPALIVES_COUNT=5
```

### Test Plan

Locally. Modified unit tests.

- [ ] `make check` passes

### Deployment Plan

N/A


datajunction-server/datajunction_server/api/graphql/middleware.py
```diff
@@ -1,6 +1,36 @@
...
+        payload = json.loads(body_bytes)
+        query = payload.get("query")
+        if not query:
+            return False
+        document = parse(query)
+        return any(
+            getattr(defn, "operation", None) == OperationType.MUTATION
+            for defn in document.definitions
+        )
+    except (GraphQLError, json.JSONDecodeError):
```
agorajek: Which of the above lines could trigger a `GraphQLError` ?

datajunction-server/datajunction_server/api/graphql/middleware.py
```diff
@@ -11,7 +41,20 @@ async def dispatch(self, request: Request, call_next):
...
-            session = get_session_manager().session()
+            body_bytes = await request.body()
+
+            # Restore request body for downstream usage
+            async def receive():
+                return {"type": "http.request", "body": body_bytes}
+
+            request._receive = receive  # type: ignore
+
+            # Set up the database session based on whether it's a mutation or not
```
agorajek, line 56: very cool

datajunction-server/datajunction_server/config.py
```diff
@@ -11,26 +11,53 @@
...
 
 
+class DatabaseConfig(BaseModel):
+    """
+    Metadata database configuration.
+    """
+
+    uri: str
+    pool_size: int = 20
+    max_overflow: int = 20
```
agorajek: I think we may still want to do `-1` here for better sleep, no?

agorajek (approved):
This is great. I think it will make our life better.

datajunction-server/datajunction_server/api/graphql/middleware.py
shangyian: If the user inputs an invalid GraphQL query, then calling `parse(query)` could result in a `GraphQLError`.

datajunction-server/datajunction_server/config.py
shangyian: I changed this default to 100

PR #1419, opened 2025-06-20
Skip deactivated nodes in namespace export.

agorajek:
### Summary

Because we should.

### Test Plan

Unit tests and some manual calls.

- [ ] `make check` passes

### Deployment Plan

Auto.


shangyian (approved):
Thanks for fixing!

PR #1420, opened 2025-06-21
Add better tests for read replica + use in docker compose

shangyian:
### Summary

This change modifies the tests to run against a server setup with both a read-write db and a read-only replica, to provide more confidence that the read-only cluster setup is working. It also adds a read-only user to the docker-compose setup to better simulate the read-only behavior of a read replica.

A few other changes were needed to get the read replica working:
* On `GET` endpoints, we don't create or update the user, and just use the user object as-is
* Removed the query request database storage for `GET` calls on various `/sql` endpoints -- we should use the actual cache setup for this. 
* Moved the creation of column attributes to app startup

### Test Plan

- [x] PR has an associated issue: follow-up to #1417 
- [x] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/main.py
+++ b/datajunction-server/datajunction_server/api/main.py
@@ -59,7 +59,7 @@ dependencies = [Depends(default_attribute_types), Depends(default_catalog)]
 
 
 @asynccontextmanager
-async def lifespan(app: FastAPI):
+async def lifespan(app: FastAPI):  # pragma: no cover
     """
     Lifespan context for initializing and tearing down app-wide resources, like the FastAPI cache
     """
--- a/datajunction-server/datajunction_server/api/main.py
+++ b/datajunction-server/datajunction_server/api/main.py
@@ -3,6 +3,8 @@ Main DJ server app.
 """
 
 import logging
+
+from fastapi.concurrency import asynccontextmanager
 from datajunction_server.api import setup_logging  # noqa
 
 from http import HTTPStatus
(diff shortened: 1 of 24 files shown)


datajunction-server/datajunction_server/api/main.py
```diff
@@ -63,7 +83,7 @@
         "name": "MIT License",
         "url": "https://mit-license.org/",
     },
-    dependencies=dependencies,
```
agorajek, line 66: After being removed from here, I don't see the `dependencies` variable used anywhere in this module anymore. Should we still define it then on line 58?

datajunction-server/datajunction_server/utils.py
```diff
@@ -175,7 +182,7 @@ async def get_session(request: Request) -> AsyncIterator[AsyncSession]:
         await session.rollback()  # pragma: no cover
         raise exc  # pragma: no cover
     finally:
-        await session.close()
+        await session.remove()
```
agorajek, line 185: Why this switch?

agorajek (approved):
The fastAPI lifespan functionality is new to me. Looks useful though.

Small question in line.

datajunction-server/datajunction_server/api/main.py
shangyian: Good point, will remove that

datajunction-server/datajunction_server/utils.py
shangyian: So `remove()` is used with scoped sessions, which is what `session` is here, since it's created with `async_scoped_session`. From the docs:
> The [scoped_session.remove()](https://docs.sqlalchemy.org/en/20/orm/contextual.html#sqlalchemy.orm.scoped_session.remove) method first calls [Session.close()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.close) on the current [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session), which has the effect of releasing any connection/transactional resources owned by the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) first, then discarding the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) itself. “Releasing” here means that connections are returned to their connection pool and any transactional state is rolled back, ultimately using the rollback() method of the underlying DBAPI connection.

PR #1427, opened 2025-06-24
Fix ability to run validation queries

shangyian:
### Summary

This fixes the interaction between the core API and query service so that we can actually run validation queries:

Screenshot 2025-06-27 at 12 40 40 AM

The primary fix is that the core service needs to pass in an `accept` header that works for the query service.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/data.py
+++ b/datajunction-server/datajunction_server/api/data.py
@@ -284,7 +284,7 @@ async def get_data_stream_for_node(
         background_tasks=background_tasks,
     )
     if query_request and query_request.query_id:
-        return EventSourceResponse(
+        return EventSourceResponse(  # pragma: no cover
             query_event_stream(
                 query=QueryWithResults(
                     id=query_request.query_id,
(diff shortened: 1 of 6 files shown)


datajunction-server/datajunction_server/api/data.py
```diff
@@ -324,11 +324,6 @@ async def get_data_stream_for_node(
         request_headers=request_headers,
     )
 
-    # Save the external query id reference
```
agorajek, line 327: Why is this not needed?

agorajek (approved):
LGTM, just wondering why you removed the query id management?

datajunction-server/datajunction_server/api/data.py
shangyian: This is saving an entry to the QueryRequest table, which is actually better off as a cache and not as a Postgres database table. Saving an entry doesn't work with the read-only db connection, and it also doesn't make sense in this case since we should just be using a cache to store the query instead.

PR #1431, opened 2025-06-26
Add custom_metadata to /metrics/<name> api.

agorajek:
### Summary

What the title says.

### Test Plan

unit tests and some clicking

- [ ] `make check` passes

### Deployment Plan

auto

### DIFF
--- a/datajunction-server/datajunction_server/models/metric.py
+++ b/datajunction-server/datajunction_server/models/metric.py
@@ -2,7 +2,7 @@
 Models for metrics.
 """
 
-from typing import List, Optional
+from typing import List, Optional, Dict
 
 from pydantic.main import BaseModel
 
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/models/metric.py
```diff
@@ -49,6 +49,8 @@ class Metric(BaseModel):
     derived_query: str
     derived_expression: str
 
+    custom_metadata: Optional[Dict] = None
```
shangyian, line 52: I wonder if this should just default to an empty dict instead of `None`?

shangyian (approved):

datajunction-server/datajunction_server/models/metric.py
agorajek: Yeah, I wonder too. I was doing that (using `{}`) in some places in other PRs and the copilot suggestion was to not send unnecessary data over the wire.

PR #1432, opened 2025-06-27
Add deactivated filter for downstreamNodes GraphQL query

shangyian:
### Summary

Adds an `include_deactivated` filter for the `downstreamNodes` GraphQL query. Most use cases will not want to query deleted nodes, but only the active ones.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/queries/dag.py
+++ b/datajunction-server/datajunction_server/api/graphql/queries/dag.py
@@ -54,6 +54,12 @@ async def downstream_nodes(
             description="The node type to filter the downstream nodes on.",
         ),
     ] = None,
+    include_deactivated: Annotated[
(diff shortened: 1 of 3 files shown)


samredai (approved):
Nice! Lgtm but is there really a case where it needs to look at deactivated nodes? Those are presented to the user as deleted so they should be "invisible" until they're restored. So I'm wondering if instead of this flag it should just be the hard coded behavior to ignore deactivated nodes.

shangyian:
@samredai I figured that since this is an API call, it's ok to expose more options to the user, in case they want to see all possible nodes, including deactivated ones. Also this type of functionality would be what enables actions like "reactivate a node". However, in the UI we should always be hiding deactivated nodes.

PR #1434, opened 2025-06-28
Hide deleted dimension nodes from dimension links

shangyian:
### Summary

When a dimension node is deactivated, it should be hidden from the dimension links for all nodes. If this dimension node is ever restored, the dimension links are still recoverable (they're only hidden, not cleared out).

If a dimension node is hard-deleted, it should be removed from the dimension links for all nodes. The node cannot be restored, and if it is recreated, the links will also need to be recreated.

### Test Plan

- [x] `make check` passes


agorajek (approved):
Agree. Thank you!

Issue #786, opened 2023-09-21
Add owner to node


shangyian:
Nodes should have an owner

PR #1436, opened 2025-06-30
Add node owners

shangyian:
### Summary

This PR introduces ownership metadata to DJ nodes. Each node can now be associated with one or more user-level owners, enabling better visibility into who is responsible for maintaining a given node.

#### Database Changes

To support this, we add a `nodeowner` association table between nodes and owners, which enables additional metadata like `ownership_type` on the relationship. While this PR does not support setting the `ownership_type` for the time being, it is something that can be easily enabled via an API call.

There is a database migration that also uses each node's `created_by` user to populate the list of owners.

#### API Changes

* `PATCH /nodes/{node}` will support passing in a list of owners, which will be used to update the owners for the node
* `POST /nodes/{nodetype}` will support passing in a list of owners at creation time. If no owners are specified, the user creating the node will automatically be assigned as the owner.
 
#### UI Changes

In the UI, the list of owners are displayed on the node info page and can be edited by selecting from a dropdown of available users.

| Info   | Edit |
| ----- | ---- |
| Screenshot 2025-07-01 at 7 19 53 PM | Screenshot 2025-07-01 at 7 22 07 PM |

### Test Plan

- [ ] PR has an associated issue: #786
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/database/nodeowner.py
+++ b/datajunction-server/datajunction_server/database/nodeowner.py
@@ -14,7 +14,7 @@ class NodeOwner(Base):
     Join table for users and nodes that represents ownership
     """
 
-    __tablename__ = "nodeowner"
+    __tablename__ = "node_owners"
     __table_args__ = (
         Index("idx_nodeowner_node_id", "node_id"),
         Index("idx_nodeowner_user_id", "user_id"),
@@ -35,7 +35,7 @@ class NodeOwner(Base):
         primary_key=True,
     )
     ownership_type: Mapped[str] = mapped_column(
-        String(50),
+        String(256),
         nullable=True,
     )
 
--- a/datajunction-server/datajunction_server/database/nodeowner.py
+++ b/datajunction-server/datajunction_server/database/nodeowner.py
@@ -1,0 +1,43 @@
+"""Node - owners database schema."""
+
+from sqlalchemy import (
+    ForeignKey,
+    Index,
+    String,
+)
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+from datajunction_server.database.base import Base
+
+
+class NodeOwner(Base):
+    """
+    Join table for users and nodes that represents ownership
+    """
+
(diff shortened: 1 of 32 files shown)


datajunction-server/datajunction_server/database/nodeowner.py
```diff
@@ -0,0 +1,43 @@
...
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+from datajunction_server.database.base import Base
+
+
+class NodeOwner(Base):
+    """
+    Join table for users and nodes that represents ownership
+    """
+
+    __tablename__ = "nodeowner"
```
agorajek: I feel like we didn't do a great job in naming tables in the db so far and I know its late in the game now. With that said I think it would be good to start somewhere... and I think this should be mostly plural and with underscores for easy read, so what would you say to: `node_owners` ?

datajunction-server/datajunction_server/database/nodeowner.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    user_id: Mapped[int] = mapped_column(
+        ForeignKey(
+            "users.id",
+            name="fk_nodeowner_user_id",
+        ),
+        primary_key=True,
+    )
+    ownership_type: Mapped[str] = mapped_column(
+        String(50),
```
agorajek: Whether you make it 50 or 1024 database will take up the same size cause its all just a variable string. I think 256 may be a safer choice.

datajunction-server/datajunction_server/database/nodeowner.py
shangyian: Hahaha, yeah agreed. Maybe at some point I'll put in a db migration to just change all the table names...

datajunction-server/datajunction_server/database/nodeowner.py
shangyian: Good to know! Will update

Issue #1448, opened 2025-07-25
Add Caching to Measures SQL endpoints


shangyian:
As an initial strategy, we can cache at the `GET /sql/measures/v2` API-call level, and compute the cache key based on the API input arguments. * The results can be cached for a long period.
* We will update the cache in an async call every time the API endpoint is called. 
* We can invalidate the cache (this can be added as a follow-up PR) every time any of the requested metrics, dimensions, or upstream nodes are updated. We could consider recomputing the cached values.

shangyian:
I was thinking it could be days though -- given that it will get updated if it ever gets called.

samredai:
> I was thinking it could be days though -- given that it will get updated if it ever gets called. What do you think? 

SGTM! I'm not too familiar with how this endpoint is used.

PR #1449, opened 2025-07-28
Add caching for sql generation

shangyian:
### Summary

This PR introduces a generic query cache manager to support caching of generated SQL. This setup supports different caching strategies (only one is implemented here) and is enabled for the `/sql/measures` endpoint. 

Changes include:
* Implemented a `CacheManager` base class with reusable cache key generation logic based on sorted, hashed request parameters.
* Added an implementation for the refresh-ahead caching strategy (`RefreshAheadCacheManager`), which serves cached results immediately and refreshes data asynchronously. It will compute results via the fallback logic if the cache is empty or bypassed with Cache-Control headers.
* Added a `QueryCacheManager` for caching SQL generation requests. At the moment this is only enabled for measures SQL, but it can be enabled more broadly for other types of SQL generation. This will build versioned cache keys based on the metric / dimension versions to ensure that query cache keys will change if any underlying definitions change.

### Test Plan

- [x] PR has an associated issue: #1448 
- [x] `make check` passes


datajunction-server/datajunction_server/database/queryrequest.py
```diff
@@ -406,3 +407,234 @@ async def to_versioned_query_request(
...
+        if "@" in node_key:
+            node_key, version = node_key.split("@", 1)
+            return VersionedNodeKey(name=node_key, version=version)
+        return VersionedNodeKey(name=node_key, version=None)
+
+
+@dataclass
+class VersionedQueryKey:
+    """
+    A versioned query request encapsulates the logic to version nodes, dimensions,
```
agorajek, line 456: I wonder why you simply didn't take the all the `QueryRequestParams` here and picked these only?

agorajek (approved):
One comment in-line, but I really like it overall. Thanks!

datajunction-server/datajunction_server/database/queryrequest.py
shangyian: The rest of the query params don't actually require any additional logic to set up versioning. Parameters like `include_all_columns` can just be included directly, whereas something like `default.num_repair_orders` should be frozen with the current version, so that we're always requesting SQL for the right version.

PR #1451, opened 2025-07-28
Revert "Revert "Fix hashing on measure decomposition. (#1404)" (#1407)"

agorajek:
This reverts commit ad99fac5bb53de71a13ed87db0945ed88cb25dd6.

### Summary

One more attempt at fixing the measure decomposition bug: https://github.com/DataJunction/dj/pull/1404

### Test Plan

Not much. See https://github.com/DataJunction/dj/pull/1404

- [ ] `make check` passes

### Deployment Plan

Auto.


datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -84,6 +84,13 @@ def extract(self) -> tuple[list[MetricComponent], ast.Query]:
             self._extracted = True
         return self._components, self._query_ast
 
+    def _short_hash(self, expression: str) -> str:
```
shangyian, line 86: If I remember correctly, this was removed due to the fact that it changes the hash on many stored columns, meaning that any existing materializations would not be backwards-compatible. Isn't this still an issue?

agorajek: Yes, I think you are right. And even if we provide an easy button, or do the refresh of the jobs automatically - there will still be a period when the new cubes are not yet ready but the old ones are not-compatible with the new code. I am not sure if there is a completely safe upgrade path here.

shangyian: Maybe we can do something like create a stable view on top of the materialized dataset, which is what users will typically interact with, and we can do whatever we need for the actual materialized tables as long as the view gets updated to point to the new columns. Thoughts?

I do think it's worth spending some time on migration options here, since over time we'll probably discover more bugs like this one and need to update the hash function in a similar way.

agorajek: Yes, the view is a cool idea to minimize potential impact, but I am not sure how to make use of it in an automated way. I think it could server us well as a manual fix, but I am afraid that we could be making bigger mess if we add views w/o thinking through all the pieces. Here is the steps I'd like to take wrt this:
1. Add Mat Refresh API call + UI button to DJ oss.
2. Make sure it works on Netflix side - with add any code needed.
3. Tel users about potential impact - and give them a date and steps to remediate.
4. Push this PR and help with cleanup.

I'll ping you off-line to talk about the views. I think they could help if used properly...

PR #1453, opened 2025-07-30
Add dimension link versioning to nodes

shangyian:
### Summary

Adding or removing a dimension link should bump the version for a node. Version tracking helps when we're trying to compute cache keys, so that we know to evict from the cache if the underlying node's dimension links change, and the generated SQL is no longer valid.

This also fixes a bug where `Node.get_by_names(...)` does not return the nodes in the order that they were requested. This means that there were cases where we were freezing query requests with the wrong node versions.

### Test Plan

Locally

- [x] `make check` passes

### Deployment Plan

N/A

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -1125,6 +1125,7 @@ def copy_existing_node_revision(old_revision: NodeRevision, current_user: User):
                 join_sql=link.join_sql,
                 join_type=link.join_type,
                 join_cardinality=link.join_cardinality,
+                role=link.role,
                 materialization_conf=link.materialization_conf,
             )
             for link in old_revision.dimension_links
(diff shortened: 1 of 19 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1972,6 +1991,27 @@ async def remove_dimension_link(
...
+async def create_new_revision_for_dimension_link_update(
+    session: AsyncSession,
+    node: Node,
+    current_user: User,
+) -> NodeRevision:
+    """
+    Create a new revision for the node to capture the dimension link changes in a new version
+    """
+    new_revision = copy_existing_node_revision(node.current, current_user)
+    new_revision.version = str(Version.parse(node.current_version).next_major_version())
```
agorajek: Yeah... I think this should be minor.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1972,6 +1991,27 @@ async def remove_dimension_link(
...
+    node.current_version = new_revision.version
+    new_revision.node = node
+
+    session.add(new_revision)
+    await session.commit()
+    await session.refresh(new_revision)
+    await session.refresh(new_revision, ["dimension_links"])
```
agorajek, line 2010: Why is this line repeated with the additional argument?

agorajek (approved):
Looks good. Just that major-2-minor version comment.

datajunction-server/datajunction_server/internal/nodes.py
shangyian: It's because the `refresh` call here doesn't automatically refresh relations like `dimension_links`, so we have to manually do the refresh

PR #1457, opened 2025-08-03
Save versioned measures

shangyian:
### Summary

Today, "concrete" measures are derived on-the-fly, meaning that they are computed when requested for a specific metric. This change stores concrete measures into the metadata database, which enables us to recover versioned measures locked to a specific metric revision.

A concrete measure represents the binding of a measure expression and aggregation rule to a concrete node revision in the data graph. It includes a name that is generated based on the hash of these fields, which is used as an output column name in any compiled SQL or materialized datasets.

Changes here include:
* Added the `concrete_measures` table, which binds a measure expression + agg to a specific `NodeRevision`.
* Added a `node_revision_concrete_measures` link table to associate Concrete Measures with metric node revisions that use them.
* Extended the node revision model to support storing a `derived_expression` when needed.
* Added a script that can backfill Concrete Measures from existing metric definitions (to be run separately).

By adding this extra layer of metadata, we can easily find the associated measures for any given cube node revision, since they will be recorded in the metadata db.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/alembic/versions/2025_08_02_1543-8eab64955a49_add_concrete_measures_for_node_revisions.py
+++ b/datajunction-server/datajunction_server/alembic/versions/2025_08_02_1543-8eab64955a49_add_concrete_measures_for_node_revisions.py
@@ -1,0 +1,75 @@
+"""
+Add concrete measures for node revisions
+
+Revision ID: 8eab64955a49
+Revises: 634fdac051c3
+Create Date: 2025-08-02 15:43:20.001874+00:00
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+from datajunction_server.database.measure import MeasureAggregationRuleType
+from datajunction_server.utils import get_settings
+
+settings = get_settings()
+
+# revision identifiers, used by Alembic.
+revision = "8eab64955a49"
+down_revision = "634fdac051c3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "concrete_measures",
+        sa.Column("id", sa.BigInteger(), nullable=False),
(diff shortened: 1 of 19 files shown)


datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -104,6 +104,8 @@ def _simple_associative_agg(self, func) -> list[MetricComponent]:
             )
 
         expression = str(arg)
+        # TODO: we should only hash on the lowercase version of the expression, but we can't
+        # do this migration until after we get versioned measures
```
agorajek, line 108: PR is ready to fix that once this is in: https://github.com/DataJunction/dj/pull/1451

datajunction-server/datajunction_server/alembic/versions/2025_08_02_1543-8eab64955a49_add_concrete_measures_for_node_revisions.py
```diff
@@ -0,0 +1,75 @@
...
+# revision identifiers, used by Alembic.
+revision = "8eab64955a49"
+down_revision = "634fdac051c3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "concrete_measures",
```
agorajek: Thank you for the nice table names.

agorajek (approved):
Thanks for quick turnover @shangyian . After some name consideration I think that `concrete` is not the best because its opposite is not great, say: abstract, theoretical, imaginary. I think we should rename this PR to use `frozen`, `versioned` or `collapsed`, but all these I think `frozen` carries the meaning best of all.

PR #1460, opened 2025-08-05
Expose materialization information better for cubes.

agorajek:
### Summary

The overarching goal here is to add ability for a controlled cube versions and their materializations. Such that a cube owner can create/experiment with a new version of the cube while their dashboards are healthy pointing to the current materialization revision.

This PR adds some visibility into different materializations and availability states for cubes. The major changes are:
1. Adding tabs under Materialization tab to show all the active/in-active materializations attached to the cube.
2. `Add Materialization` button changes to `Rebuild (latest) Materialization` if such exists.
3. Added some APIs to support retrieval of the above info.
4. The new `Rebuild (latest) Materialization` button makes a new minor cube version so the previous latest materialization is not affected.

See example screenshot below.

Minor change: put all db relate init sql files in one spot.

### Test Plan

Manual tests. Plus unit tests.

- [ ] `make check` passes

### Deployment Plan

Screenshot 2025-08-05 at 1 28 57 PM


docker-compose.yml
```diff
@@ -141,10 +141,10 @@ services:
     environment:
       - PGPASSWORD=dj
     volumes:
-      - ./datajunction-server/init_data:/init_data
+      - ./db-init:/db-init
```
shangyian, line 144: Is this just to merge the read-only user creation script and the dj user creation script into a single dir?

datajunction-server/datajunction_server/api/materializations.py
```diff
@@ -520,3 +521,55 @@ async def run_materialization_backfill(
...
+    session: AsyncSession = Depends(get_session),
+) -> List[AvailabilityStateInfo]:
+    """
+    Retrieve all availability states for a given node across all revisions.
+    """
+    # Get all revisions with their availability states
+    node = await Node.get_by_name(
+        session,
+        node_name,
+        options=[
```
shangyian, line 544: Should this endpoint optionally support only listing availability for the latest node revision?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -912,13 +914,14 @@ async def update_cube_node(
...
     save_history: Callable,
+    refresh_materialization: bool = False,
 ) -> Optional[NodeRevision]:
     """
     Update cube node based on changes
     """
     node = await Node.get_cube_by_name(session, node_revision.name)
     node_revision = node.current  # type: ignore
-    minor_changes = has_minor_changes(node_revision, data)
+    minor_changes = has_minor_changes(node_revision, data) or refresh_materialization
```
shangyian, line 924: So in this case, refreshing the materialization will result in a new node revision being created. Is that level of state tracking needed, given that nothing has changed on the node revision itself?

datajunction-server/datajunction_server/models/materialization.py
```diff
@@ -142,6 +142,31 @@ class MaterializationConfigInfoUnified(
...
+    """
+
+    id: int
+    catalog: str
+    schema_: Optional[str] = Field(alias="schema")
+    table: str
+    valid_through_ts: int
+    url: Optional[str] = None
+    links: Optional[Dict] = None
+    categorical_partitions: Optional[List[str]] = []
```
shangyian: btw, assigning this to `[]` (which is a mutable object) could cause unexpected issues, because `AvailabilityStateInfo` will share the same list object for this field (and for each of the fields below) unless overridden. 

For pydantic models you could default to this instead: `Field(default_factory=list)`

datajunction-server/datajunction_server/models/materialization.py
```diff
@@ -142,6 +142,31 @@ class MaterializationConfigInfoUnified(
...
+    temporal_partitions: Optional[List[str]] = []
+    min_temporal_partition: Optional[List[str]] = []
+    max_temporal_partition: Optional[List[str]] = []
+    partitions: Optional[List] = []
```
shangyian:
```suggestion
categorical_partitions: list[str] = Field(default_factory=list)
temporal_partitions: list[str] = Field(default_factory=list)
min_temporal_partition: list[str] = Field(default_factory=list)
max_temporal_partition: list[str] = Field(default_factory=list)
partitions: list[str] = Field(default_factory=list)
```

datajunction-ui/src/app/pages/NodePage/NodeMaterializationTab.jsx
```diff
@@ -5,20 +5,99 @@ import * as React from 'react';
...
+        }, {});
+
+        setAvailabilityStatesByRevision(availabilityGrouped);
+
+        // Filter materializations based on checkbox state
+        const filteredMaterializations = showInactive
+          ? data // Show all materializations if checkbox is checked
+          : data.filter(mat => !mat.deactivated_at); // Only active if unchecked
+
+        setMaterializations(filteredMaterializations);
```
shangyian: I think it would be better to derive filteredMaterializations on render and not store it in the state manager. Otherwise the state setup would have redundant state and added complexity in event handlers. Something like this which will memoize the results or just compute directly in the render func:
```
const filteredMaterializations = useMemo(() => {
  return showInactive
    ? rawMaterializations
    : rawMaterializations.filter(mat => !mat.deactivated_at);
}, [rawMaterializations, showInactive]);
```

datajunction-ui/src/app/pages/NodePage/NodeMaterializationTab.jsx
```diff
@@ -35,9 +114,238 @@ export default function NodeMaterializationTab({ node, djClient }) {
...
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          justifyContent: 'space-between',
+          marginBottom: '20px',
+        }}
+      >
+        <div className="align-items-center row">
+          {sortedVersions.map(version => {
+            const isCurrentVersion = version === node?.version;
```
shangyian: Maybe the rendering for each revision tab could be extracted to a separate object so that you can then just do `<RevisionTab node={node} version={version} />`

datajunction-ui/src/app/pages/NodePage/NodeMaterializationTab.jsx
```diff
@@ -35,9 +114,238 @@ export default function NodeMaterializationTab({ node, djClient }) {
...
+    // Check if latest version has active materializations
+    const hasLatestVersionMaterialization =
+      materializationsByRevision[node?.version] &&
+      materializationsByRevision[node?.version].length > 0;
+
+    // Refresh latest materialization function
+    const refreshLatestMaterialization = async () => {
+      if (
+        !window.confirm(
+          'This will create new version of the cube and build new materialization workflows. Previous version of the cube and its materialization will be accessible using specific version label. Would you like to continue?',
```
shangyian:
```suggestion
          'This will create a new version of the cube and build new materialization workflows. The previous version of the cube and its materialization will be accessible using a specific version label. Would you like to continue?',
```

datajunction-ui/src/app/pages/NodePage/NodeMaterializationTab.jsx
```diff
@@ -206,28 +515,152 @@ export default function NodeMaterializationTab({ node, djClient }) {
...
+  const renderMaterializedDatasets = availabilityStates => {
+    if (!availabilityStates || availabilityStates.length === 0) {
+      return (
+        <div className="message alert" style={{ marginTop: '10px' }}>
+          No materialized datasets available for this revision.
+        </div>
+      );
+    }
+
+    return availabilityStates.map((availability, index) => (
```
shangyian, line 494: Similar suggestion here: maybe each individual availability state rendering could be a separate object, and this could be:
```
    return availabilityStates.map((availability, index) => <AvailabilityStateBlock availability={availability} />
```

shangyian (commented):
Exciting stuff, thank you for adding!

docker-compose.yml
agorajek: Yes.

datajunction-server/datajunction_server/internal/nodes.py
agorajek: I think so, because refreshing (actually: rebuilding) materialization is a potentially time consuming task. I feel like it is a similar weight of change as e.g. a description change. Basically we don't want to loose the previous materialization by accident.

datajunction-server/datajunction_server/models/materialization.py
agorajek: Thanks for the suggestion!

datajunction-server/datajunction_server/api/materializations.py
agorajek: I didn't have a need for this call yet, but it would be easy to change when needed.

datajunction-ui/src/app/pages/NodePage/NodeMaterializationTab.jsx
agorajek: Done.

datajunction-ui/src/app/pages/NodePage/AvailabilityStateBlock.jsx
```diff
@@ -0,0 +1,67 @@
+import TableIcon from '../../icons/TableIcon';
```
shangyian, line 1: Nice!

PR #1469, opened 2025-08-21
Fix inefficiencies in finding linked nodes to dimensions

shangyian:
### Summary

This change fixes some inefficiencies in finding nodes linked to a dimension, where we're not stopping the search when we come across a node that's already been processed.

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/sql/dag.py
```diff
@@ -882,6 +882,8 @@ async def get_nodes_with_dimension(
     final_set: Set[NodeRevision] = set()
     while to_process:
         current_node = to_process.pop()
+        if current_node.name in processed:
+            continue
         processed.add(current_node.name)
```
agorajek, line 887: Wondering why you add the node to `processed` set before it is actually processing in the lines below. Or if you actually want to do that then perhaps this variable should be really called `processing_started` ?

shangyian: That pattern is because we're marking the node as something that does not need to be handled again. With this setup, if the BFS branches back into this current node, we won't have an infinite exploration, whereas if we added it to processed after, it would keep exploring. We can rename "processed" but I think it's pretty standard to do this.

shangyian: I can change it to the more standard `explored` if that helps clarify?

agorajek: I think you are right. Don't change anything.

PR #1476, opened 2025-08-25
Avoid use of scoped session in favor of AsyncSession directly

shangyian:
### Summary

This change replaces `async_scoped_session` with `async_sessionmaker`. This is because `async_scoped_session` causes various issues with connection lifecycle and garbage collection, occasionally triggering warnings about non-checked-in connections like:
```
The garbage collector is trying to clean up non-checked-in connection <AdaptedConnection <psycopg.AsyncConnection [INTRANS] 
```

The scoped session setup doesn’t interact well with async frameworks like FastAPI when sessions should be created per-request or per-GraphQL query. Debugging connection issues was harder because the scope-based session would hold connections longer than expected. 

See the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#using-asyncio-scoped-session) for more info, particularly this tidbit:
> SQLAlchemy generally does not recommend the “scoped” pattern for new development as it relies upon mutable global state that must also be explicitly torn down when work within the thread or task is complete. Particularly when using asyncio, it’s likely a better idea to pass the [AsyncSession](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncSession) directly to the awaitable functions that need it.

### Test Plan

Tested locally

- [x] `make check` passes

### Deployment Plan

ASAP

### DIFF
--- a/datajunction-server/datajunction_server/utils.py
+++ b/datajunction-server/datajunction_server/utils.py
@@ -22,7 +22,6 @@ from sqlalchemy.exc import MissingGreenlet, OperationalError
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
-    async_scoped_session,
     async_sessionmaker,
     create_async_engine,
 )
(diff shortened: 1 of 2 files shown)


agorajek (approved):
Looks good. Hope this will help!

PR #1485, opened 2025-09-02
Add support for service accounts

shangyian:
### Summary

This PR adds support for service accounts, which will enable programmatic clients to authenticate and interact with DJ. This setup has service accounts modeled as a part of the `users` table for simplicity, and they're identified with the `kind` flag. Any given service account has to be created with an authenticated identity, which is represented by the `created_by_id` field.

This change includes API endpoints for:
* Creating a new service account.
* Generating authentication tokens for service accounts. These bearer tokens have a configurable expiration period and can be used to make subsequent calls to the DJ API. 

### Test Plan

- [x] PR has an associated issue: #1486 
- [x] `make check` passes


datajunction-server/datajunction_server/database/user.py
```diff
@@ -48,6 +74,23 @@ class User(Base):
...
+        nullable=True,
+    )
+
+    created_at: Mapped[UTCDatetime | None] = mapped_column(
+        DateTime(timezone=True),
+        insert_default=partial(datetime.now, timezone.utc),
+        nullable=True,
+    )
+
+    created_by: Mapped["User"] = relationship("User")
```
agorajek, line 93: How is this different from `created_by_id` ?

agorajek (commented):
Ah, forgot to review yday... just added one question inline.

datajunction-server/datajunction_server/database/user.py
shangyian: This is the ORM reference to the `User` object, similar to how `created_nodes` works on the user class as well.

PR #1491, opened 2025-09-10
Bulk Deployment

shangyian:
### Summary

This sets up bulk deployment of a group of nodes under a namespace.

### Test Plan

- [ ] `make check` passes


datajunction-clients/python/datajunction/cli.py
```diff
@@ -19,25 +20,20 @@ def __init__(self, builder_client: DJBuilder | None = None):
         Initialize the CLI with a builder client.
         """
         self.builder_client = builder_client or DJBuilder()
+        self.deployment_service = DeploymentService(client=self.builder_client)
 
-    def deploy(self, directory: str, dryrun: bool):
+    def push(self, directory: str, namespace: str | None = None):
         """
-        Deploy nodes from the specified directory.
+        Alias for deploy without dryrun.
```
agorajek, line 27: Wondering why you are adding the `push` command? Wasn't `deploy` sufficiently clear? Is there a different alias for `deploy with dryrun` ?

datajunction-clients/python/datajunction/deployment.py
```diff
@@ -0,0 +1,199 @@
...
+        return result
+
+    def pull(
+        self,
+        namespace: str,
+        target_path: Union[str, Path],
+        ignore_existing_files: bool = False,
+    ):
+        """
+        Export a namespace to a local project.
```
agorajek, line 53: I am pretty sure we already have this functionality...

datajunction-server/datajunction_server/database/deployment.py
```diff
@@ -0,0 +1,59 @@
...
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+from sqlalchemy_utils import UUIDType
+from datajunction_server.typing import UTCDatetime
+from datajunction_server.database.user import User
+from datetime import datetime, timezone
+from functools import partial
+from datajunction_server.database.base import Base
+
+
+class Deployment(Base):
```
agorajek, line 19: Is this concept equivalent to Staging env? If so, wouldn't it be better to call it Staging or Environment? I just think of Deployment as the process rather than the state.

datajunction-server/datajunction_server/alembic/versions/2025_09_17_2107-b6398ba852b3_deployments.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def upgrade():
+    op.create_table(
+        "deployments",
+        sa.Column("uuid", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False),
+        sa.Column("namespace", sa.String(), nullable=False),
+        sa.Column(
+            "status",
+            sa.Enum("PENDING", "RUNNING", "FAILED", "SUCCESS", name="deploymentstatus"),
```
agorajek, line 28: What's the difference between PENDING and RUNNING?

datajunction-clients/python/datajunction/deployment.py
shangyian: We do, but the existing schema is a bit different. In the interest of not breaking any existing users, it felt cleaner to have this be separate.

datajunction-clients/python/datajunction/cli.py
shangyian: I thought `push` was more analogous to `pull`, but we can switch back to deploy if that's preferred.

datajunction-server/datajunction_server/database/deployment.py
shangyian: I don't think this is conceptually a staging env, it's more like a record of a bulk load of nodes and more to a namespace. We could alternatively call it `loads`, but I'm not sure that's clearer

datajunction-server/datajunction_server/alembic/versions/2025_09_17_2107-b6398ba852b3_deployments.py
shangyian: `PENDING` should be when it first gets enqueued, and `RUNNING` is after the job gets picked up and it starts processing.

PR #1497, opened 2025-09-15
Start using patch release numbers.

agorajek:
### Summary

This PR switches our release pattern from `alpha` to `patch`.

### Test Plan

Tested with manual run: https://github.com/DataJunction/dj/pull/1498

- [ ] `make check` passes

### Deployment Plan

Auto.

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -8,18 +8,18 @@ on:
         type: choice
         description: "Select version cycle"
         required: true
-        default: alpha
+        default: patch
         options:
           # Uncomment next item when ready to switch:
           - alpha
-          # - patch
+          - patch
           # - minor
           # - major
 
 jobs:
   bump:
     env:
-      bump-type: ${{ github.event.inputs.bump || 'alpha' }}  # Default to alpha
+      bump-type: ${{ github.event.inputs.bump || 'patch' }}  # Default to patch
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/version-bump.yml
```diff
@@ -85,13 +85,14 @@ jobs:
...
-      - name: Update docs (for major, minor or patch release)
-        working-directory: ./docs
-        if: ${{ env.bump-type == 'major' || env.bump-type == 'minor' || env.bump-type == 'patch'}}
-        run: |
-          ./build-docs.sh $NEW_VERSION true
+      # - name: Update docs (for major, minor or patch release)
+      #   working-directory: ./docs
+      #   if: ${{ env.bump-type == 'major' || env.bump-type == 'minor' || env.bump-type == 'patch'}}
+      #   run: |
+      #     ./build-docs.sh $NEW_VERSION true
```
agorajek: Hey @samredai and @shangyian . This command fails because there is a 3rd argument missing for BASE_URL, but this never ran for the `alpha` releases so I temporarily commented it out. I am not sure what's the best fix here, but I think I am missing how we updated the docs so far.

shangyian: @agorajek I don't believe we actually need this command. Instead, just running `npm run build` will build the docs, and `npm run build -- -b $DEPLOY_URL` will set them up to be deployed.

shangyian (approved):

.github/workflows/version-bump.yml
agorajek: Agree. I deleted this section.

Issue #1499, opened 2025-09-17
Replace (or supplement) the OSS query service pattern with dependency injected clients for major data warehouse vendors


samredai:
Today one of the hardest parts of getting ramped up on the OSS project is connecting the entire system to your data warehouse via the query service that comes in the docker compose setup. That's mainly because it requires this separate query service to run that needs to be configured in a particular way. In a production environment at a large company, that design makes sense, but since the **client** for the query service is dependency injected, we could make it a little bit easier get started using the DJ OSS project by **only** running the single main DJ server (`datajunction_server/...`).

## How would this look?

Today the main DJ server has a [QueryServiceClient](https://github.com/DataJunction/dj/blob/main/datajunction-server/datajunction_server/service_clients.py#L76) that assumes a configuration where the DJ server makes http calls out to another service (the "query service") to do a few things:
- Register tables including detecting all of the column names and types
- Refresh tables by checking the warehouse and reflecting any schema changes
- Submit SQL queries generated by the semantic layer and returning data

The query service **client** is dependency injected in the core logic of the main DJ server (i.e. `query_service_client: QueryServiceClient = Depends(get_query_service_client)`). We can provide various implementations of this QueryServiceClient directly in main DJ server that can be configured and directly injected, allowing the main service to talk directly to the data warehouse without proxying calls through another service. For example, we could have:
- BigQueryClient
- DatabricksClient
- SnowflakeClient
- TrinoClient
- SparkClient
- etc...

They would all implement some sort of `BaseQueryServiceClient` abstract base class and provide the right methods:
- `get_columns_for_table`
- `create_view`
- `submit_query`
- `get_query`
- `materialize`
- ...

The other nice side-effect of this is that custom implementations can selectively implement the methods they plan to use. For example, it's reasonable that you don't plane to submit queries and return data via the DJ API, but simply want to register tables from your warehouse with a click of a button, keep them in sync, and model metrics, dimensions, and cubes on top of them to produce valid cube SQL. In such a scenario you could **only** implement `get_columns_for_table`.

PR #1504, opened 2025-09-22
DRAFT - Allow configurable embedded query client (Issue #1499)

samredai:
#### BaseQueryServiceClient
Issue #1499 describes the motivation behind this PR. Instead of requiring a query service matching a REST specification, this offers an injection point for custom implementations that's embedded within the main DJ server. The query service client was already an injected dependency, but this adds a query client abstract base class that the existing query service client wraps. This allows other query client implementations to also wrap the base class and be injected for use within the main server. Some examples of what can be done with this:
- A query client that directly calls a vendor database (this PR adds an implementation that uses snowflake-connector-python, for example)
- A query client that calls a DJ REST spec query service (the original and now default configuration)
- A query client that calls some other non-DJ REST spec query service

#### Query Client Configuration 
Another thing this adds is a configuration framework for configuring the query client. The implementation is chosen using a specific string value for `QueryClientConfig.type` or provided via an environment variable `QUERY_CLIENT__TYPE`. The other params required by that specific client type can also be passed in via environment variables. For example, here's a snowflake configuration:
```sh
export QUERY_CLIENT__TYPE=snowflake
export QUERY_CLIENT__CONNECTION__ACCOUNT=FOOACCOUNT
export QUERY_CLIENT__CONNECTION__USER=$SNOWFLAKE_USER
export QUERY_CLIENT__CONNECTION__PASSWORD=$SNOWFLAKE_PASSWORD
export QUERY_CLIENT__CONNECTION__WAREHOUSE=FOOWAREHOUSE
export QUERY_CLIENT__CONNECTION__DATABASE=FOODB
export QUERY_CLIENT__CONNECTION__SCHEMA=FOOTABLE
```

The snowflake implementation in this PR only implements getting column types for a table which is the minimum requirement for the query service and allows for registering tables to model on top of in DJ. It seems to be working as expected but could probably use some more validation of the column type conversions (converting snowflake column types to the corresponding DJ column type object).

### Test Plan

Added tests, configured a snowflake account, and registered some tables in the DJ UI.

- [ ] PR has an associated issue: #1499 
- [ ] `make check` passes


datajunction-server/datajunction_server/config.py
```diff
@@ -93,9 +108,12 @@ class Config:
     # How long to wait when pinging databases to find out the fastest online database.
     do_ping_timeout: timedelta = timedelta(seconds=5)
 
-    # Query service
+    # Query service url (only used with "http" query client config)
     query_service: Optional[str] = None
```
samredai, line 113: I'm keeping this here for now so as not to break any existing server configuraitons, but eventually we can migrate this to use a `QueryClientConfig` of type `"http"` and provides the url as a param. That way no matter what kind of query client you're using you go through the same query client config logic.

datajunction-server/datajunction_server/query_clients/http.py
```diff
@@ -0,0 +1,169 @@
...
+from datajunction_server.models.materialization import (
+    DruidMaterializationInput,
+    GenericMaterializationInput,
+    MaterializationInfo,
+)
+from datajunction_server.models.node_type import NodeType
+from datajunction_server.models.partition import PartitionBackfill
+from datajunction_server.models.query import QueryCreate, QueryWithResults
+from datajunction_server.query_clients.base import BaseQueryServiceClient
+from datajunction_server.service_clients import QueryServiceClient
```
agorajek, line 18: Would it make sense to delete the code of `QueryServiceClient` sooner or later? Wondering if you are perhaps thinking of doing this as the next step or not doing at all and why?

shangyian: Would deleting the legacy `QueryServiceClient` keep it it backwards compatible for existing setups?

samredai: Yeah that's exactly it, I wanted to keep this as obviously backwards compatible as possible until we can migrate internally to the new config structure. Once we do that, it will be an easy follow-up to update this and what I'm planning to do is basically just rename `QueryServiceClient` -> `HttpQueryServiceClient` and make it inherit from `BaseQueryServiceClient`.

datajunction-server/Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV RELOAD ${RELOAD}
 WORKDIR /code
 COPY . /code
 RUN pip install --no-cache-dir --upgrade -r /code/requirements/docker.txt
-RUN pip install -e .
+RUN pip install -e .[all]
```
agorajek, line 9: Btw, what's the difference with `[all]` ?

datajunction-server/datajunction_server/config.py
```diff
@@ -37,6 +37,21 @@ class DatabaseConfig(BaseModel):
...
+
+    # Connection parameters (varies by client type)
+    connection: Dict[str, Any] = Field(default_factory=dict)
+
+    # Number of retries for failed requests (mainly for HTTP client)
+    retries: int = 0
+
+
 class SeedSetup(BaseModel):
     # An "default" catalog for nodes that are pure SQL and don't belong in any
```
agorajek:
```suggestion
    # A "default" catalog for nodes that are pure SQL and don't belong in any
```

datajunction-server/datajunction_server/config.py
```diff
@@ -93,9 +108,12 @@ class Settings(BaseSettings):  # pragma: no cover
     # How long to wait when pinging databases to find out the fastest online database.
     do_ping_timeout: timedelta = timedelta(seconds=5)
 
-    # Query service
+    # Query service url (only used with "http" query client config)
```
agorajek, line 111:
```suggestion
    # Query service url (only used with "http" query client config)
    # TODO: once the `QueryClientConfig` is proven out, this can be removed.
```

datajunction-server/datajunction_server/query_clients/http.py
```diff
@@ -0,0 +1,169 @@
...
+        query_id: str,
+        request_headers: Optional[Dict[str, str]] = None,
+    ) -> QueryWithResults:
+        """Get a previously submitted query from the HTTP query service."""
+        return self._client.get_query(
+            query_id=query_id,
+            request_headers=request_headers,
+        )
+
+    def materialize(
```
agorajek, line 95: Since we started leaning towards cube materialization only (for the time being), let's maybe remove this call for now, from here and from the based class? It will be easier to add it back than to remove in the future.

agorajek (approved):
Hey @samredai - this is fantastic. I made few comments in-line. Most of them minor but one would be nice to address before you push this PR. I noticed you exposed the generic `materialize` method. I wonder if we should hide it for now and only expose the `materialize_cube` one - for easier feature management in the future.

datajunction-server/Dockerfile
shangyian: This is basically saying to install the optional packages defined in pyproject.toml under:
```
all = [
    "snowflake-connector-python>=3.0.0",
]
```

datajunction-server/datajunction_server/config.py
```diff
@@ -37,6 +37,21 @@ class DatabaseConfig(BaseModel):
     keepalives_count: int = 5
 
 
+class QueryClientConfig(BaseModel):
+    """
+    Configuration for query service clients.
+    """
+
+    # Type of query client: 'http', 'snowflake', 'bigquery', 'databricks', 'trino', etc.
```
shangyian, line 45: Is it the case that most use cases will only use a single query engine? Or is it reasonable to say that people can mix and match query engines, which can be determined based on the type of engine that's registered in our `engines` table?

samredai: I think mixing and matching is reasonable, the question I have though is do we need to support that with these basic single engine query clients? I doubt we can standardize how the specific query engine would be chosen. Your idea of looking at the engine type make sense but I could see someone else wanting it to be driven by a request header. Is it reasonable for the OSS recommendation to be create a simple wrapper QueryServiceClient implementation that has whatever query routing logic someone would like? They could have a `SuperQueryClient` that makes the decision and based on that decision just instantiates and uses the right query service client. Something like if `engine: bigquery` is in the request header, use a `BigQueryClient`, if `engine: databricks` is in the request header use a `DatabricksQueryClient`, etc.

Issue #1540, opened 2025-10-24
On rare occasions when trying to edit a node in the UI, it won't allow you and show "undefined"


samredai:
It's unclear if this is a particular state the node has to get into, however users have reported that occasionally clicking the Edit button for a node would show `undefined`. Deleting and recreating the node seems to resolve it. I don't know exactly how to reproduce it.

PR #1547, opened 2025-10-27
Fix custom_metadata field bug and make some other small changes to node details edit page

samredai:
### Summary

There's a bug that's causing saving nodes in the UI to break and just show `undefined` as the error.

Screenshot 2025-10-27 at 9 35 37 AM

I narrowed it down to a bug with this custom metadata field that was recently added. It prefills as an empty JSON object `{}` but it's actually in the formik form as a null. That was causing some issues as it flows through the graphql layer and the REST API layer and does some comparisons on the backend to determine if it's a change (I updated a little bit of that to make the default value None and the comparison handler to be simpler and just see if `custom_metadata` is being passed in and if so, to set it).

This uncovered a few other issues this tries to fix such as when adding duplicate keys to the custom metadata i.e. `{"foo": "Hi", "foo" "Bye"}` it just takes the last key so `{"foo": "Bye"}`.  I also refactored the form a little bit to just use the standard Formik conventions so that it can do things like automatically disable the Save button if the form is invalid. This should cause less surprises when a user saves but the form data is invalid.

### Test Plan

Ran the ui locally with `docker compose up` and created some nodes.

- [ ] PR has an associated issue: #1540 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -1708,8 +1708,8 @@ async def create_new_revision_from_existing(
     if data and data.required_dimensions is not None:  # type: ignore
         new_revision.required_dimensions = data.required_dimensions  # type: ignore
 
-    if data and hasattr(data, "custom_metadata"):  # type: ignore
-        new_revision.custom_metadata = data.custom_metadata  # type: ignore
+    if data and data.custom_metadata is not None:  # type: ignore
+        new_revision.custom_metadata = data.custom_metadata
 
     # Link the new revision to its parents if a new revision was created and update its status
     if new_revision.type != NodeType.SOURCE:


datajunction-server/datajunction_server/api/graphql/scalars/node.py
```diff
@@ -145,7 +145,7 @@ class NodeRevision:
     mode: Optional[NodeMode]  # type: ignore
     description: str = ""
     updated_at: datetime.datetime
-    custom_metadata: Optional[JSON] = strawberry.field(default_factory=dict)
+    custom_metadata: Optional[JSON] = None
```
shangyian, line 148: How come this field defaults to `None` when in the `to_spec` call below it defaults to `{}`? I think the latter might be the cause of the failing test as well

datajunction-ui/src/app/pages/AddEditNodePage/CustomMetadataField.jsx
```diff
@@ -2,59 +2,86 @@
...
+          }
+          try {
+            const parsed = JSON.parse(value);
+
+            if (
+              typeof parsed === 'object' &&
+              parsed !== null &&
+              !Array.isArray(parsed)
+            ) {
+              const keys = Object.keys(parsed);
```
shangyian, line 39: This logic is for detecting duplicate keys right? Should this validation also be in the API layer? (Doesn't need to be in this PR, but seems like the kind of validation we should have server-side)

datajunction-ui/src/app/pages/AddEditNodePage/CustomMetadataField.jsx
```diff
@@ -2,59 +2,86 @@
...
+          }
+        }}
+      >
+        {({ field, form }) => {
+          const errorMessage = form.errors.custom_metadata;
+          const hasError = errorMessage && form.touched.custom_metadata;
+
+          return (
+            <div>
+              <textarea
```
shangyian, line 61: Maybe we should actually use the same editor for custom JSON as we do with SQL queries so that it's easier for users to get the input right. (Doesn't have to be part of this PR, just a thought)

datajunction-server/datajunction_server/api/graphql/scalars/node.py
shangyian: I just pushed a small change for this

samredai: Thanks! I'm not certain what the default should be but when I was testing this out there was a lot of juggling between trying to find out if an empty object `{}` meant the user wanted to _remove_ all the (potentially) existing custom metadata or if they just didn't want it changed. So there were some awkward checks for this that were causing some unexpected behavior like once you set metadata in the UI, then you decide you want to remove all metadata, setting `{}` in the UI wasn't enough because the backend things "this is just the default which means don't change anything".

Anyway, feel free to make any needed updates to this PR to get it through!

datajunction-ui/src/app/pages/AddEditNodePage/CustomMetadataField.jsx
samredai: Makes sense!

PR #1549, opened 2025-10-29
MAX_BY and MIN_BY functions cannot be decomposed

shangyian:
### Summary

MAX_BY/MIN_BY functions cannot be pre-aggregated because they depend on finding the row with the maximum/minimum ordering value across the entire dataset. Pre-aggregating would lose the ordering relationship between measure and ordering values, making it impossible to correctly reconstruct the result.

For example, `MAX_BY(clicked, dateint)` must be computed on the full dataset to find the 'clicked' value corresponding to the row with the highest 'dateint'.

### Test Plan

- [ ] `make check` passes


agorajek (approved):
That looks good. We should not be decomposing these two agg functions.

PR #1571, opened 2025-11-24
Bug saving invalid metric

samredai:
### Summary

This fixes a bug that someone found about a month ago where metric validation doesn't catch when the upstream columns don't exist. I narrowed this down to the fact that we rely on the type inference system to detect missing upstream columns but we short-circuit the type inference when a function is used with a known return type (which is always the case for a metric, i.e. `sum(revenue)`).

I just added an explicit validation for metric node types in the standard node validation logic but we should probably revisit the type inference column validation to see if there's anything we can do at a lower level.

### Test Plan

Added a test

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/validation.py
+++ b/datajunction-server/datajunction_server/internal/validation.py
@@ -158,7 +158,7 @@ async def validate_node_data(
         column_mapping = {}  # pragma: no cover
     node_validator.columns = []
     type_inference_failures = {}
-    for idx, col in enumerate(query_ast.select.projection):
+    for idx, col in enumerate(query_ast.select.projection):  # type: ignore
         column = None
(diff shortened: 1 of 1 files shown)


datajunction-server/datajunction_server/internal/validation.py
```diff
@@ -108,6 +108,30 @@ async def validate_node_data(
     # Add aliases for any unnamed columns and confirm that all column types can be inferred
     query_ast.select.add_aliases_to_unnamed_columns()
 
+    if validated_node.type == NodeType.METRIC and node_validator.dependencies_map:
+        all_available_columns = set()
```
shangyian: minor: could this just be a set comprehension?

samredai: Good catch! Updated

PR #1622, opened 2025-12-22
Speed up SQL generation

shangyian:
### Summary

**Dimension Path Loading**: When building dimension joins, the code used BFS traversal with individual database queries at each step. For a query with 19 dimensions and 3-hop depth, this meant ~57 separate DB queries. After this change, we're now running a single recursive CTE query to find all dimension join paths at once.

**Batch Loading of ORM Objects**

This change batch loads all required `DimensionLink` and their associated `Node` objects in a single query with eager loading, eliminating expensive query patterns where each object was loaded individually, for each link in the path.

We also reduce the number of times `get_dj_node` gets called for each table reference by introducing a `dependencies_cache`. This is populated early in the build process with a batch load of required nodes.

**Other Changes**
* Renamed `is_ancestor_of` to `has_ancestor`. The old name was confusing, since it actually checked if other was an ancestor of self, not the other way around.
* Added an inline check (`SEPARATOR in name`) to determine if a name could be a DJ node before going to the db.
* Uses dependencies cache instead of always defaulting to a database lookup for access control.

### Benchmarking

Complexity | Before (avg) | After (avg) | Speedup
-- | -- | -- | --
Complex (deep deps, multi-dim) | 5763 ms | 1502 ms | 3.8x faster
Complex (long queries) | 5644 ms | 1300 ms | 4.3x faster
Medium (moderate deps) | 1108 ms | 603 ms | 1.8x faster
Simple (few deps, 3 dims) | 357 ms | 371 ms | ~same
Simple (no dims) | 134 ms | 148 ms | ~same

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v2.py
+++ b/datajunction-server/datajunction_server/construction/build_v2.py
@@ -14,10 +14,14 @@ from typing import (
     cast,
 )
 
+from sqlalchemy import text, bindparam, select
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import joinedload, selectinload
 
 from datajunction_server.construction.utils import to_namespaced_name
 from datajunction_server.database import Engine
+from datajunction_server.database.attributetype import ColumnAttribute
+from datajunction_server.database.column import Column
 from datajunction_server.database.dimensionlink import DimensionLink
 from datajunction_server.database.node import Node, NodeRevision
 from datajunction_server.errors import (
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/construction/build_v2.py
```diff
@@ -678,8 +888,10 @@ async def build_dimension_node_joins(self, node_ast, node_alias):
         node_ast = node_ast.to_cte(node_alias)
         self.cte_mapping[self.node_revision.name] = node_ast
 
-        # Find all dimension node joins necessary for the requested dimensions and filters
-        dimension_node_joins = await self.find_dimension_node_joins()
+        # OPTIMIZATION: Find all dimension join paths in ONE recursive CTE query,
```
shangyian: note: this is O(2) DB queries instead of O(dimensions * depth) queries.

Issue #1775, opened 2026-02-19
YAML deploy of derived metrics failed when the namespace is new


anhqle:
Consider the following derived metric, `namespace.derived`, which is defined as `namespace.metric * 2`

There are 2 dependencies: `namespace.metric` and `namespace` (where `namespace` is speculatively a dependency because of the `rsplit` code to accommodate the case where it's a dimension node)

Now DJ will check if any of these dependencies is unsatisfied

If `namespace.metric` is NOT part of the deployment: 
- `namespace.metric` -> found it in external system, satisfied
- `namespace` -> is the start of `namespace.metric`, satisfied

If `namespace.metric` is part of the deployment: 
- `namespace.metric` -> satisfied
- `namespace` -> unsatisfied (but DJ is wrong) 

So we have a bug where `namespace` is wrongly declared an unsatisfied dependency

PR #1778, opened 2026-02-20
Fix nested namespace validation in deployments

shangyian:
### Summary

When deploying nodes with nested namespace structures (e.g., `analytics.dimension.user_type`, `analytics.metric.user_count`), DJ incorrectly treated namespace prefixes as missing dependencies.

**Example failure scenario:**
  - Deploy `analytics.dimension.user_type` (a dimension node) for the first time
  - Cube references `${prefix}dimension.user_type.user_type_id` (full dimension ref)
  - DJ validation extracts parent namespace `analytics.dimension`
  - DJ only checks database nodes to verify it's a namespace prefix, and on first deployment, the database is empty, so it fails with "missing dependency: analytics.dimension"

The solution is to modify dependency validation in `orchestrator.py` to check both database nodes and current deployment nodes when validating namespace prefixes.

### Test Plan

- [ ] PR has an associated issue: #1775 
- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -2914,8 +2914,6 @@ async def hard_delete_node(
 
     await session.delete(node)
     await session.commit()
-    # Expire all objects to clear stale references to deleted entities
-    session.expire_all()
     impact = []  # Aggregate all impact of this deletion to include in response
 
     # Revalidate all downstream nodes


anhqle:
Thanks for the fix! Thinking more carefully about the bug, I think the problem is not so much the fact that the namespace is nested, but that it is new and non-existent in database.

Your fix should work though, just want to clarify the nature of the bug.

PR #1781, opened 2026-02-21
Add a fix where the namespace is classified as a phantom dependency, which fails deployment

anhqle:
### Test Plan

Added a test

- [ ] `make check` passes


shangyian:
Was this for a separate issue or the same as #1775?

PR #1785, opened 2026-02-24
Few changes and fixes to cube's Materialization.

agorajek:
### Summary

Two main changes:
1. Don't bump node version when adding/refreshing materialization.
2. Add a new end point for materialization refreshes. Not critical, but nice to have.

### Test Plan

Unit tests plus by hand locally.

- [x] `make check` passes

### Deployment Plan
Auto.

### DIFF
--- a/datajunction-server/datajunction_server/query_clients/base.py
+++ b/datajunction-server/datajunction_server/query_clients/base.py
@@ -230,6 +230,32 @@ class BaseQueryServiceClient(ABC):
             f"{self.__class__.__name__} does not support cube backfill",
         )
 
+    def refresh_cube_materialization(
+        self,
+        cube_name: str,
+        cube_version: Optional[str] = None,
+        materializations: Optional[List[Dict[str, Any]]] = None,
+        request_headers: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1297,6 +1297,52 @@ async def update_cube_node(
...
+            await session.commit()
+
+            # Serialize active materializations as complete, ready-to-use dicts
+            # so dj-query can parse them directly into model inputs without merging.
+            # This avoids a callback from query service back to DJ.
+            active_mats = []
+            for mat in node_revision.materializations:
+                if mat.deactivated_at:
+                    continue  # pragma: no cover
+                mat_dict = {
```
shangyian, line 1328: Isn't there something that does this already that's used by the /cubes endpoint? Could that be reused?

agorajek: Oh let me check...

agorajek: Unfortunately these materialization configs are different so using the same pydantic class is not working well. This is a different path.

PR #1788, opened 2026-02-24
Get DJ client to not exit 0 even when deployment fails

anhqle:
Right now `dj push` still exits 0 even when the server rejects the deployment, causing CI to mistakenly show up as green

### DIFF
--- a/datajunction-clients/python/tests/examples/deploy0/roads/contractor.yaml
+++ b/datajunction-clients/python/tests/examples/deploy0/roads/contractor.yaml
@@ -19,6 +19,7 @@ dimension_links:
   - type: join
     node_column: state
     dimension_node: ${prefix}roads.us_state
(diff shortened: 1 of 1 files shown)


datajunction-clients/python/tests/examples/deploy0/roads/contractor.yaml
```diff
@@ -19,6 +19,7 @@ dimension_links:
   - type: join
     node_column: state
     dimension_node: ${prefix}roads.us_state
+    join_on: ${prefix}roads.us_state.state_abbr = ${prefix}roads.contractor.state
```
anhqle, line 22: this fixture was broken, but deployment didn't raise before this PR

shangyian (approved):
LGTM, thanks!

PR #2034, opened 2026-04-20
Fix detached-instance crash on GraphQL `dimensionLinks.foreignKeys`

shangyian:
### Summary

Querying `findNodes { current { dimensionLinks { foreignKeys } } }` was failing with a SQLAlchemy detached-instance error introduced by the per-resolver-session refactor (#2028). The `dimension_links` resolver was returning raw ORM `DimensionLink` objects and by the time Strawberry resolved the foreignKeys field, the session that loaded those objects had closed, so the lazy load of `foreign_keys` blew up.
```
Parent instance <DimensionLink at 0x7f200f9a2bd0> is not bound to a Session; lazy load operation of attribute 'node_revision' cannot proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)
```

The fix is for the `dimension_links` resolver to turn each link into a Strawberry `DimensionLink` object, including foreign_keys, so that every attribute is realized while the parent resolver's session is still open and nothing lazy crosses the session boundary.

Other resolvers in scalars/node.py that walk nested lazy relationships (e.g., availability.partitions, materializations.backfills, metric_metadata.unit) should be audited for similar errors. The general rule is that any property reaching through a non-eager-loaded relationship is now an issue under the session-per-resolver pattern.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/graphql/scalars/node.py
+++ b/datajunction-server/datajunction_server/api/graphql/scalars/node.py
@@ -223,29 +223,19 @@ class NodeRevision:
         Returns the dimension links for this node revision.
 
         By the time this resolver runs, the parent's session has closed and
-        each DimensionLink is detached from it. The `foreign_keys`
-        hybrid_property walks `link.node_revision.name` — a lazy relationship
-        that would fail on a detached instance. We already have the owning
-        NodeRevision (`self`), so pre-seed `link.node_revision` via
-        `set_committed_value` to short-circuit the lazy load, then compute
-        `foreign_keys` while we still control the call site.
+        each DimensionLink is detached. The `foreign_keys` hybrid_property
+        walks `link.node_revision.name` — a lazy relationship that would fail
+        on a detached instance. We already have the owning NodeRevision
+        (`self`), so pre-seed `link.node_revision` via `set_committed_value`
+        to short-circuit the lazy load. Strawberry duck-types the raw ORM
+        link the rest of the way.
         """
         links = []
         for link in self.dimension_links:
             if link.dimension is None or link.dimension.deactivated_at is not None:
                 continue  # hard-deleted or deactivated dimension node
             set_committed_value(link, "node_revision", self)
-            links.append(
-                DimensionLink(  # type: ignore[call-arg]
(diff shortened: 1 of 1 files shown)


datajunction-server/datajunction_server/api/graphql/scalars/node.py
```diff
@@ -221,14 +221,32 @@ def columns(
...
+                    dimension=NodeName(name=link.dimension.name),  # type: ignore[call-arg]
+                    join_type=link.join_type or JoinType_.LEFT,
+                    join_sql=link.join_sql,
+                    join_cardinality=link.join_cardinality,
+                    role=link.role,
+                    foreign_keys=link.foreign_keys,
+                    default_value=link.default_value,
+                ),
+            )
+        return links
```
jlhester: Should `self.dimension_links` be updated to be `links` at the end of the method here? What's the point of completely rebuilding each dimension link object?

datajunction-server/datajunction_server/api/graphql/scalars/node.py
```diff
@@ -221,14 +221,32 @@ def columns(
...
-            and link.dimension.deactivated_at
-            is None  # handles deactivated dimension nodes
-        ]
+        links = []
+        for link in self.dimension_links:
+            if link.dimension is None or link.dimension.deactivated_at is not None:
+                continue  # hard-deleted or deactivated dimension node
+            set_committed_value(link, "node_revision", self)
+            links.append(
+                DimensionLink(  # type: ignore[call-arg]
```
jlhester: This could be fragile if `DimensionLink` is ever updated to include more fields

shangyian: Ah yes good point... this is actually unnecessary now that we're returning a DimensionLink ORM obj with `node_revision` pre-seeded so `foreign_keys` can resolve without a lazy load.

datajunction-server/datajunction_server/api/graphql/scalars/node.py
shangyian: I ended up completely simplifying this ... the only thing that's needed is the pre-seeding of the node revision relationship

PR #2035, opened 2026-04-20
Add a default server config for non-source nodes.

### LINKED ISSUE #1913: Server-Configured Default Catalog
There should be a server-configured default catalog that gets used when a new node's catalog cannot be determined based on its upstreams (it may have no upstreams).

jlhester:
### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
+++ b/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
@@ -884,6 +884,14 @@ class DeploymentOrchestrator:
             for node_spec in self.deployment_spec.nodes
             if node_spec.node_type == NodeType.SOURCE and node_spec.catalog
         }
+
+        # Also validate that configured default catalogs exist
+        if self.deployment_spec.default_catalog:
+            spec_catalog_names.add(self.deployment_spec.default_catalog)
+        settings = get_settings()
+        if settings.seed_setup.default_catalog_name:
+            spec_catalog_names.add(settings.seed_setup.default_catalog_name)
+
         missing_catalogs = spec_catalog_names - all_catalogs_map.keys()
         if missing_catalogs:
             self.errors.append(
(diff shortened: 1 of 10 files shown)


datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -2916,12 +2916,16 @@ async def create_nodes_from_validation(
...
+        Priority: deployment spec default_catalog > server default_catalog_name.
         Used when catalog can't be derived from parent nodes (e.g., invalid nodes
         whose parents are unresolvable). Callers should further fall back to the
         virtual catalog when this returns None.
         """
         if self.deployment_spec.default_catalog:
             return self.registry.catalogs.get(self.deployment_spec.default_catalog)
+        settings = get_settings()
+        if settings.seed_setup.default_catalog_name:
+            return self.registry.catalogs.get(settings.seed_setup.default_catalog_name)
```
shangyian, line 2928: Should we make sure that the registry always loads up the default catalog in the catalogs bulk load?

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -1894,6 +1900,11 @@ async def create_new_revision_from_existing(
         ]
         if catalogs:
             new_revision.catalog_id = catalogs[0]
+        else:
+            default_catalog = await Catalog.get_default_catalog(session)
```
shangyian, line 1904: Should this use the same pattern as above (lines 542-547), i.e., "use the default catalog or fallback to virtual catalog"?

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
jlhester: When you say "loads up", do you mean actually create the default catalog if it doesn't exist, or just check if it exists during the bulk load?

shangyian: Oh, I meant just check if it exists during bulk load. I think an explicit creation step is probably cleaner than just auto-creating.

shangyian (approved):
LGTM, 🚢

jlhester:
One of the tests failed initially with a sqlalchemy distinct key conflict, but rerunning fixed it :grimacing:. If it pops up again I'll capture it in a ticket.

PR #2230, opened 2026-06-05
[RBAC PR 1] Enforce MANAGE permission on RBAC and group admin APIs

philipfweiss:
Tracking: #2234 (step 1 of the RBAC enablement sequence).

The RBAC engine already exists, but every role/group admin endpoint was auth-only, so any logged-in user could self-grant any role. That makes restrictive access policies unenforceable.

This PR enforces the `MANAGE` action on those endpoints:

- **Roles** (`api/rbac.py`): create/update/delete, add/remove scope, assign/revoke now require `MANAGE` on the resources the scope targets. You can only grant access you already manage (prevents privilege escalation).
- **Groups** (`api/groups.py`): register and membership changes require a global `MANAGE` grant.
- Adds `AccessChecker.add_scope()` helper.
---
Why this won't take effect now: DJ’s default access policy is still permissive, so the new MANAGE checks allow requests when no explicit restrictive grant/policy denies them.

When this will start taking effect: Once a deployment enables restrictive RBAC. Since this isn't enabled today, this is a no-op. Once we seed roles, this change will become relevant. 

---
Verification:

1. Started local DJ server
```bash
./dev.sh up -d
DJ  : http://localhost:8000
UI  : http://localhost:3000
GET /health/
attempt=1 status=200 body=[{"name":"database","status":"ok"}]
```

2. Enabled RBAC restrictive setting
```yaml
authorization_provider= rbac
default_access_policy= restrictive
```

3. Seeded one allowed grant for the local dj user: MANAGE on finance.*. Also seeded growth roles for denied-path tests.

```
name                         | action | scope_type | scope_value | assigned_to
manual-finance-manager       | MANAGE | NAMESPACE  | finance.*   | dj
manual-growth-assigned-role  | READ   | NAMESPACE  | growth.*    | dj
manual-growth-role           | READ   | NAMESPACE  | growth.*    |
```

4. Logged in as non-admin user `dj`
```
POST /basic/login/ -> HTTP_STATUS:200
GET /whoami/ -> {"username":"dj","is_admin":false} HTTP_STATUS:200
```

5. Verified role creation requires MANAGE on the role scopes.
```
POST /roles/ finance.* role -> HTTP_STATUS:201
POST /roles/ growth.* role -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
```

6. Verified adding, deleting, role metadata, assigning roles, revoking roles, deleting roles all require `MANAGE`

note that POST here is shorthand for the actual payload, something like
```
POST /roles/
{
  "name": "manual-finance-reader",
  "description": "Manual PR 2230 allowed role",
  "scopes": [
    {
      "action": "read",
      "scope_type": "namespace",
      "scope_value": "finance.*"
    }
  ]
}
```

```
POST /roles/manual-empty-role/scopes/ finance.* -> HTTP_STATUS:201
POST /roles/manual-empty-role/scopes/ growth.* -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
...
DELETE /roles/manual-empty-role/scopes/write/namespace/finance.* -> HTTP_STATUS:204
DELETE /roles/manual-growth-role/scopes/read/namespace/growth.* -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
...
POST /roles/manual-finance-reader/assign -> HTTP_STATUS:201
POST /roles/manual-growth-role/assign -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
...
DELETE /roles/manual-finance-reader/assignments/dj -> HTTP_STATUS:204
DELETE /roles/manual-growth-assigned-role/assignments/dj -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
...
DELETE /roles/manual-finance-reader -> HTTP_STATUS:204
DELETE /roles/manual-growth-role -> {"message":"Access denied to 1 resource(s): growth.*"} HTTP_STATUS:403
```


datajunction-server/datajunction_server/api/rbac.py
```diff
@@ -34,6 +39,26 @@
...
+    scopes,
+) -> None:
+    """
+    Require MANAGE on every resource the given scopes target.
+
+    A principal may only define, hand out, or revoke a permission it can already
+    manage on the underlying resource, which prevents privilege escalation
+    through role creation or assignment.
+    """
+    for scope in scopes:
```
shangyian, line 61: For this setup, if scopes is `[]`, the function adds nothing to `access_checker`, so does that mean anyone can manage a scope-less role? For an empty scopes list, should this require some baseline `MANAGE` grant rather than falling through as a no-op? Something like requiring MANAGE on `*` when scopes is empty, analogous to how enforce_group_administration works.

datajunction-server/datajunction_server/api/rbac.py
```diff
@@ -417,6 +455,9 @@ async def delete_scope_from_role(
         include_deleted=False,
     )
 
+    access_checker.add_scope(scope_type, scope_value, ResourceAction.MANAGE)
```
shangyian, line 466: Should this just call `enforce_scope_management`?

datajunction-server/tests/api/groups_test.py
```diff
@@ -5,6 +5,29 @@
...
+from datajunction_server.internal.access.authorization import AuthorizationService
+from datajunction_server.models import access
+
+VALIDATOR_AUTH_SERVICE = (
+    "datajunction_server.internal.access.authorization."
+    "validator.get_authorization_service"
+)
+
+
+class DenyManageAuthorizationService(AuthorizationService):
```
shangyian, line 17: It looks like this `DenyManageAuthorizationService` gets repeated in both groups_test.py and rbac_test.py. Could this be moved to a new `tests/api/conftest.py` so the tests can share?

shangyian (approved):

datajunction-server/tests/api/groups_test.py
philipfweiss: Sure! I'll add `tests/api/conftest.py`

datajunction-server/datajunction_server/api/rbac.py
philipfweiss: Good call, sure

datajunction-server/datajunction_server/api/rbac.py
philipfweiss: I thought about the tradeoffs between these approaches- I saw two reasonable paths:

1. Path 1- Empty roles are inert metadata: they grant nothing, and adding scopes later is already guarded by MANAGE on each scope. This keeps namespace managers able to create placeholders and then add scopes they manage.

2. Path 2- Role lifecycle is global RBAC administration: even scope-less roles require baseline/global MANAGE, which is stricter and avoids empty-role clutter, but makes placeholder creation global-admin-only.

I leaned toward 1 because this PR is focused on privilege escalation, and empty roles do not grant access. I didn't realiz `enforce_group_administration` uses the stricter model. 

Anyways, I'm not so opinionated here. I'll swap over to 2.

PR #2231, opened 2026-06-05
Enforce EXECUTE permission on data-running endpoints

philipfweiss:
Tracking: #2234 (step 10, optional, of the RBAC enablement sequence).

`EXECUTE` is part of the RBAC permission hierarchy but was never checked anywhere, so running queries was gated only by build-time `READ` checks (and not at all on cache hits). This adds explicit `EXECUTE` enforcement on the query-running endpoints.

- **Node data:** `GET /data/{node}/` and `GET /stream/{node}` require `EXECUTE` on the node.
- **Metrics data:** `GET /data/` and `GET /stream/` require `EXECUTE` on each requested metric.
- Because `EXECUTE` implies `READ` in the hierarchy, this also closes the cache-hit gap where SQL isn't rebuilt and the build-time `READ` check would be skipped.

Note: this is an **independent, optional gate**. In deployments where the underlying query engine already enforces data access, a DJ-level `EXECUTE` gate is a redundant second plane and can be left permissive. Semantics: a read-only user can view metadata/SQL but cannot run queries. No-op under the default `permissive` policy, so nothing breaks today.

### DIFF
--- a/datajunction-server/datajunction_server/api/data.py
+++ b/datajunction-server/datajunction_server/api/data.py
@@ -63,6 +63,25 @@ settings = get_settings()
 router = SecureAPIRouter(tags=["data"])
 
 
+async def enforce_execute(
+    access_checker: AccessChecker,
+    node_names: List[str],
+) -> None:
+    """
+    Require EXECUTE on every node a query is being run against.
+
+    EXECUTE implies READ in the permission hierarchy, so this also covers the
+    cache-hit path where SQL is not rebuilt and the build-time READ checks would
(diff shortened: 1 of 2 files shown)


philipfweiss:
Closing as not strictly necessary for the RBAC enablement MVP. EXECUTE enforcement is an independent, optional gate on query execution; where the underlying query engine already enforces data access it is redundant. The plan now focuses on metadata/write governance (see #2234). Can revive this branch later if a DJ-level execute gate is wanted.

PR #2232, opened 2026-06-05
[RBAC PR 2] Add admin break-glass bypass to RBAC authorization

philipfweiss:
Tracking: #2234 (step 2 of the RBAC enablement sequence).

The RBAC proposal's first resolution step is "if the user is an admin, allow immediately." That short-circuit didn't exist: `is_admin` was on the user model but never consulted in the authorization path, so there was no break-glass and a deny-by-default flip would lock out admins too.

This adds it:
- Threads `is_admin` into `AuthContext` (from `user.is_admin`).
- `RBACAuthorizationService.authorize` short-circuits for admins, approving all requests with `reason="admin"`.
- The bypass is a single explicit check and is **logged for audit** (username, id, and the requests granted), which also pre-pays for the audit-logging work later.

Below I verified the admin break galss:

---
Verification:

1. Started local DJ server
```
./dev.sh up -d
DJ  : http://localhost:8000
UI  : http://localhost:3000
GET /health/
attempt=1 status=200 body=[{"name":"database","status":"ok"}]
```

2. Enabled RBAC restrictive setting

```
authorization_provider= rbac
default_access_policy= restrictive
```

3. Logged in as `non-admin` user. Then, verified that they're denied. 
```
curl -sS -c /tmp/dj-pr2232.cookies \
  -X POST "http://localhost:8000/basic/login/" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data "username=dj&password=dj"
curl -sS -b /tmp/dj-pr2232.cookies "http://localhost:8000/whoami/"
```
```
curl -sS -b /tmp/dj-pr2232.cookies \
  -X POST "http://localhost:8000/nodes/source/" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "manual_admin_bypass.orders_non_admin",
    "catalog": "default",
    "schema": "public",
    "table": "orders_non_admin",
    "columns": [
      {"name": "id", "type": "int"}
    ]
  }'
```

4. Marked same user as admin
```
docker exec dj-s0-postgres psql -U dj -d dj \
  -c "update users set is_admin = true where username = 'dj';"
```

```
GET /whoami/ -> {"username":"dj","is_admin":true} HTTP_STATUS:200
```

5. Verified bypass works
```
curl -sS -b /tmp/dj-pr2232.cookies \
  -X POST "http://localhost:8000/nodes/source/" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "manual_admin_bypass.orders_admin",
[...]

PR #2233, opened 2026-06-05
[RBAC PR 3] Add configurable default-access role to RBAC

philipfweiss:
Tracking: #2234 (step 3 of the RBAC enablement sequence).

Stacked on #2232.

With `default_access_policy=restrictive`, every reques is denied unless an explicit user or group grant matches. That leaves no way to express a baseline like "everyone can read" short of flipping the whole policy back to `permissive` or hand-granting every principal.

This adds a configurable default-access role:

* New `default_access_role` setting (default `None`).
* `AuthContext.from_user` loads the named role's scopes into `AuthContext.default_scopes` (empty when the setting is unset or the role does not exist).
* `RBACAuthorizationService` resolves the principal's own scopes and the default-role scopes together; if none grant the request, it falls back to `default_access_policy`.
* No behavior change by default: with no default role configured, authorization is unchanged.

---

How I verifie:

Run against the PR branch with a restrictive RBAC config and the seeded Roads data. The user `dj` is a normal non-admin.

**1. Started the PR server, restrictive policy, no default role.**

```
AUTHORIZATION_PROVIDER=rbac \
DEFAULT_ACCESS_POLICY=restrictive \
uvicorn datajunction_server.api.main:app --port 8010   # DB env omitted

GET /health/ -> HTTP_STATUS:200
```

**2. Logged in as non-admin `dj`; baseline denies both read and write.**

```
POST /basic/login/ -> HTTP_STATUS:200
GET  /whoami/ -> {"username":"dj","is_admin":false} HTTP_STATUS:200
GET  /nodes/default.avg_repair_price/ -> HTTP_STATUS:403
POST /nodes/source/ (default namespace) -> HTTP_STATUS:403
     {"message":"Access denied to 1 resource(s): default"}
```

**3. Created a default-access role that grants read broadly.**

```
POST /roles/
{
  "name": "global-viewer",
  "scopes": [
    {"action": "read", "scope_type": "node", "scope_value": "*"},
    {"action": "read", "scope_type": "namespace", "scope_value": "*"}
  ]
}
-> HTTP_STATUS:201
```

**4. Restarted the server with the role wired in as the default.**

```
AUTHORIZATION_PROVIDER=rbac \
DEFAULT_ACCESS_POLICY=restrictive \
DEFAULT_ACCESS_ROLE=global-viewer \
uvicorn datajunction_server.api.main:app --port 8010   # DB env omitted
```

**5. Same non-admin user now gets fallback read, while writes stay denied.**

```
POST /basic/login/ -> HTTP_STATUS:200
GET  /nodes/default.avg_repair_price/ -> HTTP_STATUS:200
GET  /nodes/default.repair_orders/    -> HTTP_STATUS:200
GET  /nodes/default.contractors/      -> HTTP_STATUS:200
POST /nodes/source/ (default namespace) -> HTTP_STATUS:403
     {"message":"Access denied to 1 resource(s): default"}
```

The read flips from `403` to `200` once the default role is configured; the write stays `403` because the role grants only `READ`. Unsetting `default_access_role`, or pointing it at a role that does not exist, loads no scopes and falls back to `default_access_policy`.


shangyian (approved):
LGTM 🚢

datajunction-server/datajunction_server/internal/access/authorization/service.py
```diff
@@ -121,25 +127,70 @@ def authorize(
...
         request: ResourceRequest,
     ) -> AccessDecision:
         """
         Convert ResourceRequest to AccessDecision.
+
+        Gathers all applicable scopes (explicit grants + default-access role)
+        and approves if any grants the request. Otherwise falls back to the
+        configured default_access_policy.
         """
-        has_grant = self.has_permission(
```
ruizhang0519, line 186: `has_permission` is now orphaned after the removal of this usage here, and it is still heavily tested - should we consider remove the function definition and all related tests?

datajunction-server/datajunction_server/internal/access/authorization/context.py
```diff
@@ -70,14 +74,35 @@ async def from_user(
...
+        Load the scopes of the configured default-access role, if any.
+
+        Returns an empty list when no default role is configured or the named
+        role does not exist, so authorization simply falls through to the
+        default_access_policy.
+        """
+        role_name = settings.default_access_role
+        if not role_name:
+            return []
+        default_role = await Role.get_by_name(session, role_name)
```
ruizhang0519, line 103: Default role is re-queried **on every request**. That's an extra DB round-trip per request for a value that changes rarely. Does this present a caching opportunity (TTL cache keyed on role name)?

datajunction-server/datajunction_server/internal/access/authorization/service.py
```diff
@@ -121,25 +127,70 @@ def authorize(
...
-            resource_type=request.access_object.resource_type,
-            resource_name=request.access_object.name,
+        granted = any(
+            self._scope_grants_permission(
+                scope,
+                request.verb,
+                request.access_object.resource_type,
+                request.access_object.name,
+            )
+            for scope in self.candidate_scopes(auth_context)
```
ruizhang0519: `candidate_scopes` is recomputed per request. For an endpoint checking many nodes, that rebuilds the same list N times. The pre-existing code had the same shape, so no regression, but the refactor was a natural spot to hoist it once per `authorize()` call.

datajunction-server/datajunction_server/internal/access/authorization/service.py
philipfweiss: It's a public classmethod other providers can use. I actually need it in #2239 - `_make_explicit_grant_decision`

datajunction-server/datajunction_server/internal/access/authorization/service.py
philipfweiss: Good catch!

datajunction-server/datajunction_server/internal/access/authorization/context.py
philipfweiss: So I played around with this- I feel this might be a pre-mature optimization, and it trades immediate correctness for a staleness window on a security-relevant fallback. My preference is to keep as-is and monitor for performance issues.

ruizhang0519 (approved):

PR #2252, opened 2026-06-11
GraphQL API perf optimizations

peferron:
This PR optimizes the performance of fetching cubes via GraphQL.

* Extends the existing optimization for `name` to sibling fields. That's the largest gain; around 2x faster when running locally against a copy of our database. I didn't measure a significant perf regression for name-only use cases.
* Adds an index to `node.type`, which is useful when fetching cubes since they only represent a small fraction of the overall nodes.
* Avoids a DB round-trip & materializing a potentially long list of node IDs by using an `EXISTS` subquery instead, which the Postgres planner will optimize as needed.
* Remove an apparently obsolete `LIKE '%_DOT_%` filter which couldn't use indexes—I'll need confirmation that this is indeed safe to remove.

Also adds support for `@include` and `@skip` directives; before clients would have to alter the query string itself to get the expected perf gains, which is inconvenient.

Finally, fixes an existing bug (happening on `main`) where metrics and dimensions within revisions are empty unless non-`name` fields are also selected. For example, the query below returns empty metrics and dimensions, unless you uncomment some of the fields:

```
{
  findNodes(nodeTypes: [CUBE], limit: 1) {
    revisions {
      cubeMetrics {
        name
        # columns {
        #   name
        # }
      }
      cubeDimensions {
        name
        # attribute
      }
    }
  }
}
```

### DIFF
--- a/datajunction-server/datajunction_server/alembic/versions/2026_06_11_1840-3c7a9f1b2e4d_add_index_on_node_type.py
+++ b/datajunction-server/datajunction_server/alembic/versions/2026_06_11_1840-3c7a9f1b2e4d_add_index_on_node_type.py
@@ -1,0 +1,27 @@
+"""add index on node type
+
+Revision ID: 3c7a9f1b2e4d
+Revises: 7e1b9c4a2d3f
+Create Date: 2026-06-11 18:40:18.157912+00:00
+
+"""
+# pylint: disable=no-member, invalid-name, missing-function-docstring, unused-import, no-name-in-module
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "3c7a9f1b2e4d"
+down_revision = "7e1b9c4a2d3f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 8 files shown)


shangyian (commented):
Thanks for adding this @peferron! 

>Remove an apparently obsolete LIKE '%_DOT_% filter which couldn't use indexes—I'll need confirmation that this is indeed safe to remove.

I just ran some confirmation queries in Postgres and this looks like a reasonable assumption: 
```
dj=> SELECT cr.cube_id, col.name, col.node_revision_id, nr.type
FROM cube cr
JOIN "column" col       ON col.id = cr.cube_element_id
LEFT JOIN noderevision nr ON nr.id = col.node_revision_id
WHERE (col.name LIKE '%\_DOT\_%' ESCAPE '\')
   <> (col.node_revision_id IS NOT NULL AND nr.type = 'METRIC')
LIMIT 50;
 cube_id | name | node_revision_id | type
---------+------+------------------+------
(0 rows)
```

shangyian (approved):
This looks good to me, thanks for adding!

PR #2263, opened 2026-06-20
Add windowed lookback for trailing-window metrics

shangyian:
### Summary

Implements live-path SQL generation for metrics that use bounded trailing-window aggregations (e.g., `SUM(revenue) OVER (ORDER BY date ROWS BETWEEN 6 PRECEDING AND CURRENT ROW)`).

When you query a trailing-window metric with a date filter, DJ now:

1. Expands the fact-table scan backward by N periods so all rows needed by the window frame are present (e.g., a 7-day trailing sum filtered to Jan 31 scans Jan 25–31).
2. Densifies the date spine over the expanded range via a `LEFT JOIN` against the sequence dimension, zero-filling gaps so `ROWS BETWEEN N PRECEDING` counts exactly N positions regardless of missing source rows.
3. Applies the window function over the dense pre-aggregated series in a `__windowed` CTE.
4. Restricts the output back to the originally requested date range so callers see only the dates they asked for.

The scan bounds, densification, and output restriction are all emitted as named CTEs (`__wl_low_date_id`, `__windowed`) rather than inline subqueries, which makes the SQL readable and lets engines optimize CTE materialization.

**Supported aggregations**

Trailing-window metrics backed by fully-additive measures (`SUM`, `COUNT`) and native window functions (`AVG`, `MIN`, `MAX`) are supported. Metrics backed by `COUNT(DISTINCT ...)` on the live path are rejected with an error (DJ pre-aggregates to one row per grain before applying the window, which would double-count entities appearing on multiple days). `LAG`/`LEAD`-based period-over-period metrics are unaffected and pass through unchanged.

**Multiple windows and cross-fact queries**

When a single query mixes metrics with different window extents (e.g., 7-day and 28-day rolling), DJ takes the union of the required lookback and densifies once over the widest range. Cross-fact queries (metrics from different source nodes) are densified jointly over a shared date spine.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
@@ -542,6 +542,140 @@ def apply_live_window_lookback(ctx: "BuildContext") -> None:
             (output_col_name, output_restriction),
         )
 
+        # Record a densification spec so the per-date grain relation the window
+        # reads (the ``base_metrics`` CTE) can be rewritten into a DENSE series:
+        # the existing per-date aggregation LEFT-joined onto the sequence
+        # dimension's complete date domain over [offset_low, high], with additive
+        # measures 0-filled on gap rows. Without this, ``ROWS BETWEEN N PRECEDING``
+        # counts physical fact rows, so a sparse fact gives wrong trailing values.
+        ctx.live_window_densify_specs.append(
+            (output_col_name, dim_physical, order_col_name, offset_low, high_expr),
+        )
+
+
+def densify_window_base_metrics(
+    result: "ast.Query",
+    ctx: "BuildContext",
+) -> None:
+    """
+    Rewrite the ``base_metrics`` CTE (the per-date grain relation the live
+    window reads from) into a DENSE series so ``ROWS BETWEEN N PRECEDING``
+    counts calendar positions rather than physical fact rows.
+
+    For each densification spec recorded by :func:`apply_live_window_lookback`,
+    the existing ``base_metrics`` SELECT (the per-date aggregation) is wrapped as
+    an inner ``agg`` subquery and LEFT-joined onto a spine -- the sequence
+    dimension's complete date domain over the expanded scan range
+    ``[offset_low, high]``. The order dimension is projected from the spine (so
+    gap dates appear), and every additive measure is wrapped in
+    ``COALESCE(measure, 0)`` so missing dates contribute 0.
+
+    Resulting CTE shape::
+
+        base_metrics AS (
+          SELECT spine.<key> AS <order_col>, COALESCE(agg.<m>, 0) AS <m>, ...
+          FROM (SELECT DISTINCT <key> FROM <dim>
+                WHERE <key> BETWEEN <offset_low> AND <high>) spine
+          LEFT JOIN (<original base_metrics select>) agg
+                 ON spine.<key> = agg.<order_col>
+        )
+
+    No-op when no densification spec was registered.
+    """
+    if not ctx.live_window_densify_specs:
+        return
+
+    # Locate the base_metrics CTE by alias.
+    base_metrics_cte: Optional[ast.Query] = None
+    for cte in result.ctes:
+        if cte.alias is not None and cte.alias.name == "base_metrics":
+            base_metrics_cte = cte
+            break
+    if base_metrics_cte is None:  # pragma: no cover
+        return
+
+    spine_alias = "__spine"
+    agg_alias = "__agg"
+
+    for (
+        output_col_name,
+        spine_physical,
+        spine_key_col,
+        offset_low,
+        high_expr,
+    ) in ctx.live_window_densify_specs:
+        original_select = base_metrics_cte.select
+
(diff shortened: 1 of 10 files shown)


datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
```diff
@@ -0,0 +1,928 @@
...
+        f"SELECT (SELECT MIN(__o.{order_col_name}) FROM ("
+        f"SELECT {order_col_name} FROM {tbl} "
+        f"WHERE {order_col_name} <= {str(lower_expr)} "
+        f"ORDER BY {order_col_name} DESC LIMIT {extent + 1}"
+        f") __o)"
+    )
+    return cast(ast.Expression, parse(sql).select.projection[0])
+
+
+def _build_offset_cte(
```
anaghshineh, line 163: I love the ideas in this PR to address the issue raised on Slack. I wanted to flag a very common use case – partition pruning on the fact scan, particularly time-based partitions (e.g., day). For these cases, I don't think the generated scan will handle partition pruning well, which could be problematic for large fact tables and wider windows. I tried a query in this shape against one of our tables (names scrubbed):

```
SELECT DISTINCT f.partition_date
FROM catalog.schema.fact_table f
WHERE f.partition_date BETWEEN (
  SELECT MIN(d.date_key)
  FROM (
    SELECT date_key
    FROM catalog.schema.date_dim
    WHERE date_key <= 20240131
    ORDER BY date_key DESC
    LIMIT 7
  ) d
) AND 20240131
```

`EXPLAIN` shows the fact scan only gets the upper bound pushed down. The lower bound ends up as a post-scan filter (partition_date >= Subquery ...), not a partition predicate. So I think we'd scan all partitions through the upper bound, not just those in the window.

The ranked subquery approach makes sense when the order dimension isn't a standard temporal dim with known granularity/format.

I wonder if a hybrid approach could work... keep the ranked subquery as the fallback, but when the order dim is a standard temporal dim with known granularity and literal filter bounds, compute the lower bound at compile time with date arithmetic instead. For example, on Spark with a daily date_key:

```
WHERE f.partition_date BETWEEN
  CAST(DATE_FORMAT(DATE_ADD(TO_DATE(CAST(20240131 AS STRING), 'yyyyMMdd'), -6), 'yyyyMMdd') AS INT)
  AND 20240131
```

Spark can evaluate that lower bound at plan time and turn it into a literal, so both sides of the BETWEEN can participate in partition pruning. We'd still want the ranked subquery when calendar days aren't contiguous (business days, sparse dims, etc.), right?

Not really a correctness concern on my side, but a performance one. WDYT?

datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
```diff
@@ -0,0 +1,928 @@
...
+
+from datajunction_server.errors import DJInvalidInputException
+from datajunction_server.models.node import NodeType
+from datajunction_server.sql.parsing import ast
+from datajunction_server.sql.parsing.backends.antlr4 import parse
+
+if TYPE_CHECKING:
+    from datajunction_server.construction.build_v3.types import BuildContext
+
+ADDITIVE_WINDOW_AGGS = frozenset({"SUM", "COUNT"})
```
anaghshineh: Nice! Do you think the validation / contract is strong enough here? If I'm following fully, this would actually allow people to pass `COUNT(DISTINCT ...)` as part of the window metric definition? Hopefully people are aware of the dangers there. Worth rejecting explicitly? Also, what about guarding against parent / base metrics that have limited aggregability? Like if the window metric tries to do a `SUM(<base_metric>) OVER ...` where `<base_metric>` is `COUNT(DISTINCT ...)`?

datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
shangyian: Yeah, I think the hybrid makes sense. When the order dim is a standard temporal dim with known granularity/format and literal filter bounds, we can compute the lower bound with compile-time date arithmetic so both sides of the `BETWEEN` are literals the engine can prune on. Then fall back to the ranked subquery otherwise. Tracking as a follow-up rather than blocking this PR, since it's an optimization

datajunction-server/datajunction_server/construction/build_v3/window_lookback.py
shangyian: Good point, I added a guard for this after your comment -- the build now raises an invalid input exception for any window metric that isn't fully decomposable. Basically there is a decomposition process for each metric, and each decomposed component is assigned an aggregability level (`FULL`, `LIMITED`, or `NONE`), and `DISTINCT` aggregations get tagged `LIMITED` during decomposition. So a direct `COUNT(DISTINCT ...) OVER ...` is rejected outright.

The nested `SUM(<base_metric>) OVER ...` where the base is a `COUNT(DISTINCT)` trips the same guard, since decomposition flattens the referenced metric to its leaf components and the `DISTINCT` leaf would be marked `LIMITED`. The rejection is on that aggregability assignment rather than function name.

PR #2318, opened 2026-07-08
Add semantic-layer REST API exposing cubes as views

robinld:
Implements part of the spec from https://github.com/betodealmeida/sl-db-engine-spec/blob/5f41b523975b2547ba73fb606a0b69615fc6967e/server/SPEC.md

Expose /semantic/* so external clients can browse cubes as views and get physical SQL to execute. SQL generation reuses DJ's canonical metrics-SQL path via a shared generate_metrics_sql helper; dialect resolution is now pin-aware so a pinned cube's dialect matches the cube it builds from.

### DIFF
--- a/datajunction-server/datajunction_server/api/semantic_layer.py
+++ b/datajunction-server/datajunction_server/api/semantic_layer.py
@@ -1,0 +1,386 @@
+"""
+Semantic-layer REST API implementing the sl-db-engine-spec over DataJunction
+cubes (each cube = a "view").
+
+Spec: https://github.com/betodealmeida/sl-db-engine-spec/blob/main/server/SPEC.md
+
+DJ generates SQL; the client executes it. This router does NOT run queries —
+``/views/{view}/sql`` returns physical SQL + dialect for the client to run;
+``/views/list`` and ``/views/{view}`` are metadata. Mounted at ``/semantic``.
+"""
+
+import logging
+from typing import Any, Optional, Union
+
+from fastapi import Depends
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel, Field  # pylint: disable=no-name-in-module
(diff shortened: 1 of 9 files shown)


datajunction-server/datajunction_server/api/sql.py
shangyian: Should this logging move into `generate_metrics_sql` so that it continues to match the existing functionality in terms of logging `merged_filters` rather than unmerged `filters`?

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def _quote_value(value: Any) -> str:
+    """Render a scalar Python value as a SQL literal via sqlglot, which handles
+    quote-escaping and numeric/bool/NULL rendering for the scalars the client
+    sends (strings, numbers, bools, null)."""
+    # Imported lazily so the optional ``sqlglot`` extra isn't required at app
+    # import time — the rest of the codebase keeps sqlglot off the eager import
+    # path (e.g. transpilation.py imports it inside functions).
+    import sqlglot  # noqa: PLC0415
```
shangyian, line 121: Is it worth raising an error with an actionable message here if this import fails? 

Realistically we should probably just make sqlglot a required dependency, since we do rely on it for SQL transpilation and don't have any realistic fallbacks (but that's orthogonal to this PR!)

shangyian (approved):
Generally LGTM! 🚢 Only left some minor comments

datajunction-server/datajunction_server/api/semantic_layer.py
robinld: I'll make a separate PR to address the sqlglot dependency

PR #2331, opened 2026-07-14
Fix node column resolution for role playing dimensions

israelhill:
# Fix role-playing dimension column resolution in column-scoped endpoints

## Problem

get_node_column — the shared resolver behind the set-partition, remove-partition, and set-attributes endpoints — resolves columns strictly by bare name, with no awareness of the role suffix (dimension_column). For a node with two columns sharing a bare name but different roles (e.g. a date dim reached via two FKs), this caused two bugs:

- Silent wrong-column: a bare-name request bound the operation to whichever role sorted last in a dict comprehension — no error, silently corrupting materialization config.
- Can't target a role: a role-qualified name (…dateint[epoch_date]) matched nothing → 404, so there was no way to pin the operation to a specific role.

Surfaced on cube arc.content_migration.cade_batch_funnel, where a UI materialize intending [epoch_date] silently landed the partition on [region_date].

## Fix

Resolve by the role-qualified identity (name + dimension_column) first, fall back to bare name only when unambiguous, and raise a 422 listing the role-qualified options when a bare name is ambiguous — instead of silently picking one. Fixing the shared resolver fixes set-partition, remove-partition, and set-attributes at once. No endpoint signature or client changes.

## Tests

- Unit (get_node_column): role-qualified resolves to the right role; ambiguous bare name raises listing both options; unique bare name still works; unknown column still 404s with the original message.
- API: role-qualified POST/DELETE …/partition targets one role without touching the other; bare ambiguous POST returns 422 and sets nothing.
- Existing partition/attributes/materialize tests still pass (backward compat)

### DIFF
--- a/datajunction-server/datajunction_server/internal/nodes.py
+++ b/datajunction-server/datajunction_server/internal/nodes.py
@@ -372,15 +372,40 @@ async def create_a_cube(
 
 def get_node_column(node: Node, column_name: str) -> Column:
     """
-    Gets the specified column on a node
+    Gets the specified column on a node.
+
+    Resolution order:
+      1. Role-qualified identity: `<name>[<role>]` (i.e. name + dimension_column),
+         so a caller can target a specific role-played dimension column.
+      2. Bare name, when unambiguous (backward compatible).
+      3. Bare name that maps to multiple role-played columns -> raise, listing the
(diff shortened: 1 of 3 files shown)


shangyian (approved):
Nice catch, thanks!

PR #2339, opened 2026-07-17
[RBAC PR 1 follow-up] Fail closed for grant delegation

philipfweiss:
Tracking: #2234 (step 1 follow-up in the RBAC enablement sequence).

DJ's default RBAC policy is permissive, and the endpoints that hand out permissions (create a role, add a scope, assign a role, manage a group) rely on that same fallback. So on a default deployment, any logged-in user can `POST /roles/` granting `node:*` (write on every node) and assign it to themselves. The permission system cannot be enforced if any user can grant themselves any permission.

**Why this was missed initially:** #2230 added the `MANAGE` checks and was verified with the policy flipped to restrictive, and its endpoint tests mocked `MANAGE` denial directly. Neither path ran the real permission logic under the normal permissive default, which is exactly where the hole is. The test database is also built from ORM metadata, so it never hit the Alembic enum mismatch that this PR also fixes.

## Background: what "grant delegation" means here

The RBAC admin endpoints are how one principal hands permissions to another. Handing out access is itself a privileged action. The rule this PR enforces: you can only give away access that you already administer. Concretely, if you hold `MANAGE` on the `finance.*` namespace, you may create or assign roles scoped inside `finance.*`, but not roles that touch `growth.*` or everything.

## What this PR changes

These rules now apply to the role and group admin endpoints only. Ordinary metadata calls (creating nodes, etc.) are unchanged.

1. **No permissive fallback for granting permissions.** Creating, updating, assigning, or revoking a role, and creating or editing a group, now require the caller to actually hold a matching `MANAGE` grant. The permissive default no longer counts as authorization for these endpoints. This is the core fix: it is what stops a normal user from self-granting `node:*`.

2. **You can only delegate what you manage (containment).** Every scope in a role you create or assign must fit inside one of your own `MANAGE` scopes. With a caller who holds `MANAGE` on `namespace:finance.*`:
   - allowed: a role granting `read` on `finance.revenue.*` (inside `finance.*`)
   - denied: a role granting `read` on `growth.*` (a namespace you do not manage)
   - denied: a role granting `write` on `node:*` or `namespace:*` (everything)

3. **Invalid scope values fail request validation.** New scopes must be a single namespace/node (`finance.revenue`), a subtree (`finance.*`), or global (`*`). Blank, whitespace-only, and malformed values such as `finance*`, `.*`, `**`, or mid-string wildcards return 422. Containment also fails closed for unsupported values already stored in the database.

4. **Admins still bypass.** The existing `is_admin` break-glass path still authorizes these endpoints, so a fresh deployment can bootstrap its first grants.

5. **Migrated-database fix (found during manual verification).** On a real Alembic-migrated database, the RBAC scope enums are stored as lowercase and the `history` entity enum was missing the RBAC labels, so creating a role returned a 500. This PR serializes the enums with the values Alembic actually created and adds a migration for the missing history labels, so the endpoints work on deployed schemas, not only on test databases.

Behavior change: on the default permissive policy, the role and group admin endpoints now return 403 instead of silently allowing a self-grant. A caller needs an explicit `MANAGE` grant (or admin). Everything outside these endpoints keeps its current behavior, and the passthrough provider still approves everything for local dev and tests.

---
Verification:

1. Started this branch in slot 1, logged in as the seeded non-admin `dj` user, and confirmed the server was healthy.

```text
./dev.sh up -d
GET http://localhost:8100/health/
-> [{"name":"database","status":"ok"}] HTTP_STATUS:200

POST /basic/login/ username=dj password=dj
-> HTTP_STATUS:200
GET /whoami/
-> {"username":"dj","is_admin":false,...} HTTP_STATUS:200
```

2. As a normal user with no grants and the default permissive policy, confirmed the self-grant hole is closed: global and group-admin requests are now denied.

```text
POST /roles/ {scope_type:"node", scope_value:"*"}
-> {"message":"Access denied to 1 resource(s): *",...} HTTP_STATUS:403

POST /groups/?username=manual-team-final
-> {"message":"Access denied to 1 resource(s): *",...} HTTP_STATUS:403
```

3. Used the admin bypass to bootstrap one manager grant (`MANAGE namespace:finance.*`) assigned to `dj`, then dropped `dj` back to a non-admin user.

```text
UPDATE users SET is_admin=true WHERE username='dj';
POST /roles/ manual-finance-manager-final [MANAGE namespace:finance.*]
-> HTTP_STATUS:201
POST /roles/manual-finance-manager-final/assign {principal_username:"dj"}
-> HTTP_STATUS:201
UPDATE users SET is_admin=false WHERE username='dj';
GET /whoami/
-> {"username":"dj","is_admin":false,...} HTTP_STATUS:200
```

4. As that non-admin manager, confirmed delegation inside `finance.*` is allowed while anything outside it is denied.

```text
POST /roles/ manual-finance-reader-final [READ namespace:finance.revenue.*]
-> HTTP_STATUS:201

POST /roles/ manual-growth-reader-final [READ namespace:growth.*]
-> {"message":"Access denied to 1 resource(s): growth.*",...} HTTP_STATUS:403

POST /roles/ manual-global-node-with-finance-final [WRITE node:*]
-> {"message":"Access denied to 1 resource(s): *",...} HTTP_STATUS:403
```

### DIFF
--- a/datajunction-server/datajunction_server/internal/access/authorization/service.py
+++ b/datajunction-server/datajunction_server/internal/access/authorization/service.py
@@ -14,6 +14,7 @@ from datajunction_server.models.access import (
     ResourceAction,
     ResourceRequest,
     ResourceType,
+    parse_scope_pattern,
 )
 from datajunction_server.internal.access.authorization.context import (
     AuthContext,
@@ -327,8 +328,8 @@ class RBACAuthorizationService(AuthorizationService):
         delegated_value: str,
     ) -> bool:
         """Return whether one supported scope pattern contains another."""
-        granted_pattern = cls._parse_scope_pattern(granted_value)
-        delegated_pattern = cls._parse_scope_pattern(delegated_value)
+        granted_pattern = parse_scope_pattern(granted_value)
+        delegated_pattern = parse_scope_pattern(delegated_value)
         if granted_pattern is None or delegated_pattern is None:
             return False
 
@@ -351,24 +352,6 @@ class RBACAuthorizationService(AuthorizationService):
         return delegated_prefix == granted_prefix or delegated_prefix.startswith(
             granted_prefix + SEPARATOR,
         )
-
-    @staticmethod
-    def _parse_scope_pattern(value: str) -> tuple[str, str] | None:
-        """Parse exact, trailing-wildcard, and global scope patterns."""
-        if not value or value == "*":
-            return ("global", "")
-        if (
-            value.startswith(SEPARATOR)
-            or value.endswith(SEPARATOR)
-            or SEPARATOR * 2 in value
-        ):
-            return None
-        if "*" not in value:
-            return ("exact", value)
-        if value.count("*") != 1 or not value.endswith(f"{SEPARATOR}*"):
-            return None
-        prefix = value[: -len(f"{SEPARATOR}*")]
-        return ("subtree", prefix) if prefix else None
 
     @classmethod
     def _scope_grants_permission(
--- a/datajunction-server/datajunction_server/internal/access/authorization/service.py
+++ b/datajunction-server/datajunction_server/internal/access/authorization/service.py
@@ -64,6 +64,21 @@ class AuthorizationService(ABC):
             The same list of requests with approved=True/False set on each
         """
 
+    def authorize_explicit_grants(
+        self,
+        auth_context: AuthContext,
+        requests: list[ResourceRequest],
+    ) -> list[AccessDecision]:
+        """Deny control-plane access until a provider explicitly supports it."""
+        return [
+            AccessDecision(
+                request=request,
+                approved=False,
+                reason="explicit_grant_provider_required",
+            )
+            for request in requests
+        ]
+
 
 class RBACAuthorizationService(AuthorizationService):
     """
@@ -142,6 +157,19 @@ class RBACAuthorizationService(AuthorizationService):
             ]
         return [self._make_decision(auth_context, request) for request in requests]
 
+    def authorize_explicit_grants(
+        self,
+        auth_context: AuthContext,
+        requests: list[ResourceRequest],
+    ) -> list[AccessDecision]:
+        """Authorize from assigned roles without policy fallback."""
+        if auth_context.is_admin:
+            return self.authorize(auth_context, requests)
+        return [
+            self._make_explicit_grant_decision(auth_context, request)
+            for request in requests
+        ]
+
     def _make_decision(
         self,
         auth_context: AuthContext,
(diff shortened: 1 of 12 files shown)


datajunction-server/datajunction_server/internal/access/authorization/service.py
```diff
@@ -235,6 +290,87 @@ def has_permission(
...
+        """Return whether one supported scope pattern contains another."""
+        granted_pattern = cls._parse_scope_pattern(granted_value)
+        delegated_pattern = cls._parse_scope_pattern(delegated_value)
+        if granted_pattern is None or delegated_pattern is None:
+            return False
+
+        if granted_type != delegated_type and not (
+            granted_type == ResourceType.NAMESPACE
+            and delegated_type == ResourceType.NODE
+            and granted_pattern[0] != "global"
```
shangyian: I might be looking at this wrong, but does the logic here mean that a global namespace grant contains fewer scopes than a narrower one? E.g., `namespace:finance.*` correctly contains `node:finance.revenue`, but `namespace:*` (global) is excluded by the `!= "global"` clause, so a global namespace manager cannot delegate any node scope, while a subtree manager can. Is this a case where we just need to bring the next block up so that this block doesn't prematurely deny?
```
        if granted_kind == "global":
            return True
```

shangyian (commented):
Other than the one thing I mentioned -- rest of this looks good!

datajunction-server/datajunction_server/internal/access/authorization/service.py
philipfweiss: Nice find... you are right that the current condition prevents `namespace:*` from containing any node scope.

But I don't think we can move the global check ahead, because that would essentially mean

```
MANAGE node:*  contains  MANAGE namespace:*
```

Someone allowed to manage all nodes could then manage namespaces, groups, and global roles.

I think we should remove the global exclusion instead:

```python
if granted_type != delegated_type and not (
    granted_type == ResourceType.NAMESPACE
    and delegated_type == ResourceType.NODE
):
    return False
```
Then the later global check works as intended: `namespace:*` contains node scopes, while `node:*` does not contain namespace scopes.

shangyian (approved):
LGTM, thanks for addressing

datajunction-server/datajunction_server/internal/access/authorization/service.py
```diff
@@ -235,6 +290,86 @@ def has_permission(
...
+        if delegated_kind == "exact":
+            return delegated_prefix.startswith(granted_prefix + SEPARATOR)
+        return delegated_prefix == granted_prefix or delegated_prefix.startswith(
+            granted_prefix + SEPARATOR,
+        )
+
+    @staticmethod
+    def _parse_scope_pattern(value: str) -> tuple[str, str] | None:
+        """Parse exact, trailing-wildcard, and global scope patterns."""
+        if not value or value == "*":
```
ruizhang0519: empty string is treated as global — should we prefer rejecting empty/whitespace scope values at the API or parser?

ruizhang0519 (commented):
Overall LGTM! One comment - 

Malformed scopes currently surface as 403 via failed containment; a 400 “invalid scope” would be clearer (optional).

Where malformed scopes are rejected as “not contained”:

```python
# service.py — scope_contains_scope
granted_pattern = cls._parse_scope_pattern(granted_value)
delegated_pattern = cls._parse_scope_pattern(delegated_value)
if granted_pattern is None or delegated_pattern is None:
    return False
```

philipfweiss:
> Overall LGTM! One comment -
> 
> Malformed scopes currently surface as 403 via failed containment; a 400 “invalid scope” would be clearer (optional).
> 
> Where malformed scopes are rejected as “not contained”:
> 
> ```python
> # service.py — scope_contains_scope
> granted_pattern = cls._parse_scope_pattern(granted_value)
> delegated_pattern = cls._parse_scope_pattern(delegated_value)
> if granted_pattern is None or delegated_pattern is None:
>     return False
> ```

Sure, I can do that

datajunction-server/datajunction_server/internal/access/authorization/service.py
philipfweiss: Hmm - ok I agree. Currently empty string currently has the same meaning as *, while whitespace becomes an exact literal scope. I agree the API should be a bit stricter, and equire * for global.

PR #2341, opened 2026-07-18
feat(instrumentation): structured event logging interface

robinld:
Add events.emit(event_type, **fields), by default a no-op. Deployments install a publisher with set_publisher(). emit() never raises into application code. Publish failures are counted and logged with a rate-limited warning.

Emit a query_requested event from build_metrics_sql.

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v3/builder.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/builder.py
@@ -52,6 +52,7 @@ from datajunction_server.construction.build_v3.utils import (
 )
 from datajunction_server.database.partition import Partition
 from datajunction_server.errors import DJError, DJInvalidInputException, ErrorCode
+from datajunction_server.instrumentation import events
 from datajunction_server.models.dialect import Dialect
 from datajunction_server.models.partition import PartitionType
 from datajunction_server.sql.parsing import ast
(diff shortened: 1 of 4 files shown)


datajunction-server/datajunction_server/construction/build_v3/builder.py
```diff
@@ -533,6 +534,14 @@ async def build_metrics_sql(
         else:
             dialect = Dialect.TRINO
 
+    # Emit after dialect resolution (so engine is accurate) but before the heavy build.
+    events.emit(
+        "query_requested",
+        metrics=list(metrics),
+        dimensions=list(dimensions),
+        engine=dialect.name,
```
shangyian, line 557: Do we need the filters as well or is this enough?

shangyian (approved):

datajunction-server/datajunction_server/construction/build_v3/builder.py
robinld: I didn't do it because I was concerned about PII & size. Maybe I should just add the unique columns referenced & orderby & limit while I'm at it.

shangyian: +1 on just the unique columns referenced

PR #2346, opened 2026-07-20
feat: warn on fan-out risk in build_v3 SQL generation

### LINKED ISSUE #1531: Handle join cardinality in SQL generation to prevent double-counting
Currently, DJ collects join cardinality metadata (one-to-one, many-to-one, one-to-many, many-to-many) for dimension links, but this information is not used in SQL generation. This can lead to double-counting metrics when aggregating measures across joins, especially for one-to-many or many-to-many relationships.

We need SQL generation to respect join cardinality and pre-aggregate upstream fact nodes before the join when necessary.

#### Proposed Behavior

1. Pre-aggregate upstream measures before joining if the join cardinality is one-to-many or many-to-many. We will wrap upstream nodes in a subquery that aggregates measures at the node’s grain, and join the aggregated subquery instead of the raw fact table.
2. For many-to-one or one-to-one joins, we can keep the current behavior, as it is safe to join directly without risking double-counting.

robinld:
Emit a non-blocking FANOUT_RISK warning when a metric query joins across ONE_TO_MANY/MANY_TO_MANY dimension links that could inflate an additive (SUM) merge.

A dj.sql.fanout_warnings counter is emitted alongside the existing build telemetry.

See issue #1531


datajunction-server/datajunction_server/construction/build_v3/dimensions.py
```diff
@@ -982,3 +986,89 @@ def build_join_clause(
...
+    merges. We warn rather than error: the user may know the data does not fan out.
+    """
+    if not unsafe_links:
+        return None
+
+    # Components with fan-out risk: those with a SUM merge
+    inflated: list[tuple[Node, MetricComponent]] = [
+        (metric_node, component)
+        for metric_node, component in grain_group.components
+        if component.merge == "SUM"
```
shangyian: Is it only "SUM" where there's a fan-out risk? Does this cover cases like `ARRAY_AGG`, `MEDIAN` etc?

datajunction-server/datajunction_server/construction/build_v3/builder.py
shangyian, line 595: When metrics are served from an already-materialized cube, it looks like we skip the grain-group build, so no fan-out warning is surfaced for that query. I get that fan-out risk is really a property of how the cube was materialized (and there is a warning at materialization time) but if that warning was ignored (or the cube was materialized before this check existed), then every downstream consumer of that cube gets the inflated numbers with no signal. Should we persist the fan-out risk on the cube at materialization and re-surface it on reads, so query-time consumers are warned too?

(That could be a follow-up PR if you'd prefer to keep this one tidy)

datajunction-server/datajunction_server/api/graphql/schema.graphql
```diff
@@ -163,6 +163,7 @@ enum ErrorCode {
   INVALID_COLUMN
   QUERY_SERVICE_ERROR
   INVALID_ORDER_BY
+  FANOUT_RISK
```
shangyian, line 166: It looks like this exposes `FANOUT_RISK` in the GraphQL `ErrorCode` enum, but no GraphQL type (e.g., `GeneratedSQL`) actually surfaces a warnings field, so GraphQL consumers still can't see fan-out warnings. 

Is exposing warnings on GraphQL in scope for this PR, or intentionally deferred?

shangyian (approved):

datajunction-server/datajunction_server/construction/build_v3/builder.py
robinld: Yes I’ve been thinking about how to deal with this. I think I’ll do it in a follow up pr. I think ideally it would be a ci hook to issue the warning before those changes are merged.

datajunction-server/datajunction_server/construction/build_v3/dimensions.py
robinld: No, you're right that those have the same problem. This check only covers cases where DJ decomposes the aggregation. I think all I can do is switch to an allowlist of known safe ones.

datajunction-server/datajunction_server/api/graphql/schema.graphql
robinld: I think I did this because the graphql endpoint is using the v2 path. Should I add this warning to the v2 path as well?

shangyian: Ahh I see, if it's v2 then there's no need to add it for now. I think I forgot about this factor.

datajunction-server/tests/sql/decompose_test.py
```diff
@@ -2303,6 +2304,26 @@ def test_real_aggregations_marked_as_aggregation(self):
...
+    (SUM merge inflates, everything else is invariant), stated for MAX_BY/MIN_BY,
+    and assumed False for other non-decomposable (holistic) aggregations.
+    """
+    # Derived from a non-SUM merge -> invariant.
+    assert is_duplication_invariant(dj_functions.Max) is True
+    assert is_duplication_invariant(dj_functions.Min) is True
+    assert is_duplication_invariant(dj_functions.ApproxCountDistinct) is True
+    # Derived from a SUM merge -> inflates.
+    assert is_duplication_invariant(dj_functions.Sum) is False
+    assert is_duplication_invariant(dj_functions.Count) is False
```
shangyian: Should this also include a test for count distinct? I actually wonder if count distinct also ends up in this "inflates" bucket even though it shouldn't...

shangyian (approved):

PR #2351, opened 2026-07-21
Fix cube role playing declarative deploy

israelhill:
### Summary

A cube that reaches one dimension through two FK roles  produces two cube columns with the same bare name (no role identifier), distinguished only by dimension_column ([epoch_date] vs [region_date]). The declarative deploy path mishandled these in two ways, and the imperative cube-creation path had the same latent bug. Both stem from one root cause: cube columns were addressed by bare name / list position instead of by their role-qualified identity (name + dimension_column).

### The bugs

Declarative deploy (internal/deployment/orchestrator.py)
1. Roles dropped / mis-bound (non-deterministic). Element roles were recomputed positionally from rendered_dimensions, then indexed against the resolved-dimension list. That list silently skips references that don't resolve, so a single skip shifts every subsequent role off by one — dropping or mis-binding the dimension_column.
2. YAML partition never applied. The per-column partition map is keyed by the role-qualified spec name (…dateint[epoch_date]) but was looked up by the bare element name, so it never matched a role-played column and the partition was silently dropped.

Imperative creation (create_cube_node_revision) — same positional-role scheme via cube_element_roles(...) against validate_cube's resolved dimensions. Correct today only because its callers pass fully-resolvable dimensions; hardened defensively.

### The fix

Resolve and apply cube columns by their role-qualified identity, never by bare name or list position:

- CubeValidationData gains dimension_column_roles, kept strictly 1:1 with dimension_columns so each role travels with its resolved column. Populated in _validate_single_cube and the copy fast-path _build_copy_cube_validation_data.
- _create_cube_node_revision_from_validation_data builds element_roles from that 1:1 list (with a defensive pad for legacy/partial validation data) and keys the partition lookup by full_element_name + (dimension_column or "").
- validate_cube now returns dimension_roles 1:1 with resolved dims; create_cube_node_revision consumes it instead of cube_element_roles. The now-dead cube_element_roles helper is removed; the remaining caller (internal/sql.py) and the test mock are updated for the new return arity.

### Testing

- New regression test (test_cube_role_played_dimension_roles_and_partition): deploys a cube whose date dim is reached via [epoch_date] + [region_date] with a partition on [epoch_date] — asserts both roles survive and the partition lands on [epoch_date] only. Fails on the pre-fix code, passes now.
- Imperative two-role path is covered by existing TestCubeRoleQualifiedDimensions.

### DIFF
--- a/datajunction-server/datajunction_server/api/helpers.py
+++ b/datajunction-server/datajunction_server/api/helpers.py
@@ -462,22 +462,6 @@ def dimension_roles_in_order(dimensions: List[str]) -> List[Optional[str]]:
     return roles
 
 
-def cube_element_roles(
-    num_metrics: int,
-    dimension_refs: List[str],
-) -> List[Optional[str]]:
-    """
-    Role suffix (or None) for each cube element, aligned to the cube's
-    node_columns.
-
-    A cube's node_columns are built metrics-first, then dimensions, so the result
-    is `num_metrics` Nones (metrics carry no role) followed by one `[role]`-or-None
-    per dimension reference. Callers index this by each node_column's position.
-    """
-    metric_roles: List[Optional[str]] = [None] * num_metrics
-    return metric_roles + dimension_roles_in_order(dimension_refs)
-
-
 def dedupe_cube_elements(columns: List[Column]) -> List[Column]:
     """
     Dedupe cube element columns by column id.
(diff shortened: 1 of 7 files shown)


datajunction-server/datajunction_server/api/helpers.py
shangyian: I think this can be removed as well since there shouldn't be any callers left

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -2588,12 +2593,15 @@ def _validate_single_cube(
             # Get the actual column
             columns = {col.name: col for col in dimension_node.current.columns}
             column_name_without_role = column_name
+            role = None
             match = re.fullmatch(COLUMN_NAME_REGEX, column_name)
             if match:  # pragma: no cover
```
shangyian: Should this actually have `# pragma: no cover` removed so that we can make sure there are tests that hit this clause?

shangyian (approved):
LGTM, thanks for fixing

PR #2361, opened 2026-07-23
Authorize node/namespace mutations in the deployment path

ruizhang0519:
Tracking: #2234 (step 0, internal mutation paths). This is the deployment slice; background propagation, the availability service-identity model, and remaining internal paths are separate follow-ups.

Deployments are declarative bulk syncs: the orchestrator creates/updates nodes and namespaces and deletes whatever the spec drops. It runs in a detached background task (`asyncio.create_task`) after the HTTP response, so the actual mutations happen outside the request scope where `AccessChecker` lives. The `POST /deployments` entrypoint only authorizes `WRITE` on the deployment's root namespace — and since `WRITE` does not imply `DELETE`, a `WRITE`-only caller could delete nodes via an empty re-deploy.

This re-authorizes inside the orchestrator (fail-closed) against the deploying user (`context.current_user`), after the plan is built and before it is applied:

- `WRITE` on the deployment root and the namespace each deployed node lands in
- `DELETE` on nodes and namespaces the deploy removes

Names are the rendered (fully-qualified) names — what the deploy actually mutates and what RBAC scopes match against. The check runs inside the deployment SAVEPOINT, so a denial rolls back any setup-phase writes. Git sync inherits this via the shared `DeploymentOrchestrator`.

---

Why this won't take effect now: DJ's default access policy is permissive, so the check allows requests unless an explicit restrictive grant/policy denies them.

When this will start taking effect: once a deployment enables restrictive RBAC. Until then this is a no-op.

Out of scope (internal-path follow-ups): `_maybe_autolock_git_namespace` (namespace git-fields mutated after the orchestrator returns), per-preagg-parent authorization within a deploy, background revalidation/propagation, and the availability service-identity model.

---

Verification:

1. Denial test — a `WRITE`-but-not-`DELETE` caller passes the HTTP entrypoint (WRITE allowed), then an empty re-deploy that would delete a node is denied by the orchestrator; the deploy fails and the node survives. Fails before this change (node deleted), passes after.

### DIFF
--- a/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
+++ b/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
@@ -40,6 +40,11 @@ from datajunction_server.errors import (
     DJWarning,
     ErrorCode,
 )
+from datajunction_server.internal.access.authorization import (
+    AccessChecker,
+    AccessDenialMode,
+)
+from datajunction_server.internal.access.authorization.context import AuthContext
 from datajunction_server.internal.deployment.utils import (
     classify_parents,
     extract_node_graph,
@@ -60,6 +65,7 @@ from datajunction_server.sql.dag import get_metric_parents_map
 from datajunction_server.internal.nodes import (
     derive_frozen_measures_bulk,
 )
+from datajunction_server.models.access import ResourceAction
 from datajunction_server.models.base import labelize
 from datajunction_server.models.deployment import (
     ColumnSpec,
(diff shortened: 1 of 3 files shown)


datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -436,6 +436,14 @@ async def _plan_and_execute(self) -> DeploymentExecuteResult:
 
         self._guard_against_accidental_wipe(deployment_plan)
 
+        # Fail-closed authorization in the internal mutation path. The HTTP
```
shangyian: Could you make this comment less verbose? It's essentially trying to say that the orchestrator needs to authorize against each namespace that's being deployed into right? Arguably it doesn't even need a comment since the method name is pretty descriptive as-is

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -462,6 +470,50 @@ async def _plan_and_execute(self) -> DeploymentExecuteResult:
...
+        Fail-closed authorization for the mutations this deploy will perform.
+
+        Deploys run in a detached background task and bulk-create/update/delete
+        nodes and namespaces, so re-authorize here against the deploying user,
+        matching how the HTTP endpoints govern each resource: WRITE on the
+        namespace a node is written into, DELETE on nodes and namespaces being
+        removed. Names are the rendered (fully-qualified) names, since that is
+        what the deploy actually mutates and what RBAC scopes match against.
+        """
+        from datajunction_server.internal.access.authorization import (
```
shangyian: Do these need to be inline imports?

shangyian (approved):

PR #2370, opened 2026-07-27
Preserve role qualifying cube column identity

israelhill:
### Summary
Cubes can use the same business dimension in multiple roles. For example, one cube may contain both an order date and a ship date, even though both come from the same underlying date table.

Some DataJunction code paths currently identify these columns only as “date,” losing the role that distinguishes them. This can cause settings intended for one role to be applied to another, columns to disappear from output, or materializations to use the wrong partition. Because these failures can occur silently, they create a data-correctness and operational risk.

This pr hardens the full cube lifecycle so each role remains distinct across:
- Cube creation and updates
- Declarative deployment and export
- API and GraphQL responses
- SQL generation and cube matching
- Materialization, partitioning, and backfills
- Python clients, CLI, MCP, and the DataJunction UI

#### Expected outcome
An order-date column and a ship-date column will remain independently identifiable and configurable throughout DJ. Ambiguous requests will fail with a clear error instead of silently selecting a column.

#### Success criteria
- Role-specific column settings are applied to the intended column
- Cube updates and deployments preserve roles and partitions
- Materializations use the correct role-specific partition
- APIs, clients, and UI display all roles without collapsing them

Automated tests cover creation, updates, deployment, materialization, and user-facing workflows.

### DIFF
--- a/datajunction-clients/python/datajunction/cli.py
+++ b/datajunction-clients/python/datajunction/cli.py
@@ -143,7 +143,16 @@ class DJCLI:
                     "display_name": getattr(node, "display_name", None),
                     "query": getattr(node, "query", None),
                     "columns": [
-                        {"name": col.name, "type": col.type} for col in node.columns
+                        {
+                            "name": col.name
+                            + (
+                                (col.dimension_column or "")
+                                if node.type == "cube"
+                                else ""
+                            ),
+                            "type": col.type,
+                        }
+                        for col in node.columns
                     ]
                     if hasattr(node, "columns") and node.columns
                     else [],
--- a/datajunction-server/datajunction_server/api/graphql/resolvers/nodes.py
+++ b/datajunction-server/datajunction_server/api/graphql/resolvers/nodes.py
@@ -46,6 +46,11 @@ class _RawColumn:
         self.type = col_type
         self.order = order
 
+    @property
+    def cube_element_name(self) -> str:
+        """Role-qualified identity matching the ORM Column contract."""
+        return self.name + (self.dimension_column or "")
+
 
 async def _attach_raw_columns(session, nodes):
     """
--- a/datajunction-server/datajunction_server/database/column.py
+++ b/datajunction-server/datajunction_server/database/column.py
@@ -107,11 +107,25 @@ class Column(Base):  # type: ignore
         cascade="all,delete",
     )
 
-    def to_spec(self):
+    @property
+    def cube_element_name(self) -> str:
+        """Role-qualified identity for a column on a cube revision.
+
+        Cube columns store the dimension attribute FQN in ``name`` and the
+        role suffix (for example, ``[order_date]``) separately in
+        ``dimension_column``.  Both parts are required to distinguish two
+        role-played uses of the same dimension attribute.
+
+        ``dimension_column`` has different semantics on non-cube columns, so
+        callers should only use this property for columns owned by cubes.
+        """
+        return self.name + (self.dimension_column or "")
+
+    def to_spec(self, *, include_cube_role: bool = False):
         from datajunction_server.models.deployment import ColumnSpec
 
         return ColumnSpec(
-            name=self.name,
+            name=self.cube_element_name if include_cube_role else self.name,
             type=str(self.type),
             display_name=self.display_name,
             description=self.description,
--- a/datajunction-server/datajunction_server/database/partition.py
+++ b/datajunction-server/datajunction_server/database/partition.py
@@ -108,7 +108,7 @@ class Partition(Base):  # type: ignore
             )
         return None  # pragma: no cover
 
-    def categorical_expression(self):
+    def categorical_expression(self, column_name: Optional[str] = None):
         """
         Expression for the categorical partition
         """
@@ -117,7 +117,7 @@ class Partition(Base):  # type: ignore
         # Register a DJ function (inherits the `datajunction_server.sql.functions.Function` class)
         # that has the partition column name as the function name. This will be substituted at
         # runtime with the partition column name
-        amenable_partition_name = amenable_name(self.column.name)
+        amenable_partition_name = amenable_name(column_name or self.column.name)
         clazz = type(
             amenable_partition_name,
             (Function,),
--- a/datajunction-server/datajunction_server/internal/materializations.py
+++ b/datajunction-server/datajunction_server/internal/materializations.py
@@ -287,11 +287,11 @@ async def create_new_materialization(
                 access_checker,
                 current_user=current_user,
             )
-    materialization_name = (
-        f"{upsert.job.name.lower()}__{upsert.strategy.name.lower()}"  # type: ignore
-        + (f"__{temporal_partition[0].name}" if temporal_partition else "")
-        + ("__" if categorical_partitions else "")
-        + ("__".join([partition.name for partition in categorical_partitions]))
+    materialization_name = _materialization_name(
+        upsert,
+        current_revision,
+        temporal_partition,
+        categorical_partitions,
     )
     return Materialization(
         name=materialization_name,
@@ -300,6 +300,27 @@ async def create_new_materialization(
         schedule=upsert.schedule or "@daily",
         strategy=upsert.strategy,
         job=upsert.job.value.job_class,  # type: ignore
+    )
+
+
+def _materialization_name(
+    upsert: UpsertCubeMaterialization | UpsertMaterialization,
+    current_revision: NodeRevision,
+    temporal_partitions: list,
+    categorical_partitions: list,
+) -> str:
+    """Build a materialization identity that preserves cube partition roles."""
+    partition_names = [
+        partition.cube_element_name
+        if current_revision.type == NodeType.CUBE
+        else partition.name
+        for partition in temporal_partitions + categorical_partitions
+    ]
+    return (
+        f"{upsert.job.name.lower()}__{upsert.strategy.name.lower()}"  # type: ignore
+        + (f"__{partition_names[0]}" if temporal_partitions else "")
+        + ("__" if categorical_partitions else "")
+        + ("__".join(partition_names[len(temporal_partitions) :]))
     )
 
 
--- a/datajunction-clients/python/tests/test_cli.py
+++ b/datajunction-clients/python/tests/test_cli.py
@@ -5,6 +5,7 @@ import os
 import re
 import sys
 from io import StringIO
+from types import SimpleNamespace
 from typing import Callable
 from unittest import mock
 from unittest.mock import patch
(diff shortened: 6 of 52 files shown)


datajunction-server/datajunction_server/database/column.py
```diff
@@ -107,11 +107,25 @@ class Column(Base):  # type: ignore
         cascade="all,delete",
     )
 
-    def to_spec(self):
+    @property
+    def cube_element_name(self) -> str:
```
shangyian, line 111: Could this be named `semantic_name` so that it's aligned with other usage like https://github.com/DataJunction/dj/blob/05521c0e2393d6ed107bbdad5896a58bfda835e7/datajunction-server/datajunction_server/models/query.py#L77-L80?

datajunction-server/datajunction_server/internal/materializations.py
```diff
@@ -303,6 +303,27 @@ async def create_new_materialization(
...
+    """Build a materialization identity that preserves cube partition roles."""
+    partition_names = [
+        partition.cube_element_name
+        if current_revision.type == NodeType.CUBE
+        else partition.name
+        for partition in temporal_partitions + categorical_partitions
+    ]
+    return (
+        f"{upsert.job.name.lower()}__{upsert.strategy.name.lower()}"  # type: ignore
+        + (f"__{partition_names[0]}" if temporal_partitions else "")
```
shangyian, line 321: Does this assume at most one temporal partition? With two, `partition_names[0]` uses only the first and skips the second, so it vanishes from the identity.

datajunction-server/datajunction_server/database/partition.py
```diff
@@ -108,7 +108,7 @@ def temporal_expression(self, interval: Optional[str] = None):
             )
         return None  # pragma: no cover
 
-    def categorical_expression(self):
+    def categorical_expression(self, column_name: Optional[str] = None):
```
shangyian, line 111: Could `column_name` not be passed in if this just uses `self.column.cube_element_name` instead?

datajunction-server/datajunction_server/database/column.py
```diff
@@ -107,11 +107,25 @@ class Column(Base):  # type: ignore
...
+        role suffix (for example, ``[order_date]``) separately in
+        ``dimension_column``.  Both parts are required to distinguish two
+        role-played uses of the same dimension attribute.
+
+        ``dimension_column`` has different semantics on non-cube columns, so
+        callers should only use this property for columns owned by cubes.
+        """
+        return self.name + (self.dimension_column or "")
+
+    def to_spec(self, *, include_cube_role: bool = False):
```
shangyian, line 124: Can the flag `include_cube_role` be removed by doing something like:
```python
def to_spec(self):
    role = None
    if self.node_revision.type == NodeType.CUBE and self.dimension_column:
        role = self.dimension_column.strip("[]")
    return ColumnSpec(name=self.name, role=role, ...)
```

This is roughly similar to what `DimensionLinkSpec` does for example

datajunction-server/datajunction_server/internal/materializations.py
```diff
@@ -303,6 +303,27 @@ async def create_new_materialization(
     )
 
 
+def _materialization_name(
```
shangyian, line 306: Since the name is a persisted identity, this format change could orphan existing materializations right? I guess it should just be cases where cubes are referencing dimensions with roles, which wouldn't have been properly supported anyway. Worth doing a migration check though just in case.

datajunction-clients/python/datajunction/cli.py
```diff
@@ -143,7 +143,16 @@ def describe(self, node_name: str, format: str = "text"):
                     "display_name": getattr(node, "display_name", None),
                     "query": getattr(node, "query", None),
                     "columns": [
-                        {"name": col.name, "type": col.type} for col in node.columns
+                        {
```
shangyian, line 146: It looks like this is just in the `--format json` branch and the default text branch doesn't get the change to display cube cols with roles.

datajunction-server/datajunction_server/api/graphql/resolvers/nodes.py
```diff
@@ -46,6 +46,11 @@ def __init__(self, name, dimension_column, col_type, order):
         self.type = col_type
         self.order = order
 
+    @property
+    def cube_element_name(self) -> str:
```
shangyian, line 50: Similar comment to below -- could this be `semantic_name` instead?

datajunction-server/datajunction_server/database/column.py
israelhill: i called it cube_element_name to avoid misuse. `Column` is pretty overloaded since it can represent columns across the various node types. the `dimension_column` field behaves differently depending on the node type this column is associated with. For exmaple, it carries the referenced dimension name for a dimension node column:
```
dimension node column: orders.country_code

dimension_id -> common.country
dimension_column -> country_code
```

whereas for a cube column, `dimension_column` carries the bracketed role eg `[ship_date]`. So if the column object represents a column on a dimension node, this function would produce `country_codecountry_code`.

i do agree that it would be nice to have a more robust and canonical way of resolving column identifiers, but that change would have a wider blast radius. that's what i was alluding to in https://github.com/DataJunction/dj/issues/2371

datajunction-server/datajunction_server/database/column.py
israelhill: I like that approach but there are some cascading changes that would need to be made. There are a bunch of places where lookups/dicts are keyed on ColumnSpec.name, so we'd need to update all of those to use some new property that constructs the fully qualified name. Existing yaml might already have role in the name, so there are also backwards compatibility concerns. Im still learning the ins and outs of sql alchemy but wouldn't this change also require all callers to potentially load the node_revision relationship?

Maybe this could be another follow up?

datajunction-server/datajunction_server/database/partition.py
israelhill: similar to the point i made around `semantic_name`, this function is also used in non-cube node contexts in which case `dimension_column` has a different meaning and would produce invalid results. so i leave it to the caller to decide whether to pass the role-qualified column name or the ordinary column name in non-cube contexts.

israelhill: i think this again highlights a need to separate column name, semantic identity, role, and reference-link metadata instead of overloading name and dimension_column.

shangyian: hmm, but if it's used in a non-cube context, it should still have the role-qualified column either way right? as in, even if it's just a dimension for a transform node, it should still use the role-qualified version if possible

datajunction-server/datajunction_server/internal/materializations.py
israelhill: good catch. i'll look into this.

but yeah, it should only affect cases where the partition dimension has a role. this was specifically broken in my scenario where the cube i was migrating had a partition column that was referenced by two different roles, and since the role wasn't being propagated through the system, DJ could end up picking either of the two dimensions depending on what order they were in.

datajunction-server/datajunction_server/database/partition.py
shangyian: I do agree it's overloaded, but I'd argue that's partly because a node's column often is a semantic field, not just a physical one. A dimension node's column is a dimension attribute and a cube's column is a reference to one. So rather than keeping the accessor narrow and pushing the choice to callers, it might be worth giving `Column` a single semantic identity that's correct for every node type.

datajunction-server/datajunction_server/internal/materializations.py
israelhill: yeah, this assumption seems to be used in various other place of the materialization code. for example, get("/cubes/{name}/materialization") explicitly checks that there is a single partition
```
if len(temporal_partitions) != 1:
      raise DJInvalidInputException(
          "The cube must have a single temporal partition column set..."
      )
```

there are a handful of other places where we always grab the first partition. for example in SparkSqlMaterializationJob we always use temporal_partitions[0] for the partition name and filter.

are there scenarios where we allow more than one?

datajunction-clients/python/datajunction/cli.py
israelhill: the text path uses node.dimensions rather than node.columns. node.dimensions comes from the server’s cube_node_dimensions which is already role-qualified, so the text output receives the roles indirectly. The json path serializes node.columns directly which is why it needed this explicit change.

i'll add a test case around this.

datajunction-server/datajunction_server/database/column.py
shangyian: Hmm, so actually digging into this more, in the cases where lookups are keyed on `ColumnSpec.name`, that does needs changing regardless of the approach here, because otherwise there could be cases where they collide on a just the column name without the role. 

For example, models/deployment.py:691-695's `CubeSpec.__eq__` keys on `{col.name: col.partition}`. If we add role as a separate field and leave that dict as-is, the name stays role-less, so the role never participates in the comparison. That means that v1 of a cube with `dateint` and v2 of that cube with `dateint[order]` both reduce to `{"default.date.dateint": P}`, compare equal, and the cube change gets skipped instead of deployed. So that lookup needs updating regardless of which approach we take I think...

For the SQLAlchemy question, that issue goes away if we don't route through `node_revision` which is not strictly needed. The role should be recoverable from `dimension_column` alone, since both writers put it last: `[ship]` for cube columns and `country_code[ship]` for reference links.
```
@property
def role(self) -> str | None:
    if not self.dimension_column:
        return None
    match = re.search(r"\[([^\]]+)\]$", self.dimension_column)
    return match.group(1) if match else None
```

And on existing YAML with the role in the name, I think that set is pretty small, but also if we just add a validator to the pydantic class, that can help set it on parse:
```python
@model_validator(mode="after")
def split_role_from_name(self):
    if self.role is None:
        match = re.search(r"^(.*)\[([^\]]+)\]$", self.name)
        if match:
            self.name, self.role = match.group(1), match.group(2)
    return self
```

israelhill: im not sure i follow. as this PR currently stands, the role is already captured in ColumnSpec.name, so any keying we do on name will take role into account. so models/deployment.py:691-695 is safe.

shangyian: ahh sorry, you're right about the setup as-is, I was referring to the cascading changes from using a fully qualified name that's a separate property on the column

datajunction-clients/python/datajunction/cli.py
shangyian: sounds good, thanks!

datajunction-server/datajunction_server/database/column.py
shangyian: I see, makes sense, we can keep cube_element_name as-is

datajunction-server/datajunction_server/internal/materializations.py
shangyian: There might be cases where there's a date partition and then an hour partition, but it's true that grabbing the first partition seems to be littered everywhere else at the moment, so fine to leave this for now.

shangyian (approved):
Thanks for combing through this, I think it looks reasonable as-is

datajunction-server/datajunction_server/internal/materializations.py
israelhill: i ran an analysis that came back clean.

Issue #2249, opened 2026-06-10
Validate dimension reachability from cube metrics at deployment time


shangyian:
When a cube is created or updated via `dj push`, DJ validates that each referenced metric and dimension node exists and is individually valid, but does not check whether the dimensions are actually reachable from the metrics through the dimension link graph. This means a cube can be pushed successfully and all CI tests can pass, while the cube silently fails at SQL generation time with:
```
Cannot find join path from <transform> to dimension <dimension>.
Please create a dimension link between these nodes.
```
#### Proposed fix

During cube creation and update, validate that every dimension declared in the cube is reachable from every metric's parent transform via the dimension link graph. This is a graph reachability check, so it is cheap relative to generating an actual query.

Concretely, for all metrics in the cube:
1. Gather a unique set of the metrics' parents
2. Traverse the dimension link graph for each transform and merge
3. Confirm that the declared dimensions are reachable

shangyian:
@saisharan0103 this wasn't actually an issue, closing this one, but if you have a specific example you want to file here please do so separately

PR #2383, opened 2026-08-01
Validate cube dimension reachability during deployment

saisharan0103:
Issue #2249

Root cause: deployment cube validation checked that referenced metric nodes and dimension attributes exist, but per-cube validation did not fail when a declared dimension was absent from the dimension-link graph of a metric's ultimate non-metric parent. That allowed `dj push` to create or update a syntactically valid cube whose metric parent transforms could not join to every declared dimension, with the failure surfacing later during SQL generation.

Approach:
- Resolves each cube metric to its unique ultimate non-metric parent revisions during bulk cube validation.
- Builds one batched dimension reachability map for all cube parent revisions and declared cube dimension node names.
- Checks each declared dimension attribute, including role-qualified references, against every parent revision used by the cube metrics during per-cube validation.
- Appends `INVALID_DIMENSION` errors for missing paths before creating the cube revision.
- Keeps existing existence/status validation and cube metadata construction unchanged.

Files modified:
- `datajunction-server/datajunction_server/internal/deployment/orchestrator.py`
- `datajunction-server/tests/internal/deployment/orchestration_test.py`
- `fix_plan.json`

LOC: +82 / -30

`ruff check datajunction-server/datajunction_server/internal/deployment/orchestrator.py datajunction-server/tests/internal/deployment/orchestration_test.py` reports no new findings on the changed files.
Ran `pytest -x` locally with no new failures.


.mergepath-venv/bin/activate
```diff
@@ -0,0 +1,70 @@
+# This file must be used with "source bin/activate" *from bash*
```
shangyian, line 1: Why do we need this?

shangyian (commented):
There's a bunch of mergepath artifacts that shouldn't be in here.

PR #2388, opened 2026-08-03
chore: bump min Python to 3.11

betodealmeida:
### Summary

- Bump minimum version to 3.11 since [3.10 is EOL in 2 months](https://devguide.python.org/versions/#supported-versions).
- Bump pre-commit tests.
- Add `pyupgrade` to the pre-commit tests to modernize the code.

While this has a lot of changes, they are mostly mechanical, introduced by `pyupgrade` and `ruff`, like:

- `List` -> `list`, `Dict` -> `dict`, `Optional[foo]` -> `foo | None`, `"SomeType"` -> `SomeType`
- Reordering imports

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/sql/functions.py
```diff
@@ -112,7 +108,7 @@ def register(cls, func):
                     "All arguments must have a type annotation.",
                 )
             inner_types = [type_]
-            if get_origin(type_) == Union:
+            if get_origin(type_) in (Union, UnionType):
```
betodealmeida, line 111: This is needed because these annotations have different types:

```python
a: Union[foo, bar]  # Union
a: foo | bar        # UnionType
```

shangyian (approved):

shangyian:
fwiw I think 3.10 was mostly excised, just not from the integration tests (which we may need to actually remove or update since they're not passing atm):
```
client-integration-tests.yml:12   ['3.10', '3.11']
python-package-daily.yml:17       ['3.10', '3.11.0rc1']
```

PR #2392, opened 2026-08-04
Rbac background path authorization

ruizhang0519:
Tracking: #2234 (step 0) — the background half of "carry authorization through deployment and background mutation paths". The deployment half landed in #2361.

`derive_frozen_measures` and `save_column_level_lineage` run after the response returns, in their own session, and took only a revision id — so the mutation carried no authorization context and relied on whichever caller scheduled it having checked first.

Both now take the acting user plus the resource their scheduling endpoint governed, and re-authorize `WRITE` on it. That resource is not always the node: creation is governed on the target namespace, updates and revalidation on the node. Deriving a node request instead would deny the create path, because a namespace scope reaches a node request only through pattern matching — `namespace:finance` does not match `finance.total_events`, and `namespace:*` does not match a NODE request at all. Denials log and skip rather than raise, since these callers swallow exceptions.

No legitimate flow changes behavior; what changes is that the internal path is now independently fail-closed, so a future caller that schedules these without authorizing cannot mutate through them.

Downstream revalidation (`propagate_update_downstream`) is a deliberate exception, documented and pinned by a test: a node is only downstream because its own owner pointed it at the upstream, so gating it would either block owners whose nodes have dependents, or leave the graph asserting `VALID` for nodes that are now broken.

Two operator scripts have no request user, so they move to the system-facing `derive_frozen_measures_bulk` that the deployment path already uses. `migrate-measures.py` additionally had a stale two-argument call that the new signature would have bound silently.

No-op under the permissive default; relevant once a deployment enables restrictive RBAC.

---

Verification — same case, before and after. A caller is denied `WRITE` on the resource governing metric `bgwrite.total_events`, then derivation runs for it.

Before:

```
save_column_level_lineage(node_revision_id=...)   -> lineage written
derive_frozen_measures(node_revision_id=...)      -> measures derived and committed
```

After:

```
save_column_level_lineage(..., current_user=<denied>, access_target=namespace bgwrite)
-> WARNING Skipping saving column-level lineage for namespace bgwrite: dj lacks WRITE
-> lineage unchanged

derive_frozen_measures(..., current_user=<denied>, access_target=namespace bgwrite)
-> WARNING Skipping deriving frozen measures for namespace bgwrite: dj lacks WRITE
-> derivation body never runs
```

With `WRITE` granted, both proceed and lineage is written. Confirmed by neutralizing the check and re-running: the mutations return, then stop once it is restored.

Against the real RBAC matcher under a restrictive policy, an exact `namespace:bgwrite` `WRITE` grant — the grant that lets the create succeed — authorizes the namespace target and denies the node target, which is why the target is carried through rather than derived. Reverting creation to a node target makes that pin fail.

Shown at the function level because this path runs after the response is sent; the scheduling endpoints already return `403` before reaching it.


datajunction-server/scripts/migrate-measures.py
```diff
@@ -5,7 +5,7 @@
 from sqlalchemy.orm import joinedload, selectinload, sessionmaker
 
 from datajunction_server.database.node import Node, NodeRevision, NodeType
-from datajunction_server.internal.nodes import derive_frozen_measures
+from datajunction_server.internal.nodes import derive_frozen_measures_bulk
```
shangyian: This was just a one-time migration script -- are these changes still needed?

shangyian (approved):

datajunction-server/scripts/migrate-measures.py
ruizhang0519: Good point - this is indeed dead code and it is not referenced anywhere. I am leaning towards deleting this file. What do you think?

shangyian: Makes sense, go for it!

PR #2401, opened 2026-08-08
Evaluate scope patterns with the input grammar

ruizhang0519:
Tracking: #2234 (step 4) — "define one supported scope grammar and reject ambiguous patterns such as `finance*`, `.*`, and `**`".

Scope input has been validated against a grammar since #2339 (`*`, an exact name, or a subtree ending in `.*`), but evaluation parsed patterns separately. Values outside the grammar were therefore reinterpreted rather than rejected, and two of them silently meant "everything":

- `.*` and `**` collapsed to a global match, because evaluation stripped stars.
- A blank value was read as global in two more places — `parse_scope_pattern` returned global for it, and `_scope_grants_permission` had its own blank shortcut — even though scope input rejects blank outright, telling callers to use `*`.

The API can no longer create any of these, but rows predating that validation, or written outside the API, are still evaluated.

This makes one parser authoritative. `resource_matches_pattern` delegates to `parse_scope_pattern`, blank is outside the grammar, and the global shortcut recognises only `*`. A scope now means the same thing when it is written and when it is evaluated.

Valid patterns are unchanged, including `finance.*` still excluding the `finance` namespace object — that root mismatch is open decision 9 and is deliberately untouched. The rest of the global shortcut also stays, since it encodes that a global scope matches solely its own resource type, which is what keeps `namespace:*` from covering nodes.

---

Before and after, for the same values:

```
""          vs any resource        ->  granted  (now denied)
".*"        vs "secret"            ->  True     (now False)
"**"        vs "anything.at.all"   ->  True     (now False)
"finance*"  vs "finance.revenue"   ->  True     (now False)
"finance*"  vs "financial_reports" ->  False    (unchanged)
"finance.*" vs "finance.revenue"   ->  True     (unchanged)
"finance.*" vs "finance"           ->  False    (unchanged)
```

This withdraws three grants that were previously asserted as intended: `.*`/`**` matching everything, a blank scope granting through `has_permission`, and a blank grant delegating through scope containment. Each of those tests is replaced rather than deleted — an eight-pattern parametrized case for the matcher, and a blank-scope case asserting nothing is granted across both resource types.

Note for deployments: an existing grant whose value falls outside the grammar stops matching. Since those values currently mean "everything", that is the intended direction, but a quiet grant may disappear. Blanks are deliberately **not** migrated to `*` — that would preserve access where the rest of this change fails closed, and a blank row is more likely legacy than an intended global grant.

### DIFF
--- a/datajunction-server/datajunction_server/internal/access/authorization/service.py
+++ b/datajunction-server/datajunction_server/internal/access/authorization/service.py
@@ -265,23 +265,26 @@ class RBACAuthorizationService(AuthorizationService):
         resource_matches_pattern("marketing.revenue", "finance.*") --> False
         resource_matches_pattern("anything", "*") --> True
         resource_matches_pattern("finance", "finance.*") --> False
+        resource_matches_pattern("anything", ".*") --> False (outside the grammar)
+
+        Parsing is shared with scope input validation, so a scope means the same
+        thing when it is written and when it is evaluated. A value outside the
+        grammar grants nothing: evaluation used to strip stars and read ``.*`` or
+        ``**`` as a global match, so a row predating that validation, or written
+        outside the API, could silently grant everything.
         """
-        if pattern == "*":
-            return True  # Match everything
+        parsed = parse_scope_pattern(pattern)
+        if parsed is None:
+            return False
 
-        if "*" not in pattern:
-            return resource_name == pattern  # Exact match
-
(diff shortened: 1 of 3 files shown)


ruizhang0519:
Ci fix - https://github.com/DataJunction/dj/pull/2405/

philipfweiss (approved):

shangyian (approved):

PR #2404, opened 2026-08-10
feat: granular types for semantic layer API

betodealmeida:
### Summary

I chatted with @joellubi about the Apache Superset -> Quiver -> DJ integration, and we agreed it would be better to have more information about types in the metadata responses. Since Quiver uses Arrow, and Superset is moving towards Arrow, we decided to adopt the Arrow names in the response payload. This PR maps the DJ (Spark) types to Arrow types and returns them in `MetricInfo` and `DimensionInfo`.

For context: https://docs.google.com/document/d/1ho0FKy9ge0tUSRzebq1AFi28KR1H_utecQjHx4CNCEs/edit?tab=t.0#heading=h.ljxz9kmy8g4n

### Test Plan

Added unit tests.

### DIFF
--- a/datajunction-server/datajunction_server/api/semantic_layer.py
+++ b/datajunction-server/datajunction_server/api/semantic_layer.py
@@ -39,7 +39,7 @@ DEFAULT_ROW_LIMIT = 10000
 MAX_ROW_LIMIT = 100000
 
 DIMENSION_FALLBACK_ARROW_TYPE_NAME = "utf8"
-METRIC_FALLBACK_ARROW_TYPE_NAME = "float"
+METRIC_FALLBACK_ARROW_TYPE_NAME = "floating"
 
 DJ_TO_ARROW_TYPE_NAMES = {
     "array": "list",
@@ -49,8 +49,8 @@ DJ_TO_ARROW_TYPE_NAMES = {
     "char": "utf8",
     "date": "date",
     "decimal": "decimal",
-    "double": "float",
-    "float": "float",
+    "double": "floating",
+    "float": "floating",
     "int": "int",
     "integer": "int",
     "list": "list",
--- a/datajunction-server/datajunction_server/api/semantic_layer.py
+++ b/datajunction-server/datajunction_server/api/semantic_layer.py
@@ -38,7 +38,7 @@ DEFAULT_ROW_LIMIT = 10000
 # An explicit limit above this is rejected (400).
 MAX_ROW_LIMIT = 100000
 
-DEFAULT_ARROW_TYPE_NAME = "utf8"
+DIMENSION_FALLBACK_ARROW_TYPE_NAME = "utf8"
 METRIC_FALLBACK_ARROW_TYPE_NAME = "float"
 
 DJ_TO_ARROW_TYPE_NAMES = {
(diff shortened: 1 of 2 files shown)


datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -38,6 +38,34 @@
...
+DJ_TO_ARROW_TYPE_NAMES = {
+    "array": "list",
+    "bigint": "int",
+    "binary": "binary",
+    "boolean": "bool",
+    "char": "utf8",
+    "date": "date",
+    "decimal": "decimal",
+    "double": "float",
+    "float": "float",
```
shangyian: I thought in arrow it was `floatingpoint` and not `float`? https://github.com/apache/arrow/blob/main/format/Schema.fbs Unless you're using a different reference for the types of course...

shangyian (approved):

robinld (approved):

datajunction-server/datajunction_server/api/semantic_layer.py
betodealmeida: Ah, sorry, I should've left the reference!

@joellubi shared [this doc](https://docs.google.com/document/d/1ho0FKy9ge0tUSRzebq1AFi28KR1H_utecQjHx4CNCEs/edit?tab=t.0#heading=h.ljxz9kmy8g4n), it's the current proposal for standardizing the JSON representation of Arrow types.

joellubi: Yeah this is a good point. I left a comment in the doc pointing out that `float` was inconsistent with their naming standard. I've got to imagine they'll update the spec to be consistent, so lets update this to `floatingpoint` assuming they will too.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -38,6 +38,34 @@
 # An explicit limit above this is rejected (400).
 MAX_ROW_LIMIT = 100000
 
+DEFAULT_ARROW_TYPE_NAME = "utf8"
```
joellubi: Should this be `DIMENSION_FALLBACK_ARROW_TYPE_NAME` to match the metric fallback below, or is this more generic than that?

joellubi (commented):
DJ is also currently returning types in the `/semantic/{view}/sql` endpoint also (using [_generate_sql](https://github.com/DataJunction/dj/blob/b98b635009fd1e218933a0310ac84d4ad7e8dd83/datajunction-server/datajunction_server/api/semantic_layer.py#L330) under the hood) which Quiver calls in order to fulfill its own `/semantic/{view}/query` endpoint handling. We should probably standardize those types too in order to present them consistently.

datajunction-server/datajunction_server/api/semantic_layer.py
betodealmeida: Yeah, good point. Let me make this more explicit.

betodealmeida:
> DJ is also currently returning types in the `/semantic/{view}/sql` endpoint also (using [_generate_sql](https://github.com/DataJunction/dj/blob/b98b635009fd1e218933a0310ac84d4ad7e8dd83/datajunction-server/datajunction_server/api/semantic_layer.py#L330) under the hood) which Quiver calls in order to fulfill its own `/semantic/{view}/query` endpoint handling. We should probably standardize those types too in order to present them consistently.

I'll work on that.

PR #2408, opened 2026-08-11
[RBAC PR 5] Add governed namespace provisioning

philipfweiss:
Tracking: #2234 (step 5).

Creating a namespace currently leaves it without an RBAC owner. Setting up ownership requires separate role, scope, and assignment calls, so a failed sequence can leave an unmanaged namespace. Giving namespace writers access to those calls would also let `WRITE` grant `MANAGE`.

This PR adds an opt-in `POST /namespaces/{namespace}/provision` endpoint:
- The caller needs explicit `MANAGE` on the namespace, its namespace subtree, and its node subtree. Permissive fallback cannot authorize the request.
- The owner must be a group and receives `MANAGE`. Deployers must be service accounts and receive `DELETE`, which includes deployment writes.
- The namespace, roles, scopes, assignments, and their history entries commit in one transaction.
- A durable namespace marker and PostgreSQL transaction lock keep nested boundaries blocked after role renames and during concurrent requests.
- A deactivated governed namespace must use `POST /namespaces/{namespace}/restore/`, preserving its existing policy.

For example, provisioning `finance` assigns `namespace:finance:owners` to `finance-owners` and `namespace:finance:deployers` to `finance-deployer`.

Ordinary namespace creation is unchanged. The migration marks existing namespaces as non-boundaries; the provisioning endpoint marks new governed boundaries.

---
Verification:

Started `origin/main` at `:18080` and this branch at `:18081`, each with a fresh SQLite database and RBAC enabled. Seeded a writer, a manager with explicit boundary `MANAGE`, the `finance-owners` group, and the `finance-deployer` service account.

1. Existing behavior creates an unowned namespace.

```bash
curl -X POST -H 'X-User: writer' \
  http://127.0.0.1:18080/namespaces/manual_governed_before/

# HTTP 201
# namespace=1, generated roles=0, generated assignments=0
```

2. `WRITE` cannot provision ownership.

```bash
curl -X POST -H 'X-User: writer' -H 'Content-Type: application/json' \
  -d '{"owner_group":"finance-owners","deployer_service_accounts":["finance-deployer"]}' \
  http://127.0.0.1:18081/namespaces/manual_governed_denied/provision

# HTTP 403
# namespace=0, generated roles=0, generated assignments=0
```

3. Explicit `MANAGE` creates the complete policy.

```bash
curl -X POST -H 'X-User: manager' -H 'Content-Type: application/json' \
  -d '{"owner_group":"finance-owners","deployer_service_accounts":["finance-deployer"]}' \
  http://127.0.0.1:18081/namespaces/manual_governed_after/provision

# HTTP 201
{"namespace":"manual_governed_after","owner_role":"namespace:manual_governed_after:owners","deployer_role":"namespace:manual_governed_after:deployers"}

# namespace=1, roles=2, assignments=2
```

4. In a fresh PostgreSQL container, invoked parent and child provisioning concurrently, renamed the winning owner role, then deactivated and restored the boundary.

```text
concurrent_success=manual_governed
concurrent_overlap=Namespace boundary `manual_governed.child` overlaps governed boundary `manual_governed`.
history=namespace:1,role:2,assignment:2
boundary_marker=True
renamed_role_overlap=Namespace boundary `manual_governed.child` overlaps governed boundary `manual_governed`.
deactivated_reprovision=Node namespace `manual_governed` is deactivated. Restore it through `POST /namespaces/manual_governed/restore/`.
restored_active=True,assignments=2
```


datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -421,6 +425,197 @@ async def create_namespace(
...
+) -> NamespaceProvisionResponse:
+    """
+    Create a namespace boundary and its RBAC assignments in one transaction.
+
+    The caller authorizes the role scopes before calling this function. The
+    provisioner only accepts a group owner and service-account deployers, so it
+    never promotes the creator's WRITE access into MANAGE.
+    """
+    validate_namespace(namespace)
+    if await NodeNamespace.get(session, namespace, raise_if_not_exists=False):
```
shangyian: Hmm, so the way this logic is set up here, if a namespace is deactivated, it won't reactivate it but will instead raise this error?

shangyian (approved):
Mostly looks good, only one question around reactivation in namespace creation

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -421,6 +425,197 @@ async def create_namespace(
...
+    if deployers:
+        deployer_role_name = f"namespace:{namespace}:deployers"
+        role_names.append(deployer_role_name)
+    for role_name in role_names:
+        if await Role.get_by_name(session, role_name):
+            raise DJAlreadyExistsException(message=f"Role `{role_name}` already exists")
+
+    session.add(NodeNamespace(namespace=namespace))
+    session.add(
+        History(
```
ruizhang0519, line 622: Could we also add history entries for the roles and assignments created here? 
Right now we only log the namespace

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -421,6 +425,197 @@ async def create_namespace(
...
+                ),
+            )
+
+
+async def _overlapping_namespace_boundary(
+    session: AsyncSession,
+    namespace: str,
+) -> str | None:
+    """Return an active governed ancestor or descendant boundary."""
+    ancestor_role_names = [
```
ruizhang0519: If I understand it correctly, an owner can rename a role without losing access; this check would then miss it and allow a child namespace to become another boundary? If so, we should track governed boundaries using something more stable than the role name

datajunction-server/datajunction_server/internal/namespaces.py
philipfweiss: Yeah, provision rejects any existing row, including a deactivated one.

This was intentional because this endpoint only creates new. Deactivation preserves the generated roles and assignments, so a previously governed namespace keeps its policy and should use `POST /namespaces/{namespace}/restore/`

datajunction-server/datajunction_server/internal/namespaces.py
philipfweiss: Good call, adding.

datajunction-server/datajunction_server/internal/namespaces.py
philipfweiss: Ah good callout. Role names can change, so this check is unreliable...

I'll see if I can track directly instead of inferring from role names.

PR #2428, opened 2026-08-17
[RBAC PR 5.1] Auto-assign owners for approved namespaces

philipfweiss:
Tracking: #2234 (step 5.1).

Today, if Alice creates `personal.alice` while DJ is permissive, DJ records the namespace without giving Alice an RBAC grant. When write restrictions are later enabled for that namespace, Alice cannot manage what she created, so an administrator has to repair ownership.

This PR adds opt-in first-creator ownership for configured personal and ad hoc namespaces:
- `CREATOR_OWNED_NAMESPACE_PATTERNS` accepts exact names and trailing `.*` subtrees. It is empty by default.
- The first human creator receives `namespace:<name>:owners`, with `MANAGE` on the namespace, descendant namespaces, and contained nodes.
- The namespace, role, scopes, assignment, and namespace/RBAC history entries commit together.
- It uses the durable boundary marker and PostgreSQL transaction lock from [#2408](https://github.com/DataJunction/dj/pull/2408), so role renames and concurrent claims cannot create nested boundaries.
- Service accounts cannot claim a new boundary. Creation under an existing governed boundary does not create a nested owner role.

Patterns are first-claim pools. With `personal.*`, any user already authorized to create a namespace can claim an unused `personal.<name>`, so operators should only configure prefixes intended for self-service.

```bash
CREATOR_OWNED_NAMESPACE_PATTERNS='["personal.*", "scratch"]'
```

Nonmatching and existing namespaces keep their current behavior. This adds no migration or backfill. [#2408](https://github.com/DataJunction/dj/pull/2408) is merged, and this draft is rebased on `main`.

---
Verification:

Started a local FastAPI harness on `:18082` with a fresh SQLite database, the production namespace router and models, permissive RBAC, two principals (`alice` as `USER`, `deploy-bot` as `SERVICE_ACCOUNT`), and:

```bash
CREATOR_OWNED_NAMESPACE_PATTERNS='["personal.*"]'
```

1. A nonmatching namespace kept the existing behavior.

```bash
curl -X POST -H 'X-User: alice' \
  http://127.0.0.1:18082/namespaces/team.finance/
# HTTP 201; owner roles = 0
```

2. A matching namespace assigned its creator and all boundary scopes.

```bash
curl -X POST -H 'X-User: alice' \
  http://127.0.0.1:18082/namespaces/personal.alice/
# HTTP 201
# namespace:personal.alice:owners -> alice
# MANAGE namespace personal.alice
# MANAGE namespace personal.alice.*
# MANAGE node personal.alice.*
```

3. A service account could not claim the first boundary, and no partial rows remained.

```bash
curl -X POST -H 'X-User: deploy-bot' \
  http://127.0.0.1:18082/namespaces/personal.bot/
# HTTP 422; namespaces = 0; owner roles = 0
```

4. Creation below Alice's boundary did not generate a nested owner role.

```bash
curl -X POST -H 'X-User: deploy-bot' \
  http://127.0.0.1:18082/namespaces/personal.alice.project/
# HTTP 201; namespace:personal.alice.project:owners roles = 0
```

5. Repeated the matching path in a fresh SQLite metadata database, inspected the marker and history, renamed Alice's generated role, and created a child as the service account.

```text
created=created,boundary_marker=True
history=role:1,assignment:1
renamed_role_child_owner=None
```

### DIFF
--- a/datajunction-server/datajunction_server/api/nodes.py
+++ b/datajunction-server/datajunction_server/api/nodes.py
@@ -743,6 +743,7 @@ async def register_table(
         session=session,
         current_user=current_user,
         save_history=save_history,
+        creator_owned_namespace_patterns=settings.creator_owned_namespace_patterns,
     )
 
     # Use reflection to get column names and types
@@ -847,6 +848,7 @@ async def register_view(
         session=session,
         current_user=current_user,
         save_history=save_history,
+        creator_owned_namespace_patterns=settings.creator_owned_namespace_patterns,
     )
 
     return await create_source(
--- a/datajunction-server/datajunction_server/config.py
+++ b/datajunction-server/datajunction_server/config.py
@@ -6,16 +6,23 @@ Configuration for the datajunction server.
 import urllib.parse
 from datetime import timedelta
 from pathlib import Path
-from typing import Any
+from typing import Annotated, Any
 
 from cachelib.base import BaseCache
 from cachelib.file import FileSystemCache
 from cachelib.redis import RedisCache
 from celery import Celery
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, StringConstraints, field_validator
 from pydantic_settings import BaseSettings
 
-from datajunction_server.naming import is_valid_namespace, parse_scope_pattern
+from datajunction_server.naming import parse_scope_pattern
+
+CreatorOwnedNamespacePattern = Annotated[
+    str,
+    StringConstraints(
+        pattern=r"^[a-zA-Z][a-zA-Z0-9_]*(\.[a-zA-Z][a-zA-Z0-9_]*)*(\.\*)?$",
+    ),
+]
 
 RESTRICTIVE_SCOPE_ACTIONS = {"read", "write", "execute", "delete", "manage"}
 RESTRICTIVE_SCOPE_TYPES = {"node", "namespace"}
@@ -264,27 +271,9 @@ class Settings(BaseSettings):  # pragma: no cover
     # Exact namespaces or subtrees where a first human creator becomes the owner.
     # CREATOR_OWNED_NAMESPACE_PATTERNS uses JSON list syntax, for example
     # ["personal.*", "scratch"].
-    creator_owned_namespace_patterns: list[str] = Field(default_factory=list)
-
-    @field_validator("creator_owned_namespace_patterns")
-    @classmethod
-    def validate_creator_owned_namespace_patterns(
-        cls,
-        patterns: list[str],
-    ) -> list[str]:
-        """Require valid namespace patterns and reject global creator ownership."""
-        for pattern in patterns:
-            parsed = parse_scope_pattern(pattern)
-            if (
-                pattern != pattern.strip()
-                or parsed is None
-                or parsed[0] == "global"
-                or not is_valid_namespace(parsed[1])
-            ):
-                raise ValueError(
-                    f"`{pattern}` must be an exact namespace or a subtree ending in `.*`",
-                )
-        return patterns
+    creator_owned_namespace_patterns: list[CreatorOwnedNamespacePattern] = Field(
+        default_factory=list,
+    )
 
     # Interval in seconds with which to expire caching of any indexes
     index_cache_expire: int = 60
--- a/datajunction-server/datajunction_server/config.py
+++ b/datajunction-server/datajunction_server/config.py
@@ -15,7 +15,7 @@ from celery import Celery
 from pydantic import BaseModel, Field, field_validator
 from pydantic_settings import BaseSettings
 
-from datajunction_server.naming import parse_scope_pattern
+from datajunction_server.naming import is_valid_namespace, parse_scope_pattern
 
 RESTRICTIVE_SCOPE_ACTIONS = {"read", "write", "execute", "delete", "manage"}
 RESTRICTIVE_SCOPE_TYPES = {"node", "namespace"}
@@ -260,6 +260,31 @@ class Settings(BaseSettings):  # pragma: no cover
                     "or a subtree ending in '.*'",
                 )
         return values
+
+    # Exact namespaces or subtrees where a first human creator becomes the owner.
+    # CREATOR_OWNED_NAMESPACE_PATTERNS uses JSON list syntax, for example
+    # ["personal.*", "scratch"].
+    creator_owned_namespace_patterns: list[str] = Field(default_factory=list)
+
+    @field_validator("creator_owned_namespace_patterns")
+    @classmethod
+    def validate_creator_owned_namespace_patterns(
+        cls,
+        patterns: list[str],
+    ) -> list[str]:
+        """Require valid namespace patterns and reject global creator ownership."""
+        for pattern in patterns:
+            parsed = parse_scope_pattern(pattern)
+            if (
+                pattern != pattern.strip()
+                or parsed is None
+                or parsed[0] == "global"
+                or not is_valid_namespace(parsed[1])
+            ):
+                raise ValueError(
+                    f"`{pattern}` must be an exact namespace or a subtree ending in `.*`",
+                )
+        return patterns
 
     # Interval in seconds with which to expire caching of any indexes
     index_cache_expire: int = 60
--- a/datajunction-server/datajunction_server/config.py
+++ b/datajunction-server/datajunction_server/config.py
@@ -269,8 +269,8 @@ class Settings(BaseSettings):  # pragma: no cover
         return values
 
     # Exact namespaces or subtrees where a first human creator becomes the owner.
-    # CREATOR_OWNED_NAMESPACE_PATTERNS uses JSON list syntax, for example
-    # ["personal.*", "scratch"].
+    # A subtree excludes its root, so matching both requires JSON list syntax
+    # such as ["personal", "personal.*"].
     creator_owned_namespace_patterns: list[CreatorOwnedNamespacePattern] = Field(
         default_factory=list,
     )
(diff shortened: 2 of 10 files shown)


datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -453,6 +464,20 @@ def _namespace_boundary_scopes(
...
 
+def _matches_creator_owned_pattern(
+    namespace: str,
+    patterns: Sequence[str],
+) -> bool:
+    return any(
+        namespace == pattern
+        or (
+            pattern.endswith(f"{SEPARATOR}*")
+            and namespace.startswith(pattern.removesuffix("*"))
```
shangyian: This is the security-relevant line, and it's safe by accident rather than by construction.

`personal.*` strips to the prefix `personal.` and then it's a raw `startswith`. That works because the trailing dot survives `removesuffix("*")` — but what actually stops `personal*` from becoming the prefix `personal` and matching `personalthings` is the `StringConstraints` regex over in `config.py`. So the property is held up jointly by a validator in another file, and `test_creator_owned_namespace_patterns` is the only thing pinning it.

A segment comparison would be self-evidently correct and wouldn't depend on that regex staying exactly as it is:

```python
prefix = pattern.removesuffix(f"{SEPARATOR}*").split(SEPARATOR)
parts = namespace.split(SEPARATOR)
return len(parts) > len(prefix) and parts[:len(prefix)] == prefix
```

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -709,6 +789,12 @@ async def create_or_reactivate_namespace(
     Access control is enforced by each caller, not here, so this must only be
     reached from a path that has already authorized the write.
     """
+    validate_namespace(namespace)
```
shangyian: Moving validation earlier is right — you don't want to take a global lock on a junk namespace string. But it changes behaviour beyond this feature and isn't in the description.

`create_namespace` already calls `validate_namespace`, so on `main` the reactivation path never re-validates: an existing deactivated namespace is reactivated without its name being checked. Now it is. If `RESERVED_NAMESPACE_NAMES` or the part regex has ever tightened, a namespace created legally under the older rules can no longer be reactivated.

Probably fine, but worth confirming deliberately rather than as a side effect of an RBAC change.

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -539,6 +564,60 @@ def _assignment_creation_history(
...
+        return False
+
+    if await _overlapping_namespace_boundary(
+        session,
+        namespace,
+    ):
+        return False
+    if current_user.kind != PrincipalKind.USER:
+        raise DJInvalidInputException(
+            message="Only user principals can own creator-owned namespaces",
```
shangyian: This is a hard failure, not a fallback — worth making explicit in the description.

Once an operator configures `personal.*`, a service account creating `personal.anything` goes from 201 to 4xx. "Service accounts cannot claim a new boundary" reads like *no owner role is created*, not *the request fails*.

Defensible for a self-service prefix, but it's a config-dependent way to break existing automation, and the blast radius depends entirely on what someone puts in the pattern list. Is failing intended, or should a service account just create the namespace unowned?

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -539,6 +564,60 @@ def _assignment_creation_history(
...
+        )
+
+    owner_role_name = f"namespace:{namespace}:owners"
+    if await Role.get_by_name(session, owner_role_name, include_deleted=True):
+        raise DJAlreadyExistsException(
+            message=f"Role `{owner_role_name}` already exists",
```
shangyian: `include_deleted=True` means a soft-deleted `namespace:<ns>:owners` permanently blocks recreating that namespace.

Delete-and-recreate is exactly the lifecycle a scratch or personal namespace invites, so I'd expect someone to hit this. Is there a path back that doesn't need an administrator — reusing the existing role, or undeleting it when the same principal reclaims the namespace?

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -405,25 +406,35 @@ async def create_namespace(
...
         if not await get_node_namespace(  # pragma: no cover
             session=session,
             namespace=parent_namespace,
             raise_if_not_exists=False,
         ):
             logger.info("Created namespace `%s`", parent_namespace)
-            node_namespace = NodeNamespace(namespace=parent_namespace)
+            node_namespace = NodeNamespace(
+                namespace=parent_namespace,
+                is_governed_boundary=(
```
shangyian: Deep creation leaves an orphan. Creating `personal.alice.foo` directly with `include_parents=True` creates `personal`, `personal.alice` and `personal.alice.foo`, and marks only the leaf as the boundary.

`personal.alice` then exists, is unowned, and can never be claimed — a later create returns EXISTS rather than staging a role. So the creator owns a subtree whose parent nobody owns.

Not a security hole (the descendant check correctly prevents a nested boundary later), but is it the intended outcome, or should the boundary land on the highest namespace this creation brings into being under the pattern?

datajunction-server/datajunction_server/internal/namespaces.py
```diff
@@ -709,6 +789,12 @@ async def create_or_reactivate_namespace(
     Access control is enforced by each caller, not here, so this must only be
     reached from a path that has already authorized the write.
     """
+    validate_namespace(namespace)
+    if _matches_creator_owned_pattern(
+        namespace,
+        creator_owned_namespace_patterns,
+    ):
+        await lock_namespace_boundary_lifecycle(session)
```
shangyian: Minor: this serialises every creator-owned namespace creation against every other, including unrelated ones — `personal.alice` and `personal.bob` contend on the same lock.

Almost certainly fine at self-service volume, and correctness clearly comes first here. Just noting it becomes a single chokepoint if a pattern is ever configured over something high-throughput.

datajunction-server/datajunction_server/config.py
```diff
@@ -222,6 +229,13 @@ class Settings(BaseSettings):  # pragma: no cover
     rbac_require_admin: bool = False
     rbac_admin_users: list[str] = Field(default_factory=list)
 
+    # Exact namespaces or subtrees where a first human creator becomes the owner.
+    # CREATOR_OWNED_NAMESPACE_PATTERNS uses JSON list syntax, for example
+    # ["personal.*", "scratch"].
```
ruizhang0519: A pattern like `personal.*` only matches child namespaces; it does not match the root namespace `personal`

If both are wanted, would `personal*` be sufficient, or would `["personal", "personal.*"]` be preferred?

This could be a common use case, and it is worth having comments and examples here.

datajunction-server/datajunction_server/api/nodes.py
ruizhang0519, line 740: `create_or_reactivate_namespace` is a helper shared by three callers: `POST /namespaces/{namespace}/`, `register_table` and `register_view` 

That means whoever first creates a matching namespace becomes its owner, and “create” here is not only the namespace API. Registering a table or view can do it too -- is this intentional?

datajunction-server/datajunction_server/internal/namespaces.py
shangyian: Moving validation earlier is right — you don't want to take a global lock on a junk namespace string. But it changes behaviour beyond this feature and isn't in the description.

`create_namespace` already calls `validate_namespace`, so on `main` the reactivation path never re-validates: an existing deactivated namespace is reactivated without its name being checked. Now it is. If `RESERVED_NAMESPACE_NAMES` or the part regex has ever tightened, a namespace created legally under the older rules can no longer be reactivated.

Probably fine, but worth confirming deliberately rather than as a side effect of an RBAC change.

datajunction-server/datajunction_server/internal/namespaces.py
shangyian, line 589: With `Role.get_by_name(..., include_deleted=True)` doesn't that mean that a soft-deleted `namespace:<ns>:owners` permanently blocks recreating that namespace?

Is there a path back to recreation that doesn't need an admin, maybe by reusing the existing role, or undeleting it when the same principal reclaims the namespace?

datajunction-server/datajunction_server/internal/namespaces.py
shangyian, line 793: Minor: Doesn't this serialize every creator-owned namespace creation against every other, including unrelated ones? So for example `personal.alice` and `personal.bob` would contend on the same lock.

Although it's probably fine at self-service volume, but just noting that it could become a bottleneck if the pattern is ever configured over something high-throughput.

datajunction-server/datajunction_server/config.py
philipfweiss: `personal*` is intentionally left out because it could match, say, `personalized`. So I think latter is preferred. 

OK, I can add examples.

datajunction-server/datajunction_server/api/nodes.py
philipfweiss: Yes, this is intentional. All three callers can create a namespace and authorize WRITE first. Keeping ownership in the shared helper prevents table or view registration from creating a matching namespace without an owner.

datajunction-server/datajunction_server/internal/namespaces.py
philipfweiss: Yeah, `include_deleted=True` reserves the generated role name. 

There is no automatic reclaim path in this PR... Deactivated namespaces should use the `restore` endpoint, which preserves their role and assignments. 

Safely reclaiming a deleted role would require validating its provenance, scopes, and assignments, so hard-deleted boundaries currently require admin recovery.

datajunction-server/datajunction_server/internal/namespaces.py
philipfweiss: Yeah- I think it's fine for the curr volume. Higher throughput would require hierarchical locking or a database-enforced overlap constraint.

ruizhang0519 (approved):

PR #2429, opened 2026-08-19
[RBAC PR 4 follow-up] Add scoped restrictive enforcement

philipfweiss:
Tracking: #2234 (step 4).

With a permissive default, a missing grant allows every ordinary metadata mutation. Operators need a way to require explicit grants for `shared.main` writes while unrelated namespaces keep their current behavior.

This draft adds a `RESTRICTIVE_SCOPES` setting:
- Entries use `action:scope_type:scope_value`, such as `write:node:shared.main.*`.
- An admin bypass or explicit principal, group, or service-account grant can approve a matching request.
- A matching rule denies before the default-access role and permissive fallback are considered.
- Actions match exactly, so `WRITE`, `DELETE`, and `MANAGE` are configured separately while `READ` and `EXECUTE` remain open.
- Settings validation rejects unknown actions, resource types, and patterns outside the shared scope grammar during startup.
- Decisions identify the matched restrictive rule or fallback source in their reason.

This draft uses an explicit scope set for namespace boundaries and preserves existing scope meanings. Protecting `shared.main` for one action requires the namespace root, namespace subtree, and node subtree:

```text
write:namespace:shared.main
write:namespace:shared.main.*
write:node:shared.main.*
```

The empty default leaves existing deployments unchanged. Rollout begins only when an operator supplies rules and provisions the matching grants.

---
Verification:

Started a local FastAPI harness on `:18084` that routed HTTP requests through `RBACAuthorizationService` with `DEFAULT_ACCESS_POLICY=permissive` and this write boundary:

```text
RESTRICTIVE_SCOPES=[
  "write:namespace:shared.main",
  "write:namespace:shared.main.*",
  "write:node:shared.main.*"
]
```

1. An ungranted write inside the boundary was denied.

```bash
curl http://127.0.0.1:18084/check/write/node/shared.main.revenue
# {"approved":false,"reason":"restrictive_scope:write:namespace:shared.main.*"}
```

2. An explicit grant approved the same request. A default-role grant remained denied.

```bash
curl -H 'x-grant: explicit' http://127.0.0.1:18084/check/write/node/shared.main.revenue
# {"approved":true,"reason":"explicit_grant"}

curl -H 'x-grant: default' http://127.0.0.1:18084/check/write/node/shared.main.revenue
# {"approved":false,"reason":"restrictive_scope:write:namespace:shared.main.*"}
```

3. Read access inside the boundary and writes outside it followed the permissive baseline.

```bash
curl http://127.0.0.1:18084/check/read/node/shared.main.revenue
# {"approved":true,"reason":"default_access_policy_permissive"}

curl http://127.0.0.1:18084/check/write/node/shared.other.revenue
# {"approved":true,"reason":"default_access_policy_permissive"}
```

### DIFF
--- a/datajunction-server/tests/internal/authorization_test.py
+++ b/datajunction-server/tests/internal/authorization_test.py
@@ -3,11 +3,13 @@
 import logging
 from datetime import UTC, datetime, timedelta
 from types import SimpleNamespace
-from typing import cast
+from typing import ClassVar, cast
 
 import pytest
+from pydantic import ValidationError
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from datajunction_server.config import Settings
 from datajunction_server.database.group_member import GroupMember
 from datajunction_server.database.rbac import Role, RoleAssignment, RoleScope
 from datajunction_server.database.user import PrincipalKind, User
@@ -31,6 +33,7 @@ from datajunction_server.models.access import (
     ResourceAction,
     ResourceRequest,
     ResourceType,
+    parse_restrictive_scope_rule,
 )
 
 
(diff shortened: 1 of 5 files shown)


datajunction-server/tests/internal/authorization_test.py
```diff
@@ -1142,6 +1145,345 @@ async def test_expired_assignment_excluded_from_candidate_scopes(
...
+            self.context(),
+            [request],
+        )[0]
+
+        assert decision.approved is False
+        assert decision.reason and decision.reason.startswith("restrictive_scope:")
+
+    def test_only_configured_actions_and_boundary_are_restricted(self, mocker):
+        self.configure(mocker, rules=["write:node:shared.main.*"])
+        requests = [
```
betodealmeida: Why not `parametrize` these as well, with the expected result?

betodealmeida (approved):

datajunction-server/tests/internal/authorization_test.py
philipfweiss: Good callout, sure

philipfweiss: Updated in 1849979e7

Issue #2451, opened 2026-08-24
Filter on a self-joined dimension's column is copied onto every alias of the same table, corrupting results


shangyian:
#### Summary

When a dimension node's query self-joins the same underlying table, a filter on a column produced by one side of the join is also applied to the other side. The result is a plausible-looking but wrong aggregate, or silently missing rows with no error or warning.

#### Repro

A dimension that builds a month-to-date bridge by self-joining a date dimension:
```yaml
name: example.date_mtd_bridge
node_type: dimension
primary_key: [anchor_date, period_date]
query: |-
  SELECT
      a.date_int AS anchor_date
    , b.date_int AS period_date
  FROM example.date_dim a
  INNER JOIN example.date_dim b
      ON b.date_int BETWEEN a.first_date_of_month AND a.date_int
```

A transform joins orders to the bridge on `period_date`, and exposes `anchor_date`, so each order appears once per MTD window that contains it:
```yaml
name: example.orders_mtd
node_type: transform
primary_key: [order_id, order_date, anchor_date]
query: |-
  SELECT
      o.order_id
    , o.order_date
    , b.anchor_date
    , o.discount
  FROM example.orders o
  INNER JOIN example.date_mtd_bridge b
      ON o.order_date = b.period_date
```

The metric:
```yaml
name: example.avg_discount_mtd
node_type: metric
query: SELECT SUM(discount) * 1.0 / NULLIF(COUNT(*), 0) FROM example.orders_mtd
```

Filtering on the `anchor_date` yields this incorrect output SQL, since b.date_int maps to `period_date` not `anchor_date`:
```sql
    WHERE a.date_int IN (20180208, 20180210)
      AND b.date_int IN (20180208, 20180210)
```

The filter should constrain only `a`, leaving `b` free to range from the first of the month to the anchor date:
```sql
-- expected
WHERE a.date_int IN (20180208, 20180210)

-- actual
WHERE a.date_int IN (20180208, 20180210)
  AND b.date_int IN (20180208, 20180210)   -- this is copied onto the wrong side
```

shangyian:
https://github.com/DataJunction/dj/compare/main...shangyian:dj:bug/self-join-anchor-filter-pushdown?expand=1 provides a unit test to reproduce

PR #2452, opened 2026-08-24
Add failing repro of filter leaks across a self-joined dimension

shangyian:
### Summary

When a dimension node's query joins a table to itself, a filter on a column from one side of that join was also applied to the other side. The result was an extra, incorrect filter.

Filter pushdown into a CTE works in two steps:
1. The predicate is resolved against the CTE's projection, which is what picks the alias (e.g., filtering on `anchor_date` becomes `a.date_int`).
2. A second pass copies that predicate to other references to the same physical table, so nested subqueries and union arms scanning it are filtered too. When it copies, it swaps the alias prefix and keeps the column name.

That second step is sound as long as the other references are in deeper scopes, but a self-join breaks it, because both sides are the same physical table sitting in the same scope. The pass sees the far side as another place to spread to and emits `b.date_int`, which is the definition of `period_date`, a different output column entirely.

The fix is to stop copying within the scope the predicate was already resolved for. The alias chosen there is the one the filtered column projects from, and the other side is a join input that has to stay unconstrained. Copying into deeper scopes still happens, since that's the point of the pass.

### Test Plan

Added two tests in `tests/construction/build_v3/test_self_join_anchor_filter_pushdown.py` -- the self-join repro, plus a control that runs the identical bridge shape over two different physical tables. The control isolates "both aliases resolve to the same table" as the trigger rather than the self-join shape itself.

- [x] PR has an associated issue: #2451
- [x] `make check` passes

### DIFF
--- a/datajunction-server/tests/construction/build_v3/test_self_join_anchor_filter_pushdown.py
+++ b/datajunction-server/tests/construction/build_v3/test_self_join_anchor_filter_pushdown.py
@@ -1,0 +1,303 @@
+"""
+Filter pushdown into a dimension whose own query self-joins one table.
+
+A month-to-date bridge is built by self-joining a date dimension: the ``a`` side
+supplies the anchor date, the ``b`` side enumerates every date from the start of
+that month through the anchor. Both output columns therefore trace back to the
+same physical column on the same physical table, distinguished only by alias.
+
+Filtering on the anchor column should constrain the ``a`` side alone. The period
+side has to stay free, because it is what gives each anchor its window.
+"""
+
+import pytest
+from httpx import AsyncClient
+
+from tests.construction.build_v3 import assert_sql_equal
+
+
+async def _setup_mtd_bridge(client: AsyncClient) -> None:
+    """Create a date table, orders table, self-joined MTD bridge, transform, metric."""
+    response = await client.post(
+        "/nodes/source/",
+        json={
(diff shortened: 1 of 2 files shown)


datajunction-server/tests/construction/build_v3/test_self_join_anchor_filter_pushdown.py
```diff
@@ -0,0 +1,306 @@
...
+
+    assert_sql_equal(
+        sql,
+        """
+        WITH default_date_mtd_bridge AS (
+          SELECT a.date_int AS anchor_date, b.date_int AS period_date
+          FROM default.public.dates a
+          INNER JOIN default.public.dates b
+            ON b.date_int BETWEEN a.first_date_of_month AND a.date_int
+          WHERE a.date_int IN (20180208, 20180210)
```
betodealmeida, line 137: So before we'd also have here:

```sql
AND b.date_int IN (20180208, 20180210)
```

?

betodealmeida (approved):

datajunction-server/tests/construction/build_v3/test_self_join_anchor_filter_pushdown.py
shangyian: Yep, that's right, it would include both filters even though the second is incorrect

PR #2481, opened 2026-09-01
Add swr cache control option

emahannan:
### Summary

This PR adds the ability to pass in the new cache-control option, stale-while-revalidate. Stale-while-revalidate (SWR) is a caching strategy that immediately serves stale (expired) content from a cache while fetching a fresh update in the background. The changes in this PR include:

- Added opt-in result SWR caching in DJQS via Cache-Control: stale-while-revalidate.
- Uses a 12-hour freshness window and configurable retention via max-age (capped at one week).
- Returns retained stale results immediately while scheduling one async refresh; failed refreshes preserve stale data and release the retry lease.
- no-cache continues to force a fresh query; no-store avoids cache writes/refreshes.
- Forwarded Cache-Control from DataJunction server to DJQS without forwarding credentials.
- Added UI support for optional maxAge and staleWhileRevalidate arguments, defaulting SWR off.
- Added cache, refresh, forwarding, and UI request tests.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-query/djqs/api/queries.py
+++ b/datajunction-query/djqs/api/queries.py
@@ -5,7 +5,7 @@ Query related APIs.
 import json
 import logging
 import uuid
-from dataclasses import asdict
+from dataclasses import asdict, replace
 from http import HTTPStatus
 from typing import Any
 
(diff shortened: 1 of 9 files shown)


shangyian (approved):
The changes on the datajunction-server side look good! I think you'll need to fix some tests on the DJQS side though.

PR #2482, opened 2026-09-01
[DJ Semantic Fingerprint 1] Add semantic node fingerprints

philipfweiss:
DJ already classifies semantic and cosmetic node changes, but that decision is not exposed as a reusable node identity. Consumers otherwise have to duplicate field selection and SQL normalization, which can drift from the server.

This PR adds:
- version-dispatched local fingerprint and explicit parent-composition primitives derived from `ChangeTier` field definitions
- isolated semantic normalization and frozen builders under `semantic_fingerprints`
- an explicit version 1 field projection that new model fields cannot change implicitly
- structural SQL AST serialization under `sql/parsing`, independent of render settings
- versioned Merkle composition from sorted parent fingerprints

`local_node_fingerprint()` hashes one node definition. The lower-level `compose_node_fingerprint()` requires explicit parent fingerprints. These are construction primitives, and `NodeSpec` has no fingerprint method. [#2488](https://github.com/DataJunction/dj/pull/2488) adds `SemanticFingerprintGraph`, which owns graph snapshots, parent resolution, and graph-bound evaluation.

### Fingerprint composition example

For each node, `f_node` is its intrinsic semantic fingerprint and `h_node` is its Merkle object hash:

`h_node = H(f_node, sorted parent h values)`

```mermaid
flowchart LR
    S["source: orders<br/>fingerprint: f_orders<br/>object hash: h_orders = H(f_orders)"]
    D["dimension: customer<br/>fingerprint: f_customer<br/>object hash: h_customer = H(f_customer)"]
    T["transform: clean_orders<br/>fingerprint: f_clean_orders<br/>object hash: h_clean_orders = H(f_clean_orders, sort(h_orders, h_customer))"]
    M["metric: order_count<br/>fingerprint: f_order_count<br/>object hash: h_order_count = H(f_order_count, h_clean_orders)"]
    C["cube: orders_cube<br/>fingerprint: f_orders_cube<br/>object hash: h_orders_cube = H(f_orders_cube, sort(h_order_count, h_customer))"]

    S -->|SQL parent| T
    D -.->|dimension link| T
    T -->|SQL parent| M
    M -->|cube metric| C
    D -->|cube dimension| C

    classDef source fill:#fff1d6,stroke:#c2410c,color:#431407
    classDef dimension fill:#ecfdf5,stroke:#059669,color:#064e3b
    classDef derived fill:#dbeafe,stroke:#2563eb,color:#172554
    class S source
    class D dimension
    class T,M,C derived
```

---
Verification:

```python
from datajunction_server.api.main import app
from datajunction_server.models.deployment import SourceSpec, TransformSpec
from datajunction_server.semantic_fingerprints.engine import (
    compose_node_fingerprint,
    local_node_fingerprint,
)

source = SourceSpec(
    name="source",
    catalog="warehouse",
    schema="sales",
    table="orders",
)
changed = source.model_copy(update={"table": "orders_v2"})
compact = TransformSpec(name="orders", query="SELECT id FROM source")
formatted = TransformSpec(name="orders", query=" SELECT  id\nFROM source ")

source_hash = local_node_fingerprint(source)
original = compose_node_fingerprint(
    compact,
    parent_fingerprints=[source_hash],
)
format_only = compose_node_fingerprint(
    formatted,
    parent_fingerprints=[source_hash],
)
downstream = compose_node_fingerprint(
    compact,
    parent_fingerprints=[local_node_fingerprint(changed)],
)

print(f"version={original.version}")
print(f"format_stable={original == format_only}")
print(f"parent_change_propagates={original != downstream}")
```

Observed:

```text
version=1
format_stable=True
parent_change_propagates=True
```

---

This is PR 1 of 4:

1. [#2482: Add semantic node fingerprints](https://github.com/DataJunction/dj/pull/2482) (this PR)
2. [#2488: Evaluate fingerprints across deployment graphs](https://github.com/DataJunction/dj/pull/2488)
3. [#2483: Add semantic fingerprints to deployment impact](https://github.com/DataJunction/dj/pull/2483)

### DIFF
--- a/datajunction-server/datajunction_server/semantic_fingerprints/v1.py
+++ b/datajunction-server/datajunction_server/semantic_fingerprints/v1.py
@@ -15,10 +15,10 @@ from datajunction_server.models.deployment import (
     TransformSpec,
 )
 from datajunction_server.models.semantic_fingerprint import SemanticFingerprint
-from datajunction_server.semantic_fingerprints.canonical import (
-    canonical_field_value,
+from datajunction_server.semantic_fingerprints.normalization import (
     canonical_json,
-    canonical_value,
+    normalize_field,
+    normalize_value,
 )
 
 if TYPE_CHECKING:
@@ -70,11 +70,11 @@ def build_fingerprint(
     """Build a version 1 semantic fingerprint."""
     fingerprint_fields = semantic_fields(type(spec))
     for field in fingerprint_fields:
-        canonical_json(canonical_value(getattr(spec, field)))
+        canonical_json(normalize_value(getattr(spec, field)))
 
     rendered = spec.rendered_spec()
     fields = {
-        field: canonical_field_value(
+        field: normalize_field(
             rendered,
             field,
             resolved_columns=resolved_columns,
@@ -87,7 +87,7 @@ def build_fingerprint(
     }
     node_payload = {
         "domain": "datajunction/node-semantic",
-        "node_type": canonical_value(rendered.node_type),
+        "node_type": normalize_value(rendered.node_type),
         "fields": fields,
     }
     node_digest = hashlib.sha256(
--- a/datajunction-server/datajunction_server/semantic_fingerprints/v1.py
+++ b/datajunction-server/datajunction_server/semantic_fingerprints/v1.py
@@ -1,0 +1,106 @@
+"""Frozen semantic fingerprint version 1."""
+
+from __future__ import annotations
+
+import hashlib
+from collections.abc import Iterable
+from typing import TYPE_CHECKING
+
+from datajunction_server.models.deployment import (
+    ChangeTier,
+    CubeSpec,
+    DimensionSpec,
+    MetricSpec,
+    SourceSpec,
+    TransformSpec,
+)
+from datajunction_server.models.semantic_fingerprint import SemanticFingerprint
+from datajunction_server.semantic_fingerprints.canonical import (
+    canonical_field_value,
+    canonical_json,
+    canonical_value,
+)
+
+if TYPE_CHECKING:
+    from datajunction_server.models.deployment import ColumnSpec, NodeSpec
+
+
+_FIELDS_BY_SPEC_TYPE: dict[type[NodeSpec], tuple[str, ...]] = {
+    SourceSpec: (
+        "columns",
+        "dimension_links",
+        "primary_key",
+        "catalog",
+        "schema_",
+        "table",
+    ),
+    TransformSpec: (
+        "columns",
+        "dimension_links",
+        "primary_key",
+        "query",
+    ),
+    DimensionSpec: (
+        "columns",
+        "dimension_links",
+        "primary_key",
+        "query",
+    ),
+    MetricSpec: ("query", "required_dimensions"),
+    CubeSpec: ("metrics", "dimensions", "filters", "columns"),
+}
+
+
+def semantic_fields(spec_type: type[NodeSpec]) -> tuple[str, ...]:
+    """Return the frozen field projection for a concrete node type."""
+    try:
+        return _FIELDS_BY_SPEC_TYPE[spec_type]
+    except KeyError as exc:
+        raise TypeError(
+            f"No semantic fingerprint fields for {spec_type.__name__}",
+        ) from exc
+
+
+def build_fingerprint(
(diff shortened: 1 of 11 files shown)


datajunction-server/datajunction_server/semantic_fingerprints/v1.py
```diff
@@ -0,0 +1,106 @@
...
+def build_fingerprint(
+    spec: NodeSpec,
+    parent_fingerprints: Iterable[SemanticFingerprint],
+    *,
+    resolved_columns: list[ColumnSpec] | None,
+) -> SemanticFingerprint:
+    """Build a version 1 semantic fingerprint."""
+    fingerprint_fields = semantic_fields(type(spec))
+    for field in fingerprint_fields:
+        canonical_json(normalize_value(getattr(spec, field)))
```
shangyian: Is the result of this actually used anywhere? Or is 72-73 just dead code?

shangyian (approved):
This looks great -- liking the shift to having the logic on `NodeSpec` with `ChangeTier`.

datajunction-server/datajunction_server/semantic_fingerprints/v1.py
philipfweiss: Ah yeah it's dead. This came from an earlier pass when I moved fingerprint construction to `rendered_spec`. Removing.

betodealmeida:
In the example:

```python
source_hash = source.semantic_fingerprint()
original = compact.semantic_fingerprint(parent_fingerprints=[source_hash])
format_only = formatted.semantic_fingerprint(parent_fingerprints=[source_hash])
```

How hard would it be to not require `parent_fingerprints`, and instead just traverse the node upstream (with memoization) to compute it? So that we can simply do:

```python
source_hash = source.semantic_fingerprint()
original = compact.semantic_fingerprint()
format_only = formatted.semantic_fingerprint()
```

This way we don't have the risk of missing a fingerprint and generating an incorrect fingerprint.

(Unless this is supposed to be an internal API that the machinery from the next PRs is going to use.)

philipfweiss:
> In the example:
> 
> ```python
> source_hash = source.semantic_fingerprint()
> original = compact.semantic_fingerprint(parent_fingerprints=[source_hash])
> format_only = formatted.semantic_fingerprint(parent_fingerprints=[source_hash])
> ```
> 
> How hard would it be to not require `parent_fingerprints`, and instead just traverse the node upstream (with memoization) to compute it? So that we can simply do:
> 
> ```python
> source_hash = source.semantic_fingerprint()
> original = compact.semantic_fingerprint()
> format_only = formatted.semantic_fingerprint()
> ```
> 
> This way we don't have the risk of missing a fingerprint and generating an incorrect fingerprint.

Good question! The argless implementation is actually what I have done in the past, but I think unfortunately due to DJ's data model it doesn't work. I ran into the following failure case:

```python
orders_v1 = SourceSpec(
    namespace="sales",
    name="orders",
    catalog="warehouse",
    schema_="raw",
    table="orders_v1",
)
orders_v2 = orders_v1.model_copy(update={"table": "orders_v2"})
revenue = TransformSpec(
    namespace="sales",
    name="revenue",
    query="SELECT amount FROM ${prefix}orders",
)
current = {"sales.orders": orders_v1, "sales.revenue": revenue}
proposed = {"sales.orders": orders_v2, "sales.revenue": revenue}
assert (
    current["sales.revenue"].semantic_fingerprint()
    == proposed["sales.revenue"].semantic_fingerprint()
)  # Incorrect: the parent changed
```

The argless call never receives current or proposed, so it doesn't know which orders definition to traverse.

philipfweiss:
Hmm- ok I've thought about the issue some more. I think a better API might be:
```
current_graph["revenue"].semantic_fingerprint()
proposed_graph["revenue"].semantic_fingerprint()
```

or something instead like

```
graph = SemanticFingerprintGraph(...)
graph.fingerprint("revenue")
```

The key insight here is that the hash depends on the graph existing already, making it a property of an already computed (and frozen) graph. When it's on the node, you essentially have to supply the whole graph context anyways. If you're wrong, you can footgun and get the wrong hashes. 

Requiring the full graph first prevents omitted-parent bugs, memoizes traversal, and handles cycles/ missing parents consistently.

PR #2483, opened 2026-09-01
[DJ Semantic Fingerprint 3] Add semantic fingerprints to deployment impact

philipfweiss:
`/deployments/impact` reports operations and changed fields without a server-owned semantic identity. Metadata consumers otherwise have to repeat DJ's classification, SQL handling, and graph traversal.

Depends on [#2488](https://github.com/DataJunction/dj/pull/2488).

This PR:
- wires deployment graph fingerprints into create, update, no-op, delete, and downstream impact results
- discovers external downstream nodes across SQL, dimension-link, required-dimension, and cube-filter edges
- preserves current fingerprint state when a requested deletion is blocked
- returns additive nullable `change_tier` and `semantic_fingerprint` fields
- preserves `deploy_type` and parses old and new response shapes in the Python client

### Impact propagation example

```mermaid
flowchart LR
    X{"orders table changes"}
    S["source: orders<br/>object hash: h_orders to h_orders_next<br/>operation: update<br/>direct tier: major"]
    T["transform: clean_orders<br/>object hash: h_clean_orders to h_clean_orders_next<br/>operation: noop<br/>direct tier: none"]
    M["metric: order_count<br/>object hash: h_order_count to h_order_count_next<br/>operation: noop<br/>direct tier: none"]
    C["cube: orders_cube<br/>object hash: h_orders_cube to h_orders_cube_next<br/>operation: noop<br/>direct tier: none"]

    X --> S
    S ==>|h_orders changed| T
    T ==>|h_clean_orders changed| M
    M ==>|h_order_count changed| C

    classDef direct fill:#fff1d6,stroke:#c2410c,color:#431407
    classDef propagated fill:#dbeafe,stroke:#2563eb,color:#172554
    class S direct
    class T,M,C propagated
```

The metadata rows become `source::orders::h_orders_next`, `transform::clean_orders::h_clean_orders_next`, `metric::order_count::h_order_count_next`, and `cube::orders_cube::h_orders_cube_next`. All four CODEOWNERS are therefore included in review.

---
Verification:

Started this branch on development slot 1:

```bash
printf '1\n' > .dev-port
./dev.sh up -d
```

Then deployed a source, transform, and metric before changing only the source table and calling `POST /deployments/impact`:

```python
import copy
import json
import time

import requests

base = "http://127.0.0.1:8100"
namespace = "fingerprint_verification"
session = requests.Session()
session.post(
    f"{base}/basic/login/",
    data={"username": "dj", "password": "dj"},
).raise_for_status()

initial = {
    "namespace": namespace,
    "nodes": [
        {
            "name": "source",
            "node_type": "source",
            "catalog": "default",
            "schema": "test",
            "table": "orders",
            "columns": [{"name": "id", "type": "bigint"}],
        },
        {
            "name": "transform",
            "node_type": "transform",
            "query": "SELECT id FROM ${prefix}source",
        },
        {
            "name": "metric",
            "node_type": "metric",
            "query": "SELECT COUNT(*) FROM ${prefix}transform",
        },
    ],
}

def deploy(spec):
    deployment = session.post(f"{base}/deployments", json=spec).json()
    while True:
        result = session.get(f"{base}/deployments/{deployment['uuid']}").json()
        if result["status"] in {"success", "failed"}:
            return result
        time.sleep(0.1)

before = deploy(initial)
before_hashes = {
    result["name"]: result["semantic_fingerprint"]["digest"]
    for result in before["results"]
    if result["deploy_type"] == "node"
}
proposed = copy.deepcopy(initial)
proposed["nodes"][0]["table"] = "orders_v2"
response = session.post(f"{base}/deployments/impact", json=proposed)
response.raise_for_status()
results = [
    {
        "name": result["name"],
        "operation": result["operation"],
        "change_tier": result["change_tier"],
        "semantic_fingerprint": result["semantic_fingerprint"],
        "hash_changed": (
            result["semantic_fingerprint"]["digest"]
            != before_hashes[result["name"]]
        ),
    }
    for result in response.json()["results"]
    if result["deploy_type"] == "node"
]
print(json.dumps({"status": response.status_code, "results": results}, indent=2))
```

Observed:

```json
{
  "status": 200,
  "results": [
    {
      "name": "fingerprint_verification.transform",
      "operation": "noop",
      "change_tier": "none",
      "semantic_fingerprint": {
        "version": 1,
        "digest": "396e92df092afb466e34ef038b4bd6b2987b6ec09b1f39700f5d1cbf649e3c71"
      },
      "hash_changed": true
    },
    {
      "name": "fingerprint_verification.metric",
      "operation": "noop",
      "change_tier": "none",
      "semantic_fingerprint": {
        "version": 1,
        "digest": "3f6230e34838ea84ca8cd10ce6b4049b45b3ff79cf2e9840a419ade1731d19d2"
      },
      "hash_changed": true
    },
    {
      "name": "fingerprint_verification.source",
      "operation": "update",
      "change_tier": "major",
      "semantic_fingerprint": {
        "version": 1,
        "digest": "dd0e10eeead298cdb31af85f3ccd3dac98760564c4d1f97473eb25bc5d024e98"
      },
      "hash_changed": true
    }
  ]
}
```

---

This is PR 3 of 4:

1. [#2482: Add semantic node fingerprints](https://github.com/DataJunction/dj/pull/2482)
2. [#2488: Evaluate fingerprints across deployment graphs](https://github.com/DataJunction/dj/pull/2488)
3. [#2483: Add semantic fingerprints to deployment impact](https://github.com/DataJunction/dj/pull/2483) (this PR)

### DIFF
--- a/datajunction-clients/python/datajunction/models.py
+++ b/datajunction-clients/python/datajunction/models.py
@@ -3,8 +3,9 @@
 from __future__ import annotations
 
 import enum
+import re
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal, TypeAlias
 
 from datajunction._base import SerializableMixin
 
@@ -362,6 +363,42 @@ class BranchInfo(SerializableMixin):
 
 
 @dataclass
+class SemanticFingerprint:
+    """A semantic node digest returned by the server."""
+
+    digest: str
+    version: Literal[1] = 1
+
+    def __post_init__(self) -> None:
+        if self.version != 1:
+            raise ValueError(
+                f"Unsupported semantic fingerprint version: {self.version}",
+            )
+        if re.fullmatch(r"[0-9a-f]{64}", self.digest) is None:
+            raise ValueError(
+                "Semantic fingerprint digest must be 64 lowercase hexadecimal characters",
+            )
+
+    @classmethod
+    def from_dict(cls, d: dict) -> SemanticFingerprint:
+        return cls(
+            version=d.get("version", 1),
+            digest=d.get("digest", ""),
+        )
+
+
+SemanticFingerprintValue: TypeAlias = SemanticFingerprint | Literal["unknown"]
+
+
+def _parse_semantic_fingerprint(value: Any) -> SemanticFingerprintValue | None:
+    if value is None or value == "unknown":
+        return value
+    if not isinstance(value, dict):
+        raise ValueError("Semantic fingerprint must be an object or 'unknown'")
+    return SemanticFingerprint.from_dict(value)
+
+
+@dataclass
 class DeploymentResult:
     """A single node-level result within a deployment."""
 
@@ -370,15 +407,22 @@ class DeploymentResult:
     status: str
     message: str = ""
     changed_fields: list[str] = field(default_factory=list)
+    deploy_type: str = ""
+    change_tier: Literal["none", "minor", "major"] | None = None
+    semantic_fingerprint: SemanticFingerprintValue | None = None
 
     @classmethod
     def from_dict(cls, d: dict) -> DeploymentResult:
+        fingerprint = d.get("semantic_fingerprint")
         return cls(
             name=d.get("name", ""),
             operation=d.get("operation", ""),
             status=d.get("status", ""),
             message=d.get("message", ""),
             changed_fields=d.get("changed_fields") or [],
+            deploy_type=d.get("deploy_type", ""),
+            change_tier=d.get("change_tier"),
+            semantic_fingerprint=_parse_semantic_fingerprint(fingerprint),
         )
 
 
@@ -392,9 +436,11 @@ class DownstreamImpact:
     caused_by: list[str] = field(default_factory=list)
     depth: int = 0
     impact_type: str = ""
+    semantic_fingerprint: SemanticFingerprintValue | None = None
 
     @classmethod
     def from_dict(cls, d: dict) -> DownstreamImpact:
+        fingerprint = d.get("semantic_fingerprint")
         return cls(
             name=d.get("name", ""),
             node_type=d.get("node_type", ""),
@@ -402,6 +448,7 @@ class DownstreamImpact:
             caused_by=d.get("caused_by") or [],
             depth=d.get("depth", 0),
             impact_type=d.get("impact_type", ""),
+            semantic_fingerprint=_parse_semantic_fingerprint(fingerprint),
         )
 
 
--- a/datajunction-clients/python/datajunction/models.py
+++ b/datajunction-clients/python/datajunction/models.py
@@ -367,13 +367,9 @@ class SemanticFingerprint:
     """A semantic node digest returned by the server."""
 
     digest: str
-    version: Literal[1] = 1
+    version: int = 1
 
     def __post_init__(self) -> None:
-        if self.version != 1:
-            raise ValueError(
-                f"Unsupported semantic fingerprint version: {self.version}",
-            )
         if re.fullmatch(r"[0-9a-f]{64}", self.digest) is None:
             raise ValueError(
                 "Semantic fingerprint digest must be 64 lowercase hexadecimal characters",
--- a/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
+++ b/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
@@ -139,6 +139,7 @@ from datajunction_server.models.unit import (
     structured_to_legacy_unit,
 )
 from datajunction_server.sql.dag import get_metric_parents_map
+from datajunction_server.sql.parsing.backends.exceptions import DJParseException
 from datajunction_server.typing import UTCDatetime
 from datajunction_server.utils import (
     SEPARATOR,
@@ -1619,6 +1620,21 @@ class DeploymentOrchestrator:
             external_deps=external_dep_names,
         )
 
+    def _extract_plan_node_graph(
+        self,
+        nodes: list[NodeSpec],
+    ) -> dict[str, list[str]]:
+        if not self.dry_run:
+            return extract_node_graph(nodes)
+
+        node_graph = {}
+        for node in nodes:
+            try:
+                node_graph.update(extract_node_graph([node]))
+            except DJParseException:
+                node_graph[node.rendered_name] = []
+        return node_graph
+
     async def _create_deployment_plan(
         self,
     ) -> tuple[DeploymentPlan, list[DeploymentResult]]:
@@ -1680,9 +1696,8 @@ class DeploymentOrchestrator:
         external_deps: set[str] = set()
         if to_deploy or to_delete:
             with self._timer.phase("  plan: extract node graph") as p:
-                node_graph = extract_node_graph(
+                node_graph = self._extract_plan_node_graph(
                     [node for node in to_deploy if not isinstance(node, CubeSpec)],
-                    tolerate_parse_errors=self.dry_run,
                 )
                 p.append(f"{len(node_graph)} nodes in graph")
             with self._timer.phase("  plan: check external deps") as p:
@@ -1802,9 +1817,8 @@ class DeploymentOrchestrator:
                             existing_node.name,
                         )
 
-                node_graph = extract_node_graph(
+                node_graph = self._extract_plan_node_graph(
                     [node for node in to_deploy if not isinstance(node, CubeSpec)],
-                    tolerate_parse_errors=self.dry_run,
                 )
 
                 # Re-check external deps (should be empty now)
--- a/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
+++ b/datajunction-server/datajunction_server/internal/deployment/orchestrator.py
@@ -540,21 +540,6 @@ class DeploymentOrchestrator:
         # SAVEPOINT, so setup-phase writes roll back too.
         await self._authorize_deployment_plan(deployment_plan)
 
-        with self._timer.phase("build semantic fingerprints"):
-            (
-                self._current_semantic_fingerprints,
-                self._proposed_semantic_fingerprints,
-            ) = await build_deployment_fingerprints(
-                self.session,
-                deployment_plan.existing_specs,
-                self.deployment_spec.nodes,
-                [],
-                additional_target_names={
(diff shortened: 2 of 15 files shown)


datajunction-clients/python/datajunction/models.py
```diff
@@ -361,6 +362,42 @@ def from_dict(
...
+class SemanticFingerprint:
+    """A semantic node digest returned by the server."""
+
+    digest: str
+    version: Literal[1] = 1
+
+    def __post_init__(self) -> None:
+        if self.version != 1:
+            raise ValueError(
+                f"Unsupported semantic fingerprint version: {self.version}",
```
shangyian: With this raise, doesn't that mean that it'll raise on an unrecognized version the moment a server emits version 2? It feels cleaner to add a more graceful fallback, and keep the strict validation on the server (e.g., in case of client / server mismatch).

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -4299,9 +4411,29 @@ def filter_nodes_to_deploy(
...
+                proposed_columns = None
+                if isinstance(existing_spec, SourceSpec) and isinstance(
+                    node_spec,
+                    SourceSpec,
+                ):
+                    if not existing_spec.columns:
+                        resolved_columns = node_spec.columns
+                    if not node_spec.columns:
+                        proposed_columns = existing_spec.columns
+                changed_fields, reordered_fields = existing_spec.semantic_diff(
```
shangyian: This is the line that changes deploy behaviour rather than impact reporting — see the top-level comment. The fix looks right to me; it just needs to be visible in the PR description, since it changes version numbers for description-only and tag-only edits.

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -4027,7 +4121,13 @@ async def _validate_node_deletion(
         references: dict[str, list[str]] = {}
 
         # Query just IDs and names of nodes being deleted (more efficient than loading full objects)
-        stmt = select(Node.id, Node.name).where(Node.name.in_(list(nodes_to_delete)))
+        # Hold the target rows until commit so no concurrent transaction can add
+        # a new foreign-key reference after this validation snapshot.
+        stmt = (
+            select(Node.id, Node.name)
+            .where(Node.name.in_(list(nodes_to_delete)))
+            .with_for_update()
```
shangyian: Should this only add in `FOR UPDATE` here if it's not a dry-run? Otherwise it'll hold row locks on `Node` until the transaction ends for no benefit, since the dry-run transaction rolls back anyway.

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -850,6 +883,39 @@ async def _handle_no_changes(self) -> list[DeploymentResult]:
...
+            spec.rendered_name for spec in plan.to_delete
+        }
+        current, proposed = await build_deployment_fingerprints(
+            self.session,
+            plan.existing_specs,
+            self.deployment_spec.nodes,
+            plan.deletable_specs,
+            additional_target_names=target_names,
+        )
+        self._current_semantic_fingerprints = current
```
shangyian, line 912: I think this does a second full graph build per deployment (with the first being in `_plan_and_execute`). Is this rebuild needed, or could the first build take `deletable_specs` into account once `delete_references` is known?

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -1543,6 +1609,21 @@ async def _build_copy_plan(self) -> DeploymentPlan:
...
+        nodes: list[NodeSpec],
+    ) -> dict[str, list[str]]:
+        if not self.dry_run:
+            return extract_node_graph(nodes)
+
+        node_graph = {}
+        for node in nodes:
+            try:
+                node_graph.update(extract_node_graph([node]))
+            except DJParseException:
```
shangyian, line 1624: With this exception caught, does this mean that a dry run won't collect warnings / errors for an unparseable query, while a real deploy still raises? I think we should also collect these errors and surface them for dry-runs.

datajunction-server/datajunction_server/semantic_fingerprints/normalization.py
```diff
@@ -211,6 +211,28 @@ def normalize_field(
...
 
 
+def _normalize_comparison_field(
+    spec: NodeSpec,
+    field: str,
+    *,
+    resolved_columns: list[ColumnSpec] | None,
+    preserve_order: bool = False,
+) -> Any:
+    """Normalize change detection without altering versioned fingerprints."""
```
shangyian: Hmm, so looking at this comparison func, it now normalizes `required_dimensions` differently from the fingerprint. But doesn't this mean that two specs can compare equal even tho their fingerprints differ, so the actual deployment would result in a change tier of `NONE`? That discrepancy seems confusing... could it be avoided (and just use `canonical_required_dimensions`)?

datajunction-server/datajunction_server/internal/impact.py
```diff
@@ -191,11 +198,37 @@ async def _build_propagation_context(
...
+                & (NodeRevision.version == Node.current_version),
+            )
+            .where(Node.type == NodeType.CUBE)
+            .where(is_(Node.deactivated_at, None)),
+        )
+    ).all()
+    children_by_parent: dict[str, set[int]] = defaultdict(set)
+    for node_id, filters in rows:
+        for parent_name, _ in extract_dimension_refs_from_filters(filters or []):
+            children_by_parent[parent_name].add(node_id)
```
shangyian: This loads every non-deactivated cube's filters in the instance on each propagation, whether or not any cube is downstream. Filters are text so the predicate cannot be pushed into SQL, but the whole query could be skipped when nothing reachable is a cube.

datajunction-server/datajunction_server/internal/impact.py
```diff
@@ -207,19 +240,68 @@ async def _propagate_via_parent_graph(
...
+                .join(
+                    parent_revision,
+                    parent_revision.id == DBColumn.node_revision_id,
+                )
+                .join(
+                    metric_revision,
+                    metric_revision.id == BoundDimensionsRelationship.metric_id,
+                )
+                .join(Node, Node.id == metric_revision.node_id)
+                .where(parent_revision.node_id.in_(frontier_ids))
```
shangyian, line 287: `metric_revision` is pinned to `Node.current_version` below, but `parent_revision` is not, so a column belonging to a stale parent revision can still pull its metric into the frontier. The dimension-link query just above pins, so should this one match it in behavior?

datajunction-server/tests/api/deployments_test.py
```diff
@@ -1345,6 +1345,9 @@ async def deploy_and_wait(client, deployment_spec: DeploymentSpec):
         await asyncio.sleep(1)
         response = await client.get(f"/deployments/{deployment_uuid}")
         data = response.json()
+    for result in data.get("results", []):
+        result.pop("change_tier", None)
+        result.pop("semantic_fingerprint", None)
```
shangyian, line 1352: Is it worth actually setting up an assertion here?

datajunction-server/datajunction_server/internal/impact.py
```diff
@@ -191,11 +198,37 @@ async def _build_propagation_context(
...
 
 
+async def _cube_filter_children_by_parent(
+    session: AsyncSession,
+) -> dict[str, set[int]]:
+    from datajunction_server.internal.deployment.utils import (
+        extract_dimension_refs_from_filters,
+    )
+
+    rows = (
```
shangyian, line 208: I might be missing something, but it looks like the `rows` query will end up loading every non-deactivated cube's filters? As in, it's not filtering to whether a cube is in a set of ids (e.g., a cube is either within or downstream of the deployment). This query could be skipped if there are no downstream cubes, but also limited to the ones that are affected by the deployment.

datajunction-clients/python/datajunction/models.py
philipfweiss: Ah good catch, yes I can swap to a graceful fallback

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
philipfweiss: 👍 I'll change so we only do `FOR UPDATE` when it's not a dry run

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
philipfweiss: Oh interesting. I'll play around with it, but it should be possible in one build

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
philipfweiss: The parse error is still surfaced-the node is still marked `invalid` and the error goes through to the result message. 

I can update the tests to demonstrate that and perhaps add a comment

datajunction-server/datajunction_server/semantic_fingerprints/normalization.py
philipfweiss: Yes, we can just use `canonical_required_dimensions`.

datajunction-server/datajunction_server/internal/impact.py
philipfweiss: I think the current behavior is intentional. A metric can still depend on an older revision of a dimension column. If we only consider the dimension’s current revision, we could miss that dependency and fail to mark the metric as affected. I’ll add a test to make this behavior clear.

datajunction-server/tests/api/deployments_test.py
philipfweiss: Sure

datajunction-server/datajunction_server/internal/impact.py
philipfweiss: You’re right that this scans all active cubes. We can’t limit it to cubes already identified as downstream because this query is what discovers cubes affected only through filters... 

That being said, it shouldn't be too bad of a cost, roughly proportional to the total number of cube filters. It is one query per deployment with changes, so I expect the cost to be pretty modest.

shangyian (approved):
LGTM, thanks for adding!

PR #2484, opened 2026-09-01
feat: annotate sqlglot ast

betodealmeida:
### Summary

When transpiling between sqlglot dialects it's helpful to annotate the sqlglot AST so that the generators have more context. The example that made me work on this PR is this Spark query:

```sql
SELECT
  CAST(id AS BIGINT) AS id,
  key,
  value
FROM (
  SELECT
    id,
    explode(map_column) AS (key, value)
  FROM some_table T
)
```

Here, `map_column` is a map — in Spark `EXPLODE()` can be used both on arrays and maps. But when transpiled to Trino, the Trino generator treats it like an array, resulting in this invalid SQL:

```sql
SELECT
  TRY_CAST(id AS BIGINT) AS id,
  key,
  value
FROM (
  SELECT
    id,
    IF(_u.pos = _u_2.key, _u_2.value) AS value
  FROM some_table AS T
  CROSS JOIN UNNEST(SEQUENCE(1, GREATEST(CARDINALITY(map_column)))) AS _u(pos)
  CROSS JOIN UNNEST(map_column) WITH ORDINALITY AS _u_2(value, key)
  WHERE
    _u.pos = _u_2.key
    OR (
      _u.pos > CARDINALITY(map_column) AND _u_2.key = CARDINALITY(map_column)
    )
)
```

The query above is invalid here:

```sql
CROSS JOIN UNNEST(map_column) WITH ORDINALITY AS _u_2(value, key)
```

Since in Trino `UNNEST(map_column) with ORDINALITY` applied to a map returns 3 columns `(key, value, index)`, not the expected 2.

https://github.com/tobymao/sqlglot/pull/8235 fixed this, so that when the argument to `EXPLODE()` is annotated as a map the SQL gets generated correctly:

```sql
SELECT
  TRY_CAST("_0"."id" AS BIGINT) AS "id",
  "_0"."key" AS "key",
  "_0"."value" AS "value"
FROM (
  SELECT
    "t"."id" AS "id",
    _u_2."key" AS "key",
    _u_2."value" AS "value"
  FROM "some_table" AS "t"
  CROSS JOIN UNNEST("t"."map_column") AS _u_2("key", "value")
) AS "_0"
```

The problem is that, for the transpilation to work we need to make sure the sqlglot is annotated with types. I added logic that builds a sqlglot schema from the DJ schema, and qualifies and annotates the SQL before transpilation. This way we get the expected valid SQL.

### Test Plan

I added a unit test capturing the problem, but until we have a new release of `sqlglot` it has to be skipped. But I tested locally with the `main` branch, and the tests passed.

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/tests/transpilation_test.py
+++ b/datajunction-server/tests/transpilation_test.py
@@ -96,6 +96,25 @@ def test_sqlglot_transpile_success():
         assert result == "SELECT * FROM bar"
 
 
+def test_sqlglot_transpile_uses_schema_to_annotate_types():
+    """Schema types are attached before dialect-specific SQL generation."""
+    plugin = SQLGlotTranspilationPlugin()
+    result = plugin.transpile_sql(
+        "SELECT t.payload['name'] FROM catalog.schema.events AS t",
+        input_dialect=Dialect.SPARK,
+        output_dialect=Dialect.BIGQUERY,
+        schema={
+            "catalog": {
+                "schema": {
+                    "events": {"payload": "struct<name string>"},
+                },
+            },
+        },
+    )
+
+    assert result == "SELECT\n  t.payload.name\nFROM catalog.schema.events AS t"
+
+
 def test_default_transpile_success():
     with mock.patch(
         "datajunction_server.transpilation.settings.transpilation_plugins",
--- a/datajunction-server/tests/transpilation_test.py
+++ b/datajunction-server/tests/transpilation_test.py
@@ -129,24 +129,20 @@ def test_sqlglot_transpile_map_explode_with_schema():
     result = plugin.transpile_sql(
         """SELECT
   CAST(test_id AS BIGINT) AS test_id,
-  cell_id,
-  cell_name
+  key,
+  value
 FROM (
   SELECT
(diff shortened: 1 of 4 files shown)


datajunction-server/tests/transpilation_test.py
```diff
@@ -96,6 +98,76 @@ def test_sqlglot_transpile_success():
...
+    result = plugin.transpile_sql(
+        """SELECT
+  CAST(test_id AS BIGINT) AS test_id,
+  cell_id,
+  cell_name
+FROM (
+  SELECT
+    test_id,
+    explode(cells) AS (cell_id, cell_name)
+  FROM source.prodhive.dse.ab_test_detail_d_v2 T
```
shangyian: Can we make this example non-netflix-specific :)

shangyian (approved):
This looks good, thanks for adding!

datajunction-server/tests/transpilation_test.py
betodealmeida: Oops, forgot about this one! Will change, sorry!

PR #2488, opened 2026-09-01
[DJ Semantic Fingerprint 2] Evaluate fingerprints across deployment graphs

philipfweiss:
Fingerprint construction primitives can compose parent hashes, but deployment code has no server-owned evaluator for the full semantic parent graph. Cyclic dimension links also have no topological starting node, and recursive graph traversal can exceed Python's recursion limit on deep chains.

Depends on [#2482](https://github.com/DataJunction/dj/pull/2482).

This PR:
- adds `SemanticFingerprintGraph` for lazy, memoized evaluation within one graph snapshot
- resolves semantic parents from SQL, dimension links, required dimensions, cube members, and cube filters
- loads external ancestors and computes current and proposed Merkle fingerprints
- requires every concrete node spec type to register its parent resolver
- normalizes source columns and required-dimension identity across authored and persisted specs
- propagates unavailable fingerprints to dependent nodes without blocking deployment
- evaluates cycles with deterministic strongly connected component hashes

Callers construct one graph for each current or proposed snapshot, then use `fingerprint(name)` or `fingerprints(names)`. Missing parents return `unknown`, so an incomplete snapshot cannot produce a plausible graph-bound hash.

### Why cyclic graphs need SCC hashing

Direct Merkle evaluation cannot compute `A`, `B`, or `C` here because every object hash requires another unfinished object hash:

```mermaid
flowchart LR
    A["dimension A<br/>h_A = H(f_A, h_B)"]
    B["dimension B<br/>h_B = H(f_B, h_C)"]
    C["dimension C<br/>h_C = H(f_C, h_A)"]

    A -->|requires h_B| B
    B -->|requires h_C| C
    C -->|requires h_A| A

    classDef cycle fill:#fee2e2,stroke:#dc2626,color:#450a0a
    class A,B,C cycle
```

The evaluator uses an iterative two-pass SCC traversal. Its graph walk is `O(V + E)`, where `V` is the node count and `E` is the semantic parent edge count. Deterministic ordering adds sorting overhead. Explicit stacks avoid Python recursion limits on deep dependency chains. Parent discovery, SQL parsing, and database loading occur outside this graph-walk bound.

Each cycle becomes one atomic component in a condensation graph, which is always a DAG:

```text
h_scc = H(
  sorted(member name + intrinsic fingerprint),
  sorted(internal child + parent edges),
  sorted(external child + parent + parent hash)
)

h_A = H(f_A, h_scc)
h_B = H(f_B, h_scc)
h_C = H(f_C, h_scc)
```

```mermaid
flowchart LR
    P["external parent P<br/>object hash: h_P"]
    SCC["SCC {A, B, C}<br/>component hash: h_scc"]
    A["dimension A<br/>h_A = H(f_A, h_scc)"]
    B["dimension B<br/>h_B = H(f_B, h_scc)"]
    C["dimension C<br/>h_C = H(f_C, h_scc)"]
    D["downstream node D<br/>object hash changes"]

    P -->|external parent hash| SCC
    SCC --> A
    SCC --> B
    SCC --> C
    A --> D
    B --> D
    C --> D

    classDef parent fill:#fff1d6,stroke:#c2410c,color:#431407
    classDef component fill:#ede9fe,stroke:#7c3aed,color:#2e1065
    classDef member fill:#ecfdf5,stroke:#059669,color:#064e3b
    classDef downstream fill:#dbeafe,stroke:#2563eb,color:#172554
    class P parent
    class SCC component
    class A,B,C member
    class D downstream
```

Changing a member definition, internal edge, or external parent hash changes `h_scc`, every member hash, and downstream hashes. Acyclic singleton components keep the ordinary Merkle path. An unavailable member or external parent makes the component and its dependents unavailable.

---
Verification:

1. Checked out the exact PR head and built it in a fresh Docker slot:

```bash
git rev-parse HEAD
printf '2\n' > .dev-port
./dev.sh up -d
curl http://localhost:8200/health/
```

Observed:

```text
de5ed41cd22d6d3ea3cfc18ba94e30157b0a8c85
[{"name":"database","status":"ok"}]
```

2. Logged in as `dj` and deployed `graph_verification.external_source` so the graph evaluator could load a parent from committed database state:

```text
POST /basic/login/                              -> 200
POST /deployments                              -> 200
GET /deployments/{uuid}                        -> status=success
```

3. Ran `SemanticFingerprintGraph` inside the backend container with:

```bash
docker exec -i dj-s2 /code/.venv/bin/python -
```

The script exercised these inputs directly against the code at `/code/datajunction_server`:
- `source -> transform -> metric`, then changed only the source table
- repeated lookup from one graph snapshot to check memoization
- `first -> second -> third -> first` plus a downstream node, with reversed input order and a changed cycle member
- a 1,100-node dimension-link chain
- invalid SQL and unresolved-parent nodes with dependents
- authored and persisted forms of the same required dimension
- a proposed transform whose source parent existed only in Postgres

Observed:

```json
{
  "acyclic_propagation": {
    "all_descendants_changed": true,
    "changed": [
      "verify.metric",
      "verify.source",
      "verify.transform"
    ],
    "memoized": true
  },
  "cycle_evaluation": {
    "input_order_stable": true,
    "member_change_propagates": true,
    "members": [
      "cycle_verify.downstream",
      "cycle_verify.first",
      "cycle_verify.second",
      "cycle_verify.third"
    ]
  },
  "deep_chain": {
    "nodes": 1100,
    "target": "deep_verify.node_1099",
    "target_known": true
  },
  "external_ancestor": {
    "loaded_from_database": true,
    "matches_expected_merkle_hash": true,
    "fingerprint": {
      "version": 1,
      "digest": "ecce67b36517bb7ad6ce126902d0116909759c96838b08d5ce1b372cd96a4aac"
    }
  },
  "failure_propagation": {
    "verify.invalid": "unknown",
    "verify.invalid_dependent": "unknown",
    "verify.unresolved": "unknown",
    "verify.unresolved_dependent": "unknown"
  },
  "required_dimension_identity_stable": true
}
```

---

This is PR 2 of 4:

1. [#2482: Add semantic node fingerprints](https://github.com/DataJunction/dj/pull/2482)
2. [#2488: Evaluate fingerprints across deployment graphs](https://github.com/DataJunction/dj/pull/2488) (this PR)
3. [#2483: Add semantic fingerprints to deployment impact](https://github.com/DataJunction/dj/pull/2483)

### DIFF
--- a/datajunction-server/datajunction_server/internal/deployment/fingerprints.py
+++ b/datajunction-server/datajunction_server/internal/deployment/fingerprints.py
@@ -21,9 +21,14 @@ from datajunction_server.models.deployment import (
 )
 from datajunction_server.models.node import NodeType
 from datajunction_server.models.semantic_fingerprint import (
+    LATEST_SEMANTIC_FINGERPRINT_VERSION,
     UNKNOWN_SEMANTIC_FINGERPRINT,
     SemanticFingerprint,
     SemanticFingerprintValue,
+)
+from datajunction_server.semantic_fingerprints.engine import (
+    compose_node_fingerprint,
+    local_node_fingerprint,
 )
 from datajunction_server.semantic_fingerprints.merkle import (
     cycle_component_fingerprint,
@@ -225,13 +230,11 @@ def _resolved_proposed_specs(
 
 def _compute_merkle_fingerprints(
     specs: dict[str, NodeSpec],
-    target_names: Iterable[str],
     ignored_parse_errors: set[str],
     parent_cache: ParentCandidateCache | None = None,
+    *,
+    version: int = LATEST_SEMANTIC_FINGERPRINT_VERSION,
 ) -> FingerprintMap:
-    target_names = set(target_names)
-    if not target_names:
-        return {}
     parent_cache = parent_cache if parent_cache is not None else {}
     graph: dict[str, list[str]] = {}
     failed_names: set[str] = set()
@@ -334,9 +337,9 @@ def _compute_merkle_fingerprints(
                 try:
                     if not is_cycle:
                         component_results[component_index] = {
-                            members[0]: fingerprint_specs[
-                                members[0]
-                            ].semantic_fingerprint(
+                            members[0]: compose_node_fingerprint(
+                                fingerprint_specs[members[0]],
+                                version,
                                 parent_fingerprints=[
                                     edge[2] for edge in external_edges
                                 ],
@@ -344,7 +347,10 @@ def _compute_merkle_fingerprints(
                         }
                     else:
                         local_fingerprints = {
-                            name: fingerprint_specs[name].semantic_fingerprint()
+                            name: local_node_fingerprint(
+                                fingerprint_specs[name],
+                                version,
+                            )
                             for name in members
                         }
                         internal_edges = [
@@ -360,7 +366,9 @@ def _compute_merkle_fingerprints(
                             external_edges,
                         )
                         component_results[component_index] = {
-                            name: fingerprint_specs[name].semantic_fingerprint(
+                            name: compose_node_fingerprint(
+                                fingerprint_specs[name],
+                                version,
                                 parent_fingerprints=[component_fingerprint],
                             )
                             for name in members
@@ -384,11 +392,52 @@ def _compute_merkle_fingerprints(
     if processed != len(components):  # pragma: no cover
         raise RuntimeError("SCC condensation graph contains a cycle")
 
-    return {
-        name: component_results[component_by_name[name]][name]
-        for name in target_names
-        if name in specs
-    }
+    return {name: component_results[component_by_name[name]][name] for name in specs}
+
+
+class SemanticFingerprintGraph:
+    """Semantic fingerprints evaluated within one graph snapshot."""
+
+    def __init__(
+        self,
+        specs: dict[str, NodeSpec],
+        *,
+        ignored_parse_errors: set[str] | None = None,
+        parent_cache: ParentCandidateCache | None = None,
+        version: int = LATEST_SEMANTIC_FINGERPRINT_VERSION,
+    ):
+        self._specs = dict(specs)
+        self._ignored_parse_errors = set(ignored_parse_errors or ())
+        self._parent_cache = parent_cache if parent_cache is not None else {}
+        self._version = version
+        self._fingerprints: FingerprintMap | None = None
+
+    def fingerprint(self, name: str) -> SemanticFingerprintValue:
+        """Return one node's fingerprint in this graph snapshot."""
+        return self._evaluate()[name]
+
+    def fingerprints(
+        self,
+        names: Iterable[str] | None = None,
+    ) -> FingerprintMap:
+        """Return fingerprints for selected nodes in this graph snapshot."""
+        target_names = list(self._specs if names is None else names)
+        if not target_names:
+            return {}
+        fingerprints = self._evaluate()
+        return {
+            name: fingerprints[name] for name in target_names if name in fingerprints
+        }
+
+    def _evaluate(self) -> FingerprintMap:
+        if self._fingerprints is None:
+            self._fingerprints = _compute_merkle_fingerprints(
+                self._specs,
+                self._ignored_parse_errors,
+                self._parent_cache,
+                version=self._version,
+            )
+        return self._fingerprints
 
 
 async def build_deployment_fingerprints(
@@ -398,6 +447,7 @@ async def build_deployment_fingerprints(
     deleted_specs: Iterable[NodeSpec],
     *,
     additional_target_names: Iterable[str] = (),
+    version: int = LATEST_SEMANTIC_FINGERPRINT_VERSION,
 ) -> tuple[FingerprintMap, FingerprintMap]:
     proposed_specs = list(proposed_specs)
     additional_target_names = set(additional_target_names)
@@ -437,18 +487,21 @@ async def build_deployment_fingerprints(
         parent_cache=parent_cache,
     )
     external.update(target_specs)
-    current_graph = {**external, **existing_specs}
-    proposed_graph = {**external, **proposed}
-    current = _compute_merkle_fingerprints(
-        current_graph,
-        deleted_names | additional_target_names,
+    current_graph = SemanticFingerprintGraph(
+        {**external, **existing_specs},
         ignored_parse_errors=deleted_names,
         parent_cache=parent_cache,
+        version=version,
     )
-    proposed_hashes = _compute_merkle_fingerprints(
-        proposed_graph,
-        submitted_names | additional_target_names,
-        ignored_parse_errors=set(),
+    proposed_graph = SemanticFingerprintGraph(
+        {**external, **proposed},
         parent_cache=parent_cache,
+        version=version,
+    )
+    current = current_graph.fingerprints(
+        deleted_names | additional_target_names,
+    )
+    proposed_hashes = proposed_graph.fingerprints(
+        submitted_names | additional_target_names,
     )
     return current, proposed_hashes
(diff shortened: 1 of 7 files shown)


datajunction-server/datajunction_server/internal/deployment/fingerprints.py
```diff
@@ -0,0 +1,460 @@
...
+        raise TypeError(f"Expected cube spec, got {type(spec).__name__}")
+    return {
+        node_name
+        for node_name, _ in extract_dimension_refs_from_filters(
+            cube.rendered_filters,
+        )
+    }
+
+
+SEMANTIC_PARENT_RESOLVERS: dict[type[NodeSpec], ParentResolver] = {
```
shangyian: Could the graph here build on `extract_node_graph` in utils.py rather than assembling a second one? It already returns `name -> [parent names]` for a list of specs and caches the parsed AST on each, which looks like what   `strongly_connected_components` needs.

To be more specific, I'd like to avoid having multiple ways of computing a node's parents so that we don't introduce further inconsistencies later on. For example,  `extract_upstream_candidates` encodes the derived-metric rule (no `FROM`) and that same check reappears here at line 122 and again at 149. If deployment ordering and fingerprint propagation ever disagree about an edge, that'd be unfun to debug.

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
```diff
@@ -0,0 +1,507 @@
...
+    return set(query | extra)
+
+
+def _resolved_parent_names(
+    spec: NodeSpec,
+    specs: dict[str, NodeSpec],
+    cache: ParentCandidateCache,
+) -> tuple[list[str], list[str]]:
+    query, extra = _candidate_parts(spec, cache)
+    query_candidates = set(query)
```
shangyian, line 181: Isn't the logic here just a duplicate of the logic here https://github.com/DataJunction/dj/pull/2488/changes#diff-21f60a4720fcd7892a80d2fdf06b42988ff898fe8972ba79b1f83e2beeda5d2fR146-R154? Could it be factored out into something like `_direct_query_parents` that just finds the spec's direct query parents?

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
```diff
@@ -0,0 +1,507 @@
...
+    specs: dict[str, NodeSpec],
+    ignored_parse_errors: set[str],
+    parent_cache: ParentCandidateCache | None = None,
+    *,
+    version: int = LATEST_SEMANTIC_FINGERPRINT_VERSION,
+) -> FingerprintMap:
+    parent_cache = parent_cache if parent_cache is not None else {}
+    graph: dict[str, list[str]] = {}
+    failed_names: set[str] = set()
+    for name in sorted(specs):
```
shangyian, line 301: It looks like there's two loops that go through specs here (second one starting https://github.com/DataJunction/dj/pull/2488/changes#diff-21f60a4720fcd7892a80d2fdf06b42988ff898fe8972ba79b1f83e2beeda5d2fR263) -- can they be collapsed into one? It doesn't look like the second loop depends on the first loop's output.

datajunction-server/datajunction_server/internal/deployment/utils.py
```diff
@@ -60,6 +61,26 @@ def extract_upstream_candidates(
...
+def extract_dimension_refs_from_filters(
+    filters: list[str],
+) -> list[tuple[str, str]]:
+    """Extract namespaced dimension columns referenced by filter expressions."""
+    if not filters:
+        return []
+    combined = " AND ".join(f"({filter_})" for filter_ in filters)
+    try:
+        tree = parse(f"SELECT 1 WHERE {combined}")
+    except Exception:
```
shangyian: If the filter is unparseable for some reason, should it be omitted? Or should the bad filter actually be part of the semantic fingerprint of the node? 

Also can this be changed to `DJParseException`?

datajunction-server/datajunction_server/internal/deployment/utils.py
```diff
@@ -86,7 +107,11 @@ def classify_parents(
     return resolved, missing
 
 
-def extract_node_graph(nodes: list[NodeSpec]) -> dict[str, list[str]]:
+def extract_node_graph(
+    nodes: list[NodeSpec],
+    *,
+    tolerate_parse_errors: bool = False,
```
shangyian: Does `tolerate_parse_errors` actually get used outside of tests? I can see a use case where we still want a semantic fingerprint on an invalid / unparseable SQL query, but I don't think it's actually enabled via any call path?

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
```diff
@@ -0,0 +1,507 @@
...
+            if name in specs and specs[name].node_type == NodeType.METRIC
+        }
+        unresolved_query: set[str] = set()
+    else:
+        query_parents = query_candidates & specs.keys()
+        unresolved_query = query_candidates - specs.keys()
+
+    extra_candidates = set(extra)
+    resolved = query_parents | (extra_candidates & specs.keys())
+    unresolved = unresolved_query | (extra_candidates - specs.keys())
```
shangyian: From this setup, I think there might be false positives that end up in `extra_candidates`, causing the semantic fingerprint to be unresolved when it shouldn't be. 

For example, `_cube_parent_candidates` reads any dotted column reference in a cube's filters as `node_name.column`, and accepts `node_name` as a node as long as it has a separator in it. That works for the non-struct column case -- e.g.,  `default.hard_hat.state = 'CA'` gives us `default.hard_hat` as the node name, but if the column is a struct and we're accessing a nested field, there would be a problem: `default.hard_hat.address.city = 'LA'` -> `[('default.hard_hat.address', 'city')]`

Since `default.hard_hat.address` isn't in specs, this candidate gets put into unresolved and the cube's fingerprint will come back as `UNKNOWN`.

I think the same applies to metrics via `_metric_parent_candidates` around the handling of required dimensions. If there's a required dim written as a nested struct path, there would be the same issue with the prefix guesswork.

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
philipfweiss: Oh nice, yes it is. I'll consolidate.

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
philipfweiss: Nice find, yes I can collapse them

datajunction-server/datajunction_server/internal/deployment/utils.py
philipfweiss: Ah you're right, this was left over from an earlier pass. I'll do a scan for any dead code I might have left in from previous iterations.

datajunction-server/datajunction_server/internal/deployment/utils.py
philipfweiss: > If the filter is unparseable for some reason, should it be omitted? Or should the bad filter actually be part of the semantic fingerprint of the node?

IMO the bad filter should remain part of the node fingerprint so an invalid filter is distinguishable from having no filter. 

It already is included through the cube’s filters field. Returning `[]` here only skips parent extraction for the broken filter, it does not omit the filter text from the fingerprint. 

> Also can this be changed to DJParseException?

Sure

datajunction-server/datajunction_server/internal/deployment/fingerprints.py
philipfweiss: Oh shoot yeah this is a bug. 

I think I'll need to change the reference resolution algorithm so it finds the node-name boundary using nodes that exist in the submitted graph or database.

 Today we remove exactly one segment, so for example `default.hard_hat.address.city` becomes the guessed node `default.hard_hat.address`...
 
Instead, we could continue removing segments until we find the longest existing node name. Here that would resolve to node `default.hard_hat`, leaving `address.city` as the nested field.

We should only return unknown when none of the possible node names exists.

shangyian (approved):
Thanks for addressing comments, looks good!

PR #2496, opened 2026-09-02
Single-Pass Column Pruning

shangyian:
The v3 builder trims each CTE it builds down to the columns that are used downstream. That trimming used to happen while the query was still being built, before the CTEs that would read a given CTE existed, so it worked from a precomputed list of columns that *might* be needed. However, this list kept missing cases, resulting in generated SQL that selected unprojected columns from a source. 

This changes it so that the old logic around column pruning is removed, subsumed with a single-pass column prune that runs once after the query is fully assembled.

#### Removed

- The pruning call inside the per-CTE loop in `collect_node_ctes`.
- The `needed_columns_by_node` map and the parameter passing it to `collect_node_ctes`.
- `_add_transitive_consumer_columns`, which existed only to patch gaps in the enumeration.
- `extract_columns_referenced_from_node` and `extract_join_columns_for_node`

#### Added

Added `prune_cte_projections`, which is called after the query is assembled. It keeps track of the query scopes ordered by reads, then walks the scopes, pruning it down to the columns it needs before recording.

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v3/cte.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/cte.py
@@ -938,14 +938,29 @@ def _set_op_arms(select: ast.SelectExpression) -> list[ast.SelectExpression]:
     return arms
 
 
-def _dedups_rows(select: ast.SelectExpression) -> bool:
+def _compares_whole_rows(select: ast.SelectExpression) -> bool:
     """
-    Whether the whole projected row is the select's match key.
+    Whether this select picks its rows by comparing them against each other.
 
-    DISTINCT compares complete rows, and so does every set operation but
-    UNION ALL — the one kind that passes its arms' rows through untouched.
-    Dropping a column then folds together rows that were distinct and
-    silently changes the row count.
+    DISTINCT does — it keeps a row only if no earlier row equals it. So does
+    every set operation but UNION ALL: UNION, INTERSECT, EXCEPT, EXCEPT ALL
+    and MINUS all decide a row's fate by looking at the other arm's rows.
+    UNION ALL alone just concatenates, and is the one safe kind. Read per
+    arm, so an arm that is itself DISTINCT under a UNION ALL counts too.
+
+    This matters because pruning rests on an unread column being free to
+    drop, and a row-to-row comparison uses every column projected, read or
+    not. Drop ``drop_me`` from
+
+        SELECT grp, keep, drop_me FROM t WHERE id <= 2
+        UNION
+        SELECT grp, keep, drop_me FROM t WHERE id >= 3
+
+    and two rows fold into one, so SUM(keep) answers 30 where it owes 40.
+    Nothing raises; the number just gets quieter.
+
+    Callers put such a scope in ``unprunable``, which keeps its projection
+    whole and stops a narrowed demand reaching a source it stars.
     """
     return any(
         (arm.quantifier or "").upper() == "DISTINCT"
@@ -1037,14 +1052,14 @@ def _scopes_readers_first(
     cte_names: set[str],
 ) -> list[str]:
     """
-    Topologically sort the scopes on reads, readers ahead of what they read.
+    Topologically sort the scopes based on read order.
 
     A CTE is only reached once every scope reading it has been, so the demand
     on it is complete by then. Pruning a reader only narrows what it asks of
     its own sources, so one pass in this order settles every projection.
 
-    Returns the CTE names plus ``""`` for the outer select, which nothing
-    reads and which therefore always comes first.
+    Returns the CTE names plus ``""`` for the outer select. No scope reads
+    the outer select, so it always comes first.
     """
     reads: dict[str, set[str]] = {}
     for key, scope in scopes.items():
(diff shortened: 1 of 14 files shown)


betodealmeida:
Have you considered having DJ generate the full query, and let `sqlglot` trim before transpilation? Something like this:

```python
expression = sqlglot.parse_one(rendered, read="spark")
expression = qualify(
    expression,
    schema=_sqlglot_schema(query),
    dialect="spark",
    identify=False,
)
# pushdown_projections will remove unused columns from CTEs/subqueries, propagating requirements upstream
expression = pushdown_projections(
    expression,
    schema=_sqlglot_schema(query),
    dialect="spark",
)
return expression.sql(dialect=output_dialect)
```

I asked Claude to test your unit tests with the code above and it said it works, but I don't fully understand the tradeoffs. We might not want to use the `sqlglot` approach if we there are use cases where we never convert the DJ AST to the `sqlglot` AST. I'm not sure how much our codepath uses `sqlglot` today, if it's always present or not.

shangyian:
@betodealmeida today `sqlglot` is mostly used for transpilation and whoever is configuring the server can either choose sqlglot or their own library as the transpiler. The AST itself is completely separate... I think my concern here is that for example `_sqlglot_schema` looks like a private function, so it's not clear that we'd want this type of dependency on an external library. Separately, we could consider cutting over to sqlglot entirely rather than only using it for transpilation, but that would be a pretty big change.

betodealmeida:
> @betodealmeida today `sqlglot` is mostly used for transpilation and whoever is configuring the server can either choose sqlglot or their own library as the transpiler. The AST itself is completely separate... 

Ah, I did see the plugin for `sqlglot`, but I didn't see any other plugin so I wasn't sure if there's ever a path where it's not used.

> I think my concern here is that for example `_sqlglot_schema` looks like a private function, so it's not clear that we'd want this type of dependency on an external library.

Oh, that's the function I added here:

https://github.com/DataJunction/dj/pull/2484/changes#diff-3fb805525c6e047bf075307cafbfecbc4dcdc4a0e4c2e4482d77baa6fe985718R200

It converts from a DJ schema to a `sqlglot` schema.

> Separately, we could consider cutting over to sqlglot entirely rather than only using it for transpilation, but that would be a pretty big change.

I really like that we have our own AST, it adds complexity but makes the project more independent. It's such an integral part of DJ that I think we should use `sqlglot` just for the final transpilation.

shangyian:
@betodealmeida ah I missed that, thanks for flagging `_sqlglot_schema`. I think if we're going to retain DJ's AST, we should just enable the pruning as a part of the SQL generation process, so that if someone does want to use DJ standalone without sqlglot, that they can continue to do so.

datajunction-server/datajunction_server/construction/build_v3/cte.py
```diff
@@ -800,65 +803,357 @@ def filter_cte_projection(
...
+
+
+def _arm_sources(arm: ast.SelectExpression, cte_names: set[str]) -> dict[str, str]:
+    """
+    Map every qualifier an arm can use for a CTE to that CTE's name.
+
+    A table can be addressed by its alias or by the CTE name itself, so both
+    land in the map.
+    """
+    sources: dict[str, str] = {}
```
betodealmeida, line 1004: I don't think having a single dict here works, because aliases can be shadowed. For example, this SQL:

```sql
WITH
  a AS (SELECT id, keep FROM t1),
  b AS (SELECT marker FROM t2)
SELECT x.keep, id
FROM a AS x  -- sources["x"] = "a"
WHERE EXISTS (
  SELECT 1 FROM b AS x WHERE x.marker = 1  -- sources["x"] = "b"!
)
```

Would be pruned to:

```sql
WITH
  a AS (SELECT id FROM t1),  -- no keep!
  b AS (SELECT marker FROM t2)
SELECT x.keep, id
FROM a AS x
WHERE EXISTS (
  SELECT 1 FROM b AS x WHERE x.marker = 1
)
```

datajunction-server/datajunction_server/construction/build_v3/cte.py
```diff
@@ -800,65 +803,357 @@ def filter_cte_projection(
...
+    keeping a name a CTE does not have is a no-op.
+
+    A qualifier reaches us two ways. Dotted into the name is the parsed form,
+    and it can carry further struct segments. Hung off the column as a table is
+    what the builder produces for its own projections and GROUP BY, and there
+    the column name stands alone.
+    """
+    bare_targets = set(sources.values())
+    for part in _arm_parts(arm):
+        for column in part.find_all(ast.Column):
```
betodealmeida, line 1037: This will fail to identify a `USING`, no? Like here in this SQL:

```sql
SELECT a.keep FROM a JOIN b USING (id)
```

I think it would miss that `id` is being used.

datajunction-server/datajunction_server/construction/build_v3/cte.py
```diff
@@ -800,65 +803,357 @@ def filter_cte_projection(
...
+    """
+    Map every qualifier an arm can use for a CTE to that CTE's name.
+
+    A table can be addressed by its alias or by the CTE name itself, so both
+    land in the map.
+    """
+    sources: dict[str, str] = {}
+    for part in _arm_parts(arm):
+        for table in part.find_all(ast.Table):
+            name = table.name.identifier(quotes=False)
```
betodealmeida, line 1007: This is bit tricky if we ever want to support Snowflake (and some other DBs) in the future, where:

```
FROM Foo       -- normalized to FOO
FROM foo       -- also normalized to FOO
FROM "Foo"     -- exact, case-sensitive Foo
FROM "foo"     -- exact, case-sensitive foo
```

We might have to introduce a dialect-specific normalization function. But I think it's OK for now.

shangyian: Yeah, I think the dialect story is already a bit fuzzy with the way this AST is setup.

datajunction-server/datajunction_server/construction/build_v3/cte.py
shangyian: Let me address this in a separate PR! 

(Sorry I was trying to get a fix in as the original pruning setup is breaking some workflows)

PR #2498, opened 2026-09-02
feat: endpoint for dimensional values

betodealmeida:
### Summary

For the [semantic layer API](https://github.com/betodealmeida/sl-rest-api/blob/main/SPEC.md#post-viewsview_namevalues) we need to expose an endpoint that returns distinct values for a given dimension, in order to populate dropdowns:

```
POST /views/{view_name}/values

{
    "additional_configuration": {},
    "dimension": "sales.region",
    "filters":   []
}
```

This PR adds a new endpoint at `/sql/dimensions/v3/`, for symmetry with the existing `/sql/metrics/v3/` endpoint. The new endpoint can be used to return the distinct combination of dimension attributes from a single dimension node, producing a query like:

```sql
SELECT DISTINCT dim1, ..., dimn FROM dim_node_table;
```

A following PR will implement the actual endpoint for the semantic layer API, leveraging the code path from this PR.

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/api/djsql.py
+++ b/datajunction-server/datajunction_server/api/djsql.py
@@ -3,6 +3,8 @@ Data related APIs.
 """
 
 import logging
+from dataclasses import dataclass
+from typing import Literal
 
 from fastapi import Depends, Query, Request
 from pydantic import BaseModel
@@ -49,19 +51,39 @@ class TranslatedDJSQL(BaseModel):
     dialect: str
 
 
-def selects_from_metrics(select: ast.SelectExpression) -> bool:
-    """Check if a SELECT sources from the 'metrics' table."""
-    return (
-        select.from_ is not None
-        and len(select.from_.relations) == 1
-        and len(select.from_.relations[0].extensions) == 0
-        and str(select.from_.relations[0].primary).lower() == "metrics"
-    )
+@dataclass(frozen=True)
+class ParsedDJSQL:
+    """Validated DJ SQL components and selected pseudo-table."""
+
+    source: Literal["metrics", "dimensions"]
+    metrics: list[str]
+    dimensions: list[str]
+    filters: list[str]
+    orderby: list[str]
+    limit: int | None
+
+
+def get_pseudo_table(
+    select: ast.SelectExpression,
+) -> Literal["metrics", "dimensions"] | None:
+    """Return the single unjoined pseudo-table selected by the query."""
+    if (
+        select.from_ is None
+        or len(select.from_.relations) != 1
+        or select.from_.relations[0].extensions
+    ):
+        return None
+    table = str(select.from_.relations[0].primary).lower()
+    if table == "metrics":
+        return "metrics"
+    if table == "dimensions":
+        return "dimensions"
+    return None
 
 
 def parse_dj_sql(
     query: str,
-) -> tuple[list[str], list[str], list[str], list[str], int | None]:
+) -> ParsedDJSQL:
     """
     Parse a DJ SQL query and extract metrics, dimensions, filters, orderby, limit.
 
@@ -75,17 +97,17 @@ def parse_dj_sql(
             LIMIT 10
 
     Returns:
-        Tuple of (metrics, dimensions, filters, orderby, limit)
+        Parsed pseudo-table and query components.
 
     Note: Validation of metric/dimension nodes is delegated to build_metrics_sql.
     """
     tree = parse(query)
     select = tree.select
 
-    if not selects_from_metrics(select):
+    source = get_pseudo_table(select)
+    if source is None:
         raise DJInvalidInputException(
-            "DJ SQL queries must SELECT FROM metrics. "
-            "Example: SELECT metric1, dim1 FROM metrics GROUP BY dim1",
+            "DJ SQL queries must SELECT FROM metrics or dimensions.",
         )
 
     # Validate no unsupported clauses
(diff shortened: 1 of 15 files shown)


datajunction-server/datajunction_server/construction/build_v3/node_query.py
```diff
@@ -72,6 +79,100 @@
...
+
+    requested_refs = [parse_dimension_ref(dim) for dim in requested]
+    available = {column.name for column in dimension_node.current.columns}  # type: ignore[union-attr]
+    missing = [ref.column_name for ref in refs if ref.column_name not in available]
+    if missing:
+        raise DJInvalidInputException(
+            f"Dimension `{node_name}` does not contain columns: {missing}",
+        )
+
+    cte_pairs, _, _ = collect_node_ctes(ctx, [dimension_node])
```
shangyian, line 162: Hmm, actually the way this query is built it looks like it's selecting DISTINCT on the dimension node directly. Does that mean we don't need the check upstream that says "Metricless queries cannot select attributes"?

datajunction-server/datajunction_server/api/djsql.py
```diff
@@ -119,6 +119,9 @@ def parse_dj_sql(
         if col_ident not in dimensions:
             metrics.append(col_ident)
 
+    if not metrics:
```
shangyian: Could DJ SQL also support dimensional values queries?

datajunction-server/tests/api/cubes_test.py
```diff
@@ -1234,6 +1234,33 @@ async def test_cube_filters_merged_with_request_filters(
...
+
+    response = await client_with_repairs_cube.get(
+        "/sql/dimensions/v3/",
+        params={
+            "dimensions": ["default.dispatcher.company_name"],
+            "cube": "default.repairs_cube",
+        },
+    )
+    assert response.status_code == 422
+    assert "exactly one node" in response.json()["message"]
```
shangyian: How come this one fails -- is it because the dimension attribute is not part of the a cube? Should it still be possible to retrieve the dimensional values as long as it's reachable from the metrics in the cube?

datajunction-server/tests/api/cubes_test.py
```diff
@@ -1234,6 +1234,33 @@ async def test_cube_filters_merged_with_request_filters(
...
+    response = await client_with_repairs_cube.get(
+        "/sql/dimensions/v3/",
+        params={
+            "dimensions": ["default.hard_hat.state"],
+            "cube": "default.repairs_cube",
+        },
+    )
+    assert response.status_code == 200
+    assert "state = 'AZ'" in response.json()["sql"]
+    assert "num_repair_orders" not in response.json()["sql"]
```
shangyian: Can this assert the full sql query so that we have a record of what's generated for cubes?

datajunction-server/tests/construction/build_v3/metrics_sql_test.py
```diff
@@ -4,6 +4,147 @@
...
+
+    @pytest.mark.asyncio
+    async def test_dimension_only_query_from_transform(self, client_with_build_v3):
+        response = await client_with_build_v3.get(
+            "/sql/dimensions/v3/",
+            params={"dimensions": ["v3.order_details.status"]},
+        )
+
+        assert response.status_code == 200, response.json()
+        assert "DISTINCT" in response.json()["sql"]
```
shangyian: Can this assert the full sql comparison?

datajunction-server/datajunction_server/api/sql.py
```diff
@@ -566,6 +569,58 @@ async def get_combined_measures_sql_v3(
...
+@router.get(
+    "/sql/dimensions/v3/",
+    response_model=V3TranslatedSQL,
+    name="Get Dimensions SQL V3",
+    tags=["sql", "v3"],
+)
+async def get_dimensions_sql_v3(
+    dimensions: list[str] = Query([]),
+    filters: list[str] = Query([]),
+    cube: str | None = Query(None),
```
shangyian: fwiw, a cube can also have filters, so would these two filters end up layering on top of each other with an `AND`? Also if filters are exposed here, should it be possible to specify an arbitrary set of metrics + dims + filters, alongside a cube?

datajunction-server/datajunction_server/api/djsql.py
betodealmeida: Ah, that's a great idea! I added:

```sql
SELECT dim FROM dimensions GROUP BY dim
```

datajunction-server/datajunction_server/api/sql.py
betodealmeida: Yes, for filters we `AND` them, so that if we don't hit the materialization we preserve the cube filter and the requested filter.

Here I added `cube` as an option to narrow down the valid dimensional values given a context, but I think supporting a full set of metrics, dimensions, and filters would require this endpoint to become a general semantic-query endpoint.

datajunction-server/datajunction_server/construction/build_v3/node_query.py
betodealmeida: The guard prevents someone from passing a metric node name to the API and getting invalid/broken SQL.

datajunction-server/tests/api/cubes_test.py
betodealmeida: Yeah, the initial logic required the dimension and filter from the same node. I changed it to support reachable dims across nodes.

shangyian (approved):
LGTM! Excited for this

Issue #2245, opened 2026-06-07
Proposal: Level (and other Derived) Metrics


shangyian:
*Reframe [#1695 "Support Multi-Scope Ratio Metrics"](https://github.com/DataJunction/dj/issues/1695).*

## Problem

DJ stores a metric as a SQL string (`SELECT <expr> FROM <node>`) and recovers its structure by parsing that string back. However, using SQL as the storage format causes a number of issues:

- **Off-query-grain metrics can't be expressed.** A metric computed *coarser* than the query (share-of-total) or *finer then rolled up* (weighted median, average of daily peaks) has no representation. Both are forced into a hand-built pre-aggregation transform, which hides the grain from the semantic layer and can't be sliced freely.
- **Definitions are ambiguous.** `IF(MAX(a) >= 3, 1, 0)` means *either* "threshold the query-grain max" *or* "a per-entity boolean, then roll up". These result in different numbers, but the SQL can't say which is the right one.
- **Structure must be reverse-engineered to do anything.** Re-graining a window re-parses its `OVER`; a `ROWS` frame silently assumes dense rows at one grain; leveling a ratio means discovering the division. This is fragile and only works for shapes the parser recognizes, but...
- **The contract is far too wide.** Any valid SQL is storable which may lead to malformed or ambiguous definitions

## Proposal

A metric's structure is the definition; SQL is a compile target, not the storage format. Metric structure include *kind* (atomic or derived) and two aggregate properties (*grain*, *frame*):
```
metric
├─ atomic  = aggregate(expr) over a node            ← KIND
│     ├─ fixed_grain   (default: query grain)       ← GRAIN
│     ├─ reaggregate   (default: auto)              ← GRAIN
│     └─ frame?        (optional ordered window)    ← FRAME
└─ derived = expression over other metrics          ← KIND
      └─ frame?        (optional, e.g. LAG over a metric)
```

### Kind

- **Atomic** = exactly one aggregate over one node. The inner expression is unrestricted (`SUM(CASE …)`, `SUM(a*b)`).
- **Derived** = a pure expression over *other metrics*, no raw aggregation (`a / b`).

Disallow arithmetic that mixes aggregates inside one metric: `SUM(a)/SUM(b)` becomes atomics `a`, `b` + derived `a/b`. This removes the ambiguity, since every aggregate is a named value at the query grain, and everything on top is scalar arithmetic with exactly one meaning.

### Grain

A grain defines where an aggregate is computed, and how it rolls up. 

[...]

PR #2502, opened 2026-09-03
feat: semi-additive measures

betodealmeida:
### Summary

This PR adds metric reaggregation support for semi-additive measures, aligned with the proposal https://github.com/DataJunction/dj/issues/2245. Metrics can declare protected dimensions, and the aggregation behavior changes when the query grain drops the dimension.

For example, suppose we have this metric:

```yaml
name: v3.daily_balance
type: metric
query: SELECT SUM(line_total) FROM v3.order_details
reaggregate:
  rules:
    - dimension: v3.date.date_id[order]
      fn: last_value
```

Meaning:

> `daily_balance` is additive within a day, but **not additive across order date**. If date disappears from the query grain, collapse with the last date’s value.

If we get a request that includes the protected dimension:

```text
metrics: [v3.daily_balance]
dimensions: [v3.date.date_id[order]]
```

We generate this SQL:

```sql
WITH order_details_0 AS (
  SELECT
    order_date AS date_id_order,
    SUM(line_total) AS line_total_sum
  FROM v3.order_details
  GROUP BY order_date
)
SELECT
  date_id_order,
  SUM(line_total_sum) AS daily_balance
FROM order_details_0
GROUP BY date_id_order
```

Because `date_id_order` is still in the output grain, there is no semi-additive collapse. Normal aggregation is safe.

On the other hand, if we don't request the protected dimension we shouldn't aggregate over it:

```text
metrics: [v3.daily_balance]
dimensions: [v3.product.category]
```

The generated SQL:

```sql
WITH order_details_0 AS (
  SELECT
    product_category AS category,
    order_date AS date_id_order,
    SUM(line_total) AS line_total_sum
  FROM v3.order_details
  GROUP BY product_category, order_date
)
SELECT
  category,
  MAX_BY(line_total_sum, date_id_order) AS daily_balance
FROM order_details_0
GROUP BY category
```

Here, the user asked for `category only`, but DJ keeps `date_id_order` as a private inner grain. Then it collapses each category’s daily balances with `MAX_BY(value, date)`, meaning "take the value from the latest date."

Without this, DJ would generate something like:

```sql
SELECT
  product_category AS category,
  SUM(line_total) AS daily_balance
FROM v3.order_details
GROUP BY product_category
```

That would incorrectly sum balances across dates, which is wrong for snapshots or balances, for example.

I've updated the UI to show and allow setting the semi-additive dimension:

Screenshot 2026-09-04 at 2 23 46 PM

When editing:

Screenshot 2026-09-04 at 2 24 22 PM

Valid options:

Screenshot 2026-09-04 at 2 24 27 PM

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/construction/build_v3/builder.py
```diff
@@ -320,6 +321,14 @@ async def setup_build_context(
...
+    internal_reaggregate_dimensions = missing_reaggregate_dimensions(
+        ctx.decomposed_metrics.values(),
+        output_dimensions_after_expression_scan,
+    )
+    for dimension in internal_reaggregate_dimensions:
+        if dimension not in ctx.dimensions:
+            ctx.dimensions.append(dimension)
 
     # A second load_nodes pass is needed when either:
     # 1. metric expressions introduced dimension nodes not yet in ctx.nodes, OR
```
shangyian, line 334: If this is just a parent-column protected dimension (e.g., it's not the fully qualified node name like `v3.order_details.order_date` but just `order_date`), then it validates cleanly here but fails at query time:

```yaml
POST /nodes/metric/
{"query": "SELECT SUM(line_total) FROM v3.order_details",
 "reaggregate": {"rules": [{"dimension": "order_date", "fn": "last_value"}]}}
  -> 201, status: valid

GET /sql/metrics/v3/?metrics=v3.balance&dimensions=v3.product.category
  -> 422 "Reference `order_date` is not fully qualified. Use the `node.column` form..."
```

Should this just reject a non-fully-qualified name (and I think the UI might need to change based on that as well)?

datajunction-ui/src/app/pages/AddEditNodePage/ReaggregateFields.jsx
```diff
@@ -0,0 +1,79 @@
...
+  const { values } = useFormikContext();
+  const [dimensionField] = useField('reaggregate_dimension');
+  const [dimensionOptions, setDimensionOptions] = useState([]);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      if (values.upstream_node) {
+        const data = await djClient.node(values.upstream_node);
+        setDimensionOptions(
+          data.columns.map(col => ({
```
shangyian, line 23: This is related to the above comment -- here it's not producing fully qualified column names, just the column name itself, but then those metrics will fail to generate sql.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
```diff
@@ -1394,62 +1583,107 @@ def build_window_agg_cte_from_base_metrics(
...
+
+        if decomposed.aggregability == Aggregability.NONE:  # pragma: no cover
+            # NONE: non-additive (like AVG), need to recompute.
+            # For AVG, we need the raw sum and count, but those are in components.
+            # For now, just use the column directly (window function will handle it).
+            return metric_ref
+
+        # base_metrics exposes metric columns, not raw component/grain columns.
+        # Reaggregate leaf metrics from those projected metric values.
+        return ast.Function(ast.Name("SUM"), args=[metric_ref])
```
shangyian, line 1630: Hmm, does this make distinct counts additive? For example let's say `base_metrics` is grouped by `(date_id, category, week)` and the outer query
collapses `date_id` away. If we have this setup:

| date | visitor_count |
| --- | --- |
| 2026-01-05 | 2 |
| 2026-01-06 | 1 |

The change here will return `SUM(visitor_count)` which isn't the distinct visitor count. The removed code here had a special case for `Aggregability.LIMITED` with `COUNT(DISTINCT grain_col)`.

To be fair I think the old form was broken too, since `base_metrics` projects `visitor_count` and not `customer_id`, so `COUNT(DISTINCT base_metrics.customer_id)` referenced a missing column and would have errored.

Since this is in a path unrelated to semi-additive measures... could it be split out? We can have a separate PR that addresses this issue with its own tests.

datajunction-server/datajunction_server/construction/build_v3/measures.py
```diff
@@ -2375,32 +2458,42 @@ def process_metric_group(
         components_per_metric[decomposed.metric_node.name] = len(decomposed.components)
 
     # Analyze grain groups - split by aggregability
-    # Extract just the column names from dimensions for grain analysis
-    dim_column_names = [parse_dimension_ref(d).column_name for d in ctx.dimensions]
-    grain_groups = analyze_grain_groups(metric_group, dim_column_names)
+    output_dimensions = list(ctx.dimensions)
```
shangyian, line 2461: Hmm, it looks like a filter on the protected dimension will make the metric additive?

For example if we test with the same metric (`v3.daily_balance`) + dimension but with and without a filter:

**correct** `metrics=[v3.daily_balance] + dimensions=[v3.product.category]`

```sql
WITH
v3_order_details AS (
SELECT  o.order_date,
      oi.product_id,
      oi.quantity * oi.unit_price AS line_total
 FROM default.v3.orders o JOIN default.v3.order_items oi ON o.order_id = oi.order_id
),
v3_product AS (
SELECT  product_id,
      category
 FROM default.v3.products
),
order_details_0 AS (
SELECT  t2.category,
      t1.order_date date_id_order,
      SUM(t1.line_total) line_total_sum_e1f61696
 FROM v3_order_details t1 LEFT OUTER JOIN v3_product t2 ON t1.product_id = t2.product_id
 GROUP BY  t2.category, t1.order_date
)

SELECT  order_details_0.category AS category,
      MAX_BY(order_details_0.line_total_sum_e1f61696, order_details_0.date_id_order) AS daily_balance
 FROM order_details_0
 GROUP BY  order_details_0.category
```

**incorrect** `metrics=[v3.daily_balance] + dimensions=[v3.product.category] + filters=["v3.date.date_id[order] >= 20260101"]`

```sql
-- order_date disappears from the CTE's GROUP BY and MAX_BY becomes SUM
WITH
v3_order_details AS (
SELECT  oi.product_id,
      oi.quantity * oi.unit_price AS line_total
 FROM default.v3.orders o JOIN default.v3.order_items oi ON o.order_id = oi.order_id
 WHERE  o.order_date >= 20260101
),
v3_product AS (
SELECT  product_id,
      category
 FROM default.v3.products
),
order_details_0 AS (
SELECT  t2.category,
      SUM(t1.line_total) line_total_sum_e1f61696
 FROM v3_order_details t1 LEFT OUTER JOIN v3_product t2 ON t1.product_id = t2.product_id
 GROUP BY  t2.category
)

SELECT  order_details_0.category AS category,
      SUM(order_details_0.line_total_sum_e1f61696) AS daily_balance
 FROM order_details_0
 GROUP BY  order_details_0.category
```

So "daily balance by category" gives the latest balance per
category, and "daily balance by category, for January" gives the sum of every
daily balance in January, which is a different metric.

I think it does work for the cube path though, because `build_synthetic_grain_group` strips filter dims first.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
```diff
@@ -207,6 +353,31 @@ def get_comp_aggregability(comp_name: str) -> Aggregability:
         comp = decomposed.components[0]
```
shangyian: The reaggregate branch sits inside `if len(decomposed.components) == 1:` (just above). `_attach_reaggregate_spec` guards the base-metric case, but a derived metric inlines one component per base, so `ratio = v3.balance_a / v3.balance_b` — both semi-additive on the same parent and the same protected dimension — produces two components in one grain bucket.

`gg.reaggregate_dimension_aliases` is populated for both, `_validate_reaggregate_base_group_join_safety` returns early on `len(grain_groups) <= 1`, and the generic `deepcopy(combiner_ast)` path emits `SUM(cte.a_sum) / NULLIF(SUM(cte.b_sum), 0)` over a CTE grouped by the protected dimension — summing across it.

The sharp edge is that in the same response the two base metrics *are* collapsed, so the ratio disagrees with its own numerator and denominator. This seems like it should fail closed the way the other unsupported shapes do. Reported from reading.

datajunction-server/datajunction_server/internal/nodes.py
```diff
@@ -886,6 +896,8 @@ async def _derive_frozen_measures_impl(
             session=session,
             name=measure.name,
         )
+        if frozen_measure:
+            _raise_if_frozen_measure_conflicts(frozen_measure, measure)
         if not frozen_measure and measure.aggregation:
             frozen_measure = FrozenMeasure(
                 name=measure.name,
```
shangyian, line 903: The two frozen-measure paths (deployment vs direct metrics creation) seem to differ:
- direct metrics creation (this one) keeps the reaggregate: `rule=measure.rule`
- deployment strips it with `rule=_frozen_measure_rule(measure.rule)`

I think stripping makes more sense since the reaggregate isn't a property of the measure but rather something that can be applied on top

datajunction-server/datajunction_server/api/cubes.py
```diff
@@ -288,6 +317,7 @@ async def cube_materialization_info(
             message=f"Cube node `{name}` does not exist.",
             http_status_code=404,
         )
+    await _validate_cube_reaggregate_materialization(session, node)
```
shangyian, line 320: Should this part first call the pre-check `cube_matcher._metric_graph_has_reaggregate` (since it's already used in other APIs like in `find_matching_cube`)?

datajunction-server/datajunction_server/models/reaggregate.py
```diff
@@ -0,0 +1,113 @@
...
+    dimension: str
+    fn: ReaggregationFunction
+
+
+class ReaggregateSpec(BaseModel):
+    """
+    Declaration for how a metric rolls up from its accumulation grain.
+    """
+
+    fn: ReaggregationFunction | None = None
```
shangyian, line 75: Are`fn` and `weight` used right now in sql gen or is this meant to be for later? Wonder if it's worth dropping until there's a specific use for these fields

datajunction-server/datajunction_server/construction/build_v3/measures.py
```diff
@@ -2211,7 +2259,12 @@ def build_grain_group_sql(
     columns_metadata = []
 
     # Add dimension columns (skip filter-only dimensions as they're not in projection)
+    effective_output_dimension_refs = output_dimension_refs or {
```
shangyian: `output_dimension_refs or {...}` treats "no dimensions requested" as "no filtering", since an empty set is falsy.

With zero requested dimensions and one internal protected dimension — `GET /sql/metrics/v3/?metrics=v3.daily_balance` — `resolved_dimensions` holds only the protected dim, so it is emitted as a regular dimension in `columns_metadata` *and* again by the new `internal_dimension_aliases` loop: a duplicate column in the grain-group metadata, and a private dimension presented as user-facing.

`if output_dimension_refs is None` matches how `build_select_ast` tests the same parameter.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
```diff
@@ -50,13 +50,159 @@
...
+        return None
+    return ref.rsplit("[", 1)[1].rstrip("]")
+
+
+def _source_dimension_alias(ctx: BuildContext, dimension_ref: str) -> str | None:
+    """
+    Return the registered source column alias for a dimension ref.
+    """
+    alias = ctx.alias_registry.get_alias(dimension_ref)
+    if alias or _dimension_ref_role(dimension_ref) is not None:
```
shangyian, line 158: it looks like when the protected dimension has a role and there's no alias in the registry, we'll end up falling through to `SUM(metric_ref)`. Should this raise an "unsupported semi-additive shape" error in that case?

datajunction-server/datajunction_server/construction/build_v3/cube_matcher.py
```diff
@@ -53,6 +62,327 @@
 # parsed: coverage can't be proven, so cube matching must fail SAFE (reject the
 # cube and fall back) rather than fail open and emit invalid SQL.
 _FILTER_COVERAGE_UNKNOWN = object()
+ReaggregateRequirement = tuple[str, str, ReaggregationFunction]
+
+
+def _split_dimension_ref(ref: str) -> tuple[str, str | None]:
```
shangyian, line 68: Instead of adding another dimension-with-role parser, can this just reuse `parse_dimension_ref` from `construction/build_v3/dimensions.py` instead?

It actually looks like there's quite a few versions of this function that got added here: 
- `_dimension_ref_base` / `_dimension_ref_role` (in `build_v3/decomposition.py`)
- `_dimension_ref_role` again in  (`build_v3/metrics.py`)
- `_split_dimension_ref` here
- an inline `dim_ref.rsplit("[", 1)` in `build_v3/dimensions.py`

datajunction-query/uv.lock
```diff
@@ -1,10 +1,11 @@
 version = 1
 revision = 3
-requires-python = ">=3.10, <4.0"
+requires-python = ">=3.11, <4.0"
```
robinld, line 3: Did we intend to change this?

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -984,6 +995,12 @@ def _build_metric_data_from_cache(
         ]
 
         if metric_parents:
+            if metric_node.current.reaggregate:
+                raise DJInvalidInputException(
+                    "Semi-additive metrics must be base metrics in V1. "
```
robinld, line 1000: could we push this check into validate_node_data_v2?

datajunction-server/datajunction_server/models/reaggregate.py
betodealmeida: They're not used currently, I added because I started to implement https://github.com/DataJunction/dj/issues/2245. I agree, let's drop it for now.

datajunction-server/datajunction_server/construction/build_v3/builder.py
betodealmeida: Ah, thanks for catching this, I missed it! Will fix it here and below.

datajunction-server/datajunction_server/api/cubes.py
betodealmeida: Yeah, I don't think it needs to be (for correctness), but it would be nice for performance. Let me add the check.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
betodealmeida: I'm not super familiar with roles, but when wouldn't it be not present in the alias registry?

Regardless, raising when not present is the safe option, let me implement that.

betodealmeida: I changed the logic to not use the registry at all, we use the column names from the CTEs.

datajunction-server/datajunction_server/construction/build_v3/cube_matcher.py
betodealmeida: Yeah, good point. I think we might need a lower-level helper function, since `parse_dimension_ref` requires a fq node.column reference, while the reagg path accepts bare references. Let me refactor it, I think it will be cleaner.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
betodealmeida: Ah, good catch. I'll add a fail-closed guard to this PR, and fix the issue in a followup.

datajunction-server/datajunction_server/internal/nodes.py
betodealmeida: Ah, good point, this should be metric-level, not part of the frozen measure. Let me add the logic to strip it consistently.

datajunction-server/datajunction_server/construction/build_v3/measures.py
betodealmeida: Oi, thanks for catching this. Let me fix it.

datajunction-query/uv.lock
betodealmeida: Oops, not intentional, let me revert.

datajunction-server/datajunction_server/sql/decompose.py
betodealmeida: Ah, let me do it!

shangyian (approved):
LGTM!

PR #2505, opened 2026-09-03
feat: return metadata in the SL API

betodealmeida:
### Summary

This PR adds [richer metadata to the semantic layer API](https://github.com/betodealmeida/sl-rest-api/blob/main/SPEC.md#metadata-schema) responses, allowing clients to provide a better UX. For example, this DJ metric:

```yaml
---
type: metric
name: finance.total_revenue
display_name: Total Revenue
description: Total recognized revenue
query: SELECT SUM(revenue) FROM finance.orders
unit:
  kind: currency
  code: USD
```

Is returned as:

```json
 {
   "id": "finance.total_revenue",
   "name": "total_revenue",
   "type": "floating",
   "definition": "finance.total_revenue",
   "description": "Total recognized revenue",
   "aggregation": "OTHER",
   "metadata": {
     "display_name": "Total Revenue",
     "semantic_type": "currency",
     "unit": {
       "kind": "currency",
       "code": "USD"
     },
     "format": {
       "preset": "currency"
     },
     "filter": {
       "kind": "number",
       "operators": [
         "=",
         "!=",
         ">",
         ">=",
         "<",
         "<=",
         "IS NULL",
         "IS NOT NULL"
       ],
       "default_operator": "="
     },
     "extensions": {
       "superset": {
         "d3format": "$,.2f"
       },
       "google_sheets": {
         "numberFormat": {
           "type": "CURRENCY",
           "pattern": "$#,##0.00"
         }
       }
     }
   }
 }
```

Fields are generated conservatively based on DJ metadata, while allowing for custom overrides:

```yaml
 type: metric
 name: finance.total_revenue
 display_name: Total Revenue
 description: Total recognized revenue
 query: SELECT SUM(revenue) FROM finance.orders

 unit:
   kind: currency
   code: USD

 custom_metadata:
   semantic_layer:
     display_name: Net Revenue

     format:
       preset: currency
       precision: 0

     filter:
       kind: range
       operators:
         - "="
         - ">"
         - ">="
         - "<"
         - "<="
       default_operator: ">="

     extensions:
       superset:
         d3format: "$,.0f"

       google_sheets:
         numberFormat:
           type: CURRENCY
           pattern: "$#,##0"
```

@shangyian I wonder if it makes sense to eventually consolidate the "semantic" APIs that DJ has? 

### Test Plan

- [ ] `make check` passes


shangyian:
@betodealmeida can you clarify what you mean by consolidate semantic APIs? Do you mean the various semantic metadata APIs that we provide (e.g., under `GET /metrics`)?

datajunction-server/datajunction_server/internal/seed.py
```diff
@@ -4,17 +4,53 @@
...
+    SEMANTIC_LAYER_METADATA_KEY,
+    SEMANTIC_LAYER_METADATA_SCHEMA,
+)
 from datajunction_server.utils import get_settings
 
 logger = logging.getLogger(__name__)
 settings = get_settings()
 
 
+async def seed_default_custom_metadata_schemas(session: AsyncSession) -> None:
```
shangyian: nice!

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -102,21 +119,326 @@ def _cube_column_type_map(cube: NodeRevision) -> dict[str, str | None]:
...
+    DJ custom metadata.
+    """
+    for element in getattr(cube, "cube_elements", []):
+        revision = element.node_revision
+        if revision is None:
+            continue
+        if revision.type == NodeType.METRIC:
+            if revision.name != column_id:
+                continue
+            raw = revision.custom_metadata or {}
```
shangyian, line 301: Hmm, so looking at this part, when a node's `custom_metadata` has no `semantic_layer` key, `raw` becomes the entirety of `custom_metadata` right? Is that intentional? I thought this would only return the dict under the `semantic_layer` key, since all other items on the the `custom_metadata` could be completely unrelated.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -102,21 +119,326 @@ def _cube_column_type_map(cube: NodeRevision) -> dict[str, str | None]:
...
+            if isinstance(values, Mapping)
+        }
+        if isinstance(raw_metadata.get("extensions"), Mapping)
+        else {}
+    )
+    producer_metadata = {
+        key: value for key, value in raw_metadata.items() if key not in _METADATA_KEYS
     }
+    if producer_metadata:
+        extensions["datajunction"] = {
```
shangyian: This promotion is the mechanism behind the leak, and it is also unreachable for validated writes — the schema this PR seeds sets `additionalProperties: False` on the `semantic_layer` object, so stray keys inside the wrapper are rejected before they could ever get here.

That leaves it serving only the unwrapped path, which is the case I do not think should exist. Unless the intent is to open the schema, I would drop this block.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -102,21 +119,326 @@ def _cube_column_type_map(cube: NodeRevision) -> dict[str, str | None]:
...
+    if unit_data and "kind" in unit_data:
+        unit_kind = unit_data["kind"]
+        if unit_kind == "time":
+            return "duration"
+        if unit_kind in {
+            "currency",
+            "percentage",
+            "proportion",
+            "count",
+            "duration",
```
shangyian, line 140: minor: It looks like `"duration"` is not a member of `UnitKind`

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -102,21 +119,326 @@ def _cube_column_type_map(cube: NodeRevision) -> dict[str, str | None]:
...
+        }
+    elif format_is_explicit and preset == "number":
+        d3format = f",.{precision}f"
+        number_format = {
+            "type": "NUMBER",
+            "pattern": _fixed_decimal_pattern("", precision),
+        }
+    elif format_is_explicit and preset == "percentage":
+        d3format = f".{precision}%"
+        decimals = f".{''.join('0' for _ in range(precision))}" if precision else ""
```
shangyian, line 257: Could this part just reuse `_fixed_decimal_pattern` rather than inlining a similar expression?

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -148,17 +473,23 @@ def _metrics_payload(cube: NodeRevision) -> list["MetricInfo"]:
...
-    metadata_by_name = _cube_metadata_map(cube)
+    columns_by_name = _cube_column_map(cube)
     return [
         DimensionInfo(
             id=dim_ref,
             name=dim_ref.split(".")[-1],
             type=type_by_name.get(dim_ref) or DIMENSION_FALLBACK_ARROW_TYPE_NAME,
             definition=dim_ref,
-            description=None,
+            description=getattr(columns_by_name.get(dim_ref), "description", None),
```
shangyian: `getattr` here is doing hidden double duty as a `None` check on the dict lookup, since `columns_by_name.get(dim_ref)` can be `None`. `columns_by_name[dim_ref].description if dim_ref in columns_by_name else None` says the same thing and matches the pattern used for `metadata` a few lines below.

Same for `getattr(cube, "cube_elements", [])` in `_raw_column_metadata` — both read oddly in a PR whose main cleanup is removing exactly this kind of probing from `_generated_column_arrow_type_name`.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -283,13 +764,7 @@ class MetricInfo(BaseModel):
...
     description: str | None
     aggregation: str
-    metadata: MetricsMetadata
-
-
-class DimensionMetadata(BaseModel):
-    """Dimension-specific field metadata"""
-
-    display_name: str
+    metadata: ColumnMetadata | None = None
```
shangyian: `metadata` moving from required to optional is a breaking change for anyone reading `metric["metadata"]["display_name"]`, which was previously guaranteed to exist. `_column_metadata` returns `None` when everything is empty, and the payload builders also return `None` when the column is missing from the cube, so absence is reachable two ways.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -343,7 +820,11 @@ class GeneratedSQLResponse(BaseModel):
 # ---------------------------------------------------------------------------
 
 
-@router.post("/views/list", response_model=list[ViewSummary])
+@router.post(
+    "/views/list",
+    response_model=list[ViewSummary],
+    response_model_exclude_none=True,
```
shangyian: This drops previously-emitted `null` fields — `description` and `grain` — from the response. Combined with `metadata` becoming optional, that is a payload shape change for existing clients. Worth calling out in the description.

datajunction-server/datajunction_server/models/semantic_layer_metadata.py
```diff
@@ -0,0 +1,158 @@
...
+        },
+        "Extensions": {
+            "type": "object",
+            "additionalProperties": {"type": "object"},
+        },
+    },
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        **_COLUMN_METADATA_PROPERTIES,
```
shangyian: `additionalProperties: False` here is the right contract in my view, but it contradicts the `extensions.datajunction` promotion in `_column_metadata` and the test that asserts it — stray keys under `semantic_layer` cannot survive validation to reach that code.

Worth deciding which one wins. Keeping this strict and dropping the promotion seems cleaner than the reverse.

datajunction-server/datajunction_server/internal/seed.py
```diff
@@ -4,17 +4,53 @@
...
+            node_type=None,
+        )
+        session.add(schema)
+
+    schema.json_schema = SEMANTIC_LAYER_METADATA_SCHEMA
+    schema.value_kind = "object"
+    schema.filterable = False
+    schema.description = SEMANTIC_LAYER_METADATA_DESCRIPTION
+    schema.reserved = True
+    schema.deactivated_at = None
```
shangyian: This reassigns every field and clears `deactivated_at` on each startup, so an operator who deactivated the reserved schema deliberately gets it silently re-enabled on the next restart. Is that intended?

Also, the `commit()` below runs whether or not anything changed — worth skipping when the existing row already matches.

datajunction-server/tests/api/semantic_layer_test.py
```diff
@@ -170,6 +249,408 @@ def test_metric_and_dimension_payloads_fallback_when_type_is_unknown(self):
...
+                    "extensions": {
+                        "superset": {"d3format": "€,.2f"},
+                        "google_sheets": {
+                            "numberFormat": {
+                                "type": "CURRENCY",
+                                "pattern": "€#,##0.00",
+                            },
+                        },
+                    },
+                    "owner": "finance",
```
shangyian, line 277: If I'm reading this setup correctly, this can't actually deploy successfully due to the `owner` field which isn't part of the semantic layer JSON schema. I wonder if you could add a test that actually asserts that failure?

docs/content/0.1.0/docs/data-modeling/custom-metadata.md
```diff
@@ -58,6 +58,44 @@ curl -X POST $DJ_SERVER/metadata-schemas/ \
...
+      country:
+        filter:
+          kind: select
+          operators: [IN, NOT IN]
+          default_operator: IN
+          multi: true
+```
+
+The key is reserved so namespace-scoped schemas cannot redefine the shared
+contract. Producer-specific fields must be placed inside an object under
```
shangyian: "must be placed inside an object under `extensions`" is the behaviour I would want, but the code currently promotes stray top-level keys into `extensions.datajunction` automatically, so it is lenient where this says "must". Whichever way that lands, this sentence and `_column_metadata` should agree.

The surrounding section is otherwise clear — the metric-versus-dimension split reads well.

datajunction-server/datajunction_server/api/semantic_layer.py
```diff
@@ -102,21 +119,326 @@ def _cube_column_type_map(cube: NodeRevision) -> dict[str, str | None]:
...
+    extensions = (
+        {
+            str(namespace): cast(JSONObject, dict(values))
+            for namespace, values in raw_metadata.get("extensions", {}).items()
+            if isinstance(values, Mapping)
+        }
+        if isinstance(raw_metadata.get("extensions"), Mapping)
+        else {}
+    )
+    producer_metadata = {
```
shangyian, line 344: Does this mean that if a node has a `semantic_layer` key under `custom_metadata`, e.g.:
```yaml
custom_metadata:
  semantic_layer:
    format: {preset: currency}
    owner: finance          # a random key
````
Then `raw_metadata` becomes `{format: ..., owner: finance}`, which would fail the JSON schema validation.

However, if a node doesn't have the `semantic_layer` key, e.g.:
```yaml
custom_metadata:
  tier: 2
  something: "abcd"
```

Then `_raw_column_metadata` doesn't find the key, so `raw_metadata` just becomes everything in `custom_metadata`. I assume that's unintentional?

betodealmeida:
> @betodealmeida can you clarify what you mean by consolidate semantic APIs? Do you mean the various semantic metadata APIs that we provide (e.g., under `GET /metrics`)?

Yeah, sorry if I wasn't clear. What I mean is that we have the DJ-native API (`GET /metrics/{name}`, eg) and we have the [semantic layer REST API](https://github.com/betodealmeida/sl-rest-api/blob/main/SPEC.md) (`POST /semantic/views/{name}`), and they also return metadata in different schemas. But I think they might be different enough for us to justify both, at least for now. For example, the `POST /semantic/views/{name}` supports passing additional configuration, so it allows clients to filter out branches they don't want. It's a more user-focused API, compared to the more engineering-centric native API, I think.

shangyian:
@betodealmeida Yeah, I think for simplicity for the time being we might just want to keep both for now. If we do remove or try to consolidate, we'll need to look at existing consumers of those APIs and make sure they're migrated before we can remove.

datajunction-server/datajunction_server/api/semantic_layer.py
betodealmeida: Ah, good catch!

PR #2507, opened 2026-09-03
Fix column pruning for when table aliases are used in subqueries and for `USING` or `NATURAL JOIN`

shangyian:
### Summary

Fixes some bugs with #2496 thanks @betodealmeida!

#### Bug 1: table alias reused inside a subquery

The pass builds one lookup of table alias → block by reading the `FROM` clauses. But it reads the `FROM` clauses inside subqueries too, into the same lookup, so if a subquery reuses an alias for a different block, the subquery's version overwrites the outer one.
```sql
WITH a AS (SELECT id, keep FROM t1),
     b AS (SELECT marker FROM t2)
SELECT x.keep, id
FROM a AS x
WHERE EXISTS (SELECT 1 FROM b AS x WHERE x.marker = 1)
```
The outer query's `x` refers to `a`. The EXISTS subquery's `x` refers to `b`. The lookup ends up saying `x` means `b`, so when the pass sees `x.keep` in the outer select it credits `keep` to `b` and concludes `a` doesn't need `keep` and prunes it:
```sql
WITH a AS (SELECT id FROM t1),          -- keep is gone
     b AS (SELECT marker FROM t2)
SELECT x.keep, id                        -- but this still reads it
FROM a AS x
WHERE EXISTS (SELECT 1 FROM b AS x WHERE x.marker = 1)
```

#### Bug 2: join columns from `USING`

`USING (id)` names a column that both sides of the join must have. In the parse tree those names aren't stored as column references and the pass finds columns by scanning for column references. Example:
```sql
WITH a AS (SELECT id, keep FROM t1),
     b AS (SELECT id, marker FROM t2)
SELECT a.keep FROM a JOIN b USING (id)
```
Only `a.keep` looks used and `id` is pruned:
```
WITH a AS (SELECT keep FROM t1),        -- id is gone
     b AS (SELECT id, marker FROM t2)
SELECT a.keep FROM a JOIN b USING (id)   -- fails since it's joining on a column a doesn't have
```

The same issue comes up when using `NATURAL JOIN` -- it joins on every column name the two sides happen to share. 

### Test Plan

- [ ] `make check` passes

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v3/cte.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/cte.py
@@ -999,28 +999,37 @@ def _joins_naturally(arm: ast.SelectExpression) -> bool:
     return any(join.natural for join in arm.find_all(ast.Join))
 
 
-def _arm_sources(
+def _enclosing_selects(
+    node: ast.Node,
+    outermost: ast.SelectExpression,
+) -> list[ast.SelectExpression]:
+    """The selects a node sits in, innermost first, ending at outermost."""
+    chain: list[ast.SelectExpression] = []
+    current = node.parent
+    while current is not None:
+        if isinstance(current, ast.SelectExpression):
+            chain.append(current)
+            if current is outermost:
+                break
+        current = current.parent
+    return chain
+
+
+def _select_bindings(
     arm: ast.SelectExpression,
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Awesome sauce!

PR #2532, opened 2026-09-09
fixed_grain for level metrics

### LINKED ISSUE #2245: Proposal: Level (and other Derived) Metrics

robinld:
Implements the `fixed_grain` half of [#2245](https://github.com/DataJunction/dj/issues/2245). A base metric can declare the grain at which its aggregate is computed, and DJ broadcasts that value when the query requests a finer grain.

This PR is stacked on [#2502](https://github.com/DataJunction/dj/pull/2502), which implements the complementary `reaggregate` direction. It can be retargeted to `main` after #2502 lands.

```yaml
- name: revenue
  query: SELECT SUM(revenue) FROM sales_f

- name: total_revenue
  query: SELECT SUM(revenue) FROM sales_f
  fixed_grain: []

- name: pct_of_total
  query: SELECT revenue / total_revenue
```

`fixed_grain` has three forms:

- omitted: compute at the query grain, preserving current behavior
- `[]`: compute at the global grain
- `[dimension, ...]`: compute at the declared partition grain

For example, querying `total_revenue` at a finer grain generates the equivalent of:

```sql
SUM(SUM(sales_f_0.revenue_sum)) OVER () AS total_revenue
```

The inner aggregate computes the query-group contribution. The outer window re-merges those contributions within the declared fixed grain. A non-empty grain produces `OVER (PARTITION BY ...)` instead.

## Implementation

- `sql/decompose.py` validates that the declaration belongs to one freely aggregatable base component, propagates it through decomposition, and rejects a later window over a broadcast value.
- The broadcast is written into the combiner during decomposition, naming partition dimensions in unresolved form. Each consumer resolves them as it does any dimension reference — a CTE-qualified column for a query, a bare column for a cube — so the window reaches both query SQL and the SQL stored on a materialized cube.
- `construction/build_v3/cube_matcher.py` skips a cube that does not materialize a metric's partition dimension, mirroring `validate_cube_covers_reaggregate_dimensions`. A cube that does carry it is used normally.
- The declaration is carried through storage, deployment specs, namespace export/copy, REST and GraphQL output, loader allowlists, change classification, and semantic fingerprints.

## Current constraints

The initial implementation accepts only cases where broadcasting can be expressed without changing the metric's meaning.

| Situation | Behavior |
| --- | --- |
| `COUNT(DISTINCT ...)`, a multi-component metric (`AVG`, variance), or a non-decomposable aggregate | Refused; the result cannot be re-merged as a window. Multi-component is unimplemented rather than impossible — it needs every merge windowed, not just the first |
| A declared grain dimension is absent from the query grain | Joined and grouped on as private grain, then projected away — the same treatment a protected `reaggregate` dimension gets. The result stays the shape that was requested |
| A non-empty fixed grain is combined with another fact grain group | Refused; the joined SELECT groups over a `COALESCE`, so a partition naming one group's column is ungrouped |
| A derived metric declares its own fixed grain | Refused; the declaration belongs to the base aggregate |
| A frame accumulates a fixed-grain metric, directly or transitively | Refused; the frame would add the same broadcast value once per row. `MAX`/`AVG` over it are allowed, as is referencing it beside a window — e.g. a trailing sum divided by a global total |
| A cube does not materialize the partition dimension | Cube skipped; DJ builds from the underlying graph. A cube that carries the dimension is used normally |

## Related fixes

- Normalizes both frozen-measure creation paths so metric-level `reaggregate` and `fixed_grain` declarations do not become part of shared measure state or identity.
- Adds both declarations to the GraphQL `extractedMeasures` loader's `load_only`. Reading a field that loader omits raises `MissingGreenlet` on a cold session, which the batch handler turns into a silent null rather than an error.

## Out of scope

- Relative grain declarations such as `{hierarchy, drill_up}`
- The proposed `frame` primitive
- UI support. `reaggregate` gained create/edit fields and node-page display in #2502; the equivalent for `fixed_grain` is deferred to a follow-up, so the declaration is currently settable only through deployment specs and the API.

## Known gaps

- Direct `POST /nodes/metric` validation does not yet reject an unresolvable fixed-grain column, although deployment validation does. The invalid declaration is rejected later when a query attempts to use it.

## Testing

Coverage includes global and partitioned broadcasts, derived ratios, declaration round trips, namespace copying, semantic identity, frozen-measure reuse, GraphQL decomposition loading, window refusal, and cube matching and materialization carrying the broadcast.

### DIFF
--- a/datajunction-server/datajunction_server/construction/build_v3/measures.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/measures.py
@@ -2042,6 +2042,7 @@ def build_grain_group_sql(
     # This is needed for metrics SQL to correctly reference component columns
     component_aliases: dict[str, str] = {}
     reaggregate_dimension_aliases: dict[str, str] = {}
+    fixed_grain_dimension_aliases: dict[str, str] = {}
     internal_dimension_aliases: dict[str, str] = {}
 
     if output_dimension_refs:
@@ -2056,6 +2057,15 @@ def build_grain_group_sql(
         dimension_alias = ctx.alias_registry.register(dimension_ref)
         reaggregate_dimension_aliases[component_name] = dimension_alias
         internal_dimension_aliases[dimension_ref] = dimension_alias
+
+    # A declared partition needs a column in this CTE. It only becomes an output
+    # column if the caller also asked for it; otherwise it is private grain.
+    for _fg_metric_node, fg_component in grain_group.components:
+        for dimension_ref in fg_component.rule.fixed_grain or []:
+            dimension_alias = ctx.alias_registry.register(dimension_ref)
+            fixed_grain_dimension_aliases[dimension_ref] = dimension_alias
+            if dimension_ref not in (output_dimension_refs or ()):
+                internal_dimension_aliases[dimension_ref] = dimension_alias
 
     for metric_node, component in grain_group.components:
         metrics_covered.add(metric_node.name)
@@ -2421,6 +2431,7 @@ def build_grain_group_sql(
         metrics=list(metrics_covered),
         parent_name=grain_group.parent_node.name,
         component_aliases=component_aliases,
+        fixed_grain_dimension_aliases=fixed_grain_dimension_aliases,
         reaggregate_dimension_aliases=reaggregate_dimension_aliases,
         is_merged=grain_group.is_merged,
         component_aggregabilities=grain_group.component_aggregabilities,
--- a/datajunction-server/datajunction_server/construction/build_v3/measures.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/measures.py
@@ -73,6 +73,7 @@ from datajunction_server.construction.build_v3.utils import (
     make_name,
 )
 from datajunction_server.database.node import Node
+from datajunction_server.errors import DJInvalidInputException
 from datajunction_server.internal.scan_estimation import calculate_scan_estimate
 from datajunction_server.models.decompose import Aggregability, MetricComponent
 from datajunction_server.models.node_type import NodeType
@@ -2006,6 +2007,23 @@ def build_grain_group_sql(
         GrainGroupSQL with SQL and metadata for this grain group
     """
     parent_node = grain_group.parent_node
+    query_grain = (
+        output_dimension_refs
+        if output_dimension_refs is not None
+        else set(ctx.dimensions)
+    )
+
+    # Validate before pre-aggregation matching, whose successful path returns
+    # early. The physical source does not change the required output grain.
+    for fixed_grain_metric, component in grain_group.components:
+        for dimension_ref in component.rule.fixed_grain or []:
+            if dimension_ref not in query_grain:
+                raise DJInvalidInputException(
+                    f"Metric `{fixed_grain_metric.name}` declares "
+                    f"`fixed_grain` on `{dimension_ref}`, which is not a "
+                    "requested output dimension. Add it to the query's "
+                    "dimensions before requesting this metric.",
+                )
 
     # Check for matching pre-aggregation
     if ctx.use_materialized and ctx.available_preaggs:
@@ -2042,7 +2060,10 @@ def build_grain_group_sql(
     # This is needed for metrics SQL to correctly reference component columns
     component_aliases: dict[str, str] = {}
     reaggregate_dimension_aliases: dict[str, str] = {}
-    fixed_grain_dimension_aliases: dict[str, str] = {}
+    fixed_grain_partition_aliases: dict[str, str] = {}
+    # Protected reaggregate dimensions this CTE projects privately, under an
+    # internal alias, so the metrics layer can collapse them without
+    # exposing them as output columns.
     internal_dimension_aliases: dict[str, str] = {}
 
     if output_dimension_refs:
@@ -2058,14 +2079,11 @@ def build_grain_group_sql(
         reaggregate_dimension_aliases[component_name] = dimension_alias
         internal_dimension_aliases[dimension_ref] = dimension_alias
 
-    # A declared partition needs a column in this CTE. It only becomes an output
-    # column if the caller also asked for it; otherwise it is private grain.
-    for _fg_metric_node, fg_component in grain_group.components:
-        for dimension_ref in fg_component.rule.fixed_grain or []:
-            dimension_alias = ctx.alias_registry.register(dimension_ref)
-            fixed_grain_dimension_aliases[dimension_ref] = dimension_alias
-            if dimension_ref not in (output_dimension_refs or ()):
-                internal_dimension_aliases[dimension_ref] = dimension_alias
+    for _, component in grain_group.components:
+        for dimension_ref in component.rule.fixed_grain or []:
+            fixed_grain_partition_aliases[dimension_ref] = ctx.alias_registry.register(
+                dimension_ref
+            )
 
     for metric_node, component in grain_group.components:
         metrics_covered.add(metric_node.name)
@@ -2431,7 +2449,7 @@ def build_grain_group_sql(
         metrics=list(metrics_covered),
         parent_name=grain_group.parent_node.name,
         component_aliases=component_aliases,
-        fixed_grain_dimension_aliases=fixed_grain_dimension_aliases,
+        fixed_grain_partition_aliases=fixed_grain_partition_aliases,
         reaggregate_dimension_aliases=reaggregate_dimension_aliases,
         is_merged=grain_group.is_merged,
         component_aggregabilities=grain_group.component_aggregabilities,
@@ -2470,7 +2488,13 @@ def process_metric_group(
 
     # Analyze grain groups - split by aggregability
     output_dimensions = list(ctx.dimensions)
-    output_dimension_refs = set(output_dimensions)
+    dimensions_already_present = set(output_dimensions)
+    # A filter-only dimension sits in ctx.dimensions for join resolution but
+    # never reaches the final projection, so it isn't a real output ref --
+    # otherwise a fixed_grain dimension mentioned only in a filter would
+    # look like a legitimate output and skip the check in
+    # `build_grain_group_sql` that refuses it.
+    output_dimension_refs = dimensions_already_present - ctx.filter_dimensions
     grain_groups = analyze_grain_groups(metric_group, output_dimensions)
 
     # Merge compatible grain groups from same parent into single CTEs
@@ -2491,7 +2515,7 @@ def process_metric_group(
             for dimension_ref in dict.fromkeys(
                 grain_group.reaggregate_component_dimensions.values(),
             )
-            if dimension_ref not in output_dimension_refs
+            if dimension_ref not in dimensions_already_present
         ]
         ctx.dimensions = output_dimensions + internal_dimensions
         try:
--- a/datajunction-server/datajunction_server/construction/build_v3/metrics.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/metrics.py
@@ -403,6 +403,48 @@ def _build_metric_aggregation(
     return expr_ast
 
 
+def validate_fixed_grain_query_shape(
+    decomposed: DecomposedMetricInfo,
+    grain_group_count: int = 1,
+) -> None:
+    """
+    Refuse a query whose shape cannot express the declared partition.
+
+    The window itself is already in the combiner, put there during decomposition,
+    and a partition dimension the query did not ask for is added to the grain by
+    the same mechanism that serves `required_dimensions`. What is left is the one
+    shape that still cannot work.
+    """
+    # Read off the component rules, not the node revision: decomposition
+    # attaches the declaration to the one broadcastable component, so consumers
+    # see it without reaching back into the graph. A derived metric's components
+    # are the union across its parents, so the declaring one need not be first.
+    fixed_grain = next(
+        (
+            comp.rule.fixed_grain
+            for comp in decomposed.components
+            if comp.rule.fixed_grain
+        ),
+        None,
+    )
+    if not fixed_grain:
+        return
+
+    metric_name = decomposed.metric_node.name
+    # Across grain groups the enclosing SELECT groups positionally over a COALESCE
+    # of the shared dimensions, so a partition naming one group's column is
+    # ungrouped there and the engine rejects it.
+    if grain_group_count > 1:
+        raise DJInvalidInputException(
+            f"Metric `{metric_name}` declares a non-empty `fixed_grain` and "
+            "cannot be combined with metrics from another fact table in one "
+            "query: the joined result groups over a `COALESCE` of the shared "
+            "dimensions, so a partition naming one side's column is not grouped "
+            "and the engine rejects it. Query it on its own, or use a global "
+            "grain (`fixed_grain: []`), which partitions by nothing.",
+        )
+
+
 def _build_pre_agg_wrapper_cte(
     alias: str,
     gg: GrainGroupSQL,
@@ -814,8 +856,13 @@ def process_base_metrics(
                     column_name=col_name,
                 )
 
-            # Qualify dimension refs with this grain group's CTE alias
-            qualified_dim_refs = qualify_dimension_refs(dimension_aliases, alias)
+            # Qualify dimension refs with this grain group's CTE alias. The
+            # group's private grain is merged in so a declared partition
+            # resolves even though it is not an output dimension.
+            qualified_dim_refs = qualify_dimension_refs(
+                {**dimension_aliases, **gg.fixed_grain_dimension_aliases},
+                alias,
+            )
             replace_dimension_refs_in_ast(expr_ast, qualified_dim_refs)
             metric_exprs[metric_name] = MetricExprInfo(
                 expr_ast=expr_ast,
@@ -1994,6 +2041,13 @@ def generate_metrics_sql(
         base_grain_groups,
         skip_pre_agg=will_have_base_metrics_cte,
     )
+    # Counted over base groups only: a window grain group is a second group for
+    # the SAME fact table, which is not the cross-fact case this refuses.
+    for fixed_grain_candidate in decomposed_metrics.values():
+        validate_fixed_grain_query_shape(
+            fixed_grain_candidate,
+            len(base_grain_groups),
+        )
     _validate_reaggregate_base_group_join_safety(base_grain_groups)
 
     # Build dimension info and projection
--- a/datajunction-server/datajunction_server/construction/build_v3/metrics.py
+++ b/datajunction-server/datajunction_server/construction/build_v3/metrics.py
@@ -856,11 +856,9 @@ def process_base_metrics(
                     column_name=col_name,
                 )
 
-            # Qualify dimension refs with this grain group's CTE alias. The
-            # group's private grain is merged in so a declared partition
-            # resolves even though it is not an output dimension.
+            # Qualify dimension refs with this grain group's CTE alias.
             qualified_dim_refs = qualify_dimension_refs(
-                {**dimension_aliases, **gg.fixed_grain_dimension_aliases},
+                {**dimension_aliases, **gg.fixed_grain_partition_aliases},
                 alias,
             )
             replace_dimension_refs_in_ast(expr_ast, qualified_dim_refs)
--- a/datajunction-server/datajunction_server/internal/deployment/validation.py
+++ b/datajunction-server/datajunction_server/internal/deployment/validation.py
@@ -527,6 +527,10 @@ class NodeSpecBulkValidator:
             if reaggregate_error is not None:
                 errors.append(reaggregate_error)
 
+            fixed_grain_error = self._check_fixed_grain_dimensions(spec)
+            if fixed_grain_error is not None:
+                errors.append(fixed_grain_error)
+
             cross_fact_error = self._check_cross_fact_dimensions(spec)
             if cross_fact_error is not None:
                 errors.append(cross_fact_error)  # pragma: no cover
@@ -680,6 +684,10 @@ class NodeSpecBulkValidator:
                     if SEPARATOR in rule.dimension:
                         dim_node_name = rule.dimension.rsplit(SEPARATOR, 1)[0]
                         req_dim_node_names.add(dim_node_name)
+            for grain_dim in getattr(spec, "rendered_fixed_grain", None) or []:
+                if SEPARATOR in grain_dim:
+                    dim_node_name = grain_dim.rsplit(SEPARATOR, 1)[0]
+                    req_dim_node_names.add(dim_node_name)
 
         self._all_dim_nodes = dict(self.context.dependency_nodes)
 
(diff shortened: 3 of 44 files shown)


datajunction-server/datajunction_server/construction/build_v3/measures.py
```diff
@@ -2057,6 +2058,15 @@ def build_grain_group_sql(
         reaggregate_dimension_aliases[component_name] = dimension_alias
         internal_dimension_aliases[dimension_ref] = dimension_alias
 
+    # A declared partition needs a column in this CTE. It only becomes an output
+    # column if the caller also asked for it; otherwise it is private grain.
+    for _fg_metric_node, fg_component in grain_group.components:
```
shangyian: **Blocking.** This registers an alias for each `fixed_grain` ref, and the comment above says "A declared partition needs a column in this CTE" — but as far as I can trace, nothing acts on the registration.

- The projection loop at `measures.py:1131` iterates `resolved_dimensions`. `internal_dimension_aliases` is only read *inside* that loop, as a lookup for the output alias — it never adds anything to what the loop visits.
- A dimension reaches `resolved_dimensions` only through `ctx.dimensions`, set at `measures.py:2488-2494` to `output_dimensions + internal_dimensions`, and `internal_dimensions` is built solely from `grain_group.reaggregate_component_dimensions`.

So a fixed-grain dimension that isn't also a requested output dimension never enters the CTE. Building `v3.grain_needs_category` against `dimensions=v3.order_details.order_id` yields a CTE with no `category` column feeding `PARTITION BY order_details_0.category`.

The filter-only case needs a separate fix: `measures.py:1132` skips `ctx.filter_dimensions` *before* the `internal_dimension_aliases` check, so those can't be projected at all.

**Worth settling before patching:** even once the CTE emits the column, the outer `SELECT` has it bare in `PARTITION BY` while grouping only by the output dimensions, which Spark and Trino both reject. `reaggregate` avoids this because its private dimension is always collapsed inside an aggregate (`metrics.py:356`). So this likely isn't "add fixed_grain refs to `internal_dimensions`" — the outer layer needs equivalent treatment, and adding the dimension to the outer `GROUP BY` would change the result shape the PR is careful to preserve.

datajunction-server/tests/construction/build_v3/fixed_grain_test.py
```diff
@@ -0,0 +1,1070 @@
...
+    assert response.status_code == 200, response.json()
+    body = response.json()
+
+    # The result is exactly what was asked for — no extra column.
+    assert [col["name"] for col in body["columns"]] == [
+        "order_id",
+        "grain_needs_category",
+    ]
+    # And the partition resolved against a real CTE column, not a raw ref.
+    assert "PARTITION BY order_details_0.category" in body["sql"]
```
shangyian: This substring is present in SQL that no engine will accept — see the `measures.py` comment. The CTE it refers to has no `category` column.

The column-list assertion just above is a full equality and does its job, so the output shape is pinned; it's the runnability that isn't. Asserting the CTE projects the column, or comparing the whole statement, would have caught it.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -1174,8 +1343,59 @@ def _extract_base(
         if reaggregate is not None and dimension_reaggregate_rules(reaggregate):
             self._attach_reaggregate_spec(components, reaggregate)
 
+        if fixed_grain is not None:
+            self._attach_fixed_grain_spec(components, fixed_grain, query_ast)
+
         return components, query_ast
 
+    def _attach_fixed_grain_spec(
```
shangyian: This only runs during decomposition, and nothing decomposes at write time, so an unsupported shape is accepted and then permanently unqueryable:

```
POST /nodes/metric/
{"query": "SELECT COUNT(DISTINCT order_id) FROM v3.order_details",
 "fixed_grain": []}
```

returns 201 with `status: valid`, and every later `GET /sql/metrics/v3/` returns 422 `Unsupported fixed_grain metric shape`.

"Known gaps" mentions this for unresolvable columns, but it covers unsupported aggregate shapes too, and the node looks healthy until someone queries it. `fixed_grain_test.py:97` already describes this exact shape of problem for grain dimensions — "the first sign of trouble is a permanent failure at query time" — so one write-time hook would cover both.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -770,6 +771,104 @@ class MetricData:
 # =============================================================================
 
 
+# Window functions that return their input unchanged when every row in the
+# frame holds the same value — which is exactly what a broadcast produces. An
+# unrecognised function is treated as accumulating, so the guard fails closed.
+_IDEMPOTENT_OVER_A_CONSTANT = frozenset(
```
shangyian: These are exempted from the accumulation guard because a broadcast value is row-constant. It's constant within the *fixed-grain* partition, not across the frame's partition.

```
AVG(total_by_category) OVER (PARTITION BY region)
```

over a parent declaring `fixed_grain: [category]` passes the guard and returns a row-weighted average of per-category totals. `FIRST_VALUE` without `ORDER BY` is nondeterministic on top of that.

Comparing the frame's `PARTITION BY` against the declared grain would settle it; so would dropping the exemptions and failing closed, as the comment suggests is the intent.

datajunction-server/datajunction_server/sql/decompose.py
```diff
@@ -941,6 +1041,42 @@ async def extract(
...
+            # Checked on the extracted components, not this parent's own
+            # declaration: decomposition recurses through derived parents, so a
+            # grain declared two levels down still arrives on a component here.
+            #
+            # A derived metric is only at risk when the frame actually
+            # accumulates the parent; referencing it beside a window is fine.
+            # A base metric has no reference to match, so the hazard is its own
+            # aggregate already carrying a window — not a window anywhere in the
+            # query, which `SUM(x - LAG(x) OVER (...))` has quite legitimately.
+            windowed_over_the_grain = (
```
shangyian: Small perf thing: this is evaluated before the `fixed_grain is not None` test on the next line, so every base metric pays a second full parse of a query already parsed into `base_ast`, whether or not `fixed_grain` is involved. Swapping the two conditions makes it pay-for-use.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
```diff
@@ -403,6 +403,48 @@ def get_comp_aggregability(comp_name: str) -> Aggregability:
...
+        return
+
+    metric_name = decomposed.metric_node.name
+    # Across grain groups the enclosing SELECT groups positionally over a COALESCE
+    # of the shared dimensions, so a partition naming one group's column is
+    # ungrouped there and the engine rejects it.
+    if grain_group_count > 1:
+        raise DJInvalidInputException(
+            f"Metric `{metric_name}` declares a non-empty `fixed_grain` and "
+            "cannot be combined with metrics from another fact table in one "
```
shangyian: Just so I understand -- this is refusing to join two upstreams if they're at different grains right?

datajunction-server/datajunction_server/internal/validation.py
```diff
@@ -471,12 +482,31 @@ async def validate_node_data(
             if invalid_reaggregate_functions
             else []
         )
+        invalid_fixed_grain_dimensions_error = (
+            [
+                DJError(
+                    code=ErrorCode.INVALID_COLUMN,
+                    message=(
+                        "Node definition declares a fixed grain referencing "
```
shangyian: Does the fixed grain have to be a column that's on a parent node or can it be a any dimension that's joinable into the parent node?

datajunction-server/datajunction_server/construction/build_v3/measures.py
shangyian, line 2063: Hmm it looks like this registers an alias for each `fixed_grain` ref, and the comment above says "A declared partition needs a column in this CTE", but I don't see anything that acts on `fixed_grain_dimension_aliases`?

I don't think a fixed-grain dimension that isn't also a requested output dimension enters the CTE?

datajunction-server/tests/construction/build_v3/fixed_grain_test.py
shangyian: This substring is present in SQL that no engine will accept — see the `measures.py` comment. The CTE it refers to has no `category` column.

The column-list assertion just above is a full equality and does its job, so the output shape is pinned; it's the runnability that isn't. Asserting the CTE projects the column, or comparing the whole statement, would have caught it.

datajunction-server/datajunction_server/sql/decompose.py
shangyian, line 1351: Is it worth also running this check during metric creation? Basically instead of accepting a metric like:
```
POST /nodes/metric/
{"query": "SELECT COUNT(DISTINCT order_id) FROM v3.order_details",
 "fixed_grain": []}
```

And then saying that the unsupported shape is unqueryable later e.g. when calling `GET /sql/metrics/v3/`.

datajunction-server/datajunction_server/sql/decompose.py
shangyian, line 777: Is this still true if there's a metric like `AVG(total_by_category) OVER (PARTITION BY region)`? I think `MIN` and `MAX` should continue to work (e.g., `MIN(total_by_category) OVER (PARTITION BY region)` or smallest category total in this region). But `AVG` would be dependent on the grain right?

datajunction-server/datajunction_server/sql/decompose.py
shangyian: Small perf thing: this is evaluated before the `fixed_grain is not None` test on the next line, so every base metric pays a second full parse of a query already parsed into `base_ast`, whether or not `fixed_grain` is involved. Swapping the two conditions makes it pay-for-use.

datajunction-server/datajunction_server/construction/build_v3/metrics.py
shangyian, line 440: Just so I understand, this is refusing to join together two upstreams that are at different grains right?

datajunction-server/datajunction_server/internal/deployment/validation.py
```diff
@@ -841,6 +849,44 @@ def _check_reaggregate_dimension(self, spec: NodeSpec) -> DJError | None:
...
+            self._all_dim_nodes,
+        )
+
+        if not invalid:
+            return None
+
+        return DJError(
+            code=ErrorCode.INVALID_COLUMN,
+            message=(
+                "Node definition declares a fixed grain referencing columns "
```
shangyian, line 884: Shouldn't it be possible to declare a fixed grain if it's just any dimension attribute? As in, it doesn't have to be restricted to just the columns on the parent node right

datajunction-server/tests/construction/build_v3/fixed_grain_test.py
```diff
@@ -0,0 +1,1070 @@
+"""Tests for the `fixed_grain` metric declaration."""
```
shangyian, line 1: Can these tests assert the full SQL so that it's easier to see what the query generation is doing?

datajunction-server/datajunction_server/construction/build_v3/metrics.py
robinld: Yes

datajunction-server/datajunction_server/internal/deployment/validation.py
robinld: You're right, it actually does support this, this error is stale.

datajunction-server/datajunction_server/construction/build_v3/measures.py
robinld: The names are confusing. I'll make some changes to (hopefully) make this less confusing.

shangyian (approved):
LGTM!

PR #2547, opened 2026-09-16
Optimize semantic fingerprint computation

shangyian:
### Summary

This PR adds in a few optimizations to make computing fingerprints more efficient. The basic premise is that fingerprint computation would hash and reparse every node in the graph on push, even when only a handful of nodes changed.

This scopes the work to what actually needs recomputing:
  - Compute fingerprints only for the ancestor closure of the requested nodes, not the whole graph.
  - Skip computing a fingerprint for nodes outside the changed/downstream set (use their existing fingerprint instead).
  - Reuse existing spec objects (and their parsed ASTs / parent-candidate caches) for names known to be unchanged, and share a fingerprint cache across the current/proposed graph snapshots.
  - Skip ANTLR re-parsing when the query text is unchanged, and reuse ASTs already parsed by impact propagation.

#### Benchmarking

| Metric | Pre-PR Baseline | Post-PR |
|---|---|---|
| Fingerprint phase | 6833ms | 3370ms |
| Nodes reaching deploy pipeline | 962 of 963 | 79 of 963 |
| Deploy-nodes phase | 10659ms | 871ms |
| Total deployment time | 27771ms | 10385ms |

### Test Plan

- [ ] `make check` passes


philipfweiss (approved):
Thanks Yian!

datajunction-server/datajunction_server/internal/impact.py
```diff
@@ -89,12 +89,13 @@ async def propagate_impact(
...
     Returns:
-        List of DownstreamImpact describing each affected downstream node.
+        DownstreamImpact list, plus non-metric ASTs already parsed here,
+        reusable by the fingerprint pass.
     """
     start = time.perf_counter()
     all_root_names = changed_node_names | deleted_node_names
     if not all_root_names and not changed_link_node_names:
-        return []
+        return [], {}
```
philipfweiss: I don't usually love the `[], {}` return type, would prefer to just package into a dataclass at this point.

datajunction-server/datajunction_server/internal/impact.py
```diff
@@ -457,7 +458,7 @@ async def _revalidate_and_apply(
         by_depth[impact.depth].append(impact)
 
     if not by_depth:
-        return results
+        return results, {}
```
philipfweiss: Same here and below, I feel like there is probably a cleaner datatype here.

datajunction-server/datajunction_server/internal/impact.py
shangyian: Makes sense! Cleaning this up

PR #2548, opened 2026-09-16
CEL Checks Engine

shangyian:
### Summary

Adds a CEL expression-evaluation layer for governance checks. This PR doesn't wire in calling the checks engine at deployment time and is just focused on the engine itself.

A defined check has a name, an optional `when` guard, a `condition`, and a gate. Both clauses are [CEL](https://cel.dev/) over a declared set of bindings, which keeps governance rules in configuration rather than in DJ's code. See https://pkg.go.dev/cel.dev/cel-go/common/overloads

### Test Plan

- [ ] PR has an associated issue: #
- [ ] `make check` passes
- [ ] `make test` shows 100% unit test coverage

PR #2554, opened 2026-09-19
Declare checks and rulesets in a deployment spec

shangyian:
### Summary

Enable a deployment spec to include checks and rulesets. This PR builds on top of the CEL evaluation layer in #2548.

```yaml
checks:
  - name: demo.owner_present
    condition: "size(node.owners) >= 1"
    gate: warn

rulesets:
  - name: baseline
    checks: [demo.owner_present]
  - name: strict
    when: "node.custom_metadata.sample.size != null"
    includes: [baseline]
    checks: [demo.column_descriptions]
```

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/internal/checks/manifest.py
```diff
@@ -0,0 +1,101 @@
...
+    checks:
+      - name: demo.owner_present
+        description: Every entity has an owner.
+        condition: "size(node.owners) >= 1"
+        gate: warn
+
+      - name: demo.primary_key_set
+        description: Dimensions declare a primary key.
+        when: "node.type == 'dimension'"     # only applies to dimensions
+        condition: "size(node.primary_key) >= 1"
```
betodealmeida, line 14: One thing that might be worth adding to the docs is guidance on how to use `when` vs `condition`; I assume someone could ask why not just write:

```yaml
condition: "node.type != 'dimension' || size(node.primary_key) >= 1"
```

In the docs we could say something like:

> Avoid folding applicability into condition, because an inapplicable check should be reported as skipped rather than passed, especially for regression-based gates.

PR #2556, opened 2026-09-19
Run checks during deployment

shangyian:
### Summary

Run declared checks during deployment. This builds on #2548 (the CEL evaluation layer) and #2554 (`checks` and `rulesets` in the deployment spec).

Every node the deploy creates, updates, skips or removes is projected into a set of bindings, with the checks running before the plan is applied:
```
node          the node as declared
previous      the same node as currently deployed
dependencies  its downstreams
change.kind   create, update, delete or noop
```

```yaml
checks:
  - name: demo.primary_key_set
    when: "node.node_type == 'dimension'"
    condition: "size(node.primary_key) >= 1"
    gate: block

  - name: demo.domain_tagged
    condition: "node.tags.exists(t, t.tag_type == 'domain')"
    gate: warn
```

Setting gate to `warn` just records the failure, setting it to `block` will block the deploy, and `block_on_regression` refuses only if the check passed against the deployed state. Each check/ruleset is recorded via `DeploymentResult`.

### Test Plan

- [ ] `make check` passes


datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -624,6 +657,194 @@ async def _authorize_deployment_plan(self, plan: DeploymentPlan) -> None:
...
+                operation=operation,
+                declared=declared.properties,
+                tag_types=tag_types,
+            )
+            previous_bindings = None
+            if wants_previous and previous is not None:
+                previous_bindings = build_bindings(
+                    previous,
+                    previous=previous,
+                    dependencies=self._checked_dependencies(plan, name, {}),
```
betodealmeida: This is using dependencies from the current plan, no? Should we use dependencies from `existing_specs` instead of `plan.node_graph` here?

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -624,6 +657,194 @@ async def _authorize_deployment_plan(self, plan: DeploymentPlan) -> None:
...
+                        gate=str(result.gate),
+                    )
+                    for result in results
+                ],
+                rulesets=[
+                    NodeRulesetVerdict(ruleset=outcome.ruleset, verdict=outcome.verdict)
+                    for outcome in outcomes
+                ],
+            )
+            self.check_results.append(node_results)
```
betodealmeida, line 785: This doesn't seem to be surfaced to the API, should we add it to `DeploymentInfo`?

datajunction-server/datajunction_server/internal/deployment/checks.py
```diff
@@ -0,0 +1,364 @@
...
+    Project one node spec into the `node` binding.
+    """
+    dumped = spec.model_dump() if spec is not None else {}
+    projected = {
+        name: _cel_safe(dumped.get(name), empty)
+        for name, empty in _NODE_DEFAULTS.items()
+    }
+    # Both are excluded from the dump, so they are read off the spec.
+    projected["name"] = spec.rendered_name if spec is not None else ""
+    projected["namespace"] = (spec.namespace or "") if spec is not None else ""
```
betodealmeida: Isn't `spec.namespace` usually just the top-level namespace? Should we do something like this instead?

```python
projected["namespace"] = get_namespace_from_name(spec.rendered_name)
```

datajunction-server/datajunction_server/internal/deployment/checks.py
```diff
@@ -0,0 +1,364 @@
...
+    Set default values for CEL reads.
+    """
+    defaults: dict[str, Any] = {}
+    pending = [base]
+    while pending:
+        cls = pending.pop()
+        for name, info in cls.model_fields.items():
+            if name in skip:
+                continue
+            empty: Any = [] if _is_sequence_field(info.annotation) else ""
```
betodealmeida: Do we need type-dependent defaults? Eg, for a rule:

```
  when: node.node_type == 'metric'
  condition: node.significant_digits == null || node.significant_digits >= 1
```

datajunction-server/datajunction_server/internal/checks/validator.py
```diff
@@ -76,7 +72,7 @@ def ok(self) -> bool:
 
 def load_checks(
     checks: Iterable[CheckSpec],
-    fixtures: Sequence[Activation],
+    fixtures: Sequence[Bindings],
```
betodealmeida, line 74: I wonder if in line 85 below we should only validate `condition` when `when` is true? Otherwise the condition might depend on attributes not present in the fixture, which might also cause type mismatch (since missing attributes default to empty string).

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
```diff
@@ -624,6 +657,194 @@ async def _authorize_deployment_plan(self, plan: DeploymentPlan) -> None:
...
+        """Upstreams as (name, spec), preferring the spec this deploy will write."""
+        return [
+            (upstream, in_flight.get(upstream) or plan.existing_specs.get(upstream))
+            for upstream in plan.node_graph.get(node_name, [])
+        ]
+
+    @staticmethod
+    def _node_check_status(
+        results: Sequence[CheckResult],
+        outcomes: Sequence[RulesetOutcome],
```
robinld: looks like `outcomes` is unused here

datajunction-server/datajunction_server/internal/deployment/checks.py
```diff
@@ -0,0 +1,364 @@
...
+            # A list default wins: one subclass typing it as a list is enough.
+            if defaults.get(name) != []:
+                defaults[name] = empty
+        pending.extend(cls.__subclasses__())
+    return defaults
+
+
+# custom_metadata and tags are projected per node instead.
+_NODE_DEFAULTS = _spec_defaults(NodeSpec, frozenset({"custom_metadata", "tags"}))
```
robinld: should `skip` include `dimension_links` as well since we project it separately later?

datajunction-server/datajunction_server/internal/deployment/checks.py
shangyian: Ah you're right, fixing this!

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
shangyian: Ah yes, definitely

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
shangyian: Good catch, fixed so that it now uses the existing specs instead of `plan.node_graph`

datajunction-server/datajunction_server/internal/deployment/checks.py
shangyian: Yeah, makes sense, added that here https://github.com/DataJunction/dj/blob/160308f1a/datajunction-server/datajunction_server/internal/deployment/checks.py#L67-L80

datajunction-server/datajunction_server/internal/deployment/orchestrator.py
shangyian: good point, removed!

datajunction-server/datajunction_server/internal/deployment/checks.py
shangyian: Yeah, it was always overwritten I think, added it to https://github.com/DataJunction/dj/commit/d6426b1119a1d1ba13a088a892a90580dbb8bd41#diff-4bbe21c96f0197d0d5bca95bcde61db5ddecc8207843eae6583b356877ed38b9R102

robinld (approved):

